PR #222, opened 2019-06-20
Bugs/en 2555 header resolvers storer nonce

iulianpascalau:
HeaderResolvers are now a single struct. Any entity from Elrond Network can resolve a header (shard or metachain) by nonce.


dataRetriever/resolvers/headerResolver.go
```diff
@@ -100,12 +107,18 @@ func (hdrRes *HeaderResolver) resolveHeaderFromNonce(key []byte) ([]byte, error)
...
-
 	if hash == nil || !ok {
-		return nil, nil
+		hash, err = hdrRes.hdrNoncesStorage.Get(key)
+		if err != nil {
+			log.Debug(err.Error())
+		}
+
+		if len(hash) == 0 {
+			return nil, nil
```
SebastianMarian: This kind of situations should be resolved generally. We need to trust that if err is nil then the object is not nil and vice versa. Please add a task in Jira for this problem to be discussed and to be resolved generally

iulianpascalau: Ok, user story EN-2581 has been created

AdoAdoAdo: I think there's no sense in using the output "hash" if there was an error reported. 
We could just return nil, err. We should also avoid returning nil, nil, instead we can return nil, ErrNotFound or something similar.

dataRetriever/resolvers/headerResolver.go
```diff
@@ -123,6 +136,43 @@ func (hdrRes *HeaderResolver) resolveHeaderFromNonce(key []byte) ([]byte, error)
 	return buff, nil
 }
 
+// resolveHeaderFromHash resolves a header using its key (header hash)
+func (hdrRes *HeaderResolver) resolveHeaderFromHash(key []byte) ([]byte, error) {
+	value, ok := hdrRes.headers.Peek(key)
+	if !ok {
+		return hdrRes.hdrStorage.Get(key)
```
AdoAdoAdo: if !ok we can just return notFound, why do we need to do the get? the only thing get does in addition is updating the recently used-ness of the element.

dataRetriever/resolvers/headerResolver_test.go
```diff
@@ -315,6 +361,11 @@ func TestHeaderResolver_ProcessReceivedMessageRequestNonceTypeNotFoundInHdrNonce
 		&mock.CacherStub{},
 		headersNonces,
 		&mock.StorerStub{},
+		&mock.StorerStub{
+			GetCalled: func(key []byte) (i []byte, e error) {
+				return nil, nil
```
AdoAdoAdo: I don't see any any use in returning nil, nil, maybe change to a NotFound error?

dataRetriever/resolvers/headerResolver.go
iulianpascalau: I think we should discuss about this as using an error defined in implementation package will render the decoupling through interfaces useless by tying the higher level package to one of the implementing package. The higher level package will have something like
```GO
hash, err := implementingPackage.Get(...)
if err != nil{
     if err == implementingPackage.ErrKeyNotFound{
           //do something special here
           //return something
     }
     return nil, err
}
The problem worsens when we have more implementations. Having a "core" kind of package that will hold common errors doesn't look too appealing for me.

The idea coined with Camil is to return something in this only special case where there is no error (cacher, storer, trie, etc. perfomed ok) but the data is simply not there. We can implement a kind of "empty", not nil objects for this purpose only. In the case of []byte we could return make([]byte, 0) and in the case of interface{} we could return struct{}{}. In this way we keep our programming pattern in which we can not have situations like (nil, nil) or (object, error) and also we can easily check that the returned object is an empty one, telling that the implementation used couldn't retrieve the required data because it does not exists (not that something malfunctioned). We also need to rework the get func from level db.

```

dataRetriever/factory/metachain/resolversContainerFactory.go
```diff
@@ -114,7 +114,7 @@ func (rcf *resolversContainerFactory) generateShardHeaderResolvers() ([]string,
 		// TODO: Should fix this to ask only other shard peers
 		excludePeersFromTopic := factory.ShardHeadersForMetachainTopic + shardC.CommunicationIdentifier(shardC.SelfId())
 
-		resolver, err := rcf.createOneShardHeaderResolver(identifierHeader, excludePeersFromTopic)
+		resolver, err := rcf.createOneShardHeaderResolver(identifierHeader, excludePeersFromTopic, idx)
```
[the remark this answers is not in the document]
iulianpascalau: ok

dataRetriever/resolvers/headerResolver.go
iulianpascalau: Because the peek is done on the data pool and the get is done on the storage

dataRetriever/resolvers/headerResolver_test.go
iulianpascalau: This is actually a stub usage. 
It will be changed to return something else.

dataRetriever/factory/metachain/resolversContainerFactory.go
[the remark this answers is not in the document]
iulianpascalau: Done

dataRetriever/resolvers/headerResolver_test.go
iulianpascalau: Changed all occurrences.

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -71,35 +80,72 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithMemMessenger(t *testing.T) {
...
+	chanDone2 := make(chan struct{})
 
 	dPoolRequestor.Headers().RegisterHandler(func(key []byte) {
 		hdrStored, _ := dPoolRequestor.Headers().Peek(key)
+		fmt.Printf("Recieved hash %v\n", base64.StdEncoding.EncodeToString(key))
 
-		if reflect.DeepEqual(hdrStored, &hdr) && hdr.Signature != nil {
-			chanDone <- true
+		if reflect.DeepEqual(hdrStored, &hdr1) && hdr1.Signature != nil {
+			assert.Equal(t, hdrStored, &hdr1)
```
AdoAdoAdo: the assert seems extraneous as it is already validated by the conditional above.

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -71,35 +80,72 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithMemMessenger(t *testing.T) {
...
+			fmt.Printf("Recieved header with hash %v\n", base64.StdEncoding.EncodeToString(key))
+			chanDone1 <- struct{}{}
 		}
 
-		assert.Equal(t, hdrStored, &hdr)
-
+		if reflect.DeepEqual(hdrStored, &hdr2) && hdr2.Signature != nil {
+			assert.Equal(t, hdrStored, &hdr2)
```
AdoAdoAdo: same as above, I think the assert can be removed.

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -71,35 +80,72 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithMemMessenger(t *testing.T) {
...
+	hdrHash1 := hasher.Compute(string(hdrBuff1))
+	hdrBuff2, _ := marshalizer.Marshal(&hdr2)
+	hdrHash2 := hasher.Compute(string(hdrBuff2))
 
-	//Step 2. resolver has the header
-	dPoolResolver.Headers().HasOrAdd(hdrHash, &hdr)
-	dPoolResolver.HeadersNonces().HasOrAdd(0, hdrHash)
+	//Step 2. resolver has the headers
+	dPoolResolver.Headers().HasOrAdd(hdrHash1, &hdr1)
+	dPoolResolver.HeadersNonces().HasOrAdd(0, hdrHash1)
```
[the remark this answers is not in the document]
iulianpascalau: Done

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Right. Remained from old impl. Will change

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Assert removed

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Assert removed

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -33,7 +33,7 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithNetMessenger(t *testing.T) {
 
 	shardCoordinator := &sharding.OneShardCoordinator{}
 
-	fmt.Println("Requestor:")
+	fmt.Println("Requester:")
 	nRequestor, mesRequestor, _, resolversFinder := createNetNode(
```
SebastianMarian: requestor?

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -33,7 +33,7 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithNetMessenger(t *testing.T) {
...
 		dPoolRequestor,
```
SebastianMarian: requestor?

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -33,7 +33,7 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithNetMessenger(t *testing.T) {
...
 		storeRequestor,
```
SebastianMarian: requestor?

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -49,8 +49,8 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithNetMessenger(t *testing.T) {
 		shardCoordinator,
 	)
 
-	nRequestor.Start()
-	nResolver.Start()
+	_ = nRequestor.Start()
```
SebastianMarian: requestor?

integrationTests/singleShard/block/interceptedRequestHdr_test.go
```diff
@@ -49,8 +49,8 @@ func TestNode_GenerateSendInterceptHeaderByNonceWithNetMessenger(t *testing.T) {
...
+	_ = nResolver.Start()
 	defer func() {
 		_ = nRequestor.Stop()
```
SebastianMarian: requestor?

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Removed all occurrences from the project

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Removed all occurrences from the project

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Removed all occurrences from the project

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Removed all occurrences from the project

integrationTests/singleShard/block/interceptedRequestHdr_test.go
iulianpascalau: Removed all occurrences from the project

PR #224, opened 2019-06-21
Recreate trie and request shard headers missing

SebastianMarian:
* Fixed bug in boostrap from storage when recreate trie fails
* Fixed bug in CreateBlockHeader when checkAndRequestIfShardHeadersMissing was not called in every situation and with wrong round

### DIFF
--- a/process/sync/baseSync.go
+++ b/process/sync/baseSync.go
@@ -111,13 +111,14 @@ func (boot *baseBootstrap) loadBlocks(
 		}
 
 		if err == nil {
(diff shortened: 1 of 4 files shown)


process/sync/baseSync.go
```diff
@@ -111,15 +111,16 @@ func (boot *baseBootstrap) loadBlocks(
 		}
 
 		if err == nil {
+			err = boot.accounts.RecreateTrie(boot.blkc.GetCurrentBlockHeader().GetRootHash())
```
raduchis: In the current version, the error is always nil. Does this change anything in the logic that calls this method?

raduchis: ok

PR #225, opened 2019-06-24
Feat/en 2488 add data gas price gas limit to tx

ccorcoveanu:
Passed gasLimit, gasPrice from REST endpoints up to the node.
This PR also contains the option to choose on which port we start the REST webserver and the option to explicitly set it OFF.


api/transaction/routes.go
```diff
@@ -119,7 +119,7 @@ func SendTransaction(c *gin.Context) {
 		return
 	}
 
-	txHash, err := ef.SendTransaction(gtx.Nonce, gtx.Sender, gtx.Receiver, gtx.Value, gtx.Data, signature)
+	txHash, err := ef.SendTransaction(gtx.Nonce, gtx.Sender, gtx.Receiver, gtx.Value, 0, 0, gtx.Data, signature)
```
sasurobert: change 0 to gtx.GasPrice and gtx.GasLimit

ccorcoveanu: Done, also removed some unused code

PR #226, opened 2019-06-24
Feat/en 2595 bootstrap from a given round

SebastianMarian:


cmd/node/main.go
```diff
@@ -158,6 +158,12 @@ VERSION:
 		Usage: "The file containing the secret keys which ...",
 		Value: "./config/initialNodesSk.pem",
 	}
+	// boostrapRoundIndex defines a flag that specifies the maximum number of nodes which will be used from the initialNodes
```
sasurobert: change the comment

process/block/metablock.go
```diff
@@ -396,6 +396,7 @@ func (mp *metaProcessor) CommitBlock(
 		return err
 	}
 
+	//TODO: Should be analyzed if this action is really necessary or not (right now there is no action of removing them)
```
sasurobert: it is necessary to save into storage. should add remove from pool as well.

process/block/shardblock.go
```diff
@@ -465,6 +465,7 @@ func (sp *shardProcessor) CommitBlock(
 		return err
 	}
 
+	//TODO: Should be analyzed if this action is really necessary or not (right now there is no action of removing them)
```
sasurobert: it is necessary to save into storage. should add remove from pool as well.

process/sync/baseSync.go
```diff
@@ -97,22 +98,25 @@ func (boot *baseBootstrap) loadBlocks(
 
 		for i := highestNonceInStorer - blockFinality - lastBlocksToSkip; i <= highestNonceInStorer; i++ {
 			if i > highestNonceInStorer-lastBlocksToSkip {
-				errNotCritical := boot.removeBlock(i, blockUnit, hdrNonceHashDataUnit)
-				if errNotCritical != nil {
-					log.Info(errNotCritical.Error())
-				}
+				err = boot.removeBlock(i, blockUnit, hdrNonceHashDataUnit)
```
sasurobert: this error became not critical. Is it ok ?

process/sync/baseSync.go
```diff
@@ -135,13 +139,10 @@ func (boot *baseBootstrap) applyBlock(
 		return err
 	}
 
-	headerNoncePool := boot.headersNonces
-	if headerNoncePool == nil {
-		return process.ErrNilDataPoolHolder
+	if header.GetRound() > boot.boostrapRoundIndex {
```
sasurobert: the previous for does not know about the bootstrapRoundIndex. Is it blockFinality ? If yes, then rename the variable.

process/sync/shardblock.go
```diff
@@ -172,6 +174,10 @@ func (boot *ShardBootstrap) applyNotarizedBlock(nonce uint64, notarizedHdrNonceH
 		return err
 	}
 
+	if header.GetRound() > boot.boostrapRoundIndex {
```
sasurobert: is this called only in case of bootstrap ?

cmd/node/main.go
SebastianMarian: Changed

process/block/metablock.go
SebastianMarian: Yes, in storer is necessary but I was referring here if it is necessary to put it in the pool. Changed the comment of the TODO

process/block/shardblock.go
SebastianMarian: Yes, in storer is necessary but I was referring here if it is necessary to put it in the pool. Changed the comment of the TODO

process/sync/baseSync.go
SebastianMarian: Changed back as it was a problem here anyway. The removing mechanism is good for cleaning the storer but if something goes wrong with it this should not be critical as anyway the node will request the missing info from network and than it will committed it back to the storer (overwrite if it is the same)

process/sync/shardblock.go
SebastianMarian: Yes, this mechanism is called from the StartSync() method which is called only one time when the node starts and only is it has something in the storage

process/sync/baseSync.go
SebastianMarian: bootstrapRoundIndex is a global variable of baseBoostrap structure and it is known in whole class. It is passed from main initialization and it's came from the given flag

SebastianMarian: It represents the round index until which the node will bootstrap from storage. After this it will request the needed info from network and it will also delete all the info with a greater round than this parameter which is stored in its data base

process/sync/baseSync.go
```diff
@@ -121,6 +123,15 @@ func (boot *baseBootstrap) loadBlocks(
 		}
 	}
 
+	firstNonceToDeleteFromStorer := highestNonceInStorer - lastBlocksToSkip + 1
+
+	for i := firstNonceToDeleteFromStorer; i <= highestNonceInStorer; i++ {
```
raduchis: firstNonceToDeleteFromStorer is actually the i from the previous for

process/sync/baseSync.go
```diff
@@ -121,6 +123,15 @@ func (boot *baseBootstrap) loadBlocks(
...
+		errNotCritical := boot.removeBlock(i, blockUnit, hdrNonceHashDataUnit)
+		if errNotCritical != nil {
+			log.Info(fmt.Sprintf("remove block with nonce %d: %s\n", i, err.Error()))
+		}
+	}
+
```
raduchis: I would split this entire method in 3 smaller methods:
1. computeHighestNonce
2. computeLatestCorrectBlock
3. removeAfterLatestCorrectBlock

process/sync/baseSync.go
```diff
@@ -227,9 +237,13 @@ func (boot *baseBootstrap) loadNotarizedBlocks(blockFinality uint64,
 			return process.ErrBoostrapFromStorage
 		}
 
-		for i := highestNonceInStorer - blockFinality - lastBlocksToSkip; i <= highestNonceInStorer-lastBlocksToSkip; i++ {
+		firstNonceToApplyFromStorer := highestNonceInStorer - blockFinality - lastBlocksToSkip
```
raduchis: looks very similar to the loadBlocks method. Can you merge them some how?

process/sync/baseSync.go
```diff
@@ -95,24 +96,25 @@ func (boot *baseBootstrap) loadBlocks(
 			return process.ErrBoostrapFromStorage
```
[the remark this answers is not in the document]

process/sync/baseSync.go
```diff
@@ -95,24 +96,25 @@ func (boot *baseBootstrap) loadBlocks(
...
+			err = boot.applyBlock(i, getHeader, getBlockBody)
+			if err != nil {
+				log.Info(fmt.Sprintf("apply block with nonce %d: %s\n", i, err.Error()))
+				lastBlocksToSkip++
+				break
 			}
 		}
 
 		if err == nil {
 			err = boot.accounts.RecreateTrie(boot.blkc.GetCurrentBlockHeader().GetRootHash())
```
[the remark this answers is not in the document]

process/sync/baseSync.go

process/sync/baseSync.go

process/sync/baseSync.go
SebastianMarian: Refactored, and extract the common code

process/block/shardblock.go
```diff
@@ -821,25 +822,15 @@ func (sp *shardProcessor) computeMissingHeaders(header *block.Header) [][]byte {
...
-	metaBlockCache := sp.dataPool.MetaBlocks()
-	if metaBlockCache == nil {
-		return missingHeaders
-	}
-
 	for i := 0; i < len(header.MetaBlockHashes); i++ {
-		obj, ok := metaBlockCache.Peek(header.MetaBlockHashes[i])
-		if !ok {
+		hdr, _ := process.GetMetaHeaderFromPool(header.MetaBlockHashes[i], sp.dataPool.MetaBlocks())
+		if hdr == nil {
```
sasurobert: use the returned error instead.

process/sync/baseSync.go
```diff
@@ -72,75 +73,104 @@ func (boot *baseBootstrap) loadBlocks(
...
+					err.Error()))
 				continue
 			}
 
 			break
 		}
 	}
 
+	if currentNonce <= blockFinality {
+		return process.ErrBoostrapFromStorage
```
sasurobert: return a more specific error.

process/sync/baseSync.go
```diff
@@ -72,75 +73,104 @@ func (boot *baseBootstrap) loadBlocks(
...
+
+func (boot *baseBootstrap) cleanupStorage(
+	removeBlockBody func(uint64, dataRetriever.UnitType, dataRetriever.UnitType) error,
+	nonce uint64,
+	blockUnit dataRetriever.UnitType,
+	hdrNonceHashDataUnit dataRetriever.UnitType) {
+
+	errNotCritical := removeBlockBody(nonce, blockUnit, hdrNonceHashDataUnit)
+	if errNotCritical != nil {
+		log.Info(fmt.Sprintf("remove block body with nonce %d: %s\n", nonce, errNotCritical.Error()))
```
sasurobert: log.debug ?

process/sync/baseSync.go
```diff
@@ -240,9 +259,55 @@ func (boot *baseBootstrap) loadNotarizedBlocks(blockFinality uint64,
 		}
 	}
 
+	if currentNonce <= blockFinality {
+		return process.ErrBoostrapFromStorage
```
sasurobert: return a more specific error

process/sync/baseSync.go
```diff
@@ -240,9 +259,55 @@ func (boot *baseBootstrap) loadNotarizedBlocks(blockFinality uint64,
...
+	}
+
+	for i := currentNonce + 1; i <= highestNonceInStorer; i++ {
+		boot.cleanupNotarizedStorage(i, hdrNonceHashDataUnit)
```
sasurobert: no need for this function. it could be done here.

process/sync/baseSync.go
```diff
@@ -240,9 +259,55 @@ func (boot *baseBootstrap) loadNotarizedBlocks(blockFinality uint64,
...
+	}
+
+	return nil
+}
+
+func (boot *baseBootstrap) cleanupNotarizedStorage(nonce uint64, hdrNonceHashDataUnit dataRetriever.UnitType) {
+	errNotCritical := boot.removeNotarizedBlockHeader(nonce, hdrNonceHashDataUnit)
+	if errNotCritical != nil {
+		log.Info(fmt.Sprintf("remove notarized block header with nonce %d: %s\n", nonce, errNotCritical.Error()))
```
sasurobert: log.debug ?

cmd/node/config/config.toml
```diff
@@ -56,7 +56,7 @@
         FilePath = "PeerData"
         Type = "LvlDBSerial"
         BatchDelaySeconds = 30
-        MaxBatchSize = 10
```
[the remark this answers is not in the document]
SebastianMarian: Forget to put them back. I made some tests with different configurations. Still need to determine the sweet spot for these. Until then I restored the old value back.

process/block/preprocess/transactions.go
```diff
@@ -270,6 +270,28 @@ func (txs *transactions) SaveTxBlockToStorage(body block.Body) error {
...
 }
 
+// getTransaction gets the transaction with a given sender and destination shard id and a hash, from pool or from storage
+func (txs *transactions) getTransaction(
+	senderShardID uint32,
+	destShardID uint32,
+	txHash []byte,
+) *transaction.Transaction {
+	tx := txs.getTransactionFromPool(senderShardID, destShardID, txHash)
+	if tx != nil {
```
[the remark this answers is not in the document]
SebastianMarian: Yep, I created this method only for some tests. Anyway I removed it as it is not used.

process/block/preprocess/transactions.go
```diff
@@ -270,6 +270,28 @@ func (txs *transactions) SaveTxBlockToStorage(body block.Body) error {
 	return nil
 }
 
+// getTransaction gets the transaction with a given sender and destination shard id and a hash, from pool or from storage
+func (txs *transactions) getTransaction(
```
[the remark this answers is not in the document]
SebastianMarian: Yes, removed it

process/block/shardblock.go

process/sync/baseSync.go

process/sync/baseSync.go
SebastianMarian: It can stay like this, as bootstrap from storage is not an frequently event and when we do this we need these info if there would be some problems.

process/sync/baseSync.go

process/sync/baseSync.go
SebastianMarian: I just wanted to mentain the same format used in loadBlocks, just in case that in the future we would need to clean more stuff

process/sync/baseSync.go
SebastianMarian: It can stay like this, as bootstrap from storage is not an frequently event and when we do this we need these info if there would be some problems.

PR #229, opened 2019-06-24
Bugs/en 2588 fix direct sender

iulianpascalau:


p2p/libp2p/directSender.go
```diff
@@ -161,27 +160,33 @@ func (ds *directSender) Send(topic string, buff []byte, peer p2p.PeerID) error {
...
 
 	return nil
 }
 
+// SendAsync will asynchronously send a direct message to the connected peer
+func (ds *directSender) SendAsync(topic string, buff []byte, peer p2p.PeerID) {
+	go func() {
+		err := ds.Send(topic, buff, peer)
+		if err != nil {
+			log.Debug(err.Error())
```
SebastianMarian: I would change this to log.Info to see if we still have some errors here

iulianpascalau: changed to log.Info

p2p/libp2p/netMessenger.go
```diff
@@ -452,7 +452,9 @@ func (netMes *networkMessenger) UnregisterMessageProcessor(topic string) error {
 
 // SendToConnectedPeer sends a direct message to a connected peer
 func (netMes *networkMessenger) SendToConnectedPeer(topic string, buff []byte, peerID p2p.PeerID) error {
-	return netMes.ds.Send(topic, buff, peerID)
+	netMes.ds.SendAsync(topic, buff, peerID)
+
+	return nil
```
raduchis: Do we still need the error?

iulianpascalau: Not quite. It just satisfy the interface.

PR #232, opened 2019-06-25
Bug/fix goroutine leak

ccorcoveanu:
Fixed a goroutine leak caused by not closing the response body every time we made requests to the elastic API.
A big TODO for next sprint when we can focus on improving the code testability is to cover this struct with better unit tests even if it means to rethink how it was written.


core/indexer/elasticsearch.go
```diff
@@ -162,6 +164,8 @@ func (ei *elasticIndexer) createIndex(index string, body io.Reader) error {
 		res, err = ei.db.Indices.Create(index)
 	}
 
+	defer closeESResponseBody(res)
```
SebastianMarian: I would move this after the next check anyway to keep the same format as it used in the other modifications

PR #237, opened 2019-06-27
added factory for preprocessor creation.

sasurobert:
factory for preprocessor creation.


process/factory/shard/preProcessorsContainerFactory.go
```diff
@@ -0,0 +1,159 @@
...
+		txProcessor:       txProcessor,
+		accounts:          accounts,
+		scProcessor:       scProcessor,
+		scResultProcessor: scResultProcessor,
+		interProcessor:    interProcessor,
+		requestHandler:    requestHandler,
+	}, nil
+}
+
+func (ppcm *preProcessorsContainerFactory) Create() (process.PreProcessorsContainer, error) {
```
ccorcoveanu: Comment missing on exported method

process/factory/shard/preProcessorsContainerFactory.go
```diff
@@ -0,0 +1,159 @@
...
+	container := containers.NewPreProcessorsContainer()
+
+	key, preproc, err := ppcm.createTxPreProcessor()
```
ccorcoveanu: Why is key returned from this function? It is not actually used there, so it's not useful. You can just simply return two values (preproc and err) and do `err = container.Add(block.TxBlock, preproc)`

process/factory/shard/preProcessorsContainerFactors_test.go
```diff
@@ -0,0 +1,400 @@
+package shard
```
ccorcoveanu: Typo in the file's name (Factors)

process/factory/shard/preProcessorsContainerFactory.go
sasurobert: done

process/factory/shard/preProcessorsContainerFactory.go
sasurobert: done

process/factory/shard/preProcessorsContainerFactors_test.go
sasurobert: done

PR #241, opened 2019-07-01
Fixed bugs with fork choice and notarized meta blocks

SebastianMarian:
Fork choice detected (After more than 12 sec of latency in which one proposer tried to write on disk and then to broadcast the valid block with nonce n, the network already proposed another block with the same nonce n, and also the next proposer broadcast the block with nonce n + 1. Before the rest of the network to sync this block with nonce n + 1, they also received the block with nonce n from the slow proposer. Because they didn't sync yet the block with nonce n + 1 and the final block is still n - 1 (as k finality is 1) they will start a fork choice on the block with nonce n (as they have now two block with nonce n but with different round inside). As the rule is that when two VALID blocks have the same nonce, the one with the lower round will be selected, they will try to request the first block with nonce n. The problem is that the only one who got this block is the slower proposer, but because he also received the new block with nonce n + 1 which has been broadcast by the proposer of that round,  this caused a fail of validity when it verified it against his own block with nonce n, and for this reason he considered that he is also on the fork and rolled back to the final block with nonce n - 1, and removed its own proposed block with nonce n. From now nobody will have this block anymore and all the network is stuck, as nobody can resolve the request of this first block with nonce n. The solution for this case is that nobody from network should accept blocks, even they are valid, with the round inside them, lower than the current round - k finality. In this case any slow proposer has time (its round + next one) to broadcast the block to the network as this block to be taken into consideration.

Problem with the meta blocks which are added into node's shard pool but not notarized until the first cross shard transaction will happen. (this will cause that if there are many meta headers not notarized between two cross shard transactions with destination in node's shard, there is no enough time (in 4 seconds) to compute the validity checks for all of them and for this reason no more cross shard txs could be validated form that point (all will remain in state pending).The solution is that all the hashes from the validated metablocks, until that point, to be added in the creation of each shard block even if there are no cross shard transactions with destination in self shard. In this way the pool is emptied each round and the validation of the next cross shard transaction has to be fast as there are no many meta headers remained to be validated.

### DIFF
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -291,6 +291,11 @@ func (mp *metaProcessor) RestoreBlockIntoPools(headerHandler data.HeaderHandler,
 
 		headerPool.Put([]byte(hdrHash), &hdr)
 
+		err = mp.store.GetStorer(dataRetriever.BlockHeaderUnit).Remove([]byte(hdrHash))
+		if err != nil {
+			return err
+		}
+
 		shardMBHeaderCounterMutex.Lock()
 		shardMBHeadersTotalProcessed -= len(hdr.MiniBlockHeaders)
 		shardMBHeaderCounterMutex.Unlock()
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -541,7 +541,9 @@ func (mp *metaProcessor) createLastNotarizedHdrs(header *block.MetaBlock) error 
 	}
 
 	for i := uint32(0); i < mp.shardCoordinator.NumberOfShards(); i++ {
-		mp.finalNotarizedHdrs[i] = tmpNotedHdrs[i]
+		if tmpNotedHdrs[i].GetNonce() != mp.lastNotarizedHdrs[i].GetNonce() {
+			mp.finalNotarizedHdrs[i] = tmpNotedHdrs[i]
+		}
 	}
 
 	return nil
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -277,23 +277,31 @@ func (mp *metaProcessor) RestoreBlockIntoPools(headerHandler data.HeaderHandler,
 		hdrHashes = append(hdrHashes, shardData.HeaderHash)
 	}
 
-	hdrsBuff, err := mp.store.GetAll(dataRetriever.BlockHeaderUnit, hdrHashes)
-	if err != nil {
-		return err
-	}
-
-	for hdrHash, hdrBuff := range hdrsBuff {
-		hdr := block.Header{}
-		err = mp.marshalizer.Unmarshal(&hdr, hdrBuff)
+	for _, hdrHash := range hdrHashes {
+		buff, err := mp.store.Get(dataRetriever.BlockHeaderUnit, hdrHash)
 		if err != nil {
-			return err
+			log.Error(err.Error())
+			continue
+		}
+
+		hdr := block.Header{}
+		err = mp.marshalizer.Unmarshal(&hdr, buff)
+		if err != nil {
+			log.Error(err.Error())
+			continue
 		}
 
 		headerPool.Put([]byte(hdrHash), &hdr)
 
 		err = mp.store.GetStorer(dataRetriever.BlockHeaderUnit).Remove([]byte(hdrHash))
 		if err != nil {
-			return err
+			log.Error(err.Error())
+		}
+
+		nonceToByteSlice := mp.uint64Converter.ToByteSlice(hdr.Nonce)
+		err = mp.store.GetStorer(dataRetriever.ShardHdrNonceHashDataUnit).Remove(nonceToByteSlice)
+		if err != nil {
+			log.Error(err.Error())
 		}
 
 		shardMBHeaderCounterMutex.Lock()
@@ -427,9 +435,28 @@ func (mp *metaProcessor) CommitBlock(
 		log.LogIfError(errNotCritical)
 	}
 
+	err = mp.createLastNotarizedHdrs(header)
+	if err != nil {
+		return err
+	}
+
 	_, err = mp.accounts.Commit()
 	if err != nil {
 		return err
+	}
+
+	log.Info(fmt.Sprintf("metaBlock with nonce %d and hash %s has been committed successfully\n",
+		header.Nonce,
+		core.ToB64(headerHash)))
+
+	errNotCritical = mp.removeBlockInfoFromPool(header)
+	if errNotCritical != nil {
+		log.Info(errNotCritical.Error())
+	}
+
+	errNotCritical = mp.forkDetector.AddHeader(header, headerHash, process.BHProcessed)
+	if errNotCritical != nil {
+		log.Info(errNotCritical.Error())
 	}
 
 	err = chainHandler.SetCurrentBlockBody(body)
@@ -443,21 +470,6 @@ func (mp *metaProcessor) CommitBlock(
 	}
 
 	chainHandler.SetCurrentBlockHeaderHash(headerHash)
-
-	err = mp.createLastNotarizedHdrs(header)
-	if err != nil {
-		return err
-	}
-
-	errNotCritical = mp.removeBlockInfoFromPool(header)
-	if errNotCritical != nil {
-		log.Info(errNotCritical.Error())
-	}
-
-	errNotCritical = mp.forkDetector.AddHeader(header, headerHash, process.BHProcessed)
-	if errNotCritical != nil {
-		log.Info(errNotCritical.Error())
-	}
 
 	if mp.core != nil && mp.core.TPSBenchmark() != nil {
 		mp.core.TPSBenchmark().Update(header)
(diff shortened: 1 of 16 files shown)


process/block/shardblock.go
```diff
@@ -489,16 +489,19 @@ func (sp *shardProcessor) CommitBlock(
...
-	processedMetaHdrs, errNotCritical := sp.getProcessedMetaBlocksFromPool(body)
+	processedMetaHdrs, errNotCritical := sp.getProcessedMetaBlocksFromPool(body, header)
 	if errNotCritical != nil {
 		log.Debug(errNotCritical.Error())
 	}
 
-	err = sp.saveLastNotarizedHeader(sharding.MetachainShardId, processedMetaHdrs)
-	if err != nil {
-		return err
+	errNotCritical = sp.saveLastNotarizedHeader(sharding.MetachainShardId, processedMetaHdrs)
```
sasurobert: It is better to keep this critical. saveLastNotarizedHeader could return an error could not processed if lastNoterized.Nonce != maxNonce from processedMetaHdrs.

process/block/shardblock.go
```diff
@@ -489,16 +489,19 @@ func (sp *shardProcessor) CommitBlock(
...
+	if errNotCritical != nil {
+		log.Debug(errNotCritical.Error())
 	}
 
+	log.Info(fmt.Sprintf("last notarized block nonce from metachain is %d\n",
```
sasurobert: this has to be protected by mutex.

process/block/shardblock.go
```diff
@@ -528,11 +531,44 @@ func (sp *shardProcessor) CommitBlock(
 	return nil
 }
 
+// getMaxNonceInMetaBlocksUsed returns the maximum nonce of meta blocks included in the given shard header
+func (sp *shardProcessor) getMaxNonceInMetaBlocksUsed(header *block.Header) uint64 {
```
sasurobert: getMaxNonceFromUsedMetaBlocks

process/constants.go
```diff
@@ -26,4 +26,5 @@ const (
 
 const ShardBlockFinality = 1
 const MetaBlockFinality = 1
+const BlockFinality = 1
```
sasurobert: rename to something with ForkBlockFinality ?

process/sync/basicForkDetector.go
```diff
@@ -113,6 +113,9 @@ func (bfd *basicForkDetector) checkBlockValidity(header data.HeaderHandler, stat
 	if int32(header.GetRound()) > bfd.rounder.Index() {
 		return ErrHigherRoundInBlock
 	}
+	if int32(header.GetRound()) < bfd.rounder.Index()-process.BlockFinality {
```
sasurobert: make a variable with a good name.

process/block/shardblock.go
SebastianMarian: Changed

process/block/shardblock.go

process/block/shardblock.go

process/constants.go
SebastianMarian: Done

process/block/metablock.go
```diff
@@ -277,20 +277,33 @@ func (mp *metaProcessor) RestoreBlockIntoPools(headerHandler data.HeaderHandler,
...
 
 		headerPool.Put([]byte(hdrHash), &hdr)
 
+		err = mp.store.GetStorer(dataRetriever.BlockHeaderUnit).Remove([]byte(hdrHash))
+		if err != nil {
+			log.Error(err.Error())
+		}
+
+		nonceToByteSlice := mp.uint64Converter.ToByteSlice(hdr.Nonce)
+		err = mp.store.GetStorer(dataRetriever.ShardHdrNonceHashDataUnit).Remove(nonceToByteSlice)
```
sasurobert: put it into the shardhdrnonces pool as well.

process/block/shardblock.go
```diff
@@ -351,24 +358,53 @@ func (sp *shardProcessor) RestoreBlockIntoPools(headerHandler data.HeaderHandler
...
+		}
+
+		metaBlockPool.Put(metaBlockHash, &metaBlock)
+
+		err = sp.store.GetStorer(dataRetriever.MetaBlockUnit).Remove(metaBlockHash)
+		if err != nil {
+			log.Error(err.Error())
+		}
+
+		nonceToByteSlice := sp.uint64Converter.ToByteSlice(metaBlock.Nonce)
```
sasurobert: put into the metahdr nonce pool as well.

process/block/shardblock.go
```diff
@@ -473,32 +505,51 @@ func (sp *shardProcessor) CommitBlock(
...
-	err = sp.saveLastNotarizedHeader(sharding.MetachainShardId, processedMetaHdrs)
+	sp.mutNotarizedHdrs.RLock()
+	lastNotarizedNonce := sp.lastNotarizedHdrs[sharding.MetachainShardId].GetNonce()
+	sp.mutNotarizedHdrs.RUnlock()
+
+	log.Info(fmt.Sprintf("last notarized block nonce from metachain is %d and max nonce used in current header is %d\n",
+		lastNotarizedNonce,
+		maxNonce))
+
+	if maxNonce > 0 && lastNotarizedNonce != maxNonce {
```
sasurobert: add this verification inside the saveLastNotarizedHeader function and make it return a critical error.

process/block/shardblock.go
```diff
@@ -551,6 +627,8 @@ func (sp *shardProcessor) getProcessedMetaBlocksFromPool(body block.Body) ([]dat
 		miniBlockHashes[i] = mbHash
 	}
 
+	log.Info(fmt.Sprintf("cross nini blocks in body: %d\n", len(miniBlockHashes)))
```
sasurobert: nini typo

process/block/shardblock.go
```diff
@@ -1170,6 +1256,13 @@ func (sp *shardProcessor) createAndProcessCrossMiniBlocksDstMe(
 		}
 
 		maxTxRemaining := uint32(maxTxInBlock) - nrTxAdded
```
sasurobert: move this line before calling the createAndprocessMiniBlockFromHeader function

process/block/shardblock_test.go
```diff
@@ -2777,6 +2777,12 @@ func TestShardProcessor_GetProcessedMetaBlockFromPoolShouldWork(t *testing.T) {
 		createDummyMetaBlock(destShardId, destShards[2], miniblockHashes[4], miniblockHashes[5]),
 	)
 
+	//hashes := make([][]byte, 0)
```
sasurobert: delete commented code

process/block/shardblock_test.go
```diff
@@ -3188,8 +3201,10 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrNoDstMB(t *testing.T) {
 	shardHdr := &block.Header{Round: 15}
 	shardBlock := block.Body{}
 
+	//shardHeader := &block.Header{}
```
sasurobert: delete commented code

process/block/shardblock_test.go
```diff
@@ -3206,7 +3221,11 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrNoDstMB(t *testing.T) {
 	// wrong header type in pool and defer called
 	dataPool.MetaBlocks().Put(currHash, shardHdr)
 
-	processedMetaHdrs, err = sp.GetProcessedMetaBlocksFromPool(shardBlock)
+	//hashes := make([][]byte, 0)
```
sasurobert: delete commented code

process/block/shardblock_test.go
```diff
@@ -3223,7 +3242,12 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrNoDstMB(t *testing.T) {
 	dataPool.MetaBlocks().Put(currHash, currHdr)
 	dataPool.MetaBlocks().Put(prevHash, prevHdr)
 
-	processedMetaHdrs, err = sp.GetProcessedMetaBlocksFromPool(shardBlock)
+	//hashes = make([][]byte, 0)
```
sasurobert: delete commented code

process/block/shardblock_test.go
```diff
@@ -3372,7 +3396,12 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrNotAllMBFinished(t *tes
 	dataPool.MetaBlocks().Put(currHash, currHdr)
 	dataPool.MetaBlocks().Put(prevHash, prevHdr)
 
-	processedMetaHdrs, err := sp.GetProcessedMetaBlocksFromPool(shardBlock)
+	//hashes := make([][]byte, 0)
```
sasurobert: delete commented code

process/block/shardblock_test.go
```diff
@@ -3497,7 +3526,13 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrAllMBFinished(t *testin
 	dataPool.MetaBlocks().Put(prevHash, prevHdr)
 	dataPool.MetaBlocks().Put([]byte("shouldNotRemove"), &block.MetaBlock{Nonce: 47})
 
-	processedMetaHdrs, err := sp.GetProcessedMetaBlocksFromPool(shardBlock)
+	//hashes := make([][]byte, 0)
```
sasurobert: delete commented code

process/block/shardblock.go

process/block/shardblock.go
SebastianMarian: Fixed

process/block/metablock.go

process/block/shardblock.go

process/block/shardblock.go

PR #251, opened 2019-07-05
Bug/en 2838 delete from branch node correctly

BeniaminDrasovean:
The issue appeared when you had a branchNode that had only 2 children, a leafNode and an extensionNode. If you deleted the leafNode, then the branchNode should have been replaced with an extensionNode leading directly to the child of the extensionNode, but the bug was that the branchNode was replaced with an extensionNode leading to the extensionNode, which is an incorrect type of construction.

To resolve this, I changed the reduceNode method accordingly, and changed the way that it was called.

### DIFF
--- a/data/trie/branchNode.go
+++ b/data/trie/branchNode.go
@@ -436,15 +436,9 @@ func (bn *branchNode) delete(key []byte, db data.DBWriteCacher, marshalizer mars
 		if err != nil {
 			return false, nil, err
 		}
-		if childPos != 16 {
-			newNode := bn.reduceNode(pos)
-			return true, newNode, nil
-		}
-		child := bn.children[pos]
-		if child, ok := child.(*leafNode); ok {
-			return true, newLeafNode([]byte{byte(pos)}, child.Value), nil
-		}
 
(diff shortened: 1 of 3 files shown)


data/trie/branchNode.go
```diff
@@ -436,26 +436,17 @@ func (bn *branchNode) delete(key []byte, db data.DBWriteCacher, marshalizer mars
...
-			newNode := bn.reduceNode(pos)
-			return true, newNode, nil
-		}
-		child := bn.children[pos]
-		if child, ok := child.(*leafNode); ok {
-			return true, newLeafNode([]byte{byte(pos)}, child.Value), nil
-		}
 
+		newNode := bn.children[pos].reduceNode(pos)
+		return true, newNode, nil
```
sasurobert: add empty line before return

data/trie/branchNode.go
```diff
@@ -436,26 +436,17 @@ func (bn *branchNode) delete(key []byte, db data.DBWriteCacher, marshalizer mars
...
 	}
 
 	bn.dirty = dirty
 	return true, bn, nil
```
sasurobert: add empty line before return

PR #255, opened 2019-07-07
enabled iele vm to be used by the node

iulianpascalau:
changed transaction.Data from []byte to string
minor code changes: updated/added comments, used _ for error handling in tests

### DIFF
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -58,7 +58,6 @@ import (
 	"github.com/ElrondNetwork/elrond-go/process/factory"
 	"github.com/ElrondNetwork/elrond-go/process/factory/metachain"
 	"github.com/ElrondNetwork/elrond-go/process/factory/shard"
-	"github.com/ElrondNetwork/elrond-go/process/mock"
 	"github.com/ElrondNetwork/elrond-go/process/smartContract"
 	"github.com/ElrondNetwork/elrond-go/process/smartContract/hooks"
 	processSync "github.com/ElrondNetwork/elrond-go/process/sync"
@@ -68,6 +67,8 @@ import (
 	"github.com/ElrondNetwork/elrond-go/storage"
 	"github.com/ElrondNetwork/elrond-go/storage/memorydb"
 	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
+	ielecommon "github.com/ElrondNetwork/elrond-vm/iele/common"
+	"github.com/ElrondNetwork/elrond-vm/iele/elrond/node/endpoint"
 	"github.com/btcsuite/btcd/btcec"
 	libp2pCrypto "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/urfave/cli"
@@ -1275,9 +1276,12 @@ func newShardBlockProcessorAndTracker(
 	if err != nil {
 		return nil, nil, err
 	}
-	//TODO: change the mock
+
+	cryptoHook := &hooks.VMCryptoHook{}
+	ieleVM := endpoint.NewElrondIeleVM(vmAccountsDB, cryptoHook, ielecommon.Default)
+
 	scProcessor, err := smartContract.NewSmartContractProcessor(
-		&mock.VMExecutionHandlerStub{},
+		ieleVM,
 		argsParser,
 		core.Hasher,
 		core.Marshalizer,
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -1277,7 +1277,8 @@ func newShardBlockProcessorAndTracker(
 		return nil, nil, err
 	}
 
-	cryptoHook := &hooks.VMCryptoHook{}
+	//TODO replace this with a vm factory
+	cryptoHook := hooks.NewVMCryptoHook()
 	ieleVM := endpoint.NewElrondIeleVM(vmAccountsDB, cryptoHook, ielecommon.Default)
 
 	scProcessor, err := smartContract.NewSmartContractProcessor(
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -775,7 +775,8 @@ func startStatisticsMonitor(file *os.File, config config.ResourceStatsConfig, lo
 }
 
 func createApiResolver(vmAccountsDB vmcommon.BlockchainHook) (facade.ApiResolver, error) {
-	cryptoHook := &hooks.VMCryptoHook{}
+	//TODO replace this with a vm factory
+	cryptoHook := hooks.NewVMCryptoHook()
 	ieleVM := endpoint.NewElrondIeleVM(vmAccountsDB, cryptoHook, ielecommon.Default)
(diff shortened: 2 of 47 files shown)


cmd/node/factory/structs.go
```diff
@@ -1275,9 +1276,12 @@ func newShardBlockProcessorAndTracker(
 	if err != nil {
 		return nil, nil, err
 	}
-	//TODO: change the mock
+
+	cryptoHook := &hooks.VMCryptoHook{}
```
sasurobert: should make a constructor for those hooks.

cmd/node/main.go
```diff
@@ -766,11 +774,11 @@ func startStatisticsMonitor(file *os.File, config config.ResourceStatsConfig, lo
 	return nil
 }
 
-func createApiResolver() (facade.ApiResolver, error) {
-	//TODO replace this with a vm factory
-	vm := &mock.VMExecutionHandlerStub{}
+func createApiResolver(vmAccountsDB vmcommon.BlockchainHook) (facade.ApiResolver, error) {
+	cryptoHook := &hooks.VMCryptoHook{}
```
sasurobert: leave the TODO as the factory is a must :P

integrationTests/frontend/wallet/txInterception_test.go
```diff
@@ -129,20 +147,14 @@ func testInterceptedTxFromFrontendGeneratedParams(
 		sig, _ := hex.DecodeString(frontendSignature)
 		assert.Equal(t, txRecovered.Signature, sig)
```
sasurobert: delete empty line.

process/smartContract/process.go
```diff
@@ -242,6 +242,11 @@ func (sc *scProcessor) DeploySmartContract(
 		return err
 	}
 
+	vmInput.ContractCode, err = hex.DecodeString(string(vmInput.ContractCode))
```
sasurobert: put these lines inside sc.createVMDeployInput

process/transaction/process_test.go
```diff
@@ -4,10 +4,11 @@ import (
 	"bytes"
 	"crypto/rand"
 	"errors"
-	"github.com/ElrondNetwork/elrond-go/process/smartContract"
 	"math/big"
 	"testing"
 
+	"github.com/ElrondNetwork/elrond-go/process/smartContract"
+
```
sasurobert: goimports

cmd/node/factory/structs.go
iulianpascalau: Right, changed everywhere in the project.

cmd/node/main.go
iulianpascalau: Re-added TODO

integrationTests/frontend/wallet/txInterception_test.go
iulianpascalau: deleted

process/smartContract/process.go
iulianpascalau: Can not do that as the transformation should be done only when deploy is done. When the SC is called, the func names should be plain text as there are more easily read and debugged. Can be changed in the future if we really want.

process/transaction/process_test.go
iulianpascalau: done

cmd/node/main.go
```diff
@@ -766,11 +774,12 @@ func startStatisticsMonitor(file *os.File, config config.ResourceStatsConfig, lo
 	return nil
 }
 
-func createApiResolver() (facade.ApiResolver, error) {
+func createApiResolver(vmAccountsDB vmcommon.BlockchainHook) (facade.ApiResolver, error) {
 	//TODO replace this with a vm factory
-	vm := &mock.VMExecutionHandlerStub{}
+	cryptoHook := hooks.NewVMCryptoHook()
+	ieleVM := endpoint.NewElrondIeleVM(vmAccountsDB, cryptoHook, ielecommon.Default)
```
ccorcoveanu: Why do we create a new VM here and another one for `NewSmartContractProcessor` in `structs.go`?

data/transaction/transaction.go
```diff
@@ -88,7 +88,7 @@ func TransactionGoToCapn(seg *capn.Segment, src *Transaction) capnp.TransactionC
 	dest.SetSndAddr(src.SndAddr)
 	dest.SetGasPrice(src.GasPrice)
 	dest.SetGasLimit(src.GasLimit)
-	dest.SetData(src.Data)
+	dest.SetData([]byte(src.Data))
```
ccorcoveanu: I would also change the schema instead of converting the string to `Data` format

integrationTests/frontend/wallet/txInterception_test.go
```diff
@@ -28,6 +28,8 @@ func TestInterceptedTxFromFrontendGeneratedParamsWithoutData(t *testing.T) {
 }
 
 func TestInterceptedTxFromFrontendGeneratedParams(t *testing.T) {
+	t.Skip("repair this test")
```
ccorcoveanu: We can generate this params and un-skip this tests before merge

process/interface.go
```diff
@@ -184,7 +184,7 @@ type PreProcessorsContainerFactory interface {
 	Create() (PreProcessorsContainer, error)
 }
 
-// PreProcessorsContainer defines an PreProcessors holder data type with basic functionality
+// IntermediateProcessorContainer defines an PreProcessors holder data type with basic functionality
```
ccorcoveanu: The name is updated but the comment is still related to `PreProcessors`

process/interface.go
```diff
@@ -195,7 +195,7 @@ type IntermediateProcessorContainer interface {
 	Keys() []block.Type
 }
 
-// PreProcessorsContainerFactory defines the functionality to create an PreProcessors container
+// IntermediateProcessorsContainerFactory defines the functionality to create an PreProcessors container
```
ccorcoveanu: The name is updated but the comment is still related to `PreProcessors`

process/smartContract/process.go
iulianpascalau: Done

cmd/node/main.go
iulianpascalau: Because we need 2 instances: one for the tx processor and the other one for resolving routes requests (get methods). Since there is a state when executing the vm code, we could not use the same instance as it would corrupt the state when 2 calls would have been made in the same time.

integrationTests/frontend/wallet/txInterception_test.go
iulianpascalau: done

data/transaction/transaction.go
iulianpascalau: Done

process/interface.go
iulianpascalau: done

process/interface.go
iulianpascalau: done

PR #269, opened 2019-07-10
Prometheus integration

bogdan-rosianu:


api/api.go
```diff
@@ -2,7 +2,11 @@ package api
 
 import (
 	"fmt"
+	"net/http"
 	"reflect"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
AdoAdoAdo: remove this empty line and run goimports again

api/api.go
```diff
@@ -34,12 +40,34 @@ func Start(elrondFacade MainApiHandler) error {
 	if err != nil {
 		return err
 	}
-	registerRoutes(ws, elrondFacade)
+	registerRoutes(ws, elrondFacade, elrondFacade.PrometheusMonitoring())
```
AdoAdoAdo: PrometheusMonitoring is accessible with elrondFacade which is already provided, so you could probably remove it as a parameter.

api/api.go
```diff
@@ -34,12 +40,34 @@ func Start(elrondFacade MainApiHandler) error {
...
+func joinMonitoringSystem(port string, prometheusJoinURL string) error {
+	req, err := http.NewRequest("POST", prometheusJoinURL, strings.NewReader(port))
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	err = resp.Body.Close()
+	if err != nil {
```
AdoAdoAdo: you can directly return the err, without any extra conditional

cmd/node/main.go
```diff
@@ -509,9 +518,21 @@ func startNode(ctx *cli.Context, log *logger.Logger) error {
 
 	ef := facade.NewElrondNodeFacade(currentNode, externalResolver)
 
+	prometheusJoinURLAvailable := true
+	prometheusJoinUrl, err := getPrometheusJoinURL(ctx.GlobalString(serversConfigurationFile.Name))
+	if err != nil {
```
AdoAdoAdo: can you do a combined conditional err!=nil || prometheusJoinUrl== ""  since these have the same outcome

api/api.go
bogdan-rosianu: done

api/api.go
bogdan-rosianu: done

api/api.go
bogdan-rosianu: done

cmd/node/main.go
bogdan-rosianu: fixed

api/api.go
```diff
@@ -52,6 +76,10 @@ func registerRoutes(ws *gin.Engine, elrondFacade middleware.ElrondHandler) {
 	txRoutes.Use(middleware.WithElrondFacade(elrondFacade))
 	transaction.Routes(txRoutes)
 
+	if elrondFacade.(MainApiHandler).PrometheusMonitoring() {
```
ccorcoveanu: We should extract `elrondFacade.(MainApiHandler)` in a variable, something like `apiHandler, ok := elrondFacade.(MainApiHandler)`. Then check if ok is true, otherwise this code will panic if elrondFacade can't be casted to MainApiHandler

cmd/node/main.go
```diff
@@ -159,6 +160,13 @@ VERSION:
 		Usage: "The port on which the rest API will start on",
 		Value: "8080",
 	}
+
+	// flag for monitoring using Prometheus
+	prometheus = cli.BoolFlag{
```
ccorcoveanu: If this is a bool flag, we should maybe call it something like `with-prometheus` or `use-prometheus`

config/config.go
```diff
@@ -156,6 +156,12 @@ type ExplorerConfig struct {
 // ServersConfig will hold all the confidential settings for servers
 type ServersConfig struct {
 	ElasticSearch ElasticSearchConfig
+	Prometheus    PrometheusConfig
+}
+
+// PrometheusConfig will hold configuration for prometheus, such as the join URL
+type PrometheusConfig struct {
+	PrometheusJoinURL string
```
ccorcoveanu: The variable here and in the config should be `PrometheusURL`. We should always use a base url and go from there

cmd/node/main.go
```diff
@@ -546,6 +564,20 @@ func startNode(ctx *cli.Context, log *logger.Logger) error {
 	return nil
 }
 
+func getPrometheusJoinURL(serversConfigurationFileName string) (string, error) {
+	serversConfig, err := core.LoadServersPConfig(serversConfigurationFileName)
+	if err != nil {
+		return "", err
+	}
+	joinURL := serversConfig.Prometheus.PrometheusJoinURL
+	statusURL := strings.Replace(joinURL, "/join", "/status", 1)
```
ccorcoveanu: See comment below, regarding `PrometheusJoinURL`. This is one of the situations why it's better to have a base url, otherwise we will get into situations like this where we need to do a lot of tricks to build a correct url

api/api.go
bogdan-rosianu: done

cmd/node/main.go

config/config.go
bogdan-rosianu: done

cmd/node/main.go

cmd/node/main.go
```diff
@@ -569,13 +569,14 @@ func getPrometheusJoinURL(serversConfigurationFileName string) (string, error) {
...
 		return "", err
 	}
-	joinURL := serversConfig.Prometheus.PrometheusJoinURL
-	statusURL := strings.Replace(joinURL, "/join", "/status", 1)
-	_, err = http.Get(statusURL)
-	if err != nil {
+	baseURL := serversConfig.Prometheus.PrometheusBaseURL
+	statusURL := baseURL + serversConfig.Prometheus.StatusRoute
+	resp, err := http.Get(statusURL)
+	if err != nil || resp.StatusCode == http.StatusNotFound {
```
ccorcoveanu: Maybe it's better to have two different conditions here. If you will get 404 `err` can be nil

bogdan-rosianu: fixed

api/api.go
```diff
@@ -2,11 +2,14 @@ package api
...
 import (
 	"fmt"
+	"net/http"
 	"reflect"
+	"strings"
 
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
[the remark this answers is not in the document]
bogdan-rosianu: done

PR #275, opened 2019-07-11
Feat/en 2893 session info add other fields

raduchis:
added txSignPK address to session.info
added -destination-shard-as-observer cli flag


cmd/node/main.go
```diff
@@ -449,6 +459,18 @@ func startNode(ctx *cli.Context, log *logger.Logger, version string) error {
 		return err
 	}
 
+	output := fmt.Sprintf("%s:%s\n%s:%s\n%s:%s\n%s:%v\n%s:%s\n%s:%s",
```
sasurobert: end it with new line

cmd/node/main.go
```diff
@@ -367,6 +375,19 @@ func startNode(ctx *cli.Context, log *logger.Logger, version string) error {
 	}
 	log.Info("Starting with public key: " + factory.GetPkEncoded(pubKey))
 
+	destinationShardAsObserverString := ctx.GlobalString(destinationShardAsObserver.Name)
+	if destinationShardAsObserverString != metachainShardName {
```
SebastianMarian: You can compare this in insensitive mode using this line:

if !strings.EqualFold(destinationShardAsObserverString, metachainShardName)
or
if strings.ToLower(destinationShardAsObserverString) != strings.ToLower(metachainShardName)

cmd/node/main.go
raduchis: fixed

cmd/node/main.go
raduchis: fixed

PR #291, opened 2019-07-19
Feat/en 3155 heartbeat refactor

bogdan-rosianu:


node/heartbeat/hearbeatMessageInfo.go
```diff
@@ -5,8 +5,12 @@ import "time"
 // heartbeatMessageInfo retain the message info received from another node (identified by a public key)
 type heartbeatMessageInfo struct {
 	maxDurationPeerUnresponsive time.Duration
-	peerHeartbeats              map[string]*PeerHeartbeat
 	getTimeHandler              func() time.Time
+	timeStamp                   time.Time
+	maxInactiveTime             Duration
+	isActive                    bool
+	shardID                     uint32
+	alreadyAccessed             bool
```
iulianpascalau: Why do we have alreadyAccessed?
Should not update once per execution the data inside the struct.

node/heartbeat/monitor.go
```diff
@@ -120,7 +123,7 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P) error {
 			m.heartbeatMessages[string(hb.Pubkey)] = pe
```
iulianpascalau: lines 110-114 are now useless. Also peerMessenger is useless and func PeerAddress from interface.go

node/heartbeat/monitor.go
```diff
@@ -34,6 +35,7 @@ func NewMonitor(
 	marshalizer marshal.Marshalizer,
 	maxDurationPeerUnresponsive time.Duration,
 	pubKeyList []string,
+	shardID uint32,
```
iulianpascalau: this is not used properly. Should drop the shardId parameter and field. This should be sent by the message originator (the message will have the shardId set by the one who outputs the hearbeat message)
As written now, all messages received by one node will have the same shard ID as the node on which we are connected to.

node/node.go
```diff
@@ -681,6 +681,7 @@ func (n *Node) StartHeartbeat(config config.HeartbeatConfig) error {
 		n.privKey,
 		n.marshalizer,
 		HeartbeatTopic,
+		n.shardCoordinator.SelfId(),
```
iulianpascalau: not used properly. Should have sent n.shardCoordinator.

node/heartbeat/sender.go
```diff
@@ -15,6 +15,7 @@ type Sender struct {
 	privKey       crypto.PrivateKey
 	marshalizer   marshal.Marshalizer
 	topic         string
+	shardID       uint32
```
iulianpascalau: this should have been a sharding.Coordinator instance instead of the shardId

node/heartbeat/sender.go
```diff
@@ -45,6 +47,7 @@ func NewSender(
 		privKey:       privKey,
 		marshalizer:   marshalizer,
 		topic:         topic,
+		shardID:       shardID,
 	}
 
 	return sender, nil
```
iulianpascalau: line 61: should insert:
```GO
ShardID: s.shardCoordinator.SelfId(),
```
We can not start the heartbeat sender with a predefined shard ID as there is no guarantee that in time the node's shard ID is going to stay the same. That is why we need the shard coordinator instance that "knows" all the time the current time

node/heartbeat/hearbeatMessageInfo.go
sasurobert: +1

node/heartbeat/hearbeatMessageInfo.go
```diff
@@ -30,49 +35,28 @@ func (hbmi *heartbeatMessageInfo) clockTime() time.Time {
...
-			phb.MaxInactiveTime.Duration = crtDuration
-		}
+	crtDuration := hbmi.getTimeHandler().Sub(hbmi.timeStamp)
+	hbmi.isActive = crtDuration < hbmi.maxDurationPeerUnresponsive
+	if hbmi.maxInactiveTime.Duration < crtDuration {
+		hbmi.maxInactiveTime.Duration = crtDuration
 	}
 }
 
 // HeartbeatReceived processes a new message arrived from a p2p address
```
sasurobert: change the comment as it does not match what the function does.

node/heartbeat/monitor.go
sasurobert: +1

node/heartbeat/hearbeatMessageInfo.go
bogdan-rosianu: removed

node/heartbeat/monitor.go
bogdan-rosianu: removed

node/heartbeat/monitor.go
bogdan-rosianu: done

node/node.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/hearbeatMessageInfo.go
bogdan-rosianu: done

node/heartbeat/hearbeatMessageInfo.go
```diff
@@ -37,26 +40,22 @@ func (hbmi *heartbeatMessageInfo) clockTime() time.Time {
...
-		hbmi.shardID = shardID
-		hbmi.isActive = true
-		hbmi.timeStamp = crtTime
-		hbmi.maxInactiveTime = Duration{Duration: 0}
-	}
-	crtDuration := crtTime.Sub(hbmi.timeStamp)
-	if hbmi.maxInactiveTime.Duration < crtDuration {
+func (hbmi *heartbeatMessageInfo) updateMaxInactiveTimeDuration() {
+	crtDuration := hbmi.getTimeHandler().Sub(hbmi.timeStamp)
+	if hbmi.maxInactiveTime.Duration < crtDuration && hbmi.timeStamp != (time.Time{}) {
```
sasurobert: you should set maxInactive to maxDuration, han you do not need the second part of the condition

node/heartbeat/sender.go
```diff
@@ -6,16 +6,17 @@ import (
...
-	privKey       crypto.PrivateKey
-	marshalizer   marshal.Marshalizer
-	topic         string
-	shardID       uint32
+	peerMessenger    PeerMessenger
+	singleSigner     crypto.SingleSigner
+	privKey          crypto.PrivateKey
+	marshalizer      marshal.Marshalizer
+	topic            string
+	shardCoordinator *sharding.Coordinator
```
sasurobert: no need for pointer here. Coordinator is an interface

node/heartbeat/sender.go
```diff
@@ -25,7 +26,7 @@ func NewSender(
 	privKey crypto.PrivateKey,
 	marshalizer marshal.Marshalizer,
 	topic string,
-	shardID uint32,
+	shardCoordinator *sharding.Coordinator,
```
sasurobert: no need for pointer here.

node/heartbeat/sender.go
```diff
@@ -58,6 +59,7 @@ func (s *Sender) SendHeartbeat() error {
 
 	hb := &Heartbeat{
 		Payload: []byte(fmt.Sprintf("%v", time.Now())),
+		ShardID: (*s.shardCoordinator).SelfId(),
```
sasurobert: no need of pointer if you do the suggested changes

node/heartbeat/sender_test.go
```diff
@@ -2,15 +2,21 @@ package heartbeat_test
...
 	"github.com/ElrondNetwork/elrond-go/crypto"
 	"github.com/ElrondNetwork/elrond-go/node/heartbeat"
 	"github.com/ElrondNetwork/elrond-go/node/mock"
-	"github.com/pkg/errors"
+	"github.com/ElrondNetwork/elrond-go/sharding"
 	"github.com/stretchr/testify/assert"
 )
 
+func getShardCoordinatorMock() *sharding.Coordinator {
+	shardCoord := (sharding.Coordinator)(mock.ShardCoordinatorMock{})
```
sasurobert: no pointer pls

node/node.go
```diff
@@ -681,7 +681,7 @@ func (n *Node) StartHeartbeat(config config.HeartbeatConfig) error {
 		n.privKey,
 		n.marshalizer,
 		HeartbeatTopic,
-		n.shardCoordinator.SelfId(),
+		&n.shardCoordinator,
```
sasurobert: no pointer

node/heartbeat/hearbeatMessageInfo.go
bogdan-rosianu: not possible because maxInactive will be compared to the difference between current time and timestamp and the 'if' clause. The default value for maxDurationPeerUnresponsive is 1 min. Also, providing a maximum duration value would not be ok. 
However, I added a variable called emptyTimestamp so it would be more readable. Also, the second condition of the _if_ won't be calculated very often, because most of the time the first condition won't pass

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/sender_test.go
bogdan-rosianu: done

node/node.go
bogdan-rosianu: done

node/heartbeat/monitor_test.go
```diff
@@ -271,7 +239,111 @@ func TestMonitor_ProcessReceivedMessageShouldWork(t *testing.T) {
...
+				(obj.(*heartbeat.Heartbeat)).Pubkey = rcvdHb.Pubkey
+				(obj.(*heartbeat.Heartbeat)).ShardID = rcvdHb.ShardID
+				return nil
+			},
+		},
+		time.Second*1000,
+		[]string{"pk1"},
+	)
+
+	// First send from pk1 from shad 0
```
BeniaminDrasovean: Typo: shad

api/node/routes_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
```
BeniaminDrasovean: Could you arrange the imports? And also, remove errs nickname from errors package.

node/heartbeat/monitor_test.go
bogdan-rosianu: done

api/node/routes_test.go
bogdan-rosianu: done

PR #298, opened 2019-07-22
Bugs/en 3179 data trie loading bugfix

iulianpascalau:


data/state/accountsDB.go
```diff
@@ -159,24 +200,26 @@ func (adb *AccountsDB) SaveDataTrie(accountHandler AccountHandler) error {
...
+	err = accountHandler.SetRootHashWithJournal(root)
 	if err != nil {
 		return err
 	}
 
-	err = accountHandler.SetRootHashWithJournal(root)
+	trackableDataTrie.ClearDataCaches()
+
+	//append a journal entry as the data needs to be updated in its trie
+	entry, err := NewBaseJournalEntryData(accountHandler, newDataTrie)
```
sasurobert: can you move this into a new function, that same as other saveStuffwithJournal ?

data/state/accountsDB.go
```diff
@@ -123,18 +134,48 @@ func (adb *AccountsDB) LoadDataTrie(accountHandler AccountHandler) error {
...
+
+	for i := len(adb.entries) - 1; i >= 0; i-- {
+		dataTrieEntry, ok := adb.entries[i].(*BaseJournalEntryData)
+		if !ok {
+			continue
+		}
+
+		dataRootHash, err := dataTrieEntry.trie.Root()
+		if err != nil {
+			return nil, errors.New("corrupted trie found in journal while computing root hash: " + err.Error())
```
sasurobert: do not make a new error. Add it into a constant one.

data/state/accountsDB.go
```diff
@@ -123,18 +134,48 @@ func (adb *AccountsDB) LoadDataTrie(accountHandler AccountHandler) error {
...
+		}
+
+		if bytes.Equal(dataRootHash, rootHash) {
+			return dataTrieEntry.trie.DeepClone()
+		}
+	}
+
+	return nil, NewErrMissingTrie(rootHash)
```
sasurobert: if missing you could log.Debug the missing roothash, and return a constant error.

data/state/accountsDB.go
```diff
@@ -112,7 +112,18 @@ func (adb *AccountsDB) LoadDataTrie(accountHandler AccountHandler) error {
...
 	}
 
-	dataTrie, err := adb.mainTrie.Recreate(accountHandler.GetRootHash())
+	dataTrie, err := adb.searchDataTrieInJournal(accountHandler.GetRootHash())
+	if err == nil {
+		accountHandler.SetDataTrie(dataTrie)
+		return nil
+	}
+
+	_, isErrMissingTrie := err.(*ErrMissingTrie)
```
sasurobert: can you change these tests ? i do not like handling different type of errors in different ways.

data/state/accountsDB.go
raduchis: +1

integrationTests/vm/iele/vmAgar_test.go
```diff
@@ -191,6 +192,189 @@ func TestAgarioContractTopUpAnfWithdrawShouldWork(t *testing.T) {
 	assert.Equal(t, newValue, getIntValueFromSC(accnts, scAddressBytes, "balanceOf", userAddress))
 }
 
+func TestAgarioContractJoinGameReward(t *testing.T) {
```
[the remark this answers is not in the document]
raduchis: +1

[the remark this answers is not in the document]
iulianpascalau: Agree, will be refactored along with all integrationTests subpackages. First, the branch bugs/EN-3188-agario-withdraw-cross-shard has to be merged in development branch (will soon open a PR for that)

data/state/accountsDB.go
iulianpascalau: done. saveDataTrieWithJournal

data/state/accountsDB.go
iulianpascalau: Disagree, will be next to impossible to debug here the error generated. All you will see might be a "the node is nil" and since we also have a trie behind AccountState and n tries in accounts it will be hard to understand who output the error.

data/state/accountsDB.go
iulianpascalau: We should shift from reusing the same error vars (that do not explain too much what is wrong) to more comprehensible error types (non constant errors). Again, this helps the debugging process and it is go-idiomatic.

data/state/accountsDB.go
iulianpascalau: I follow the rule: output only (pointer, nil) or (nil, error). So, even though they are not pretty, they conform to how our program has been written. (In other cases I have output an "empty" kind of value for pointer but here should not be applied)

PR #300, opened 2019-07-23
En 3107 test 1mil accounts

BeniaminDrasovean:
Should output memory occupied by a 1 million loaded accounts accountsDB (before commit and after commit).

Should output memory occupied by changing 30.000 (nonce + balance) accounts from 1 million loaded accounts (before commit and after commit). Might also do a benchmark on how much time it takes to do the required changes.

Should measure how much time it takes to load 30.000 accounts having a completely collapsed trie.


integrationTests/state/stateTrie_test.go
```diff
@@ -1025,6 +1032,154 @@ func TestAccountsDB_ExecALotOfBalanceTxOKorNOK(t *testing.T) {
 	adbPrintAccount(acntDest.(*state.Account), "Destination")
 }
 
+func TestCreateOneMillionAccountsWithMockDB(t *testing.T) {
```
sasurobert: Make it as benchmark. So we can run when running go test with benchmark

BeniaminDrasovean: Done.

integrationTests/state/stateTrie_test.go
```diff
@@ -1025,6 +1032,152 @@ func TestAccountsDB_ExecALotOfBalanceTxOKorNOK(t *testing.T) {
...
+	)
+
+	_, _ = adb.Commit()
+
+	runtime.ReadMemStats(&rtm)
+	fmt.Printf("Mem after committing %v accounts with mockDB: go mem - %s, sys mem - %s \n",
+		nrOfAccounts,
+		core.ConvertBytes(rtm.Alloc),
+		core.ConvertBytes(rtm.Sys),
+	)
```
iulianpascalau: Maybe add a new runtime.GC() and a new print here.
Also, after the last GC, try to iterate the trie (maybe with the use of trie.String()) and check if all nodes are iterated. If 1 million accounts is too much, try a smaller test with 10 accounts (in this test mem usage will not matter, just to be sure the nodes exist)

integrationTests/state/stateTrie_test.go
```diff
@@ -1025,6 +1032,152 @@ func TestAccountsDB_ExecALotOfBalanceTxOKorNOK(t *testing.T) {
...
+}
+
+func BenchmarkCreateOneMillionAccounts(b *testing.B) {
+	nrOfAccounts := 1000000
+	nrTxs := 15000
+	txVal := 100
+	balance := nrTxs * txVal
+	persist := mock2.NewCountingDB()
+
+	adb, addr := createAccounts(b, nrOfAccounts, balance, persist)
```
iulianpascalau: This test is not relevant as it uses memory db. What does it test?

BeniaminDrasovean: This test benchmarks the time needed for executing 15000 txs on a partially collapsed trie, and on a totally collapsed one, and shows the total number of nodes in a trie and the number of nodes that changed after the 15000txs are executed. Will change the memory measurements so it will show the difference between a collapsed trie and the same trie after the exec of 15000txs.

integrationTests/state/stateTrie_test.go
```diff
@@ -1025,6 +1032,152 @@ func TestAccountsDB_ExecALotOfBalanceTxOKorNOK(t *testing.T) {
 	adbPrintAccount(acntDest.(*state.Account), "Destination")
 }
 
+func BenchmarkCreateOneMillionAccountsWithMockDB(b *testing.B) {
+	nrOfAccounts := 1000000
+	nrTxs := 15000
+	txVal := 100
+	balance := nrTxs * txVal
```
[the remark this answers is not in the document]
BeniaminDrasovean: Changed.

integrationTests/state/stateTrie_test.go
BeniaminDrasovean: Done.

PR #301, opened 2019-07-23
Bug/en 3192 fix bug when a wrong/unused fork requested header is received

SebastianMarian:
This implementation should fix a situation when all peers would been replaced in their headerNonceHash pool a good/used header in their blockchain construction, with a wrong/unused header on which they didn't construct, but which came after a late broadcast from a valid proposer.

### DIFF
--- a/process/sync/metablock.go
+++ b/process/sync/metablock.go
@@ -11,7 +11,6 @@ import (
 	"github.com/ElrondNetwork/elrond-go/data/state"
 	"github.com/ElrondNetwork/elrond-go/data/typeConverters/uint64ByteSlice"
 	"github.com/ElrondNetwork/elrond-go/dataRetriever"
-	"github.com/ElrondNetwork/elrond-go/dataRetriever/dataPool"
 	"github.com/ElrondNetwork/elrond-go/hashing"
 	"github.com/ElrondNetwork/elrond-go/marshal"
 	"github.com/ElrondNetwork/elrond-go/process"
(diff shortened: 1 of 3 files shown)


process/sync/metablock.go
```diff
@@ -376,6 +377,20 @@ func (boot *MetaBootstrap) doJobOnSyncBlockFail(hdr *block.MetaBlock, err error)
 			log.Info(errNotCritical.Error())
 		}
 	}
+
+	// The below section of code fixed a situation when all peers would been replaced in their headerNonceHash pool a
```
AdoAdoAdo: would *have* replaced

process/sync/shardblock.go
```diff
@@ -487,6 +488,20 @@ func (boot *ShardBootstrap) doJobOnSyncBlockFail(hdr *block.Header, err error) {
 			log.Info(errNotCritical.Error())
 		}
 	}
+
+	// The below section of code fixed a situation when all peers would been replaced in their headerNonceHash pool a
```
AdoAdoAdo: would have replaced

sasurobert: +1

process/sync/metablock.go
sasurobert: +1

process/sync/metablock.go
```diff
@@ -376,6 +377,20 @@ func (boot *MetaBootstrap) doJobOnSyncBlockFail(hdr *block.MetaBlock, err error)
 			log.Info(errNotCritical.Error())
 		}
 	}
+
```
AdoAdoAdo: I can see there is a lot of duplicated code between these two files, if there is common functionality, we should extract into a common interface and not have the code duplicated.

For now the hotfix can go in like this, but please open a refactor issue for this.

process/sync/metablock.go
SebastianMarian: Done

process/sync/shardblock.go
SebastianMarian: Done

process/sync/metablock.go
AdoAdoAdo: Just took a look on the function, and at line 370 "isForkDetected" - this variable name seems misleading, why is a fork detected when we have a timeout error? this seems not logical, so maybe use a better name for this.

SebastianMarian: Ok. Added task EN-3195 in jira for this

SebastianMarian: Agree. Renamed the variable

PR #306, opened 2019-07-24
Bug/en 3214 fix bug in "Has" method from nonceSyncMapCacher

SebastianMarian:
* Fixed "Has" method from nonceSyncMapCacher
* ErrInvalidBlockHash has been split in two different errors (one for current shard blocks and one for notarized blocks) to help debugging

### DIFF
--- a/dataRetriever/dataPool/nonceSyncMapCacher.go
+++ b/dataRetriever/dataPool/nonceSyncMapCacher.go
@@ -119,7 +119,11 @@ func (nspc *nonceSyncMapCacher) RemoveShardId(nonce uint64, shardId uint32) {
 		return
 	}
 
-	syncMap := val.(*ShardIdHashSyncMap)
+	syncMap, ok := val.(*ShardIdHashSyncMap)
+	if !ok {
+		return
+	}
+
 	syncMap.Delete(shardId)
 }
 
@@ -143,9 +147,19 @@ func (nspc *nonceSyncMapCacher) callAddedDataHandlers(nonce uint64, shardId uint
 	nspc.mutAddedDataHandlers.RUnlock()
 }
 
-// Has returns true if a map is found for provided nonce
-func (nspc *nonceSyncMapCacher) Has(nonce uint64) bool {
-	_, ok := nspc.cacher.Peek(nspc.nonceConverter.ToByteSlice(nonce))
+// Has returns true if a map is found for provided nonce ans shardId
+func (nspc *nonceSyncMapCacher) Has(nonce uint64, shardId uint32) bool {
+	val, ok := nspc.cacher.Peek(nspc.nonceConverter.ToByteSlice(nonce))
+	if !ok {
+		return false
+	}
 
-	return ok
+	syncMap, ok := val.(*ShardIdHashSyncMap)
+	if !ok {
+		return false
+	}
+
+	_, exists := syncMap.Load(shardId)
+
+	return exists
 }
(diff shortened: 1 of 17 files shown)


dataRetriever/dataPool/nonceSyncMapCacher.go
```diff
@@ -143,9 +147,19 @@ func (nspc *nonceSyncMapCacher) callAddedDataHandlers(nonce uint64, shardId uint
 	nspc.mutAddedDataHandlers.RUnlock()
 }
 
-// Has returns true if a map is found for provided nonce
-func (nspc *nonceSyncMapCacher) Has(nonce uint64) bool {
-	_, ok := nspc.cacher.Peek(nspc.nonceConverter.ToByteSlice(nonce))
+// Has returns true if a map is found for provided nonce ans shardId
+func (nspc *nonceSyncMapCacher) Has(nonce uint64, shardId uint32) bool {
+	val, ok := nspc.cacher.Peek(nspc.nonceConverter.ToByteSlice(nonce))
```
raduchis: Just an observation: Has does everything like the Get method. If there are calls to Has and exactly after Get, use only get to not load everything twice.

process/block/baseProcess.go
```diff
@@ -211,7 +211,7 @@ func (bp *baseProcessor) isHdrConstructionValid(currHdr, prevHdr data.HeaderHand
 	}
 
 	if !bytes.Equal(currHdr.GetPrevHash(), prevHeaderHash) {
-		return process.ErrInvalidBlockHash
+		return process.ErrInvalidNotarizedBlockHash
```
raduchis: ErrInvalidNotarizedBlockHash? Is it invalid, or they do not match?

SebastianMarian: Actually it could be ErrBlockHashDoesNotMatch

dataRetriever/dataPool/nonceSyncMapCacher.go
SebastianMarian: I verified this and it is used only in one single place where I also found this bug.

process/block/baseProcess.go
SebastianMarian: Renamed

dataRetriever/dataPool/nonceSyncMapCacher.go
raduchis: ok

process/block/baseProcess.go
raduchis: ok

integrationTests/consensus/consensus_test.go
```diff
@@ -49,7 +49,7 @@ func initNodesAndTest(numNodes, consensusSize, numInvalid uint32, roundTime uint
 		for i := uint32(0); i < numInvalid; i++ {
 			nodes[i].blkProcessor.ProcessBlockCalled = func(blockChain data.ChainHandler, header data.HeaderHandler, body data.BodyHandler, haveTime func() time.Duration) error {
 				fmt.Println("process block invalid ", header.GetRound(), header.GetNonce(), getPkEncoded(nodes[i].pk))
```
AdoAdoAdo: should we also change the logged message to match the returned error?

SebastianMarian: I think that this print is something generic here and anyway the returned error will be printed in the caller side to spot the exactly error occurred

PR #315, opened 2019-07-29
refactor displayBlock.go

BeniaminDrasovean:
create transactionCounter struct, and remove global parameters from displayBlock.go


process/block/displayBlock.go
```diff
@@ -51,32 +63,34 @@ func DisplayLogInfo(
 	numShards uint32,
 	selfId uint32,
 	dataPool dataRetriever.PoolsHolder,
+	txCounter *transactionCounter,
```
iulianpascalau: this func could have been made part of the transactionCounter structure. And all methods could have been made unexported.

process/block/displayBlock.go
```diff
@@ -106,10 +120,10 @@ func createDisplayableShardHeaderAndBlockBody(
 	return tableHeader, shardLines
 }
 
-func displayTxBlockBody(lines []*display.LineData, body block.Body) []*display.LineData {
-	txCounterMutex.Lock()
-	txsCurrentBlockProcessed = 0
-	txCounterMutex.Unlock()
+func displayTxBlockBody(lines []*display.LineData, body block.Body, txCounter *transactionCounter) []*display.LineData {
```
iulianpascalau: this func could have been made part of the transactionCounter structure.

process/block/shardblock.go
```diff
@@ -1398,3 +1408,9 @@ func (sp *shardProcessor) DecodeBlockHeader(dta []byte) data.HeaderHandler {
 
 	return &header
 }
+
+// TransactionCounter returns the object that keeps track of the total nr. of processed txs,
+// and txs processed in this block
+func (sp *shardProcessor) TransactionCounter() *transactionCounter {
```
iulianpascalau: Might remove this, we do not want to be able to use it (yet)

process/block/displayBlock.go
BeniaminDrasovean: Changed

process/block/displayBlock.go
BeniaminDrasovean: done

process/block/shardblock.go
BeniaminDrasovean: done.

raduchis: Not done yet. It is used only in the tests.

process/block/displayBlock.go
```diff
@@ -106,10 +120,10 @@ func createDisplayableShardHeaderAndBlockBody(
...
+	txCounter.mutex.Lock()
+	txCounter.currentBlockTxs = 0
```
raduchis: Use local variables for this. There is no need for the locks and unlocks in the for. One lock and one unlock after the for should be enough.

BeniaminDrasovean: Done.

process/block/shardblock.go
BeniaminDrasovean: Was done, but the changes were not pushed yet.

PR #316, opened 2019-07-29
Feat/add transaction index flag

ccorcoveanu:
Added an extra flag for disabling the transaction indexing on nodes


cmd/node/main.go
```diff
@@ -698,12 +706,14 @@ func processDestinationShardAsObserver(settingsConfig config.GeneralSettingsConf
...
 // authentication for the server is using the username and password
 func CreateElasticIndexer(
+	ctx *cli.Context,
 	serversConfigurationFileName string,
 	url string,
 	coordinator sharding.Coordinator,
 	marshalizer marshal.Marshalizer,
 	hasher hashing.Hasher,
 	log *logger.Logger,
+
```
sasurobert: delete empty line

ccorcoveanu: Done

cmd/node/main.go
```diff
@@ -698,6 +706,7 @@ func processDestinationShardAsObserver(settingsConfig config.GeneralSettingsConf
 // CreateElasticIndexer creates a new elasticIndexer where the server listens on the url,
 // authentication for the server is using the username and password
 func CreateElasticIndexer(
+	ctx *cli.Context,
```
raduchis: There is no need for the entire context here. Couldn't you just pass the bool indexTransactions?

SebastianMarian: +1

ccorcoveanu: @raduchis, @SebastianMarian - I can do that. The reasoning was that if the `NewElasticIndexer` will have more flag options, we will end up removing the bool and passing the context so we will reduce the number of parameters. Do you still think it's better to have only one bool until then?

PR #320, opened 2019-07-29
Bug/en 3249 fix bug when node resolves requests with bad unused headers

SebastianMarian:
• Fixed a bug when node resolves requests with bad/unused headers on which it didn't constructed. The new approach is to check first if it has the requested data in cache, afterwards in storage and finally in pool. As the cache for headers could be big enough (as it does not take to much memory), the performance would not be affected, but the correctness of the information delivered to the requester would be finally the same with its own. Even if some malicious nodes will overwrite the node data stored in pool with bad/unused data, this will not affect the response of the node to the other requests.

• Fixed problem in config.toml where ShardHdrNonceHashStorage / MetaHdrNonceHashStorage would not been committed for the current nonce n, because of the MaxBatchSize value, but the BlockHeaderStorage / MetaBlockStorage would be. This will cause that the bootstrap from storage would be stuck, as the node will request nonce n, but when the interceptor would received it, it would found its hash in the storage and would not announced the node about it. This will cause the node to request forever for this nonce n. Also the order of the "Put" command has been changed, firstly the info will be put in storage for ShardHdrNonceHashStorage / MetaHdrNonceHashStorage and afterwards for BlockHeaderStorage / MetaBlockStorage. This is done for the same reason explained above, and would avoid some edge cases when the node would be stopped in the middle of these put actions.

• Replaced FIFOSharded with LRU in config.toml on [TxDataPool], as there are some problems and should be analyzed. One of them is: Why the txs which are received and put it in some FIFO shards would be replaced very quick by others txs received meantime. This case is happened when the node would not received from one request all the missing txs and let say only 90% of them. In the next iteration when is expected that the node would request only for the remaining 10%, actually it would always request much more. Beside these 10% which has not been received in the first iteration, it would request for some txs which are expected to be already in its pool. This will happen because of some evicted actions which have been done meantime by the new txs received from the network, which are not in its interest for this round/block.

• Extracted some duplicated/common code from process block/sync and moved to process/common

• Replaced some wrong calls of GetShardHeader / GetMetaHeader with GetShardHeaderFromPool / GetMetaHeaderFromPool


cmd/node/config/config.toml
```diff
@@ -151,8 +151,7 @@
 
 [TxDataPool]
     Size = 100000
-    Type = "FIFOSharded"
-    Shards = 128
+    Type = "LRU"
```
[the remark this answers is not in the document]
SebastianMarian: I would suggest  then to reduce the number of shards to 16 and to increase the pool size for txs to 250k as the memory used in this new configuration would be 32.42 MB instead 12.96 MB:
[TxStorage.Cache]
250000 => 250000 * 136 bytes/tx = 32.42 MB
In this way each shard has a space of 15625 txs, so we should have, in the worst case scenario,  at least two rounds all the received txs not evicted.

dataRetriever/resolvers/headerResolver.go
```diff
@@ -101,33 +101,32 @@ func (hdrRes *HeaderResolver) ProcessReceivedMessage(message p2p.MessageP2P) err
 func (hdrRes *HeaderResolver) resolveHeaderFromNonce(key []byte) ([]byte, error) {
 	//key is now an encoded nonce (uint64)
 
-	//Step 1. decode the nonce from the key
-	nonce, err := hdrRes.nonceConverter.ToUint64(key)
+	//Step 1. search the nonce-key pair in cache-storage
```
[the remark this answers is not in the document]
SebastianMarian: Done

dataRetriever/resolvers/headerResolver.go
```diff
@@ -101,33 +101,32 @@ func (hdrRes *HeaderResolver) ProcessReceivedMessage(message p2p.MessageP2P) err
...
-				hash = existingHash
-				return false
-			}
-
-			return true
-		})
+		log.Debug(err.Error())
 	}
 
+	//Step 2. search the nonce-key pair in data pool
```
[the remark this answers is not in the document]
SebastianMarian: Done

dataRetriever/resolvers/headerResolver.go
```diff
@@ -101,33 +101,32 @@ func (hdrRes *HeaderResolver) ProcessReceivedMessage(message p2p.MessageP2P) err
...
-		hash, err = hdrRes.hdrNoncesStorage.Get(key)
+		nonce, err := hdrRes.nonceConverter.ToUint64(key)
 		if err != nil {
-			log.Debug(err.Error())
+			return nil, dataRetriever.ErrInvalidNonceByteSlice
+		}
+
+		value, ok := hdrRes.hdrNonces.Get(nonce)
+		if ok {
+			value.Range(func(shardId uint32, existingHash []byte) bool {
```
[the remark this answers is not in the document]
SebastianMarian: Range method does not return anything, and the inline func is passed to the Range method and its return is used inside the Range method implementation

process/common.go
```diff
@@ -183,3 +205,319 @@ func GetMarshalizedHeaderFromStorage(
 
 	return buffHdr, nil
 }
+
+// GetMetaHeaderWithNonce method returns a meta block header with a given nonce
```
iulianpascalau: Please check if the duplicated code can be reduced

process/common.go
```diff
@@ -183,3 +205,319 @@ func GetMarshalizedHeaderFromStorage(
...
+	hdr, err := GetShardHeaderFromStorage(hash, marshalizer, storageService)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return hdr, hash, nil
+}
+
+// GetTransaction gets the transaction with a given sender/receiver shardId and txHash
+func GetTransaction(
```
iulianpascalau: Maybe rename to GetTransactionHandler

process/common.go
```diff
@@ -183,3 +205,319 @@ func GetMarshalizedHeaderFromStorage(
 
 	return buffHdr, nil
 }
+
```
iulianpascalau: These new functions should have their own unit tests

dataRetriever/resolvers/headerResolver.go
```diff
@@ -101,33 +101,32 @@ func (hdrRes *HeaderResolver) ProcessReceivedMessage(message p2p.MessageP2P) err
...
-
-			return true
-		})
+		log.Debug(err.Error())
 	}
 
+	//Step 2. search the nonce-key pair in data pool
 	if hash == nil {
-		hash, err = hdrRes.hdrNoncesStorage.Get(key)
+		nonce, err := hdrRes.nonceConverter.ToUint64(key)
```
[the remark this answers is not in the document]
SebastianMarian: Done

dataRetriever/resolvers/headerResolver.go
```diff
@@ -101,33 +101,32 @@ func (hdrRes *HeaderResolver) ProcessReceivedMessage(message p2p.MessageP2P) err
...
+					hash = existingHash
+					return false
+				}
+
+				return true
+			})
 		}
 
-		if len(hash) == 0 {
+		if hash == nil {
```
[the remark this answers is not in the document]
SebastianMarian: Reverted

process/common.go
SebastianMarian: Renamed

process/common.go
SebastianMarian: Refactored

cmd/node/config/config.toml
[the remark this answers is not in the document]
sasurobert: even then you never know which transaction is evicted. I am thinking about a solution at sync, like waiting for longer time when asking for some transactions.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/360EntSecGroup-Skylar/excelize v1.4.1
 	github.com/ElrondNetwork/concurrent-map v0.1.2
+	github.com/ElrondNetwork/elrond-proxy-go v1.0.0
```
sasurobert: why did you add proxy-go ?

node/node.go
```diff
@@ -604,7 +604,7 @@ func (n *Node) SendTransaction(
 	return txHexHash, nil
 }
 
-//GetTransaction gets the transaction
+//GetTransactionHandler gets the transaction
```
sasurobert: comment does not match function

go.mod
SebastianMarian: Removed

node/node.go
SebastianMarian: Changed

PR #321, opened 2019-07-30
Bugs/en 3228 sc processor refund gas

iulianpascalau:
Fixed bug when refunding gas to the tx sender when executing an intra-shard sc call tx and VM changed the balance of the sender
Added integration test


integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/ElrondNetwork/elrond-go/data/transaction"
+	"github.com/ElrondNetwork/elrond-go/integrationTests"
+	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/stretchr/testify/assert"
+)
+
+var agarioFile = "agarioV2.hex"
+var stepDelay = time.Second
+
+func TestShouldProcessBlocksWithScTxsJoinAndRewardTheOwner(t *testing.T) {
```
sasurobert: function too long

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+func incrementAndPrintRound(round uint32) uint32 {
+	round++
+	fmt.Printf("#################################### ROUND %d BEGINS ####################################\n\n", round)
+
+	return round
+}
+
+func mintAllNodes(nodes []*integrationTests.TestProcessorNode, value *big.Int) {
+	for _, n := range nodes {
+		if n.ShardCoordinator.SelfId() == sharding.MetachainShardId {
```
sasurobert: delete this if, it does not make sense

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+			continue
+		}
+
+		for _, n2 := range nodes {
```
sasurobert: delete this for, you are minting every node N times - N = number of nodes - number of metanodes

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+	time.Sleep(stepDelay)
+	fmt.Println(integrationTests.MakeDisplayTable(nodes))
+}
+
+func nodeDoesJoinGame(
+	nodes []*integrationTests.TestProcessorNode,
+	idxNode int,
+	joinGameVal *big.Int,
+	scAddress []byte) {
```
sasurobert: add enter after []byte

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+	fmt.Println(integrationTests.MakeDisplayTable(nodes))
+}
+
+func nodeCallsRewardAndSend(
+	nodes []*integrationTests.TestProcessorNode,
+	idxNodeOwner int,
+	idxNodeUser int,
+	prize *big.Int,
+	scAddress []byte) {
```
sasurobert: add enter after []byte

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+		topUpValue,
+		hardCodedScResultingAddress,
+	)
+
+	nodeCallsRewardAndSend(nodes, idxProposer, idxProposer, withdrawValue, hardCodedScResultingAddress)
+	proposeBlock(nodes, idxProposer, round)
+	syncBlock(t, nodes, idxProposer, round)
+	round = incrementAndPrintRound(round)
+
+	proposeBlock(nodes, idxProposer, round)
```
sasurobert: why do you need to propose and sync again ?

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+func nodeCallsRewardAndSend(
+	nodes []*integrationTests.TestProcessorNode,
+	idxNodeOwner int,
+	idxNodeUser int,
+	prize *big.Int,
+	scAddress []byte) {
+
+	fmt.Println("Calling SC.rewardAndSendToWallet...")
+	txDeploy := createTxRewardAndSendToWallet(nodes[idxNodeOwner], nodes[idxNodeUser], prize, scAddress)
```
sasurobert: this is not a deploy tx

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func nodeDoesJoinGame(
+	nodes []*integrationTests.TestProcessorNode,
+	idxNode int,
+	joinGameVal *big.Int,
+	scAddress []byte) {
+
+	fmt.Println("Calling SC.joinGame...")
+	txDeploy := createTxJoinGame(nodes[idxNode], joinGameVal, scAddress)
```
sasurobert: this is not a deploy tx

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: reduced to 47 chars

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: actually it will break the code otherwise because metachain nodes will try to mint addresses

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: will keep it as all nodes in all shards will have to mint all nodes from the same shard :). Will rework this function

iulianpascalau: done, moved it to testInitializer

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: done

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: done

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: Good catch! Removed.

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: Indeed, fixed also multishard/block

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: fixed

integrationTests/singleShard/block/executingMiniblocksSc_test.go
```diff
@@ -0,0 +1,342 @@
...
+			assert.Fail(t, err.Error())
+			return
+		}
+	}
+
+	time.Sleep(stepDelay)
+	fmt.Println(integrationTests.MakeDisplayTable(nodes))
+}
+
+func nodeDoesJoinGame(
```
ccorcoveanu: I would name it `nodeJoinsGame`, when I first saw a call on it I thought it was something boolean, like "does it join the game?". It would be easier to read the test

integrationTests/singleShard/block/executingMiniblocksSc_test.go
ccorcoveanu: 👍, I was about to ask the same thing, maybe a comment would be needed here as an explanation

process/smartContract/process.go
```diff
@@ -403,6 +403,19 @@ func (sc *scProcessor) processVMOutput(
 		return nil, err
 	}
 
+	//we need to reload the sender account as in the case of refunding the exact account that was previously
+	//modified in saveSCOutputToCurrentState, the modifications done there should be visible here
+	//this should be applied only on sender accounts from current shard
+	if acntSnd != nil && !acntSnd.IsInterfaceNil() {
```
ccorcoveanu: Maybe refactor a bit since we have 3 levels here?

integrationTests/singleShard/block/executingMiniblocksSc_test.go
sasurobert: not the function name. but the number of the lines in the function

integrationTests/singleShard/block/executingMiniblocksSc_test.go
iulianpascalau: good one, changed. Also in multishard/block

process/smartContract/process.go
iulianpascalau: Done, changed to a function. Code should be more readable.

PR #322, opened 2019-07-30
Bug/en 3244 compare hashes in verify proof

BeniaminDrasovean:
Check if the hashes match when verifying a proof.


data/trie/patriciaMerkleTrie_test.go
```diff
@@ -327,6 +327,23 @@ func emptyTrie() data.Trie {
 	return tr
 }
 
+func TestPatriciaMerkleTrie_VerifyProofFromDifferentTrieShouldNotWork(t *testing.T) {
```
raduchis: This unit test does not test for extensionNode, branchNoce and leafNode. There is only one test and cannot test all three cases. 
Please add further tests.

data/trie/patriciaMerkleTrie.go
```diff
@@ -163,13 +163,23 @@ func (tr *patriciaMerkleTrie) VerifyProof(proofs [][]byte, key []byte) (bool, er
 	tr.mutOperation.RLock()
 	defer tr.mutOperation.RUnlock()
 
+	wantHash, err := tr.Root()
```
raduchis: There is no unit test for err here.

data/trie/patriciaMerkleTrie.go
```diff
@@ -163,13 +163,23 @@ func (tr *patriciaMerkleTrie) VerifyProof(proofs [][]byte, key []byte) (bool, er
...
+
 	key = keyBytesToHex(key)
 	for i := range proofs {
 		encNode := proofs[i]
 		if encNode == nil {
 			return false, nil
 		}
 
+		hash := tr.hasher.Compute(string(encNode))
+		if !bytes.Equal(wantHash, hash) {
```
raduchis: Are there multiple proofs for the same rootHash?

data/trie/patriciaMerkleTrie_test.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
raduchis: nope... the wantHash changes

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: tr.Root() is tested elsewhere, and if the wantHash is wrong, some tests will fail.

data/trie/patriciaMerkleTrie_test.go
raduchis: ok

PR #324, opened 2019-07-30
Feat/index unsigned tx

ccorcoveanu:
Added SC result transactions to elasticsearch


core/indexer/elasticsearch.go
```diff
@@ -347,28 +348,50 @@ func (ei *elasticIndexer) buildTransactionBulks(
 			}
 
 			currentTx, ok := currentTxHandler.(*transaction.Transaction)
-			if !ok {
-				ei.logger.Warn("elasticsearch found tx in pool but of wrong type")
+			if ok && currentTx != nil {
+				bulks[currentBulk] = append(bulks[currentBulk], &Transaction{
```
sasurobert: separate in two functions accordint to transaction type. You can create a function getTransactionType it gets input a data.transactionhandler and returns an uint. switch case according to those types.

core/indexer/elasticsearch_test.go
```diff
@@ -284,6 +304,35 @@ func TestElasticIndexer_buildTransactionBulks(t *testing.T) {
...
+			if tx.Hash == hex.EncodeToString([]byte(testSCKey)) {
+				foundSc = true
+				break
+			}
+		}
+		if foundSc {
+			break
+		}
+	}
+	assert.True(t, foundSc)
```
sasurobert: add an empty line before

core/statistics/tpsBenchmark_test.go
```diff
@@ -309,6 +309,8 @@ func TestTpsBenchmark_EmptyBlocksShouldNotUpdateMultipleTimes(t *testing.T) {
 }
 
 func TestTpsBenchmark_Concurrent(t *testing.T) {
+	t.Skip("this test is failing. Task EN-3206")
```
sasurobert: add a TODO here as well.

dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -90,7 +90,7 @@ func (rrh *ResolverRequestHandler) RequestTransaction(destShardID uint32, txHash
 }
 
 func (rrh *ResolverRequestHandler) requestByHashes(destShardID uint32, hashes [][]byte, topic string) {
-	log.Debug(fmt.Sprintf("Requesting %d transactions from shard %d from network...\n", len(hashes), destShardID))
+	log.Info(fmt.Sprintf("Requesting %d transactions from shard %d from network on topic %s...\n", len(hashes), destShardID, topic))
```
sasurobert: leave it to log.Debug

core/statistics/tpsBenchmark_test.go
raduchis: +1

core/indexer/elasticsearch.go
```diff
@@ -358,28 +359,50 @@ func (ei *elasticIndexer) buildTransactionBulks(
...
+					GasLimit:      currentTx.GasLimit,
+					Data:          currentTx.Data,
+					Signature:     hex.EncodeToString(currentTx.Signature),
+					Timestamp:     time.Duration(header.GetTimeStamp()),
+					Status:        mbTxStatus,
+				})
+				continue
+			}
+
+			currentSc, ok := currentTxHandler.(*smartContractResult.SmartContractResult)
```
raduchis: Do these txs include the transactions for the rewards?

core/indexer/elasticsearch.go
ccorcoveanu: I've done it a bit different than suggested, because having a simple `getTransactionType` would mean I would have to do the type assertions twice (once to find the type, second in the process of building the resulted transaction since I would need access to the fields that are not on the interface).
I've implemented a getTransactionByType function that calls another two functions depending on the condition branch.

core/indexer/elasticsearch_test.go
ccorcoveanu: Done

core/statistics/tpsBenchmark_test.go
ccorcoveanu: Task 3206 was merged. I will remove this line and run the tests to see if it's fixed

dataRetriever/requestHandlers/requestHandler.go
ccorcoveanu: Done

core/indexer/elasticsearch.go
```diff
@@ -358,28 +359,50 @@ func (ei *elasticIndexer) buildTransactionBulks(
...
+					Hash:          hex.EncodeToString(txHash),
+					MBHash:        hex.EncodeToString(mbHash),
+					BlockHash:     hex.EncodeToString(blockHash),
+					Nonce:         currentSc.Nonce,
+					Value:         currentSc.Value,
+					Receiver:      hex.EncodeToString(currentSc.RcvAddr),
+					Sender:        hex.EncodeToString(currentSc.SndAddr),
+					ReceiverShard: mb.ReceiverShardID,
+					SenderShard:   mb.SenderShardID,
+					GasPrice:      0,
```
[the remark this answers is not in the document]
ccorcoveanu: Now we don't the fields on the implemented data structure. But good question, I think it should, I will talk with Robert and see if it makes sense to leave a TODO here to update it when needed

core/indexer/elasticsearch.go
ccorcoveanu: No they are not implemented yet. Maybe a more uniform solution can be implemented when we have all tx types

core/indexer/elasticsearch_test.go
```diff
@@ -284,6 +304,35 @@ func TestElasticIndexer_buildTransactionBulks(t *testing.T) {
...
+
+	foundSc := false
+	for _, bulk := range bulks {
+		for _, tx := range bulk {
+			if tx.Hash == hex.EncodeToString([]byte(testSCKey)) {
+				foundSc = true
+				break
+			}
+		}
+		if foundSc {
```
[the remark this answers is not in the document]
ccorcoveanu: This was a trick because we don't have `break X` in go. It breaks out of the first loop if foundSc will be set to true in the second loop. And no, it cannot be changed, it is a local variable for this test, declared on line 321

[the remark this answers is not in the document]
raduchis: yes. My bad

PR #332, opened 2019-08-01
Fix data race conditions

camilbancioiu:
Fixes a couple of race conditions in the MemP2P mock.


p2p/memp2p/memp2p_network.go
```diff
@@ -159,6 +159,15 @@ func (network *Network) LogMessage(message p2p.MessageP2P) {
 	network.messageLogMutex.Unlock()
 }
 
+// GetMessageCount returns the number of messages logged internally by the
+// Network.
+func (network *Network) GetMessageCount() int {
+	network.messageLogMutex.Lock()
```
AdoAdoAdo: I think you can do a RLock and RUnlock here as the messages are only read.

camilbancioiu: Fixed.

PR #337, opened 2019-08-02
EN-3321 - added uptime and downtime to heatbeat

bogdan-rosianu:


node/heartbeat/hearbeatMessageInfo.go
```diff
@@ -27,9 +29,10 @@ func newHeartbeatMessageInfo(maxDurationPeerUnresponsive time.Duration) (*heartb
 		maxInactiveTime:             Duration{0},
 		isActive:                    false,
 	}
-
 	hbmi.getTimeHandler = hbmi.clockTime
 	hbmi.timeStamp = emptyTimestamp
```
AdoAdoAdo: can we update timestamp with current time - maxDurationPeerUnresponsive instead of emptyTimestamp

PR #342, opened 2019-08-07
EN-3410: Heartbeat improvements

bogdan-rosianu:


cmd/node/config/config.toml
```diff
@@ -10,6 +10,9 @@
    # StatusPollingIntervalSec represents the no of seconds between multiple polling for the status for AppStatusHandler
    StatusPollingIntervalSec = 10
 
+   # NodeDisplayName represents the friendly name a user can pick for his node in the status monitor
+   NodeDisplayName = "undefined"
```
iulianpascalau: Might leave it blank as to not have 100 nodes with undefined

cmd/node/main.go
```diff
@@ -174,6 +174,14 @@ VERSION:
 		Value: "",
 	}
 
+	// nodeDisplayName defines the friendly name used by a node in the public monitoring tools. If set, will override
+	// the NodeDisplayName from config.toml
+	nodeDisplayName = cli.StringFlag{
+		Name:  "display-name",
+		Usage: "This will represent the friendly name in the public monitoring tools. Will override the config.toml one",
+		Value: "undefined",
```
iulianpascalau: also here, might leave it empty

node/heartbeat/sender.go
```diff
@@ -58,8 +64,10 @@ func NewSender(
 func (s *Sender) SendHeartbeat() error {
 
 	hb := &Heartbeat{
-		Payload: []byte(fmt.Sprintf("%v", time.Now())),
-		ShardID: s.shardCoordinator.SelfId(),
+		Payload:         []byte(fmt.Sprintf("%v", time.Now())),
```
ccorcoveanu: This should be better done in the receive part, in `ProcessReceivedMessage`. Every node should have a representation of when a peer sends a hartbeat from their perspective, not from the sender perspective. Even though this means that you would get different results questioning different nodes for a given peer, it's more correct in a decentralised setting. Consider that I can send any timestamp I want here (past or future) and every node will accept this. The equivalent in a centralised setting is sending a hartbeat to a server (like we have in our cryptobubbles game), the player sends a ping to the server and the server notarises the timestamp relative to itself.

cmd/node/config/config.toml
bogdan-rosianu: done

cmd/node/main.go

node/heartbeat/sender.go
iulianpascalau: Partially agree. I would leave this field on, and add a new one (computed=time.Now()) in the heartbeat structure. The difference between the computed one and the received one (if positive) will represent the node's latency. If the difference is negative, an output of 0ms latency can be done.

bogdan-rosianu: Corco, as we discussed, there is no need to remove it because that pay load represents the message which is signed by the sender when sending a heartbeat in order to authenticate him when another peer receives a heartbeat.
However, we can use Iulian's idea

ccorcoveanu: 👍

PR #343, opened 2019-08-07
modified shard meta headers resolvers to query only metachain nodes

iulianpascalau:
created a new sub-component for topic resolvers that will output the list of peers to whom will send the requests
modified shard meta headers resolvers to query only metachain nodes
added 1 integration test to prove this is done correctly

Requirements: https://docs.google.com/spreadsheets/d/1PM2-rt7KH5fESrVkrA_8JG2jlvYo2xiCywGlkeb9bt8/edit#gid=979668013 , sheet: Processor, section B.6 (B.6 - B.6.4)


dataRetriever/errors.go
```diff
@@ -184,3 +184,6 @@ var ErrEmptyHeaderRequestTopic = errors.New("empty header request topic")
 
 // ErrInvalidMaxTxRequest signals that max tx request is too small
 var ErrInvalidMaxTxRequest = errors.New("max tx request number is invalid")
+
+// ErrNilPeersListCreator signals that a nil peers list creator implementation has been provided
+var ErrNilPeersListCreator = errors.New("nil peers list creator provided")
```
ccorcoveanu: Can change everywhere to `PeerList` (peer - singular) since List implies multiple items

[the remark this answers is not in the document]
sasurobert: +1

dataRetriever/resolvers/topicResolverSender/diffPeersListCreator.go
```diff
@@ -0,0 +1,84 @@
...
+type DiffPeersListCreator struct {
+	messenger             dataRetriever.MessageHandler
+	mainTopic             string
+	excludePeersFromTopic string
+}
+
+// NewDiffPeersListCreator is the constructor for DiffPeersListCreator
+func NewDiffPeersListCreator(
+	messenger dataRetriever.MessageHandler,
+	mainTopic string,
```
sasurobert: you could test to not have these 2 topics the same.

dataRetriever/errors.go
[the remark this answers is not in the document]
iulianpascalau: right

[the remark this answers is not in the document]
iulianpascalau: done

dataRetriever/interface.go
```diff
@@ -140,6 +140,11 @@ type Notifier interface {
 	RegisterHandler(func(key []byte))
 }
 
+// PeersListCreator is used to create peers list based on some criteria
```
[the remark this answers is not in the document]
iulianpascalau: removed on some criteria. It is an interface, I can not go into much details

dataRetriever/resolvers/topicResolverSender/diffPeersListCreator.go
iulianpascalau: I can do that as a slight optimization but the remaining code should be the same (maybe some one will use the struct with 2 different topics that hold the exactly the same same set of peers)

iulianpascalau: This will actually make the code more messy as there is a situation when I call createShardHeaderResolver in metachain/resolversContainerFactory.go and this call will have to be guarded by an if as to not have the situation when the func is called with the same topic, example: shardHeadersForMetachain_1

dataRetriever/interface.go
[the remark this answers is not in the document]
ccorcoveanu: 👍 I agree, I was only referring to that expression

integrationTests/multiShard/block/interceptedBlocks_test.go
```diff
@@ -174,7 +174,7 @@ func TestMetaHeadersAreRequstedOnlyFromMetachain(t *testing.T) {
 	}
 
 	fmt.Println("Delaying for nodes p2p bootstrap...")
-	time.Sleep(stepDelay)
+	time.Sleep(time.Second * 5)
```
ccorcoveanu: is this resilient enough so we won't have to change it back and forth? I don't have a better solution now. Maybe we can have a pattern of `sleepDuration * len(nodes)` if bootstraping more nodes takes more time

iulianpascalau: It is empirical. I have suggested @BeniaminDrasovean in his next PR (integrationTests refactoring) to introduce an exported variable called DelayP2PBroadcast

PR #344, opened 2019-08-07
feat/EN-3483-merge-header-interceptors

camilbancioiu:
The following structs can be merged into a single struct, since they do the same thing:

* `HeaderInterceptor`
* `HeaderInterceptorBase`
* `ShardHeaderInterceptor`

This pull request does the following:
1. Merge the code from `HeaderInterceptorBase` into `HeaderInterceptor`, then delete `HeaderInterceptorBase`
2. Delete `ShardHeaderInterceptor`
3. Replace `metablock.NewShardHeaderInterceptor` with `interceptors.NewHeaderInterceptor` in the only place it was called.


process/block/interceptors/headerInterceptor.go
```diff
@@ -34,38 +38,81 @@ func NewHeaderInterceptor(
...
+		headers:             headers,
+		headersNonces:       headersNonces,
+	}
+
+	return hdrInterceptor, nil
+}
+
+// ParseReceivedMessage will transform the received p2p.Message in an InterceptedHeader.
+// If the header hash is present in storage it will output an error
+func (hi *HeaderInterceptor) ParseReceivedMessage(message p2p.MessageP2P) (*block.InterceptedHeader, error) {
```
sasurobert: improve code coverage, copy test functions from shard and base test to headerInterceptor_test to test every bit.

camilbancioiu: Damn. Forgot that the tests needed merging as well.

process/block/interceptors/headerInterceptor.go
```diff
@@ -75,12 +122,20 @@ func (hi *HeaderInterceptor) ProcessReceivedMessage(message p2p.MessageP2P) erro
 	return nil
 }
 
+// CheckHeaderForCurrentShard checks if the header is for current shard
+func (hi *HeaderInterceptor) CheckHeaderForCurrentShard(interceptedHdr *block.InterceptedHeader) bool {
```
iulianpascalau: Maybe make this unexported?

process/block/interceptors/headerInterceptor.go
camilbancioiu: Did the merging. Now `headerInterceptor_test.go` contains the union of all the tests for these 3 structs, but without any duplicates. I'm not entirely sure that I unified the tests correctly - I don't understand the `headerInterceptor` 100%.

process/block/interceptors/headerInterceptor.go
camilbancioiu: Changed it.

process/block/interceptors/headerInterceptor.go
iulianpascalau: Tests seems fine, should add more tests for the contructor

process/block/interceptors/headerInterceptor.go
iulianpascalau: Ok, please update the comment also

process/block/interceptors/headerInterceptor.go
camilbancioiu: Added extra tests for the constructor.

process/block/interceptors/headerInterceptor.go
camilbancioiu: Done.

PR #346, opened 2019-08-08
Add pprof routes to node when enabled by flag

ccorcoveanu:
Added pprof routes over gin. It can be used with pprof tool:
https://golang.org/pkg/net/http/pprof/
https://jvns.ca/blog/2017/09/24/profiling-go-with-pprof/

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -122,6 +122,17 @@ VERSION:
 		Value: 0,
 	}
 	// profileMode defines a flag for profiling the binary
+	// If enabled, it will open the pprof routes over the default gin rest webserver.
+	// There are several routes that will be available for profiling (profiling can be analyzed with: go tool pprof):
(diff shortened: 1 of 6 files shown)


cmd/node/main.go
```diff
@@ -123,10 +122,9 @@ VERSION:
 		Value: 0,
 	}
 	// profileMode defines a flag for profiling the binary
-	profileMode = cli.StringFlag{
+	profileMode = cli.BoolFlag{
 		Name:  "profile-mode",
-		Usage: "Profiling mode. Available options: cpu, mem, mutex, block",
-		Value: "",
+		Usage: "Boolean profiling mode option. If set to true, the /debug/pprof routes will be available on the node for profiling the application.",
```
iulianpascalau: Maybe explain a little more how to use it as to not have to to go back and forth on different sites to re-learn how to use profiling tools

PR #350, opened 2019-08-08
Bugfix/added gasLimit check when creating cross shard miniBlocks

sasurobert:
Added gasLimit check when creating a cross shard miniblock. Example of the problem: shard 0 creates a miniblock with 1000 smart contract calls, all the calls are for a SC in shard 1. shard 1 when reading the from metachain the needed miniblocks to process, tries to do it, but fails as he has no time to process all those transactions.


process/block/preprocess/transactions.go
```diff
@@ -148,10 +149,6 @@ func (txs *transactions) RestoreTxBlockIntoPools(
 
 	txsRestored := 0
 
-	if miniBlockPool == nil {
```
raduchis: Good spotting this

process/block/preprocess/transactions.go
```diff
@@ -417,11 +434,21 @@ func (txs *transactions) CreateAndProcessMiniBlock(sndShardId, dstShardId uint32
...
 	for index := range orderedTxes {
 		if !haveTime() {
 			break
 		}
 
+		currTxGasLimit := minGasLimitForTx
+		if isTxRcvAddressOfSmartContract(orderedTxes[index].RcvAddr) {
+			currTxGasLimit = orderedTxes[index].GasLimit
+
+			if addedGasLimitPerCrossShardMiniblock+currTxGasLimit > process.MaxGasLimitPerMiniBlock {
```
raduchis: Do you want to test that the maxGasLimitPerMiniblock is exceeded just for SC txs? What happens if a if SC is added that uses MaxGasLimitPerMiniBlock - 1 and another 1000 normal txs are added: you would exceed the MaxGasLimitPerMiniblock?

process/block/preprocess/transactions_test.go
```diff
@@ -540,6 +543,138 @@ func TestTransactionPreprocessor_RemoveBlockTxsFromPoolOK(t *testing.T) {
...
+		txHash, _ := core.CalculateHash(marshalizer, hasher, newTx)
+		txPool.AddData(txHash, newTx, strCache)
+
+		addedTxs = append(addedTxs, newTx)
+	}
+
+	mb, err := txs.CreateAndProcessMiniBlock(sndShardId, dstShardId, process.MaxItemsInBlock, haveTimeTrue, 10)
+	assert.Nil(t, err)
+
+	assert.Equal(t, len(addedTxs), len(mb.TxHashes))
```
raduchis: Why does this assert work? You add 10 txs, each with maxGasLimitPerMiniblock/5 => you add in the miniblock 2*maxGasLimitPerMiniblock. Is that ok?

integrationTests/longTests/executingSCTransactions_test.go
```diff
@@ -392,7 +388,6 @@ func runMultipleRoundsOfTheGame(
 	topUpValue *big.Int,
 	hardCodedScResultingAddress []byte,
 	round uint64,
-	nrShards uint32,
```
raduchis: Why did this parameter dissapear?

integrationTests/testInitializer.go
```diff
@@ -622,3 +622,69 @@ func extractUint64ValueFromTxHandler(txHandler data.TransactionHandler) uint64 {
...
+	idxProposers []int,
+	round uint64,
+) uint64 {
+
+	// propose until pool is cleared
+	for i := numInTxs; i != 0; {
+		startTime := time.Now()
+		ProposeBlock(nodes, idxProposers, round)
+		elapsedTime := time.Since(startTime)
+		fmt.Printf("Block Created in %s\n", elapsedTime)
```
raduchis: There is a Sleep in the ProposeBlock. Is it really the time the block was created in or it does not really matter?

process/block/preprocess/transactions_test.go
```diff
@@ -2,7 +2,10 @@ package preprocess
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/core"
```
AdoAdoAdo: goimports

integrationTests/longTests/executingSCTransactions_test.go
sasurobert: it is not used. as the propose and sync function uses shardcoordinator.numberOfShards.

integrationTests/testInitializer.go
sasurobert: good catch. Moving the print exactly when the block is created, it is a good measure to see how much time was used.

process/block/preprocess/transactions_test.go
sasurobert: because these txs are not smart contract calls, normal transactions are processed fast, it is enough to add the constant.

integrationTests/testInitializer.go
```diff
@@ -622,3 +622,69 @@ func extractUint64ValueFromTxHandler(txHandler data.TransactionHandler) uint64 {
...
+
+		SyncBlock(t, nodes, idxProposers, round)
+		round = IncrementAndPrintRound(round)
+
+		for _, idProposer := range idxProposers {
+			proposedNode := nodes[idProposer]
```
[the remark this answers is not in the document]
sasurobert: done

integrationTests/testInitializer.go
```diff
@@ -622,3 +622,69 @@ func extractUint64ValueFromTxHandler(txHandler data.TransactionHandler) uint64 {
...
+func ProposeAndSyncBlocks(
+	t *testing.T,
+	numInTxs int,
+	nodes []*TestProcessorNode,
+	idxProposers []int,
+	round uint64,
+) uint64 {
+
+	// propose until pool is cleared
+	for i := numInTxs; i != 0; {
```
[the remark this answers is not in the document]
sasurobert: There is not repeat until, or do while. So the propose is needed to be done at least once. Than to continue proposing until the pools are empty.

integrationTests/testInitializer.go
```diff
@@ -622,3 +622,69 @@ func extractUint64ValueFromTxHandler(txHandler data.TransactionHandler) uint64 {
...
+		startTime := time.Now()
+		ProposeBlock(nodes, idxProposers, round)
+		elapsedTime := time.Since(startTime)
+		fmt.Printf("Block Created in %s\n", elapsedTime)
+
+		SyncBlock(t, nodes, idxProposers, round)
+		round = IncrementAndPrintRound(round)
+
+		for _, idProposer := range idxProposers {
```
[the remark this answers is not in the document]
sasurobert: This is a function to continue proposing blocks until there are no more transactions in the pool.

integrationTests/testInitializer.go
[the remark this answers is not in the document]
sasurobert: the loop below is needed, to propagate all the modifications through all the shards. After the last transactions were proposed, it needs 5 rounds to reach the destination shard, because of finality and asynchronous processing through metachain.

process/block/preprocess/transactions_test.go
sasurobert: done.

process/block/preprocess/transactions.go
```diff
@@ -388,6 +385,26 @@ func (txs *transactions) getAllTxsFromMiniBlock(
 	return transactions, txHashes, nil
 }
 
+//TODO move this constant to txFeeHandler
+const minGasLimitForTx = uint64(5)
+
+const numZerosForSCAddress = 10
+
+//TODO move this to smart contract address calculation component
+func isTxRcvAddressOfSmartContract(rcvAddress []byte) bool {
```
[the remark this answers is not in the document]
sasurobert: done

process/block/preprocess/transactions.go
```diff
@@ -388,6 +385,26 @@ func (txs *transactions) getAllTxsFromMiniBlock(
 	return transactions, txHashes, nil
 }
 
+//TODO move this constant to txFeeHandler
+const minGasLimitForTx = uint64(5)
+
+const numZerosForSCAddress = 10
```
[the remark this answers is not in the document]
sasurobert: done

process/block/preprocess/transactions.go
sasurobert: ok. condition moved outside. good catch.

integrationTests/longTests/executingSCTransactions_test.go
raduchis: ok

integrationTests/testInitializer.go
raduchis: ok

process/block/preprocess/transactions.go
raduchis: ok

process/block/preprocess/transactions_test.go
raduchis: but you still have more gas than the maxGasLImitPerMiniBlock... it shouldn't allow the gasLimit to be more than the max

raduchis: ok

integrationTests/testInitializer.go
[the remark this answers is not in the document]
AdoAdoAdo: then add a comment for the second loop why it is needed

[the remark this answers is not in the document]
sasurobert: added

integrationTests/testInitializer.go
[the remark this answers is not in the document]
sasurobert: added comment

PR #354, opened 2019-08-09
Feat/en 2674 add account pk in nodesSetup.json

SebastianMarian:


cmd/node/main.go
```diff
@@ -744,14 +744,14 @@ func createNodesCoordinator(
...
-	initNodesPubKeys := nodesConfig.InitialNodesPubKeys()
+	initNodesInfo := nodesConfig.InitialNodesInfo()
 	initValidators := make(map[uint32][]sharding.Validator)
 
-	for shardId, pubKeyList := range initNodesPubKeys {
+	for shardId, nodeInfoList := range initNodesInfo {
 		validators := make([]sharding.Validator, 0)
-		for _, pubKey := range pubKeyList {
+		for _, nodeInfo := range nodeInfoList {
 			// TODO: the stake needs to be associated to the staking account
```
sasurobert: delete the TODO

integrationTests/consensus/testInitializer.go
```diff
@@ -91,7 +91,7 @@ func genValidatorsFromPubKeys(pubKeysMap map[uint32][]string) map[uint32][]shard
 	for shardId, shardNodesPks := range pubKeysMap {
 		shardValidators := make([]sharding.Validator, 0)
 		for i := 0; i < len(shardNodesPks); i++ {
-			v, _ := sharding.NewValidator(big.NewInt(0), 1, []byte(shardNodesPks[i]))
+			v, _ := sharding.NewValidator(big.NewInt(0), 1, []byte(shardNodesPks[i]), []byte(shardNodesPks[i]))
```
sasurobert: BLS pubkeys are not the same length as schnorr.

sharding/nodesSetup.go
```diff
@@ -139,35 +168,69 @@ func (ns *NodesSetup) processShardAssignment() {
...
 	}
 }
 
-// InitialNodesPubKeys - gets initial public keys
+// InitialNodesPubKeys - gets initial nodes public keys
 func (ns *NodesSetup) InitialNodesPubKeys() map[uint32][]string {
-	return ns.allNodesPubKeys
+	allNodesPubKeys := make(map[uint32][]string, 0)
+	for shardId, nodesInfo := range ns.allNodesInfo {
+		pubKeys := make([]string, 0)
```
AdoAdoAdo: I think it would be more efficient if you allocate the right size and just add on the position instead of append, when you already know the size

sharding/nodesSetup.go
```diff
@@ -139,35 +168,69 @@ func (ns *NodesSetup) processShardAssignment() {
...
+		return nil, ErrShardIdOutOfRange
+	}
+	if len(ns.allNodesInfo[shardId]) == 0 {
+		return nil, ErrNoPubKeys
+	}
+
+	nodesInfo := ns.allNodesInfo[shardId]
+	pubKeys := make([]string, 0)
+	for i := 0; i < len(nodesInfo); i++ {
+		pubKeys = append(pubKeys, string(nodesInfo[i].pubKey))
```
AdoAdoAdo: same as above if you already know the size would be more efficient to allocate the right size and add on position instead of append.

sharding/nodesSetup_test.go
```diff
@@ -14,7 +14,8 @@ func createNodesSetupOneShardOneNode() *sharding.NodesSetup {
 	ns.MinNodesPerShard = 1
 	ns.InitialNodes = make([]*sharding.InitialNode, 1)
 	ns.InitialNodes[0] = &sharding.InitialNode{}
-	ns.InitialNodes[0].PubKey = "5126b6505a73e59a994caa8f556f8c335d4399229de42102bb4814ca261c7419"
+	ns.InitialNodes[0].PubKey = "52f3bf5c01771f601ec2137e267319ab6716ef6ff5dfddaea48b42d955f631167f2ce19296a202bb8fd174f4e94f8c85f619df85a7f9f8de0f3768e5e6d8c48187b767deccf9829be246aa331aa86d182eb8fa28ea8a3e45d357ed1647a9be020a5569d686253a6f89e9123c7f21f302e82f67d3e3cd69cf267b9910a663ef32"
```
AdoAdoAdo: can we add these pubKeys and addresses as constants in the file, maybe in an array testPubKeys and testAddresses and use the constants instead for initialization in tests as it doesn't read too well. same for the other tests in this file that make use of these.

sasurobert: +1

cmd/node/main.go
SebastianMarian: Done

integrationTests/consensus/testInitializer.go
SebastianMarian: Yes, this is a dummy address anyway. Changed to be at least different compared to pubkey parameter.

sharding/nodesSetup.go
SebastianMarian: Done

sharding/nodesSetup.go
SebastianMarian: Done

sharding/nodesSetup_test.go
SebastianMarian: Refactored

SebastianMarian: +2 :)

integrationTests/consensus/testInitializer.go
```diff
@@ -91,7 +92,8 @@ func genValidatorsFromPubKeys(pubKeysMap map[uint32][]string) map[uint32][]shard
 	for shardId, shardNodesPks := range pubKeysMap {
 		shardValidators := make([]sharding.Validator, 0)
 		for i := 0; i < len(shardNodesPks); i++ {
-			v, _ := sharding.NewValidator(big.NewInt(0), 1, []byte(shardNodesPks[i]), []byte(shardNodesPks[i]))
+			v, _ := sharding.NewValidator(big.NewInt(0), 1, []byte(shardNodesPks[i]),
+				[]byte("addr_"+strconv.Itoa(int(shardId))+"_"+strconv.Itoa(i)))
```
AdoAdoAdo: can you use sprintf instead? then you don't need all the string conversions and it's more readable.

sharding/nodesSetup_test.go
```diff
@@ -37,20 +57,12 @@ func createNodesSetupTwoShardTwoNodes() *sharding.NodesSetup {
 	ns.ConsensusGroupSize = 1
 	ns.MinNodesPerShard = 2
 	ns.InitialNodes = make([]*sharding.InitialNode, 4)
```
AdoAdoAdo: I would create a variable for initial nodes number, and use that in the memory allocation and loop and anywhere else it is needed, instead of using same magic value in multiple places, then it would be easy to change.

Applies to other tests in this file.

integrationTests/consensus/testInitializer.go
SebastianMarian: Done

sharding/nodesSetup_test.go
SebastianMarian: Done

PR #355, opened 2019-08-09
Test ProcessShardAssignment

drdelphi:
The shards may not be assigned correctly if one of the pubKeys is null.


AdoAdoAdo:
Thanks for reporting, this is valid issue.

It will probably be better to return an error in case of nil instead of proceeding with a badly configured nodes setup, same for metachain assignment.

PR #358, opened 2019-08-09
bugfix/make transaction coordinator call processors in strict order when processing miniblocks

sasurobert:
There was a TODO in the code that it should be investigated if processing with multiple different processors can be done in parallel or not. The order in which the transactions for a selected account has to be processed are strict - increasing nonce one by one - , so the parallel processing would make this impossible. 
Furthermore, there is an undefined behavior if the same account is updated from 2 different places at the same time - both of them want to update the value. First they got the account from the trie with the initial balance, do some calculations and both save - for example both add 5 to the balance, because of parallel execution the balance of the account will be initial + 5, but it should be initial + 5 + 5.
Another problem is both processors wants to save into the same data field, it depends which one was called faster.  Multiple nodes might reach to different state roothashes because the order was different.

### DIFF
--- a/process/coordinator/process.go
+++ b/process/coordinator/process.go
@@ -143,7 +143,7 @@ func (tc *transactionCoordinator) RequestBlockTransactions(body block.Body) {
 
 	for key, value := range separatedBodies {
 		go func(blockType block.Type, blockBody block.Body) {
-			preproc := tc.getPreprocessor(blockType)
+			preproc := tc.getPreProcessor(blockType)
 			if preproc == nil {
 				wg.Done()
 				return
@@ -173,7 +173,7 @@ func (tc *transactionCoordinator) IsDataPreparedForProcessing(haveTime func() ti
 
 	for key, value := range tc.requestedTxs {
 		go func(blockType block.Type, requestedTxs int) {
-			preproc := tc.getPreprocessor(blockType)
+			preproc := tc.getPreProcessor(blockType)
 			if preproc == nil {
 				wg.Done()
 
@@ -211,7 +211,7 @@ func (tc *transactionCoordinator) SaveBlockDataToStorage(body block.Body) error 
 
 	for key, value := range separatedBodies {
 		go func(blockType block.Type, blockBody block.Body) {
-			preproc := tc.getPreprocessor(blockType)
+			preproc := tc.getPreProcessor(blockType)
 			if preproc == nil {
 				wg.Done()
 				return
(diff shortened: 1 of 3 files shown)


process/coordinator/process.go
```diff
@@ -325,39 +334,27 @@ func (tc *transactionCoordinator) RemoveBlockDataFromPool(body block.Body) error
...
-			if preproc == nil {
-				wg.Done()
-				return
-			}
-
-			err := preproc.ProcessBlockTransactions(blockBody, round, haveTime)
-			if err != nil {
-				log.Debug(err.Error())
+	// processing has to be done in order, as the order of different type of transactions over the same account is strict
+	for blockType, blockBody := range separatedBodies {
```
iulianpascalau: This is still wrong as separatedBodies is a map. Iterating through it might lead different results. Make use of the sorted keys.

process/coordinator/process.go
```diff
@@ -487,38 +484,30 @@ func (tc *transactionCoordinator) CreateMbsAndProcessTransactionsFromMe(
...
 	miniBlocks := make(block.MiniBlockSlice, 0)
 
-	tc.mutInterimProcessors.RLock()
-
-	resMutex := sync.Mutex{}
-	// TODO: think if it is good in parallel or it is needed in sequences
-	wg := sync.WaitGroup{}
-	wg.Add(len(tc.interimProcessors))
+	// processing has to be done in order, as the order of different type of transactions over the same account is strict
+	for _, blockType := range tc.keysInterimProcs {
```
iulianpascalau: Might think to guard by mutex tc.keysInterimProcs. Might refactor this by adding a function that will give a pointer-list copy of the available interim procs and iterate freely on it. Valid also for preprocessors

process/coordinator/process.go
sasurobert: agreed. changed.

process/coordinator/process.go
sasurobert: made function to get a copy of the keys, which is protected by mutex, than the iterating is more easy to understand. For the pointer-list, every element would be of blockType and the actual preprocessor, a lot of duplicated data.

sasurobert: Keys are not used in concurrent setting. ProcessBlockTransaction and CreateBlock are never called by 2 processes at the same time.

iulianpascalau: Ok, since the keys are built only on the constructor we might leave it as it is

PR #361, opened 2019-08-10
Feat/en 3280 improve set of last check point in fork detector

SebastianMarian:
This pull request does the following:

1) Replace methods RemoveNonce and RemoveShardId from nonceSyncMapCacher with only one, called Remove, to avoid the ambiguity of selection or to wrongly select the first one when the action of removing should be done only for a specific shard and for this action should be called the second one (one wrong call was done in method removeProcessedMetablocksFromPool from shardProcessor).

2) Replace the variables lastNotarizedHdrs and finalNotarizedHdrs from block processors with only one, lastNotarizedHdrs, which now hold a list of notarized headers, so if a roll back is necessary during a fork for more than 1 block (block finality) now it could be done

3) Remove method computeHeaderHash from baseProcessor as already there is one called CalculateHash in core package

4) Add some prints to easily know what is notarized for the other chains at each commit

5) Add in shard header block displayed, the meta hashes, so now is easily to see how many meta hashes has been used/added.

6) Refactor some prints from block processors

7) Add functionality of taking into consideration in fork detector mechanism the shard headers notarized in metachain and to give the possibility for node to roll back if necessary more than 1 block if the notarized header from meta is not the same with the one on which it constructed. Also the finality in shards is given by the last notarized header (highest nonce) from the shard which has been seen in metachain blocks.

8) Refactor forkDetector by extracting some variables in structs and adjust the functionality of it, to manage the new approach of block finality and fork detection mechanism. Also fixed an edge case when an old header which would be received concomitantly with the good one but in different order could set the different fork state for nodes.


process/block/metablock.go
```diff
@@ -1041,6 +1046,7 @@ func (mp *metaProcessor) createPeerInfo() ([]block.PeerData, error) {
 
 // CreateBlockHeader creates a miniblock header list given a block body
 func (mp *metaProcessor) CreateBlockHeader(bodyHandler data.BodyHandler, round uint64, haveTime func() bool) (data.HeaderHandler, error) {
+	log.Info(fmt.Sprintf("starting create block header in round %d\n", round))
```
sasurobert: "started creating" - do we really need this log ?

process/block/metablock_test.go
```diff
@@ -1254,17 +1253,22 @@ func TestMetaProcessor_CreateShardInfoShouldWorkNoHdrAddedNotFinal(t *testing.T)
 
 	prevRandSeed := []byte("prevrand")
 
-	lastNodesHdrs := mp.LastNotarizedHdrs()
+	lastNotarizedHdrs := mp.LastNotarizedHdrs()
 	for i := uint32(0); i < shardNr; i++ {
```
sasurobert: extrac to a separate function. as it is used in the next tests, a few more times.

process/block/shardblock.go
```diff
@@ -154,6 +154,10 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	log.Info(fmt.Sprintf("starting process block with round %d and nonce %d\n",
```
sasurobert: I think these log.Info's are not really needed. Put into debug and for internal testing we could start with log flags of debug

process/block/shardblock.go
```diff
@@ -528,6 +532,7 @@ func (sp *shardProcessor) restoreMetaBlockIntoPool(miniBlockHashes map[int][][]b
 // CreateBlockBody creates a a list of miniblocks by filling them with transactions out of the transactions pools
 // as long as the transactions limit for the block has not been reached and there is still time to add transactions
 func (sp *shardProcessor) CreateBlockBody(round uint64, haveTime func() bool) (data.BodyHandler, error) {
+	log.Info(fmt.Sprintf("starting create block body in round %d\n", round))
```
sasurobert: log.Debug

process/block/shardblock.go
```diff
@@ -558,6 +563,10 @@ func (sp *shardProcessor) CommitBlock(
 		return err
 	}
 
+	log.Info(fmt.Sprintf("starting commit block with round %d and nonce %d\n",
```
sasurobert: log.Debug

process/block/shardblock.go
```diff
@@ -650,11 +652,24 @@ func (sp *shardProcessor) CommitBlock(
...
 		log.Debug(errNotCritical.Error())
 	}
 
+	errNotCritical = sp.forkDetector.AddHeader(header, headerHash, process.BHProcessed, finalHeader, finalHeaderHash)
+	if errNotCritical != nil {
+		log.Debug(errNotCritical.Error())
+	}
+
+	hdrsToAttestFinality := uint32(header.Nonce - finalHeader.Nonce)
+	sp.removeNotarizedHdrsBehindFinal(hdrsToAttestFinality)
```
sasurobert: it would be cleaner if you send the finalHeader.Nonce than in remove you iterate and remove all headers with nonce smaller that that.

process/block/shardblock.go
```diff
@@ -650,11 +652,24 @@ func (sp *shardProcessor) CommitBlock(
 		log.Debug(errNotCritical.Error())
 	}
 
-	errNotCritical = sp.forkDetector.AddHeader(header, headerHash, process.BHProcessed)
+	finalHeader, errNotCritical := sp.getHighestHdrForOwnShardFromMetachain(header.Round)
```
sasurobert: you could put finalHeaderHsah calculation inside getHighestHdrForOwnShardFromMetachain function.

process/block/shardblock.go
```diff
@@ -868,6 +955,8 @@ func (sp *shardProcessor) receivedMetaBlock(metaBlockHash []byte) {
 			requestedBlockHeaders := sp.requestFinalMissingHeaders()
 			if requestedBlockHeaders == 0 {
 				areFinalAttestingHdrsInCache = true
+			} else {
+				log.Info(fmt.Sprintf("requested %d missing final meta headers\n", requestedBlockHeaders))
```
sasurobert: we really have to think what we put into log.Info and log.Debug

process/block/shardblock.go
```diff
@@ -1290,6 +1379,7 @@ func (sp *shardProcessor) createMiniBlocks(
 
 // CreateBlockHeader creates a miniblock header list given a block body
 func (sp *shardProcessor) CreateBlockHeader(bodyHandler data.BodyHandler, round uint64, haveTime func() bool) (data.HeaderHandler, error) {
+	log.Info(fmt.Sprintf("starting create block header in round %d\n", round))
```
sasurobert: log.Debug

process/sync/basicForkDetector.go
```diff
@@ -17,13 +18,16 @@ type headerInfo struct {
...
+	nonce uint64
+	round uint64
+}
+
 type forkInfo struct {
-	checkpointNonce      uint64
-	lastCheckpointNonce  uint64
-	checkpointRound      int64
-	lastCheckpointRound  int64
+	checkpoint           []*checkpointInfo
```
sasurobert: the structures looks good :)

process/sync/basicForkDetector.go
```diff
@@ -71,55 +82,58 @@ func (bfd *basicForkDetector) AddHeader(header data.HeaderHandler, hash []byte,
...
+				nonce: finalHeader.GetNonce(),
+				round: finalHeader.GetRound(),
+				hash:  finalHeaderHash,
+				state: process.BHNotarized,
+			})
+		}
+
+		bfd.addCheckpoint(&checkpointInfo{nonce: header.GetNonce(), round: header.GetRound()})
+
 		bfd.removePastHeaders()
```
sasurobert: create one function which calls these 3 methods.

process/sync/basicForkDetector.go
```diff
@@ -71,55 +82,58 @@ func (bfd *basicForkDetector) AddHeader(header data.HeaderHandler, hash []byte,
...
+	nextRound := bfd.rounder.Index() + 1
 
 	if roundDif <= 0 {
 		return ErrLowerRoundInBlock
 	}
 	if nonceDif <= 0 {
 		return ErrLowerNonceInBlock
 	}
-	if int64(header.GetRound()) > bfd.rounder.Index() {
+	if int64(header.GetRound()) > nextRound {
```
sasurobert: should not be this >= ? as nextRound is Index()+1

process/sync/basicForkDetector.go
```diff
@@ -71,55 +82,58 @@ func (bfd *basicForkDetector) AddHeader(header data.HeaderHandler, hash []byte,
...
 		return ErrHigherRoundInBlock
 	}
-	if header.GetNonce() == bfd.fork.checkpointNonce {
-		roundTooOld := int64(header.GetRound()) < bfd.rounder.Index()-process.ForkBlockFinality
-		if roundTooOld {
-			return ErrLowerRoundInBlock
-		}
+	roundTooOld := int64(header.GetRound()) < bfd.rounder.Index()-process.ForkBlockFinality
```
sasurobert: move roundTooOld calculation to the top of the function as roundDif and nonceDif

process/sync/basicForkDetector.go
```diff
@@ -265,11 +307,15 @@ func (bfd *basicForkDetector) append(hdrInfo *headerInfo) {
 
 	for _, hdrInfoStored := range hdrInfos {
 		if bytes.Equal(hdrInfoStored.hash, hdrInfo.hash) {
-			if hdrInfoStored.state != process.BHProcessed && hdrInfo.state == process.BHProcessed {
-				// if the stored and received headers processed at the same time have equal hashes, that the old record
-				// will be replaced with the processed one. This nonce is marked at bootsrapping as processed, but as it
-				// is also received through broadcasting, the system stores as received.
-				hdrInfoStored.state = process.BHProcessed
+			if hdrInfoStored.state != process.BHProcessed {
+				// If the old appended headers have the same hashes with the new comes than the state of the old records
```
sasurobert: delete "comes"

process/block/baseProcess.go
```diff
@@ -240,23 +227,47 @@ func (bp *baseProcessor) checkHeaderTypeCorrect(shardId uint32, hdr data.HeaderH
 	return nil
 }
 
+func (bp *baseProcessor) removeNotarizedHdrsBehindFinal(hdrsToAttestFinality uint32) {
```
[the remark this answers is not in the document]
SebastianMarian: As I wanted to make only one method for both shardProcessor and metaProcessor there is hard to send the finalNonce now. As for meta I should send many final nonces depending of the number of shards.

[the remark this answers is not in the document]
SebastianMarian: So I chose to send the number of headers which should be constructed upon one header to consider that one as a final header in each chain (shard chain = last shard header notarized by meta or meta chain = 1)

process/block/baseProcess_test.go
```diff
@@ -254,6 +256,38 @@ func initStore() *dataRetriever.ChainStorer {
 	return store
 }
 
+func createTestDataPool() dataRetriever.PoolsHolder {
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/block/baseProcess.go
```diff
@@ -240,23 +227,47 @@ func (bp *baseProcessor) checkHeaderTypeCorrect(shardId uint32, hdr data.HeaderH
...
+func (bp *baseProcessor) lastNotarizedHdrForShard(shardId uint32) data.HeaderHandler {
+	notarizedHdrsCount := len(bp.lastNotarizedHdrs[shardId])
+	if notarizedHdrsCount > 0 {
+		return bp.lastNotarizedHdrs[shardId][notarizedHdrsCount-1]
+	}
+
+	return nil
+}
+
+func (bp *baseProcessor) createLastNotarizedHdrs(shardId uint32, processedHdrs []data.HeaderHandler) error {
```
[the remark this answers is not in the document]
SebastianMarian: Renamed both of them (metaProcessor and shardProcessor) from create.. to save..

process/block/metablock.go
```diff
@@ -405,6 +410,10 @@ func (mp *metaProcessor) CommitBlock(
 		return err
 	}
 
+	log.Info(fmt.Sprintf("starting commit block with round %d and nonce %d\n",
```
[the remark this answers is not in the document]
SebastianMarian: It helps on sync when we don't know if the sync of one header has not been done because of an error from processBlock or from commitBlock, as there are many common errors which could be returned from these both methods. It will be only one log per round in a normal case.  But for now I will put them on Debug level.

process/block/metablock.go
SebastianMarian: Renamed

process/block/metablock_test.go
SebastianMarian: Done

process/block/shardblock.go

process/block/shardblock.go

process/block/shardblock.go

process/block/shardblock.go
SebastianMarian: Explained above. We can do this, but we should make two separate methods, one for each processor.

process/block/shardblock.go

process/block/shardblock.go

SebastianMarian: This info is also useful to follow if the requesting final headers mechanism is properly working how it has been designed to. We can do a refactor of all the log.Info when all these things will be stable.

process/block/shardblock.go

process/sync/basicForkDetector.go
SebastianMarian: the trick is to throw stuff away :)

process/sync/basicForkDetector.go
SebastianMarian: Done

process/sync/basicForkDetector.go
SebastianMarian: Firstly it was like this, but I have seen that there are some cases when a quick proposed header, is refused here because of some little out of synchronization in terms of milliseconds between nodes clock. Actually this could happen only for blocks which are comes with flag BHProposed, and they will count only for probable highest nonce. We should check if a fast consensus could also generate this situation also for BHReceived blocks. This could be happen if the proposed header will receive very quick all the signatures needed and the proposer will broadcast this block before some nodes will advance in the next round because of the time clock differences in terms of milliseconds. For now I will keep it like this with a TODO there to analyze if this could generate some attack vectors. Anyway any header received will be checked below for signature and randome seed.

process/sync/basicForkDetector.go
SebastianMarian: Done

process/sync/basicForkDetector.go
SebastianMarian: Changed

process/block/baseProcess.go
```diff
@@ -483,3 +479,19 @@ func checkProcessorNilParameters(
 
 	return nil
 }
+
+func (bp *baseProcessor) displayLastNotarized(lastNotarizedHdrForShard data.HeaderHandler, shardId uint32) {
```
ccorcoveanu: can't this function be in the display related file?

process/block/shardblock.go
```diff
@@ -679,11 +689,84 @@ func (sp *shardProcessor) CommitBlock(
...
 }
 
+// getHighestHdrForOwnShardFromMetachain calculates the highest shard header notarized by metachain
+func (sp *shardProcessor) getHighestHdrForOwnShardFromMetachain(round uint64) (*block.Header, []byte, error) {
+	highestNonceOwnShIdHdr := &block.Header{}
+	highestNonceOwnShIdHdrHash, _ := core.CalculateHash(sp.marshalizer, sp.hasher, highestNonceOwnShIdHdr)
+
+	orderedMetaBlocks, err := sp.getOrderedMetaBlocks(round)
+	if err != nil {
+		return highestNonceOwnShIdHdr, highestNonceOwnShIdHdrHash, err
```
ccorcoveanu: Consider this scenario: Something errors, we return `&block.Header{}`.
Then, we call `AddHeader` into the fork detector with this empty header as final.

In `AddHeader` we reach the **else** case and will `setFinalCheckpoint` with nonce 0 and round 0.

dataRetriever/dataPool/nonceSyncMapCacher.go
```diff
@@ -107,13 +107,8 @@ func (nspc *nonceSyncMapCacher) copySyncMap(nonce uint64, dest dataRetriever.Sha
 	})
 }
 
-// RemoveNonce removes the provided nonce from the cache.
-func (nspc *nonceSyncMapCacher) RemoveNonce(nonce uint64) {
-	nspc.cacher.Remove(nspc.nonceConverter.ToByteSlice(nonce))
```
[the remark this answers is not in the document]
SebastianMarian: Actually nonce is the key in the map which holds nonces as key and ShardIdHashSyncMap as value for that nonce. If somebody wants to remove a nonce it should specify for which shard it wants to do this. This is why we need two parameters for any remove done.

process/block/baseProcess.go
```diff
@@ -73,11 +72,10 @@ func (bp *baseProcessor) RevertAccountState() {
...
 }
 
-// SetLastNotarizedHdr sets the last notarized header
-func (bp *baseProcessor) SetLastNotarizedHdr(shardId uint32, processedHdr data.HeaderHandler) {
+// AddLastNotarizedHdr adds the last notarized header
+func (bp *baseProcessor) AddLastNotarizedHdr(shardId uint32, processedHdr data.HeaderHandler) {
 	bp.mutNotarizedHdrs.Lock()
-	bp.finalNotarizedHdrs[shardId] = bp.lastNotarizedHdrs[shardId]
-	bp.lastNotarizedHdrs[shardId] = processedHdr
+	bp.lastNotarizedHdrs[shardId] = append(bp.lastNotarizedHdrs[shardId], processedHdr)
```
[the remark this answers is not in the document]
SebastianMarian: Actually this is a list which hold the history of notarized headers in the other chain at the moment of each new commit of the new header of its own chain. It would often happen that this to have the same header added many times one after one. If we take the case when let's say a node from shard 0 commits block with nonce 10 and let's say the highest meta block nonce used to validate this shard block has nonce 8 at the time of  the commit, it will append at this notarized list the block with nonce 8. If for the next two headers committed (11,12) nothing new will be used from meta chain, than would be another two appends with the same header with nonce 8. So the list will keep 3 nonces with value of 8. If the node will do a roll back because of any reason (fork detection for example), and it will go behind with let's say two blocks, than it will have the correct metachain notarized nonce set to 8 after it will remove the last two items from this list.

process/block/baseProcess.go
```diff
@@ -240,23 +227,47 @@ func (bp *baseProcessor) checkHeaderTypeCorrect(shardId uint32, hdr data.HeaderH
 	return nil
 }
 
+func (bp *baseProcessor) removeNotarizedHdrsBehindFinal(hdrsToAttestFinality uint32) {
+	bp.mutNotarizedHdrs.Lock()
+	for shardId := range bp.lastNotarizedHdrs {
+		notarizedHdrsCount := uint32(len(bp.lastNotarizedHdrs[shardId]))
+		if notarizedHdrsCount > hdrsToAttestFinality {
+			finalIndex := notarizedHdrsCount - 1 - hdrsToAttestFinality
+			bp.lastNotarizedHdrs[shardId] = bp.lastNotarizedHdrs[shardId][finalIndex:]
```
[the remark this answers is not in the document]
SebastianMarian: Yep, renamed

process/block/baseProcess.go
```diff
@@ -240,23 +227,47 @@ func (bp *baseProcessor) checkHeaderTypeCorrect(shardId uint32, hdr data.HeaderH
...
+		}
+	}
+	bp.mutNotarizedHdrs.Unlock()
+}
+
 func (bp *baseProcessor) restoreLastNotarized() {
```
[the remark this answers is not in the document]
SebastianMarian: Yes, actually it removes the last notarized. Renamed it.

process/block/baseProcess.go
SebastianMarian: Moved it

process/block/shardblock.go
SebastianMarian: Actually this will not happen. This is why there is on else the condition if finalHeader.GetNonce() > core.GenesisBlockNonce

SebastianMarian: As I wanted to keep the ForkDetector functional for both (metachain and shardchain), and for metachain the final block is calculated inside the metachain shard, than for metachain finalHeader and finalHash will be set to nil on AddHeader calls. But for shardchain it should be send something not nil, this is why it is created the &block.Header{} always. If there was an error the nonce is 0 and this will not set the final as 0 is the nonce of the genesis block and this block could not be received by network as everybody creates it. So, only in the case of a header with nonce greater than 0 the final will be set.

process/block/baseProcess.go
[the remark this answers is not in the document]
ccorcoveanu: Ok, got it.

process/block/shardblock.go
ccorcoveanu: Ok, got it.

PR #362, opened 2019-08-11
Development

longhuynh2811:
I encountered a sync validation error when resetting the node at the time. After resetting the node I was forced to run synchronously at nonce 56322, the front panel was synchronized, I was repeated 3 times before, currently I am in shard 1. is there a way to fix the error? this?


longhuynh2811:
> I encountered a sync validation error when resetting the node at the time. After resetting the node I was forced to run synchronously at nonce 56322, the front panel was synchronized, I was repeated 3 times before, currently I am in shard 1. is there a way to fix the error? this?
> 
> 
> 
>

PR #373, opened 2019-08-14
En 3563 update term ui

iulianpascalau:
updated gin to start in debug mode only if termui is not enabled
code refactoring in termui render
added version label in termui render


api/api.go
```diff
@@ -33,6 +33,7 @@ type prometheus struct {
 // MainApiHandler interface defines methods that can be used from `elrondFacade` context variable
 type MainApiHandler interface {
 	RestApiPort() string
+	GinDebugMode() bool
```
camilbancioiu: Could rename `Gin` to `RestAPIServer`.

cmd/node/main.go
```diff
@@ -642,7 +643,7 @@ func startNode(ctx *cli.Context, log *logger.Logger, version string) error {
 		return err
 	}
 
-	ef := facade.NewElrondNodeFacade(currentNode, apiResolver)
+	ef := facade.NewElrondNodeFacade(currentNode, apiResolver, !useTermui)
```
camilbancioiu: For readability, you could write above: `ginDebugMode := !useTermui` then pass it as the argument to `NewElrondNodeFacade`.

statusHandler/termuiStatusHandler_test.go
```diff
@@ -25,25 +25,14 @@ func TestTermuiStatusHandler_TermuiShouldPass(t *testing.T) {
...
-}
-
 func TestTermuiStatusHandler_TestIncrement(t *testing.T) {
 	t.Parallel()
 
 	var metricKey = core.MetricNonce
 
 	termuiStatusHandler := statusHandler.NewTermuiStatusHandler()
 
+	termuiStatusHandler.SetUInt64Value(metricKey, 0)
```
camilbancioiu: For readability, delete `var metricKey` and use `core.MetricNonce` directly as the arguments to `SetUInt64Value` here.

api/api.go
iulianpascalau: done

cmd/node/main.go
iulianpascalau: ok

statusHandler/termuiStatusHandler_test.go
iulianpascalau: done

PR #385, opened 2019-08-20
EN-3605 - heartbeat shard ID split into received and computed

bogdan-rosianu:
Heartbeat shard ID divided into received shard id and computed shard id in order to enable checking the validity of this information

### DIFF
--- a/core/constants.go
+++ b/core/constants.go
@@ -70,7 +70,7 @@ const MetricCountAcceptedBlocks = "erd_count_accepted_blocks"
 const MetricNodeType = "erd_node_type"
 
 // MetricLiveValidatorNodes is the metric for monitoring live validators on the network
-const MetricLiveValidatorNodes = "erd_max_validator_nodes"
+const MetricLiveValidatorNodes = "erd_live_validator_nodes"
 
 // MetricConnectedNodes is the metric for monitoring total connected peers on the network
 const MetricConnectedNodes = "erd_connected_nodes"
--- a/node/heartbeat/monitor.go
+++ b/node/heartbeat/monitor.go
@@ -66,7 +66,7 @@ func NewMonitor(
 				return nil, err
 			}
 
-			mhbi.shardID = shardId
+			mhbi.computedShardID = shardId
 			mon.heartbeatMessages[pubkey] = mhbi
 		}
 	}
@@ -121,7 +121,7 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P) error {
 			m.heartbeatMessages[string(hb.Pubkey)] = pe
 		}
 
-		pe.HeartbeatReceived(hb.ShardID, hb.VersionNumber, hb.NodeDisplayName)
+		pe.HeartbeatReceived(hb.ReceivedShardID, hb.VersionNumber, hb.NodeDisplayName)
 		m.updateAllHeartbeatMessages()
 	}(message, hbRecv)
 
@@ -175,7 +175,8 @@ func (m *Monitor) GetHeartbeats() []PubKeyHeartbeat {
 			TimeStamp:       v.timeStamp,
 			MaxInactiveTime: v.maxInactiveTime,
 			IsActive:        v.isActive,
-			ShardID:         v.shardID,
+			ReceivedShardID: v.receivedShardID,
+			ComputedShardID: v.computedShardID,
 			TotalUpTime:     v.totalUpTime,
 			TotalDownTime:   v.totalDownTime,
 			VersionNumber:   v.versionNumber,
--- a/node/heartbeat/monitor.go
+++ b/node/heartbeat/monitor.go
@@ -25,7 +25,8 @@ type Monitor struct {
 	marshalizer                 marshal.Marshalizer
 	heartbeatMessages           map[string]*heartbeatMessageInfo
 	mutHeartbeatMessages        sync.RWMutex
-	mutShardIdComputation       sync.RWMutex
+	pubKeysMap                  map[uint32][]string
+	mutPubKeysMap               sync.RWMutex
 	appStatusHandler            core.AppStatusHandler
 }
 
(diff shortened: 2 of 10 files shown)


core/constants.go
```diff
@@ -70,7 +70,7 @@ const MetricCountAcceptedBlocks = "erd_count_accepted_blocks"
 const MetricNodeType = "erd_node_type"
 
 // MetricLiveValidatorNodes is the metric for monitoring live validators on the network
-const MetricLiveValidatorNodes = "erd_max_validator_nodes"
+const MetricLiveValidatorNodes = "erd_live_validator_nodes"
```
iulianpascalau: :+1:

node/heartbeat/sender.go
```diff
@@ -36,16 +36,22 @@ func NewSender(
 	if peerMessenger == nil {
 		return nil, ErrNilMessenger
 	}
+
```
iulianpascalau: These new lines might be deleted

node/heartbeat/sender.go
```diff
@@ -66,7 +72,7 @@ func (s *Sender) SendHeartbeat() error {
 
 	hb := &Heartbeat{
 		Payload:         []byte(fmt.Sprintf("%v", time.Now())),
-		ShardID:         s.shardCoordinator.SelfId(),
+		ReceivedShardID: s.shardCoordinator.SelfId(),
```
iulianpascalau: Since this is awkward, might leave the Heartbeat's field plain 'ShardId' and have only ReceivedShardId and ComputedShardId in heartbeat message info

node/heartbeat/monitor.go
```diff
@@ -66,7 +66,7 @@ func NewMonitor(
 				return nil, err
 			}
 
-			mhbi.shardID = shardId
+			mhbi.computedShardID = shardId
```
iulianpascalau: Please add here a TODO as to not forget to recompute "computedShardID" field each time we receive a new message from the peer. Or even better, add a call to a func on line 124 and that func will have this TODO

node/heartbeat/monitor_test.go
```diff
@@ -92,6 +92,29 @@ func TestNewMonitor_OkValsShouldCreatePubkeyMap(t *testing.T) {
...
+
+	pksPerShards := map[uint32][]string{
+		0: {"pk0"},
+		1: {"pk1"},
+	}
+	mon, err := heartbeat.NewMonitor(
+		&mock.SinglesignMock{},
+		&mock.KeyGenMock{},
+		&mock.MarshalizerMock{},
+		1,
```
iulianpascalau: Might add a var for this

node/heartbeat/monitor.go
sasurobert: computedShardID has to be updated from peer processor / node coordinator, it will be updated at the end of the epoch / start of the new.

sasurobert: Might add a value which is treated as unallocated. so peer if not allocated to any shard.

node/heartbeat/sender.go
```diff
@@ -36,16 +36,22 @@ func NewSender(
...
 	}
+
 	if singleSigner == nil {
 		return nil, ErrNilSingleSigner
 	}
+
 	if privKey == nil {
 		return nil, ErrNilPrivateKey
 	}
+
```
sasurobert: leave as it was before. without empty lines between nil checks

node/heartbeat/sender.go
sasurobert: +1

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/monitor.go
bogdan-rosianu: done. made a function which has that TODO

node/heartbeat/monitor_test.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: done

node/heartbeat/monitor.go
```diff
@@ -121,13 +121,20 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P) error {
...
 		m.updateAllHeartbeatMessages()
 	}(message, hbRecv)
 
 	return nil
 }
 
+func (m *Monitor) computeShardID(pubkey string) uint32 {
+	// TODO : the shard ID will be recomputed at the end of an epoch / beginning of a new one.
+	//  For the moment, just return the initial computed shard ID
+	return m.heartbeatMessages[pubkey].computedShardID
```
iulianpascalau: this should definitely protected by a mutex

node/heartbeat/monitor_test.go
```diff
@@ -99,11 +99,13 @@ func TestNewMonitor_ShouldComputeShardId(t *testing.T) {
 		0: {"pk0"},
 		1: {"pk1"},
 	}
+
+	maxDuration := 1 * time.Millisecond
```
iulianpascalau: maybe remove 1 *

bogdan-rosianu: I think it is more readable this way.

node/heartbeat/monitor.go
bogdan-rosianu: done + tested with race conditions

node/heartbeat/monitor_test.go
sasurobert: no need for the 1 *. it is remove from the whole project.

bogdan-rosianu: ok, removed

node/heartbeat/monitor.go
bogdan-rosianu: redone with a mutex protected map of pub keys

PR #386, opened 2019-08-20
EN-3596 - feat/interceptors

iulianpascalau:
Refactored interceptors except RewardInterceptor which will be done in a separate branch


camilbancioiu:
Does this branch have a corresponding JIRA issue?

PR #399, opened 2019-08-26
EN-2890 : remove second generation of the genesis block

miiu96:
remove from main.go  CreateShardGenesisBlock and CreateMetaGenesisBlock and set in structs.go GenesisHeader and GenesisHeaderHash

### DIFF
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -512,7 +512,7 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
 func prepareGenesisBlock(args *processComponentsFactoryArgs, shardsGenesisBlocks map[uint32]data.HeaderHandler) error {
 	genesisBlock, ok := shardsGenesisBlocks[args.shardCoordinator.SelfId()]
 	if !ok {
-		return errors.New("genesis block does not exits")
+		return errors.New("genesis block does not exists")
 	}
 
 	genesisBlockHash, err := core.CalculateHash(args.core.Marshalizer, args.core.Hasher, genesisBlock)
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -512,7 +512,7 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
 func prepareGenesisBlock(args *processComponentsFactoryArgs, shardsGenesisBlocks map[uint32]data.HeaderHandler) error {
 	genesisBlock, ok := shardsGenesisBlocks[args.shardCoordinator.SelfId()]
 	if !ok {
-		return errors.New("genesis block not exits")
+		return errors.New("genesis block does not exits")
 	}
 
 	genesisBlockHash, err := core.CalculateHash(args.core.Marshalizer, args.core.Hasher, genesisBlock)
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -85,8 +85,8 @@ const (
 	MaxTxsToRequest = 100
 )
 
-// ErrShardGenesisBlockNotExits signals that genesis block not exits
-var ErrShardGenesisBlockNotExits = errors.New("genesis block not exits")
+// errShardGenesisBlockNotExits signals that genesis block not exits
+var errShardGenesisBlockNotExits = errors.New("genesis block not exits")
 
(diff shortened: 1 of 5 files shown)


cmd/node/factory/structs.go
```diff
@@ -480,6 +483,22 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
...
+		return nil, err
+	}
+
+	err = args.data.Blkc.SetGenesisHeader(genesisBlock)
+	if err != nil {
+		return nil, err
+	}
+
+	args.data.Blkc.SetGenesisHeaderHash(genesisBlockHash)
+
```
sasurobert: treat error

cmd/node/main.go
```diff
@@ -1090,20 +1090,12 @@ func createNode(
 		if err != nil {
 			return nil, err
 		}
-		err = nd.CreateShardGenesisBlock()
```
sasurobert: delete unused functions from node.go

miiu96: Done. Also a test was needed to be removed.

node/node.go
```diff
@@ -329,46 +328,6 @@ func (n *Node) StartConsensus() error {
...
-	}
-
-	blockHeaderHash := n.hasher.Compute(string(marshalizedHeader))
-
-	return n.setGenesis(header, blockHeaderHash)
-}
-
-// CreateMetaGenesisBlock creates the meta genesis block
-func (n *Node) CreateMetaGenesisBlock() error {
-	header, err := genesis.CreateMetaGenesisBlock(uint64(n.genesisTime.Unix()), n.initialNodesPubkeys)
```
iulianpascalau: please add comment to core/genesis/genesisBlockCreator.go CreateMetaGenesisBlock func

node/node.go
```diff
@@ -329,46 +328,6 @@ func (n *Node) StartConsensus() error {
...
-		return err
-	}
-
-	blockHeaderHash := n.hasher.Compute(string(marshalizedHeader))
-	err = n.store.Put(dataRetriever.MetaBlockUnit, blockHeaderHash, marshalizedHeader)
-	if err != nil {
-		return err
-	}
-
-	return n.setGenesis(header, blockHeaderHash)
```
iulianpascalau: n.setGenesis is no longer used

node/node.go
```diff
@@ -329,46 +328,6 @@ func (n *Node) StartConsensus() error {
...
-// CreateMetaGenesisBlock creates the meta genesis block
-func (n *Node) CreateMetaGenesisBlock() error {
-	header, err := genesis.CreateMetaGenesisBlock(uint64(n.genesisTime.Unix()), n.initialNodesPubkeys)
-	marshalizedHeader, err := n.marshalizer.Marshal(header)
-	if err != nil {
-		return err
-	}
-
-	blockHeaderHash := n.hasher.Compute(string(marshalizedHeader))
-	err = n.store.Put(dataRetriever.MetaBlockUnit, blockHeaderHash, marshalizedHeader)
```
iulianpascalau: This is not done anymore. Also, shard genesis block is not written.

node/node.go
miiu96: Done. Comment added.

node/node.go
miiu96: Done. Method was removed.

node/node.go
miiu96: I don't think is necessary to to write shard genesis block.

sasurobert: write the created genesis blocks into the storage.

miiu96: Done. Block written in storage.

cmd/node/factory/structs.go
```diff
@@ -85,9 +85,6 @@ const (
 	MaxTxsToRequest = 100
 )
 
-// ErrShardGenesisBlockNotExits signals that genesis block not exits
-var ErrShardGenesisBlockNotExits = errors.New("genesis block not exits")
```
iulianpascalau: Might leave this err and use it on line 515. Also. Might change the error string to "genesis block **does** not exi**sts**"

miiu96: The error string has been changed with the suggested one. But not created an error variable because in structs.go all returned errors are created with errors.New().

iulianpascalau: ok

PR #410, opened 2019-08-31
En 3828/smart contract results for own shard

sasurobert:
smart contract results for own shard. Treated as receipts, added into the block.


process/smartContract/process_test.go
```diff
@@ -1717,11 +1717,11 @@ func TestScProcessor_processSCOutputAccounts(t *testing.T) {
 
 	outAccs, err := sc.ProcessSCOutputAccounts(outputAccounts)
 	assert.Nil(t, err)
-	assert.Equal(t, 0, len(outAccs))
+	assert.Equal(t, len(outAccs), len(outAccs))
```
bogdan-rosianu: refactor these asserts as they will always be true because you compare the same stuff

sasurobert: resolved

bogdan-rosianu: node done yet. plase fix lines 1720 and 1724 also

PR #412, opened 2019-09-02
EN-3836-fix bug with status is synchornized while it shouldn't have been

miiu96:
fix bug with status is synchronized while it shouldn't


cmd/node/main.go
```diff
@@ -538,6 +538,7 @@ func startNode(ctx *cli.Context, log *logger.Logger, version string) error {
 	coreComponents.StatusHandler.SetStringValue(core.MetricConsensusState, "")
 	coreComponents.StatusHandler.SetStringValue(core.MetricConsensusRoundState, "")
 	coreComponents.StatusHandler.SetStringValue(core.MetricCrossCheckBlockHeight, "")
+	coreComponents.StatusHandler.SetUInt64Value(core.MetricIsSyncing, uint64(1))
```
sasurobert: what does uint64(1) means ?

iulianpascalau: Should have a constant for that

miiu96: Done.

statusHandler/presenter/presenterStatusHandler_test.go
```diff
@@ -103,14 +103,15 @@ func TestPresenterStatusHandler_Write(t *testing.T) {
 func TestPresenterStatusHandler_GetLogLine(t *testing.T) {
 	t.Parallel()
 
+	waitTimeToWrite := 10 * time.Millisecond
```
sasurobert: waitTimeBeforeRead

miiu96: Done. Name was changed.

PR #415, opened 2019-09-03
En 3828/index scrs to elastic search

sasurobert:
Added smart contract results to elastic search indexer as well.


process/block/preprocess/intermediateResults.go
```diff
@@ -237,6 +237,25 @@ func (irp *intermediateResultsProcessor) CreateMarshalizedData(txHashes [][]byte
...
+
+	scrPool := make(map[string]data.TransactionHandler)
+	for txHash, txInfo := range irp.interResultsForBlock {
+		if txInfo.receiverShardID != irp.shardCoordinator.SelfId() {
+			continue
+		}
+		if txInfo.senderShardID != irp.shardCoordinator.SelfId() {
+			continue
+		}
+		scrPool[txHash] = txInfo.tx
```
raduchis: You add just the transactions that are intrashard? The ones that have both sender and receiver as SelfId?

process/block/preprocess/intermediateResults.go
```diff
@@ -237,6 +237,25 @@ func (irp *intermediateResultsProcessor) CreateMarshalizedData(txHashes [][]byte
 	return mrsTxs, nil
 }
 
+// GetAllCurrentUsedTxs returns the cached finalized transactions for current round
+func (irp *intermediateResultsProcessor) GetAllCurrentUsedTxs() map[string]data.TransactionHandler {
```
iulianpascalau: GetAllIntraShardGeneratedScResults? as this reflect better what it does

process/block/preprocess/intermediateResults.go
iulianpascalau: Might add some bool vars like:
```
isDestInOtherShard := txInfo.receiverShardID != irp.shardCoordinator.SelfId()
isSenderInOtherShard := txInfo.senderShardID != irp.shardCoordinator.SelfId()
```

process/coordinator/process.go
```diff
@@ -628,13 +628,29 @@ func (tc *transactionCoordinator) CreateMarshalizedData(body block.Body) (map[ui
...
+		tc.mutPreProcessor.RUnlock()
 		return nil
 	}
 
-	return tc.txPreProcessors[blockType].GetAllCurrentUsedTxs()
+	txPool := tc.txPreProcessors[blockType].GetAllCurrentUsedTxs()
+	tc.mutPreProcessor.RUnlock()
+
+	tc.mutInterimProcessors.RLock()
+	if _, ok := tc.interimProcessors[blockType]; !ok {
```
iulianpascalau: Please make this a two liner

process/block/preprocess/intermediateResults.go
iulianpascalau: Func not tested :)

sasurobert: this implements an interface. it will not be necessary sc result.

process/block/preprocess/intermediateResults.go
sasurobert: wanted to add only those which are finalized, otherwise, they will get added 2 times.

process/block/preprocess/intermediateResults.go
sasurobert: done

process/coordinator/process.go
sasurobert: done

process/block/preprocess/intermediateResults.go
sasurobert: if I add those bool's I have to change on the whole project.

raduchis: ok

PR #417, opened 2019-09-04
En 3833 observers get out of mem

iulianpascalau:
Added a new implementation of DataPacker that calculates less precise but more optimum the packed size in bytes.

### DIFF
--- a/core/partitioning/simpleDataPacker.go
+++ b/core/partitioning/simpleDataPacker.go
@@ -1,0 +1,71 @@
+package partitioning
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/marshal"
+)
+
+// SimpleDataPacker can split a large slice of byte slices in chunks <= maxPacketSize
+// If one element still exceeds maxPacketSize, it will be returned alone
+// It does the marshaling of the resulted (smaller) slice of byte slices
+type SimpleDataPacker struct {
+	marshalizer marshal.Marshalizer
+}
+
+// NewSimpleDataPacker creates a new SizeDataPacker instance
+func NewSimpleDataPacker(marshalizer marshal.Marshalizer) (*SimpleDataPacker, error) {
+	if marshalizer == nil || marshalizer.IsInterfaceNil() {
+		return nil, core.ErrNilMarshalizer
+	}
+
+	return &SimpleDataPacker{
+		marshalizer: marshalizer,
+	}, nil
+}
+
+// PackDataInChunks packs the provided data into smaller chunks
+// limit is expressed in bytes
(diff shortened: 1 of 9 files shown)


core/partitioning/simpleDataPacker.go
```diff
@@ -0,0 +1,71 @@
...
+// SimpleDataPacker can split a large slice of byte slices in chunks <= maxPacketSize
+// If one element still exceeds maxPacketSize, it will be returned alone
+// It does the marshaling of the resulted (smaller) slice of byte slices
+type SimpleDataPacker struct {
+	marshalizer marshal.Marshalizer
+}
+
+// NewSimpleDataPacker creates a new SizeDataPacker instance
+func NewSimpleDataPacker(marshalizer marshal.Marshalizer) (*SimpleDataPacker, error) {
+	if marshalizer == nil || marshalizer.IsInterfaceNil() {
```
sasurobert: use your core func of isNil :)

core/partitioning/simpleDataPacker.go
```diff
@@ -0,0 +1,71 @@
...
+		isBuffToLarge := lenElements+len(element) >= limit
+		elementsNotEmpty := len(elements) > 0
+		if isBuffToLarge && elementsNotEmpty {
+			marshaledElements, _ := sdp.marshalizer.Marshal(elements)
+			returningBuff = append(returningBuff, marshaledElements)
+			elements = make([][]byte, 0)
+			lenElements = 0
+		}
+
+		elements = append(elements, element)
```
sasurobert: rename suggestion currentChunk

core/partitioning/simpleDataPacker.go
```diff
@@ -0,0 +1,71 @@
+package partitioning
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/marshal"
+)
+
+// SimpleDataPacker can split a large slice of byte slices in chunks <= maxPacketSize
```
sasurobert: you have the same commentary as sizeDataPacker - what is the difference - explain it in the comments. exaplin that it is not doing continuous marshalling. Add comment in the sizeDataPacker why is it doing continuous marshalling - precision, etc.

core/partitioning/simpleDataPacker.go
iulianpascalau: Agree

iulianpascalau: Not available yet. It is on feat/interceptors branch.

core/partitioning/simpleDataPacker.go
iulianpascalau: done

core/partitioning/simpleDataPacker.go
iulianpascalau: added explanations

PR #426, opened 2019-09-07
Update redirectStderrLinux.go

LucianMincu:
fixed build on arm64 as Dup2 unavailable


core/logger/redirectStderrLinux.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // redirectStderr redirects the output of the stderr to the file passed in
 func redirectStderr(f *os.File) error {
-	err := syscall.Dup2(int(f.Fd()), int(os.Stderr.Fd()))
+	err := syscall.Dup3(int(f.Fd()), int(os.Stderr.Fd()), os.O_CREATE|os.O_APPEND|os.O_WRONLY)
```
iulianpascalau: This doesn't work. Will think of a solution.

iulianpascalau: As found by @sasurobert ,  might replace `os.O_CREATE|os.O_APPEND|os.O_WRONLY` with `0`

PR #431, opened 2019-09-10
En 3877/bug fix restore blocks into pool set processed mbs

sasurobert:
When restoring a metablock from the storage set all the miniblocks to being processed, as we save into the storage only when the metablock is fully processed.


process/block/shardblock_test.go
```diff
@@ -4842,3 +4842,84 @@ func TestShardProcessor_GetHighestHdrForOwnShardFromMetachaiMetaHdrsWithOwnHdrSt
 	assert.NotNil(t, hdr)
 	assert.Equal(t, ownHdr.GetNonce(), hdr.GetNonce())
 }
+
+func TestShardProcessor_restoreMetaBlockIntoPoolVerifyMiniblocks(t *testing.T) {
```
BeniaminDrasovean: Could change restore to Restore.

process/block/shardblock.go
```diff
@@ -484,6 +484,11 @@ func (sp *shardProcessor) restoreMetaBlockIntoPool(miniBlockHashes map[string]ui
 			continue
 		}
 
+		processedMiniBlocks := metaBlock.GetMiniBlockHeadersWithDst(sp.shardCoordinator.SelfId())
```
raduchis: Should this happen also on the metaBlock RestoreFromPool?

process/block/shardblock_test.go
sasurobert: done

process/block/shardblock.go
sasurobert: No. Metachain notarizes and processes headers. He does need to look into miniblocks.

raduchis: ok

SebastianMarian: I would use the same variable name as it is used in line 526: crossMiniBlockHashes as actually this is why it represents

PR #443, opened 2019-09-13
En 3924 fix bug when nil notarized header is added and displayed

SebastianMarian:
* Fixed a situation when two or more blocks with the same nonce and the same round would be proposed by the same proposer and fork detector would not detected
them as a problem (added additional checks in fork detector when round and nonce are equal. If this would be happened, then the the lowest hash would be chosen, so all the nodes would have selected the same block as the wright one)


process/block/baseProcess.go
```diff
@@ -271,7 +271,7 @@ func (bp *baseProcessor) removeLastNotarized() {
 	bp.mutNotarizedHdrs.Lock()
 	for shardId := range bp.notarizedHdrs {
 		notarizedHdrsCount := len(bp.notarizedHdrs[shardId])
-		if notarizedHdrsCount > 0 {
+		if notarizedHdrsCount > 1 {
```
iulianpascalau: I'm against these types of changes without a proper test. Also, notarizedHdrsCount > 1 could have been extracted in a bool var to better understand the case.

process/sync/basicForkDetector.go
```diff
@@ -81,16 +81,23 @@ func (bfd *basicForkDetector) AddHeader(
...
-			bfd.setFinalCheckpoint(&checkpointInfo{nonce: finalHeader.GetNonce(), round: finalHeader.GetRound()})
-			bfd.append(&headerInfo{
-				nonce: finalHeader.GetNonce(),
-				round: finalHeader.GetRound(),
-				hash:  finalHeaderHash,
-				state: process.BHNotarized,
-			})
+		} else {
+			for i := 0; i < len(finalHeaders); i++ {
+				if finalHeaders[i].GetNonce() > bfd.GetHighestFinalBlockNonce() {
```
iulianpascalau: Should extract finalHeaders[i].GetNonce() > bfd.GetHighestFinalBlockNonce() into a var

process/block/baseProcess.go
sasurobert: It was tested, try to launch a node in a public testnet with a wrong config - which results in a different starting genesis block. The problem is we should never revert the first put block. Not a written test. But here was a logical error. Notarized list should never be nil.

process/sync/basicForkDetector.go
SebastianMarian: Done. Also extracted all the code from for in new method

process/block/baseProcess.go
SebastianMarian: Added test

process/sync/basicForkDetector.go
```diff
@@ -362,14 +368,26 @@ func (bfd *basicForkDetector) CheckFork() (bool, uint64, []byte) {
 				continue
 			}
 			if hdrInfos[i].state == process.BHNotarized {
-				lowestRoundInForkNonce = 0
-				forkHeaderHash = hdrInfos[i].hash
+				if lowestRoundInForkNonce > 0 {
+					lowestRoundInForkNonce = 0
+					forkHeaderHash = hdrInfos[i].hash
+				} else {
+					if strings.Compare(string(hdrInfos[i].hash), string(forkHeaderHash)) < 0 {
```
iulianpascalau: Might use bytes.Compare instead if strings.Compare

process/sync/basicForkDetector.go
```diff
@@ -362,14 +368,26 @@ func (bfd *basicForkDetector) CheckFork() (bool, uint64, []byte) {
...
 
-			if hdrInfos[i].round < lowestRoundInForkNonce {
-				lowestRoundInForkNonce = hdrInfos[i].round
-				forkHeaderHash = hdrInfos[i].hash
+			if hdrInfos[i].round <= lowestRoundInForkNonce {
+				if hdrInfos[i].round < lowestRoundInForkNonce {
+					lowestRoundInForkNonce = hdrInfos[i].round
+					forkHeaderHash = hdrInfos[i].hash
+				} else {
+					if strings.Compare(string(hdrInfos[i].hash), string(forkHeaderHash)) < 0 {
```
iulianpascalau: Also here

process/sync/basicForkDetector.go
```diff
@@ -362,14 +368,26 @@ func (bfd *basicForkDetector) CheckFork() (bool, uint64, []byte) {
 				continue
 			}
 			if hdrInfos[i].state == process.BHNotarized {
-				lowestRoundInForkNonce = 0
-				forkHeaderHash = hdrInfos[i].hash
+				if lowestRoundInForkNonce > 0 {
```
iulianpascalau: Might refactor this multi-level if instructions

process/sync/basicForkDetector.go
SebastianMarian: Done

process/sync/basicForkDetector.go
SebastianMarian: Done

process/sync/basicForkDetector.go
SebastianMarian: Done

PR #453, opened 2019-09-18
EN-4050: added starting parameters to the session.info file

bogdan-rosianu:


cmd/node/main.go
```diff
@@ -567,10 +567,18 @@ func startNode(ctx *cli.Context, log *logger.Logger, version string) error {
 		"GenesisTimeStamp", startTime.Unix(),
 	)
 
+	sessionInfoFileOutput += "\nStarted with parameters:\n"
```
SebastianMarian: You can also use here Sprintf even if there are no parameters for now

PR #457, opened 2019-09-18
En 4077 refactor shard processor constructor

miiu96:
Refactor shard processor constructor.
All constructor parameters was moved in a new struct.
Modify failing tests.

### DIFF
--- a/cmd/node/factory/structs.go
+++ b/cmd/node/factory/structs.go
@@ -1503,7 +1503,7 @@ func newShardBlockProcessorAndTracker(
 		return nil, nil, err
 	}
 
-	argumentsBaseProcessor := block.ArgsBaseProcessor{
+	argumentsBaseProcessor := block.ArgBaseProcessor{
 		Accounts:         state.AccountsAdapter,
 		ForkDetector:     forkDetector,
 		Hasher:           core.Hasher,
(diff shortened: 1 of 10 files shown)


process/block/argsProcessor.go
```diff
@@ -0,0 +1,37 @@
+package block
```
iulianpascalau: As to be consistent, should rename the file and structs to argProcessor (no s)

process/block/baseProcess_test.go
```diff
@@ -314,24 +314,38 @@ func (wr *wrongBody) IsInterfaceNil() bool {
 	return false
 }
 
+func CreateMockArguments() blproc.ArgsShardProcessor {
+	arguments := blproc.ArgsShardProcessor{
+		ArgsBaseProcessor: &blproc.ArgsBaseProcessor{
```
iulianpascalau: Should instantiate the fields with the most used variants as seen in shardBlock_test.go

cmd/node/factory/structs.go
```diff
@@ -1503,21 +1503,25 @@ func newShardBlockProcessorAndTracker(
...
+		RequestHandler:   requestHandler,
+		Core:             coreServiceContainer,
+	}
+	arguments := block.ArgsShardProcessor{
+		ArgsBaseProcessor: &argumentsBaseProcessor,
+		DataPool:          data.Datapool,
+		BlocksTracker:     blockTracker,
+		TxCoordinator:     txCoordinator,
+	}
+	blockProcessor, err := block.NewShardProcessor(arguments)
```
sasurobert: empty line before

process/block/argsProcessor.go
miiu96: Done.

cmd/node/factory/structs.go
miiu96: Done.

process/block/baseProcess_test.go
miiu96: Done.

PR #461, opened 2019-09-20
En 3676 rewards for metachain

AdoAdoAdo:


process/errors.go
```diff
@@ -46,6 +46,9 @@ var ErrNilBlockChain = errors.New("nil block chain")
 // ErrNilMetaBlockHeader signals that an operation has been attempted to or with a nil metablock
 var ErrNilMetaBlockHeader = errors.New("nil metablock header")
 
+// ErrNilMetaBlockHashes signals that a nil array of metablock header hashes was provided
+var ErrNilMetaBlockHashes = errors.New("nil metablock header hashes")
```
BeniaminDrasovean: Unused variable.

process/block/preprocess/rewardsHandler.go
```diff
@@ -345,23 +345,50 @@ func (rtxh *rewardsHandler) createRewardFromFees() []data.TransactionHandler {
...
+func (rtxh *rewardsHandler) createProtocolRewardsForMeta() []data.TransactionHandler {
+	metaRewardsData := rtxh.address.ConsensusMetaRewardData()
+	consensusRewardTxs := make([]data.TransactionHandler, 0)
+
+	for _, metaConsensusSet := range metaRewardsData {
+		for _, address := range metaConsensusSet.Addresses {
+			shardId, err := rtxh.address.ShardIdForAddress([]byte(address))
+			if err != nil {
+				log.Error(err.Error())
+				continue
```
sasurobert: if this is an error, should not it break ?

process/block/shardblock.go
```diff
@@ -232,6 +226,16 @@ func (sp *shardProcessor) ProcessBlock(
 		}
 	}()
 
+	processedMetaHdrs, err := sp.getProcessedMetaBlocks(body, header.MetaBlockHashes)
```
sasurobert: use function getProcessedMetablocksFromHeader, it gives you all the metablocks which are fully processed.

process/interface.go
```diff
@@ -127,12 +127,15 @@ type SpecialAddressHandler interface {
 	SetElrondCommunityAddress(elrond []byte)
 	ElrondCommunityAddress() []byte
 	SetConsensusData(consensusRewardAddresses []string, round uint64, epoch uint32)
-	ConsensusRewardAddresses() []string
+	ConsensusShardRewardData() *data.ConsensusRewardData
+	SetMetaConsensusData(rewardAddresses []string, round uint64, epoch uint32)
```
sasurobert: can we simplify this interface ? like sending the shardID for with set and get consensus data ?

process/block/shardblock.go
```diff
@@ -250,6 +254,44 @@ func (sp *shardProcessor) ProcessBlock(
 	return nil
 }
 
+func (sp *shardProcessor) setShardConsensusData(headerHandler data.HeaderHandler) error {
```
sasurobert: this functions should be in a separate component.

process/block/shardblock.go
```diff
@@ -250,6 +254,44 @@ func (sp *shardProcessor) ProcessBlock(
...
+	if err != nil {
+		return err
+	}
+
+	sp.SetConsensusData(consensusAddresses, headerHandler.GetRound())
+
+	return nil
+}
+
+func (sp *shardProcessor) setMetaConsensusData(finalizedMetaBlocks []data.HeaderHandler) error {
```
sasurobert: this functions should be in a separate component.

process/block/shardblock.go
```diff
@@ -1363,17 +1483,22 @@ func (sp *shardProcessor) createMiniBlocks(
 		log.Info(err.Error())
 	}
 
+	processedMetaHdrs, errNotCritical := sp.getProcessedMetaBlocks(destMeMiniBlocks, usedMetaHdrsHashes)
+	if errNotCritical != nil {
+		log.Debug(errNotCritical.Error())
+	}
+
+	err = sp.setMetaConsensusData(processedMetaHdrs)
```
sasurobert: a separate component could get the usedMetaHdrHashes and the dstMeMiniBlocks and set all the needed data to the special address handler. Should me moved out from blockprocessor.

process/block/shardblock.go
AdoAdoAdo: processMetablocksFromHeader also does some setting of flags which is not needed when only processing, and this function is used as well in createBlock where there is no header available.

process/interface.go
AdoAdoAdo: I think it's better to have two different functions instead, as it becomes clearer.

data/address/specialAddresses.go
```diff
@@ -63,34 +63,62 @@ func (sp *specialAddresses) BurnAddress() []byte {
 }
 
 // SetConsensusData sets the consensus rewards addresses for the round
-func (sp *specialAddresses) SetConsensusData(consensusRewardAddresses []string, round uint64, epoch uint32) {
-	sp.consensusRewardAddresses = consensusRewardAddresses
-	sp.round = round
-	sp.epoch = epoch
+func (sp *specialAddresses) SetConsensusData(rewardAddresses []string, round uint64, epoch uint32) {
```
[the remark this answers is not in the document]
AdoAdoAdo: done

data/address/specialAddresses.go
```diff
@@ -63,34 +63,62 @@ func (sp *specialAddresses) BurnAddress() []byte {
...
+	sp.shardConsensusData = &data.ConsensusRewardData{
+		Round:     round,
+		Epoch:     epoch,
+		Addresses: rewardAddresses,
+	}
+}
+
+// ConsensusShardRewardAddresses provides the consensus reward addresses
```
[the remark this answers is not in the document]
AdoAdoAdo: done

data/address/specialAddresses.go
```diff
@@ -63,34 +63,62 @@ func (sp *specialAddresses) BurnAddress() []byte {
...
+		Addresses: rewardAddresses,
+	})
+}
+
+// ClearMetaConsensusData clears the previously set addresses for rewarding metachain nodes
+func (sp *specialAddresses) ClearMetaConsensusData() {
+	sp.metaConsensusData = make([]*data.ConsensusRewardData, 0)
+}
+
+func (sp *specialAddresses) ConsensusMetaRewardData() []*data.ConsensusRewardData {
```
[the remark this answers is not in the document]
AdoAdoAdo: done

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/integrationTests"
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/opentracing/opentracing-go/log"
```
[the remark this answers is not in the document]
AdoAdoAdo: the loging was only used temporarily while doing some debugging, should have been removed

process/errors.go
AdoAdoAdo: removed

process/block/export_test.go
```diff
@@ -45,6 +45,7 @@ func (sp *shardProcessor) CreateMiniBlocks(noShards uint32, maxItemsInBlock uint
 }
 
 func (sp *shardProcessor) GetProcessedMetaBlocksFromHeader(header *block.Header) ([]data.HeaderHandler, error) {
+
```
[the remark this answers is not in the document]
AdoAdoAdo: done

process/block/preprocess/rewardsHandler.go
AdoAdoAdo: Normally this should not happen when the staking contract is available, because the address must be valid to register with the contract.

While we use config files, if an invalid address was added as a staking address, or the address is missing, then the shard ID for that address cannot be calculated and this translates in not being able to create a reward transaction for that address. 

The question is if whenever such a node is part of the consensus group if we should cancel the respective round, or we can just ignore the wrong address.

process/block/shardblock.go
sasurobert: common logic of the 2 functions should be moved together. making a base function, and one which sets the processed miniblocks as well. Even if that means iterating 2 times over the same set of data, it is only metablocks, which is at most a few elements.

AdoAdoAdo: done.

process/block/shardblock.go
AdoAdoAdo: the logic for selecting the right addresses etc was moved to the address handler, these simply call the interface, so there is no need of creating a separate component.

process/block/shardblock.go
AdoAdoAdo: the logic for selecting the right addresses etc was moved to the address handler, these simply call the interface, so there is no need of creating a separate component.

process/block/shardblock.go
AdoAdoAdo: the logic for selecting the right addresses etc was moved to the address handler, these simply call the interface, so there is no need of creating a separate component.

process/block/shardblock.go
sasurobert: There is still logic inside the function, which could have its own component

sasurobert: We should thrive to make blockprocessor smaller and to call interfaces.

AdoAdoAdo: The processor should be refactored in a separate task.
There is no need of a component that only passes data from one component to the other, it makes no sense.

data/mock/addressConverterMock.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	return hex.EncodeToString(addressContainer.Bytes()), nil
+}
+
+func (acm *AddressConverterMock) CreateAddressFromHex(hexAddress string) (state.AddressContainer, error) {
+	if acm.Fail {
+		return nil, errFailure
+	}
+
```
sasurobert: implement mock functions, panic is not preferred. use a mock default implementation, if the "Called" function is not implemented.

data/mock/addressConverterMock.go
```diff
@@ -0,0 +1,66 @@
...
+	CreateAddressFromPublicKeyBytesRetErrForValue []byte
+}
+
+func (acm *AddressConverterMock) CreateAddressFromPublicKeyBytes(pubKey []byte) (state.AddressContainer, error) {
+	if acm.Fail {
+		return nil, errFailure
+	}
+
+	if acm.CreateAddressFromPublicKeyBytesRetErrForValue != nil {
+		if bytes.Equal(acm.CreateAddressFromPublicKeyBytesRetErrForValue, pubKey) {
```
sasurobert: This is confusing. Use the typical function"Called" method from other mocks.

AdoAdoAdo: this was mock copied over from process I think.

data/mock/addressConverterMock.go
AdoAdoAdo: This was mock copied over from process.

PR #464, opened 2019-09-20
En 4096 merge development interceptors

iulianpascalau:
fixes after merge between development and feat/interceptors
Integrated TxValidator in new interceptors structures


process/transaction/interceptedTransaction.go
```diff
@@ -100,6 +105,7 @@ func (inTx *InterceptedTransaction) CheckValidity() error {
 func (inTx *InterceptedTransaction) processFields(txBuff []byte) error {
 	inTx.hash = inTx.hasher.Compute(string(txBuff))
 
+	var err error
```
sasurobert: var not needed. initialize with :=

iulianpascalau: Golang is not happy with 
```GO
inTx.sndAddr, err := inTx.addrConv.CreateAddressFromPublicKeyBytes(inTx.tx.SndAddr)
```
(assigning to an existing struct member and creating a new var instance on the same line)

process/dataValidators/txValidator_test.go
```diff
@@ -214,3 +215,16 @@ func TestTxValidator_IsTxValidForProcessingTxIsOkShouldReturnTrue(t *testing.T)
...
 }
+
+//------- IsInterfaceNil
+
+func TestTxValidator_IsInterfaceNil(t *testing.T) {
+	t.Parallel()
+
+	accounts := getAccAdapter(0, big.NewInt(0))
+	shardCoordinator := createMockCoordinator("_", 0)
+	txValidator, _ := dataValidators.NewTxValidator(accounts, shardCoordinator)
```
AdoAdoAdo: what is the purpose of creating a txValidator if you set it to nil afterwards.

PR #469, opened 2019-09-21
Dev merge into economics 21 09 2019

AdoAdoAdo:


process/block/preprocess/rewardTxPreProcessor.go
```diff
@@ -2,6 +2,7 @@ package preprocess
 
 import (
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/core"
```
sasurobert: goimports

PR #478, opened 2019-09-24
EN-3754: added tests for metachain rewards

bogdan-rosianu:


integrationTests/multiShard/block/executingRewardMiniblocks_test.go
```diff
@@ -297,7 +303,37 @@ func updateNumberTransactionsProposed(
...
+		}
+	}
+}
+
+func verifyRewardsForMetachain(
+	t *testing.T,
+	mapRewardsForMeta map[string]uint32,
+	nodes map[uint32][]*integrationTests.TestProcessorNode,
+) {
+
```
sasurobert: delete empty line

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
```diff
@@ -297,7 +303,37 @@ func updateNumberTransactionsProposed(
...
+	// TODO this should be read from protocol config
+	rewardValue := uint32(1000)
```
sasurobert: expose a function for test where you can change the reward value, than you can set anything.

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
```diff
@@ -297,7 +303,37 @@ func updateNumberTransactionsProposed(
...
+
+	for metaAddr, numOfTimesRewarded := range mapRewardsForMeta {
+		addrContainer, _ := integrationTests.TestAddressConverter.CreateAddressFromPublicKeyBytes([]byte(metaAddr))
+		acc, err := nodes[0][0].AccntState.GetExistingAccount(addrContainer)
+		assert.Nil(t, err)
+
+		expectedBalance := big.NewInt(int64(numOfTimesRewarded * rewardValue))
+		assert.Equal(t, expectedBalance, acc.(*state.Account).Balance)
```
sasurobert: test should be done on value, not on pointer. Use big.Int compare function

AdoAdoAdo: assert.Equal should be fine as it checks the value, otherwise also the test would fail because the expected balance is a new instance.

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
bogdan-rosianu: done

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
bogdan-rosianu: done

integrationTests/multiShard/block/executingRewardMiniblocks_test.go
bogdan-rosianu: agree. didn't change

PR #480, opened 2019-09-25
En 4102 improve fork detector

SebastianMarian:
**process/block**

* Added and refactored some print info which were confusing
* Added a new header request by hash on process block, if the new block does not pass the check block validity because of its previous hash
* Replaced the old method getOrderedHdrs with a new one getFinalityAttestingHeaders which is more efficient and solves one edge case, when requested final header has a higher round than current one at which the node reached with its bootstrap mechanism

**process/block/preprocess and process/coordinator**

* Removed some unused old code and refactored methods which were affected by it
* Changed the old mechanism which creates many miniblocks from the same sender to the same receiver (now the txs pool is sorted once per round)

**process/sync**

* Fixed an edge case in fork detector when nodes which are bootstrapping would not add blocks in theirs fork detector mechanism because of the too old round. Added unit tests for this.
* Added the same method which checks block validity in metaForkDetector, also in the shardForkDetector, to reject blocks which came from rounds older than last round (this fix would avoid intentionally attacks with valid blocks which are not broadcast by proposer fore some rounds)

### DIFF
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -112,28 +112,28 @@ func (bp *baseProcessor) checkBlockValidity(
 				return nil
 			}
 
-			log.Info(fmt.Sprintf("hash not match: local block hash is %s and node received block with previous hash %s\n",
+			log.Info(fmt.Sprintf("hash does not match: local block hash is %s and node received block with previous hash %s\n",
 				core.ToB64(chainHandler.GetGenesisHeaderHash()),
 				core.ToB64(headerHandler.GetPrevHash())))
 
 			return process.ErrBlockHashDoesNotMatch
 		}
 
-		log.Info(fmt.Sprintf("nonce not match: local block nonce is 0 and node received block with nonce %d\n",
+		log.Info(fmt.Sprintf("nonce does not match: local block nonce is 0 and node received block with nonce %d\n",
 			headerHandler.GetNonce()))
 
 		return process.ErrWrongNonceInBlock
 	}
 
 	if headerHandler.GetRound() <= currentBlockHeader.GetRound() {
-		log.Info(fmt.Sprintf("round not match: local block round is %d and node received block with round %d\n",
+		log.Info(fmt.Sprintf("round does not match: local block round is %d and node received block with round %d\n",
 			currentBlockHeader.GetRound(), headerHandler.GetRound()))
 
 		return process.ErrLowerRoundInBlock
 	}
 
 	if headerHandler.GetNonce() != currentBlockHeader.GetNonce()+1 {
-		log.Info(fmt.Sprintf("nonce not match: local block nonce is %d and node received block with nonce %d\n",
+		log.Info(fmt.Sprintf("nonce does not match: local block nonce is %d and node received block with nonce %d\n",
 			currentBlockHeader.GetNonce(), headerHandler.GetNonce()))
 
 		return process.ErrWrongNonceInBlock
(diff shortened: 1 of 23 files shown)


process/block/preprocess/transactions.go
```diff
@@ -424,8 +433,27 @@ func (txs *transactions) CreateAndProcessMiniBlock(sndShardId, dstShardId uint32
...
+	var (
+		err             error
+		orderedTxs      []*transaction.Transaction
+		orderedTxHashes [][]byte
+	)
+
 	timeBefore := time.Now()
-	orderedTxes, orderedTxHashes, err := SortTxByNonce(txStore)
+
+	txs.mutOrderedTxs.RLock()
```
sasurobert: this could be moved inside a separate method. It simply returns the orderedTxs, first time it orders, second time it just copies.

process/block/preprocess/transactions.go
```diff
@@ -434,22 +462,29 @@ func (txs *transactions) CreateAndProcessMiniBlock(sndShardId, dstShardId uint32
 	}
 
 	if !haveTime() {
-		log.Info(fmt.Sprintf("time is up after ordered %d txs in %v sec\n", len(orderedTxes), timeAfter.Sub(timeBefore).Seconds()))
+		log.Info(fmt.Sprintf("time is up after ordered %d txs in %v sec\n", len(orderedTxs), timeAfter.Sub(timeBefore).Seconds()))
 		return nil, process.ErrTimeIsOut
 	}
 
-	log.Debug(fmt.Sprintf("time elapsed to ordered %d txs: %v sec\n", len(orderedTxes), timeAfter.Sub(timeBefore).Seconds()))
+	defer func() {
```
sasurobert: no need for this, if taking the ordered txs is in another method.

process/block/preprocess/transactions.go
```diff
@@ -424,8 +433,27 @@ func (txs *transactions) CreateAndProcessMiniBlock(sndShardId, dstShardId uint32
 	strCache := process.ShardCacherIdentifier(sndShardId, dstShardId)
 	txStore := txs.txPool.ShardDataStore(strCache)
 
+	var (
```
AdoAdoAdo: for some reason I don't like this construction inside methods, can you add a 'var' for each of the variables?

process/block/preprocess/transactions.go
SebastianMarian: Yes, but when you return from this method yous should set the global variable with the local ones as the next time to just copy them and not to sort them again

process/sync/metaForkDetector.go
```diff
@@ -79,7 +79,13 @@ func (mfd *metaForkDetector) AddHeader(
 	return nil
 }
 
-func (mfd *metaForkDetector) checkMetaBlockValidity(header data.HeaderHandler) error {
+func (mfd *metaForkDetector) checkMetaBlockValidity(header data.HeaderHandler, state process.BlockHeaderState) error {
```
AdoAdoAdo: maybe change the name to something like shouldConsiderBlockInForkDetector, as the block might be valid, but we don't want to accept it because of the nonce

same for shard fork detector

process/sync/baseForkDetector_test.go
```diff
@@ -761,3 +761,67 @@ func TestBasicForkDetector_GetProbableHighestNonce(t *testing.T) {
 	hInfos = bfd.GetHeaders(3)
 	assert.Equal(t, uint64(3), bfd.GetProbableHighestNonce(hInfos))
 }
+
+func TestShardForkDetector_CheckShardBlockValidityShouldWork(t *testing.T) {
```
AdoAdoAdo: should work in the name not exactly valid for ErrLowerRoundInBlock

process/block/preprocess/transactions.go
SebastianMarian: Done

process/block/metablock.go
```diff
@@ -746,6 +758,54 @@ func (mp *metaProcessor) checkShardHeadersFinality(header *block.MetaBlock, high
...
+		hdr, ok := val.(*block.Header)
+		if !ok {
+			continue
+		}
+
+		if highestNonceHdrs[hdr.ShardId] == nil {
+			continue
+		}
+
+		if hdr.GetNonce() <= highestNonceHdrs[hdr.ShardId].GetNonce() ||
```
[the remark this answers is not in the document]

process/block/preprocess/transactions.go
SebastianMarian: Done

process/sync/metaForkDetector.go
SebastianMarian: Done

process/sync/baseForkDetector_test.go
SebastianMarian: Changed

process/block/baseProcess.go
```diff
@@ -198,38 +198,49 @@ func (bp *baseProcessor) isHdrConstructionValid(currHdr, prevHdr data.HeaderHand
 	// special case with genesis nonce - 0
 	if currHdr.GetNonce() == 0 {
 		if prevHdr.GetNonce() != 0 {
-			return process.ErrWrongNonceInBlock
+			return process.ErrWrongNonceInOtherChainBlock
```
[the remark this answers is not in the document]
SebastianMarian: Changed the error messages to generic ones and added shard id as additional info in these prints

process/block/baseProcess.go
```diff
@@ -198,38 +198,49 @@ func (bp *baseProcessor) isHdrConstructionValid(currHdr, prevHdr data.HeaderHand
...
-			return process.ErrRootStateMissmatch
+			return process.ErrRootStateDoesNotMatchInOtherChainBlock
 		}
 		return nil
 	}
 
 	//TODO: add verification if rand seed was correctly computed add other verification
 	//TODO: check here if the 2 header blocks were correctly signed and the consensus group was correctly elected
 	if prevHdr.GetRound() >= currHdr.GetRound() {
+		log.Info(fmt.Sprintf("round does not match in other chain: local block round is %d and node received block with round %d\n",
```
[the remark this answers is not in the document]
SebastianMarian: Changed the error messages to generic ones and added shard id as additional info in these prints

PR #482, opened 2019-09-25
EN-4112: Termui bug fixes

bogdan-rosianu:
Fixed bugs:
- [invalid key] appeared instead of public key for signing
- log list did not update when resizing
- random characters which appeared would have remained so a periodic cleanup was added

Please test on all operating systems

### DIFF
--- a/statusHandler/view/termuic/termuiConsole.go
+++ b/statusHandler/view/termuic/termuiConsole.go
@@ -3,6 +3,8 @@ package termuic
 import (
 	"os"
 	"os/signal"
+	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -16,6 +18,10 @@ import (
 //refreshInterval is used for a ticker that refresh termui console at a specific interval
 const refreshInterval = time.Second
 
+// numOfTicksBeforeRedrawing represents the number of ticks which have to pass until a fake resize will be made
+// in order to clean the unwanted appeared characters
+const numOfTicksBeforeRedrawing = 10
+
 var log = logger.DefaultLogger()
 
 // TermuiConsole data where is store data from handler
@@ -23,6 +29,7 @@ type TermuiConsole struct {
 	presenter     view.Presenter
 	consoleRender TermuiRender
 	grid          *termuiRenders.DrawableContainer
+	mutRefresh    *sync.RWMutex
 }
 
 //NewTermuiConsole method is used to return a new TermuiConsole structure
@@ -32,7 +39,8 @@ func NewTermuiConsole(presenter view.Presenter) (*TermuiConsole, error) {
 	}
 
 	tc := TermuiConsole{
-		presenter: presenter,
+		presenter:  presenter,
+		mutRefresh: &sync.RWMutex{},
 	}
 
 	return &tc, nil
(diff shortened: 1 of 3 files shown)


statusHandler/view/termuic/termuiConsole.go
```diff
@@ -95,14 +101,43 @@ func (tc *TermuiConsole) eventLoop() {
...
 		ui.Close()
 		StopApplication()
 		return
 	}
 }
+
+func (tc *TermuiConsole) doChanges(counter *uint32) {
+	atomic.AddUint32(counter, 1)
+	if atomic.LoadUint32(counter) > numOfTicksBeforeRedrawing {
+		tc.doFakeResize()
```
miiu96: Maybe you dont need method doFakeResize. I think its ok just call method doResize, and maybe you can change name of method doChange in doFakeResizeIfNeeded.

statusHandler/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -273,10 +273,15 @@ func (wr *WidgetsRender) prepareListWithLogsForDisplay() {
 	wr.lLog.WrapText = true
 }
 
-func (wr *WidgetsRender) prepareLogLines(logData []string, size int) []string {
+func (wr *WidgetsRender) prepareLogLines(logData []string, maxSize int) []string {
 	logDataLen := len(logData)
-	if logDataLen > size {
-		return logData[logDataLen-size : logDataLen]
+	maxSize = maxSize - 2 // decrease 2 units as the total maxSize of the log list includes also the header and the footer
+	if maxSize < 0 {
```
miiu96: You can change condition here.  maxSize <= 0 you can  return  an empty slice of strings (return []string{}). Because when maxSize its equals with 0 you will return logData[(logDataLen - 0):logDataLen] and this means and empty slice.

bogdan-rosianu: done

statusHandler/view/termuic/termuiConsole.go
bogdan-rosianu: remove that method. Related to the name of `doChanges` I would leave it like this for now because if I would name it `doFakeResizeIfNeeded` the name of the method would shadow the event that happens more often, the refresh

statusHandler/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -273,10 +273,15 @@ func (wr *WidgetsRender) prepareListWithLogsForDisplay() {
 	wr.lLog.WrapText = true
 }
 
-func (wr *WidgetsRender) prepareLogLines(logData []string, size int) []string {
+func (wr *WidgetsRender) prepareLogLines(logData []string, maxSize int) []string {
```
SebastianMarian: I would keep here the variable name "size" instead "maxSize", and I would modify line 278 in maxSize := size - 2

statusHandler/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -273,10 +273,15 @@ func (wr *WidgetsRender) prepareListWithLogsForDisplay() {
...
 	logDataLen := len(logData)
-	if logDataLen > size {
-		return logData[logDataLen-size : logDataLen]
+	maxSize = maxSize - 2 // decrease 2 units as the total maxSize of the log list includes also the header and the footer
+	if maxSize <= 0 {
+		return []string{} // there isn't place for any log line
+	}
+
+	if logDataLen > maxSize {
+		return logData[(logDataLen - maxSize):logDataLen]
```
SebastianMarian: Are you sure that this is correct? If logDataLen is higher than maxSize and you want to keep only the last maxSize characters from logData, then the correct format of this line would be: return logData[(logDataLen - maxSize):maxSize] or more simple
return logData[(logDataLen - maxSize):],  as the first parameter of this slice is the index from which we want to copy and the second one is the numbers of characters we want to copy from that index (including the character placed on that index)

statusHandler/view/termuic/termuiRenders/widgetsRender.go
bogdan-rosianu: done

statusHandler/view/termuic/termuiRenders/widgetsRender.go
bogdan-rosianu: done with `logData[(logDataLen - maxSize):]`

PR #486, opened 2019-09-26
EN-4155-Fix-bug-with-processed-mini-blocks-from-meta-blocks

SebastianMarian:


process/sync/shardForkDetector.go
```diff
@@ -82,8 +82,8 @@ func (sfd *shardForkDetector) AddHeader(
 func (sfd *shardForkDetector) addFinalHeaders(finalHeaders []data.HeaderHandler, finalHeadersHashes [][]byte) {
 	finalCheckpointWasSet := false
 	for i := 0; i < len(finalHeaders); i++ {
-		isFinalHeaderNonceHigherThanCurrent := finalHeaders[i].GetNonce() > sfd.GetHighestFinalBlockNonce()
-		if isFinalHeaderNonceHigherThanCurrent {
+		isFinalHeaderNonceHigherThanGenesis := finalHeaders[i].GetNonce() > process.GenesisBlockNonce
+		if isFinalHeaderNonceHigherThanGenesis {
 			if !finalCheckpointWasSet {
 				sfd.setFinalCheckpoint(&checkpointInfo{nonce: finalHeaders[i].GetNonce(), round: finalHeaders[i].GetRound()})
```
sasurobert: It does not make sense to set a final checkpoint with nonce lower than current final nonce.

process/block/shardblock.go
```diff
@@ -1588,3 +1595,59 @@ func (sp *shardProcessor) IsInterfaceNil() bool {
 	}
 	return false
 }
+
+func (sp *shardProcessor) addProcessedMiniBlock(metaBlockHash []byte, miniBlockHash []byte) {
```
sasurobert: should be moved into a separate components - like MiniBlocksVerifier / Processor

process/block/shardblock.go
```diff
@@ -36,6 +37,9 @@ type shardProcessor struct {
 	currHighestMetaHdrNonce    uint64
 	allNeededMetaHdrsFound     bool
 
+	processedMiniBlocks    map[string][][]byte
```
sasurobert: you could do map[string]map[string]bool - where you put in the second map the hashes, it is faster for constant searching

sasurobert: and you have to clean it as well. as it would consume a lot of memory. use lru

process/block/shardblock.go
```diff
@@ -1588,3 +1595,59 @@ func (sp *shardProcessor) IsInterfaceNil() bool {
...
+	if !ok {
+		return
+	}
+
+	for _, hash := range miniBlockHashes {
+		if bytes.Equal(hash, miniBlockHash) {
+			continue
+		}
+
+		preservedMiniBlockHashes = append(preservedMiniBlockHashes, hash)
```
AdoAdoAdo: I think you can still have some race here.

process/block/shardblock.go
```diff
@@ -1588,3 +1595,59 @@ func (sp *shardProcessor) IsInterfaceNil() bool {
...
+func (sp *shardProcessor) isMiniBlockProcessed(metaBlockHash []byte, miniBlockHash []byte) bool {
+	sp.mutProcessedMiniBlocks.RLock()
+	miniBlockHashes, ok := sp.processedMiniBlocks[string(metaBlockHash)]
+	sp.mutProcessedMiniBlocks.RUnlock()
+
+	if !ok {
+		return false
+	}
+
+	for _, hash := range miniBlockHashes {
```
AdoAdoAdo: you might still have some race here.

process/sync/shardForkDetector.go
SebastianMarian: There would be no such case, but with the old code there was a problem when a node rolled back and reapply a final header list. Each time the final header will increase with one because of the check.

process/block/shardblock.go
SebastianMarian: Refactored and added remove action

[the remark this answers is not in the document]

process/block/shardblock.go
SebastianMarian: Resolved them

process/block/shardblock.go
SebastianMarian: Resolved them

process/block/shardblock.go
SebastianMarian: I would make this in other task. Added one task in Jira for this (EN-4168)

PR #488, opened 2019-09-27
Trailing comma in cross block height string

bogdan-rosianu:
- added a trailing comma in the string representing the blocks heights notarized by metachain for all shards. This will help fetching the data further

### DIFF
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -598,6 +598,9 @@ func (mp *metaProcessor) saveMetricCrossCheckBlockHeight() {
 
 		crossCheckBlockHeight += fmt.Sprintf("%d: %d", i, valueStored)
 	}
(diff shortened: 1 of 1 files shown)


process/block/metablock.go
```diff
@@ -598,6 +598,9 @@ func (mp *metaProcessor) saveMetricCrossCheckBlockHeight() {
 
 		crossCheckBlockHeight += fmt.Sprintf("%d: %d", i, valueStored)
```
iulianpascalau: Better:
remove 595-597 lines
replace this with:
```GO
crossCheckBlockHeight += fmt.Sprintf("%d: %d, ", i, valueStored)
```

bogdan-rosianu: done

PR #489, opened 2019-09-27
EN-4178: StorageUnit's Put() method fix

bogdan-rosianu:


storage/storageUnit/storageunit.go
```diff
@@ -172,6 +166,7 @@ func (s *Unit) Has(key []byte) error {
 // it updates the cache either way
 // it returns if the value was originally found
 func (s *Unit) HasOrAdd(key []byte, value []byte) error {
+	// TODO : refactor this method as not all edge cases seem to be treated
```
sasurobert: more explanation is needed in this TODO as it does not say anything.

iulianpascalau: I rather delete this func from interface and impls and associated tests if it is not really used in production code (and might never be used)

bogdan-rosianu: I will delete it. The problem is : inside the `if` for the bloom filter if the key exists in the persister will return, without writing to the cache

PR #494, opened 2019-09-30
En 4184 fix print info and throttle size

SebastianMarian:


core/computers.go
```diff
@@ -12,6 +12,14 @@ func Max(a uint32, b uint32) uint32 {
 	return b
 }
 
+// Min returns the minimum number between two given
+func Min(a uint32, b uint32) uint32 {
```
ccorcoveanu: Isn't there a `Min` function in the math package?

process/block/preprocess/transactions.go
```diff
@@ -480,6 +479,9 @@ func (txs *transactions) CreateAndProcessMiniBlock(
 		}
 
 		if addedGasLimitPerCrossShardMiniblock+currTxGasLimit > process.MaxGasLimitPerMiniBlock {
```
ccorcoveanu: Can we extract this in a variable or a function returning a boolean with a descriptive name?

process/sync/shardForkDetector.go
```diff
@@ -82,8 +82,8 @@ func (sfd *shardForkDetector) AddHeader(
 func (sfd *shardForkDetector) addFinalHeaders(finalHeaders []data.HeaderHandler, finalHeadersHashes [][]byte) {
 	finalCheckpointWasSet := false
 	for i := 0; i < len(finalHeaders); i++ {
-		isFinalHeaderNonceHigherThanGenesis := finalHeaders[i].GetNonce() > process.GenesisBlockNonce
-		if isFinalHeaderNonceHigherThanGenesis {
+		isFinalHeaderNonceNotLowerThanCurrent := finalHeaders[i].GetNonce() >= sfd.finalCheckpoint().nonce
```
ccorcoveanu: Can you please explain why this was needed? And maybe add a unit test to cover this change?

process/sync/baseForkDetector.go
```diff
@@ -185,10 +185,8 @@ func (bfd *baseForkDetector) RemoveHeaders(nonce uint64, hash []byte) {
 
 	var preservedHdrInfos []*headerInfo
 
-	bfd.mutHeaders.RLock()
+	bfd.mutHeaders.Lock()
```
iulianpascalau: Good catch :+1:

core/computers.go
SebastianMarian: Yes it is but only for float type, and I should cast the values from int to float which is more expensive in terms of speed i gas

process/block/preprocess/transactions.go
SebastianMarian: Done

process/sync/shardForkDetector.go
SebastianMarian: This was an old check but it was incomplete. As the finalHeader should always increase because these headers come from meta blocks and in the normal cases they would always be higher than the last ones. But one problem could appear in a roll back situation, when the final header could be set with a lower value despite the fact that this action would not be permitted.

core/computers.go
SebastianMarian: Yes it is but only for float type, and I should cast the values from int to float which is more expensive in terms of speed i ga

SebastianMarian: I added in Jira task EN-4194 for an integration test for this case

process/sync/shardForkDetector.go
SebastianMarian: I added in Jira task EN-4194 for an integration test for this case

SebastianMarian: Added unit test

process/block/preprocess/intermediateResults.go
```diff
@@ -93,7 +94,13 @@ func (irp *intermediateResultsProcessor) CreateAllInterMiniBlocks() map[uint32]*
 				return bytes.Compare(miniBlocks[i].TxHashes[a], miniBlocks[i].TxHashes[b]) < 0
 			})
 
+			mbSpaceRemained := int(maxMbSpaceRemained) - mbAdded
```
sasurobert: in case of intermediate processor you should add all the created miniblocks. otherwise it will break the whole logic of smart contract results and rewardProcessor. Everything which was created during this round, has to be added here, no exception. The next round you do not know about this data.

process/block/preprocess/transactions.go
```diff
@@ -524,16 +529,24 @@ func (txs *transactions) computeOrderedTxs(
...
-	txStore := txs.txPool.ShardDataStore(strCache)
+	txShardPool := txs.txPool.ShardDataStore(strCache)
 
 	txs.mutOrderedTxs.RLock()
 	orderedTxs := txs.orderedTxs[strCache]
 	orderedTxHashes := txs.orderedTxHashes[strCache]
 	txs.mutOrderedTxs.RUnlock()
 
+	if txShardPool == nil || txShardPool.Len() == 0 {
+		return orderedTxs, orderedTxHashes, err
```
sasurobert: return nil here for err

process/block/preprocess/intermediateResults.go
SebastianMarian: Fixed

core/computers.go
ccorcoveanu: Just asking, would it be a good idea to name them MinUint32 in case we need it for other types?

process/block/preprocess/transactions.go
SebastianMarian: Done

core/computers.go
SebastianMarian: Yes, I will do this in my next pull request which is already in review

PR #499, opened 2019-10-01
EN-4207 : fix tests that fail with race condition

miiu96:
fix failing tests


process/common_test.go
```diff
@@ -39,16 +40,16 @@ func TestEmptyChannelShouldWorkOnNotBufferedChannel(t *testing.T) {
...
-	assert.Equal(t, 0, readsCnt)
+	assert.Equal(t, int32(0), readsCnt)
 
 	wg := sync.WaitGroup{}
 	wgChanWasWritten := sync.WaitGroup{}
-	numConcurrentWrites := 100
-	wg.Add(numConcurrentWrites)
-	wgChanWasWritten.Add(numConcurrentWrites)
-	for i := 0; i < numConcurrentWrites; i++ {
+	numConcurrentWrites := int32(100)
```
iulianpascalau: unnecessary cast to int32, should have had the cast only in assert func (less lines modified)

miiu96: Done.

PR #506, opened 2019-10-02
EN-4222 : refactor MetaProcessor move all constructor parameters in a…

miiu96:
- Refactor MetaProcessor 
  - Move all constructor parameters in a argument struct.
  - Modify all tests that fail after that modification.

- Remove TODO with unused parameter nodeConfig.


cmd/node/factory/structs.go
```diff
@@ -1753,21 +1748,26 @@ func newMetaBlockProcessorAndTracker(
...
+		Store:                 data.Store,
+		ShardCoordinator:      shardCoordinator,
+		NodesCoordinator:      nodesCoordinator,
+		SpecialAddressHandler: specialAddressHandler,
+		Uint64Converter:       core.Uint64ByteSliceConverter,
+		StartHeaders:          shardsGenesisBlocks,
+		RequestHandler:        requestHandler,
+		Core:                  coreServiceContainer,
+	}
+	arguments := block.ArgMetaProcessor{
```
ccorcoveanu: I don't thing there should be a discrepancy between how we pass the `ArgBaseProcessor` and `ArgMetaProcessor`. What I mean by this, is that `ArgMetaProcessor` is passed to the constructor by value, and `ArgBaseProcessor` is a pointer. I would use the same approach for both, either pointer, either value. Since they are structures containing only interface members and pointers, I would pass them both by value

miiu96: Done. Now that structures are passed by value

process/block/metablock.go
```diff
@@ -44,43 +39,29 @@ type metaProcessor struct {
...
-	hasher hashing.Hasher,
-	marshalizer marshal.Marshalizer,
-	store dataRetriever.StorageService,
-	startHeaders map[uint32]data.HeaderHandler,
-	requestHandler process.RequestHandler,
-	uint64Converter typeConverters.Uint64ByteSliceConverter,
-) (*metaProcessor, error) {
+func NewMetaProcessor(arguments ArgMetaProcessor) (*metaProcessor, error) {
 
 	err := checkProcessorNilParameters(
```
bogdan-rosianu: you can change this function to send the entire structure to it and test the parameters used in shardblock and metablock.

process/block/metablock_test.go
```diff
@@ -20,6 +20,29 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func CreateMockMetaArguments() blproc.ArgMetaProcessor {
```
bogdan-rosianu: this func can be made unexported

process/block/metablock.go
miiu96: Done.

process/block/metablock_test.go
miiu96: Done.

process/block/metablock.go
```diff
@@ -20,16 +20,12 @@ import (
...
-	dataPool                    dataRetriever.MetaPoolsHolder
-	currHighestShardHdrsNonces  map[uint32]uint64
-	requestedShardHdrsHashes    map[string]bool
-	allNeededShardHdrsFound     bool
-	mutRequestedShardHdrsHashes sync.RWMutex
-	shardsHeadersNonce          *sync.Map
-	nextKValidity               uint32
-	chRcvAllHdrs                chan bool
-	headersCounter              *headersCounter
+	core     serviceContainer.Core
```
ccorcoveanu: Can you please reformat the code?

process/block/shardblock.go
```diff
@@ -24,16 +24,15 @@ const maxCleanTime = time.Second
 // shardProcessor implements shardProcessor interface and actually it tries to execute block
 type shardProcessor struct {
```
ccorcoveanu: Can you please reformat the code in this struct also?

process/block/metablock.go
miiu96: Done.

process/block/shardblock.go
miiu96: Done.

PR #507, opened 2019-10-02
En 4217 bugfixing tx flood

iulianpascalau:
bugfixing transactions flood:
1. added an allowed window on nonce for received txs
2. added a throttler on tx interceptor
3. increased "seen" messages var from pubsub to 10 minutes


process/transaction/interceptor.go
```diff
@@ -81,10 +87,17 @@ func NewTxInterceptor(
 // ProcessReceivedMessage will be the callback func from the p2p.Messenger and will be called each time a new message was received
 // (for the topic this validator was registered to)
 func (txi *TxInterceptor) ProcessReceivedMessage(message p2p.MessageP2P) error {
+	canProcess := txi.throttler.CanProcess()
```
AdoAdoAdo: todo: check if throttler needs to be applied also on the processtransaction go routine.

iulianpascalau: Added todo near the go routine

process/block/shardblock.go
```diff
@@ -251,6 +251,11 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	log.Info(fmt.Sprintf("processed metaBlock %d out of %d", len(processedMetaHdrs), len(header.MetaBlockHashes)))
```
SebastianMarian: Remove these prints which were uses only for quick debugging purpose

process/block/shardblock.go
```diff
@@ -1569,6 +1574,11 @@ func (sp *shardProcessor) createMiniBlocks(
 		log.Debug(errNotCritical.Error())
 	}
 
+	log.Info(fmt.Sprintf("processed metaBlock %d out of %d", len(processedMetaHdrs), len(usedMetaHdrsHashes)))
```
SebastianMarian: Remove these prints which were uses only for quick debugging purpose

process/block/shardblock.go
AdoAdoAdo: you can remove these prints now

process/block/shardblock.go
AdoAdoAdo: you can remove these prints now.

process/interface.go
```diff
@@ -399,6 +399,14 @@ type PoolsCleaner interface {
 	IsInterfaceNil() bool
 }
 
+// InterceptorThrottler can determine if the a new joc can or cannot be started
```
SebastianMarian: if the a new joc?

process/block/shardblock.go
iulianpascalau: done

process/block/shardblock.go
iulianpascalau: done

process/block/shardblock.go
iulianpascalau: done

process/block/shardblock.go
iulianpascalau: done

process/interface.go
iulianpascalau: Rephrased

PR #512, opened 2019-10-04
EN-4247 : add timestamp in round info

miiu96:
add Timestamp in roundInfo to save sorted data


process/block/shardblock.go
```diff
@@ -545,7 +552,14 @@ func (sp *shardProcessor) indexBlockIfNeeded(
 			continue
 		}
 		signersIndexes = sp.nodesCoordinator.GetValidatorsIndexes(publicKeys)
-		go sp.core.Indexer().SaveRoundInfo(int64(i), shardId, signersIndexes, true)
+		roundInfo = indexer.RoundInfo{
+			SignersIndexes:   signersIndexes,
+			BlockWasProposed: true,
+			ShardId:          shardId,
+			Timestamp:        time.Duration(header.GetTimeStamp() - (currentBlockRound - i)),
```
AdoAdoAdo: I would add the correct round time by computing the round duration and round time from current and previous block  header. These should normally come from rounder, but seems it is not available to block processor.

Maybe add a TODO to inject the rounder interface to block processor

consensus/spos/commonSubround/subroundStartRound.go
```diff
@@ -201,7 +201,15 @@ func (sr *SubroundStartRound) indexRoundIfNeeded(pubKeys []string) {
...
-	go sr.indexer.SaveRoundInfo(round, shardId, signersIndexes, false)
+
+	roundInfo := indexer.RoundInfo{
+		SignersIndexes:   signersIndexes,
+		BlockWasProposed: false,
+		ShardId:          shardId,
+		Timestamp:        time.Duration(sr.RoundTimeStamp.Unix()),
+	}
+
+	go sr.indexer.SaveRoundInfo(round, roundInfo)
```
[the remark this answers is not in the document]
miiu96: Done. First parameter was added inside the structure.

process/block/shardblock.go
miiu96: Done.

core/indexer/elasticsearch.go
```diff
@@ -261,7 +255,7 @@ func (ei *elasticIndexer) SaveRoundInfo(round int64, shardId uint32, signersInde
 
 	req := esapi.IndexRequest{
 		Index:      roundIndex,
-		DocumentID: strconv.Itoa(int(round)),
+		DocumentID: strconv.Itoa(int(roundInfo.Index)),
```
ccorcoveanu: You can use `strconv.FormatInt(roundInfo.Index, 10)`, this way we prevent errors if the index is larger than 32 bits

miiu96: Done

PR #514, opened 2019-10-07
En 4246 trie snapshot functionality

BeniaminDrasovean:
Every time the snapshot method of the trie is called, it will create a new database that contains the current state of the trie.


data/trie/node_test.go
```diff
@@ -640,3 +646,144 @@ func TestTrieAddHashesToOldHashes(t *testing.T) {
...
+		snapshotDbCfg: cfg,
+		marshalizer:   msh,
+		hasher:        hsh,
+	}
+
+	for _, testVal := range testVals {
+		_ = tr.Update(testVal.key, testVal.value)
+		_ = tr.Snapshot()
+
+		snapshotId := strconv.Itoa(tr.snapshotId - 1)
```
bogdan-rosianu: maybe `fmt.Sprintf()` instead of `strconv` ?

data/trie/node_test.go
```diff
@@ -640,3 +646,144 @@ func TestTrieAddHashesToOldHashes(t *testing.T) {
...
+		snapshotPath := path.Join(tr.snapshotDbCfg.FilePath, snapshotId)
+		f, _ := os.Stat(snapshotPath)
+		assert.True(t, f.IsDir())
+	}
+
+	assert.Equal(t, 3, tr.snapshotId)
```
bogdan-rosianu: use a length of testVals instead of magic number `3`

data/trie/patriciaMerkleTrie.go
```diff
@@ -400,7 +428,45 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
+
+	err = tr.root.commit(true, 0, db, tr.marshalizer, tr.hasher)
+	if err != nil {
+		return err
+	}
+
+	tr.snapshots = append(tr.snapshots, db)
+
+	if len(tr.snapshots) > maxSnapshots {
+		dbUniqueId := strconv.Itoa(tr.snapshotId - len(tr.snapshots))
```
bogdan-rosianu: `fmt.Sprintf()` ?

cmd/node/config/config.toml
```diff
@@ -147,6 +147,13 @@
         MaxBatchSize = 1
         MaxOpenFiles = 10
 
+[TrieSnapshotDB]
+     FilePath = "TrieSnapshot"
+     Type = "LvlDBSerial"
+     BatchDelaySeconds = 2
```
sasurobert: should not be more than a round ? same question for evistion waitinglist

cmd/node/factory/structs.go
```diff
@@ -673,7 +687,9 @@ func getTrie(cfg config.StorageConfig, evictionWaitingListCfg config.EvictionWai
 		return nil, errors.New("error creating evictionDb: " + err.Error())
 	}
 
-	return trie.NewTrie(accountsTrieStorage, marshalizer, hasher, evictionDb, int(evictionWaitingListCfg.Size))
+	snapshotDbCfg.FilePath = filepath.Join(uniqueID, snapshotDbCfg.FilePath)
+
+	return trie.NewTrie(accountsTrieStorage, marshalizer, hasher, evictionDb, int(evictionWaitingListCfg.Size), snapshotDbCfg)
```
sasurobert: line too long.

data/mock/trieStub.go
```diff
@@ -135,3 +136,11 @@ func (ts *TrieStub) AppendToOldHashes(hashes [][]byte) {
 		ts.AppendToOldHashesCalled(hashes)
 	}
 }
+
+func (ts *TrieStub) Snapshot() error {
+	if ts.SnapshotCalled != nil {
+		return ts.SnapshotCalled()
+	}
+
```
sasurobert: delete empty line. In a case of stub return nil.

data/trie/patriciaMerkleTrie.go
```diff
@@ -40,6 +50,7 @@ func NewTrie(
 	hsh hashing.Hasher,
 	evictionDb storage.Persister,
 	evictionWaitListSize int,
+	snapshotDbCfg config.DBConfig,
```
sasurobert: Config parsing should happen on main / struct.go - this constructor should get the result of the config - like the actual file in this case.

data/trie/patriciaMerkleTrie.go
```diff
@@ -400,7 +428,45 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
+func (tr *patriciaMerkleTrie) Snapshot() error {
+	tr.mutOperation.Lock()
+	defer tr.mutOperation.Unlock()
+
+	err := tr.root.commit(false, 0, tr.db, tr.marshalizer, tr.hasher)
+	if err != nil {
+		return err
+	}
+
+	db, err := storageUnit.NewDB(
```
sasurobert: these files could be precreated and you could do complete delete and commit on them.

data/trie/branchNode.go
```diff
@@ -244,18 +244,18 @@ func (bn *branchNode) hashNode(marshalizer marshal.Marshalizer, hasher hashing.H
 	return encodeNodeAndGetHash(bn, marshalizer, hasher)
 }
 
-func (bn *branchNode) commit(level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
+func (bn *branchNode) commit(force bool, level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
```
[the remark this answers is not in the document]
BeniaminDrasovean: I agree, it's not good to use boolean parameters. But in this case, we have recursion, and not using a bool parameter will result in a lot of duplicated code. Also, this func is unexported.

data/trie/extensionNode.go
```diff
@@ -165,17 +165,17 @@ func (en *extensionNode) hashNode(marshalizer marshal.Marshalizer, hasher hashin
...
 
-func (en *extensionNode) commit(level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
+func (en *extensionNode) commit(force bool, level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
 	level++
 	err := en.isEmptyOrNil()
 	if err != nil {
 		return err
 	}
-	if !en.dirty {
+	if !en.dirty && !force {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Extrected this in a variable named 'shouldNotCommit' to make things more clearer.

data/trie/leafNode.go
```diff
@@ -120,12 +120,12 @@ func (ln *leafNode) hashNode(marshalizer marshal.Marshalizer, hasher hashing.Has
...
 }
 
-func (ln *leafNode) commit(level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
+func (ln *leafNode) commit(force bool, level byte, db data.DBWriteCacher, marshalizer marshal.Marshalizer, hasher hashing.Hasher) error {
 	err := ln.isEmptyOrNil()
 	if err != nil {
 		return err
 	}
-	if !ln.dirty {
+	if !ln.dirty && !force {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Extrected this in a variable named 'shouldNotCommit' to make things clearer.

data/trie/node_test.go
BeniaminDrasovean: Just benchmarked `fmt.Sprintf()` and `strconv`. `fmt.Sprintf()` is 3 times more slower than `strconv`.

data/trie/node_test.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: `fmt.Sprintf()` is slower than `strconv`

cmd/node/factory/structs.go
BeniaminDrasovean: Refactored.

cmd/node/config/config.toml
BeniaminDrasovean: Updated to 15. Updated the eviction waiting list to BatchDelaySeconds = 5. The waiting list needs to be stored in a database as soon as possible, so the timeframe of a round should be ok.

data/mock/trieStub.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: I can't pass a database to the constructor of the trie, because each snapshot is stored in a separate database. Each time snapshot is called, a new database needs to be created, so I need the config parameters to be able to create a new database.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: You mean another component that creates the database, and each time the snapshot function is called, a new database should be passed as a parameter?

data/trie/node_test.go
bogdan-rosianu: ok. good to know

data/trie/patriciaMerkleTrie.go
sasurobert: As you do not want to have infinite amount of snapshots, the snapshot files can be predetermined. So you can create N files, and write in those like in a sliding window. So you precreate N storageUnits - you could have a container for it, then get one out, change the snapshot there and that's it. You might not need a separate component, but a container with all the storageUnits for the tries might be sent in the constructor and not the config.

PR #525, opened 2019-10-09
EN-4291-Refactor-mini-blocks-request-from-sync

SebastianMarian:
* Fix problem on request missing mini blocks when syncing. Now the bootstrap is 10x faster

### DIFF
--- a/dataRetriever/resolvers/genericBlockBodyResolver.go
+++ b/dataRetriever/resolvers/genericBlockBodyResolver.go
@@ -83,7 +83,7 @@ func (gbbRes *GenericBlockBodyResolver) resolveBlockBodyRequest(rd *dataRetrieve
 		return nil, err
 	}
 
-	miniBlocks, _ := gbbRes.getMiniBlocks(hashes)
+	miniBlocks, _ := gbbRes.GetMiniBlocks(hashes)
 	if len(miniBlocks) == 0 {
 		return nil, dataRetriever.ErrEmptyMiniBlockSlice
 	}
@@ -141,17 +141,7 @@ func (gbbRes *GenericBlockBodyResolver) RequestDataFromHashArray(hashes [][]byte
 
 // GetMiniBlocks method returns a list of deserialized mini blocks from a given hash list either from data pool or from storage
 func (gbbRes *GenericBlockBodyResolver) GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
-	return gbbRes.getMiniBlocks(hashes)
-}
-
-// GetMiniBlocks method returns a list of deserialized mini blocks from a given hash list from data pool
-func (gbbRes *GenericBlockBodyResolver) GetMiniBlocksFromPool(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
-	return gbbRes.getMiniBlocksFromPool(hashes)
-}
-
-// getMiniBlocks method returns a list of serialized mini blocks from a given hash list either from data pool or from storage
-func (gbbRes *GenericBlockBodyResolver) getMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
-	miniBlocks, missingMiniBlocksHashes := gbbRes.getMiniBlocksFromPool(hashes)
+	miniBlocks, missingMiniBlocksHashes := gbbRes.GetMiniBlocksFromPool(hashes)
 	if len(missingMiniBlocksHashes) == 0 {
 		return miniBlocks, missingMiniBlocksHashes
 	}
@@ -162,8 +152,8 @@ func (gbbRes *GenericBlockBodyResolver) getMiniBlocks(hashes [][]byte) (block.Mi
 	return miniBlocks, missingMiniBlocksHashes
 }
 
-// getMiniBlocksFromPool returns a list of mini blocks from cache and a list of missing hashes
-func (gbbRes *GenericBlockBodyResolver) getMiniBlocksFromPool(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
+// GetMiniBlocksFromPool method returns a list of deserialized mini blocks from a given hash list from data pool
+func (gbbRes *GenericBlockBodyResolver) GetMiniBlocksFromPool(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
 	miniBlocks := make(block.MiniBlockSlice, 0)
 	missingMiniBlocksHashes := make([][]byte, 0)
 
--- a/dataRetriever/resolvers/genericBlockBodyResolver.go
+++ b/dataRetriever/resolvers/genericBlockBodyResolver.go
@@ -83,7 +83,7 @@ func (gbbRes *GenericBlockBodyResolver) resolveBlockBodyRequest(rd *dataRetrieve
 		return nil, err
 	}
 
-	miniBlocks, _ := gbbRes.GetMiniBlocks(hashes)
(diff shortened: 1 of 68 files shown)


dataRetriever/resolvers/genericBlockBodyResolver.go
```diff
@@ -141,89 +140,88 @@ func (gbbRes *GenericBlockBodyResolver) RequestDataFromHashArray(hashes [][]byte
...
-	miniBlocks := gbbRes.getMiniBlocksFromCache(hashes)
+func (gbbRes *GenericBlockBodyResolver) getMiniBlocks(hashes [][]byte) (map[string][]byte, [][]byte) {
+	marshalizedMiniBlocks, missingMiniBlocksHashes := gbbRes.getMiniBlocksFromCache(hashes)
+	if len(missingMiniBlocksHashes) == 0 {
+		return marshalizedMiniBlocks, missingMiniBlocksHashes
+	}
 
-	if miniBlocks != nil {
-		return miniBlocks
+	marshalizedMiniBlocksFromStorer, missingMiniBlocksHashes := gbbRes.getMiniBlocksFromStorer(missingMiniBlocksHashes)
```
sasurobert: why would some of the miniblocks be already in store ? At rollback you get everything from store and put in miniblock pool. Is there an other edgecase where an unprocessed miniblock might get saved into the store and not restored ?

dataRetriever/resolvers/genericBlockBodyResolver.go
```diff
@@ -141,89 +140,88 @@ func (gbbRes *GenericBlockBodyResolver) RequestDataFromHashArray(hashes [][]byte
...
 	}
 
-	return expandedMiniBlocks
+	return miniBlocks, missingMiniBlocksHashes
 }
 
 // getMiniBlocks method returns a list of serialized mini blocks from a given hash list either from data pool or from storage
-func (gbbRes *GenericBlockBodyResolver) getMiniBlocks(hashes [][]byte) [][]byte {
-	miniBlocks := gbbRes.getMiniBlocksFromCache(hashes)
+func (gbbRes *GenericBlockBodyResolver) getMiniBlocks(hashes [][]byte) (map[string][]byte, [][]byte) {
```
sasurobert: why not return here the already demarshalized miniblocks ? when taking from pool you are marshalizing, then on GetMiniBlocks function you unmarshal to the block. So return list of miniblocks here. You can unmarshal those from the storer in the getMiniBlocksFromStorer function.

dataRetriever/resolvers/genericBlockBodyResolver.go
```diff
@@ -141,89 +140,88 @@ func (gbbRes *GenericBlockBodyResolver) RequestDataFromHashArray(hashes [][]byte
...
-			return nil
+	for i := 0; i < len(hashes); i++ {
+		miniBlock, ok := gbbRes.miniBlockPool.Peek(hashes[i])
+		if !ok {
+			missingMiniBlocksHashes = append(missingMiniBlocksHashes, hashes[i])
+			continue
 		}
 
-		buff, err := gbbRes.marshalizer.Marshal(cachedMB)
+		buff, err := gbbRes.marshalizer.Marshal(miniBlock)
```
sasurobert: no need to marshal, send direct miniblock back.

dataRetriever/resolvers/genericBlockBodyResolver.go
```diff
@@ -141,89 +140,88 @@ func (gbbRes *GenericBlockBodyResolver) RequestDataFromHashArray(hashes [][]byte
...
 		buff, err := gbbRes.miniBlockStorage.Get(hashes[i])
 		if err != nil {
 			log.Debug(err.Error())
-			return nil
+			missingMiniBlocksHashes = append(missingMiniBlocksHashes, hashes[i])
+			continue
 		}
 
-		miniBlocks[i] = buff
+		marshalizedMiniBlocks[string(hashes[i])] = buff
```
sasurobert: unmarshal here and return miniblocks.

process/block/baseProcess.go
```diff
@@ -417,12 +414,15 @@ func (bp *baseProcessor) requestHeadersIfMissing(sortedHdrs []data.HeaderHandler
...
 		}
+
+		highestHdr = currHdr
 	}
 
 	// ask for headers, if there most probably should be
-	if len(missingNonces) == 0 && !isLastNotarizedCloseToOurRound {
-		startNonce := prevHdr.GetNonce() + 1
-		for nonce := startNonce; nonce < startNonce+process.MaxHeaderRequestsAllowed; nonce++ {
+	if maxRound > highestHdr.GetRound() {
```
sasurobert: this if is always true. I do not agree with the changes did in this function. Now missingNonces may have duplicates and you will ask for duplicates. keep only unique elements in missingNonces.

process/block/baseProcess.go
```diff
@@ -417,12 +414,15 @@ func (bp *baseProcessor) requestHeadersIfMissing(sortedHdrs []data.HeaderHandler
...
+		nbHeaderRequests := maxRound - highestHdr.GetRound()
```
sasurobert: +1 nice resolve.

process/sync/baseSync.go
```diff
@@ -562,3 +565,41 @@ func isRandomSeedValid(header data.HeaderHandler) bool {
 
 	return !isPrevRandSeedNilOrEmpty && !isRandSeedNilOrEmpty
 }
+
+func (boot *baseBootstrap) requestHeadersFromNonceIfMissing(nonce uint64, hdrRes dataRetriever.HeaderResolver) {
+	var err error
+	nbRequestedHdrs := 0
+	maxNonce := core.MinUint64(nonce+maxHeadersToRequestInAdvance-1, boot.forkDetector.ProbableHighestNonce())
+	for currentNonce := nonce; currentNonce <= maxNonce; currentNonce++ {
+		if boot.shardCoordinator.SelfId() == sharding.MetachainShardId {
```
sasurobert: I think it would be easier to understand it this verification is base on header type. ShardHeader / Metaheader.

dataRetriever/resolvers/genericBlockBodyResolver.go
SebastianMarian: This method is also called  from resolveBlockBodyRequest when someone else requested for a mini block. In this case, if the requested mini block is too old (when somebody is syncing) , the only place where it can still be found is in storer.

dataRetriever/resolvers/genericBlockBodyResolver.go
SebastianMarian: Good catch. Refactored

process/sync/baseSync.go
iulianpascalau: This func should be moved in the implementations (shard/meta) as to avoid this messy if.

process/sync/metablock.go
```diff
@@ -451,6 +451,9 @@ func (boot *MetaBootstrap) SyncBlock() error {
 		return err
 	}
 
+	// request in advance next meta header if missing
```
iulianpascalau: The best comment is the one that you do not write - Uncle Bob.

dataRetriever/resolvers/genericBlockBodyResolver.go
SebastianMarian: Refactored

dataRetriever/resolvers/genericBlockBodyResolver.go
SebastianMarian: Refactored

process/sync/baseSync.go

PR #529, opened 2019-10-10
En 4296 limit go routines num

BeniaminDrasovean:


p2p/libp2p/netMessenger.go
```diff
@@ -376,22 +388,34 @@ func (netMes *networkMessenger) OutgoingChannelLoadBalancer() p2p.ChannelLoadBal
 
 // BroadcastOnChannelBlocking tries to send a byte buffer onto a topic using provided channel
 // It is a blocking method. It needs to be launched on a go routine
-func (netMes *networkMessenger) BroadcastOnChannelBlocking(channel string, topic string, buff []byte) {
+func (netMes *networkMessenger) BroadcastOnChannelBlocking(channel string, topic string, buff []byte) error {
```
sasurobert: these errors are not used in any way, only to log in the components which call this. So why return it ? You could directly log it here, the error message and return simply. The same way it was before.

BeniaminDrasovean: The point of returning an error, was so that the function can be properly tested (see p2p/libp2p/netMessenger_test.go -> TestLibp2pMessenger_BroadcastOnChannelBlockingShouldLimitNumberOfGoRoutines). I can't deterministically test this function without the error returning.

PR #530, opened 2019-10-10
 EN-4313-Fix-bug-on-request-missing-headers-by-nonce

SebastianMarian:
* Fixed problem in SyncMapCacher when an empty syncMap for a nonce would not be removed from cache
* Refactored the name of the pool which hold meta blocks
* Fixed a problem in CommitBlock, when committed headers would not be removed from pools


process/block/metablock.go
```diff
@@ -864,9 +864,12 @@ func (mp *metaProcessor) requestMissingFinalityAttestingHeaders() uint32 {
 				mp.dataPool.HeadersNonces())
 
 			if err != nil {
-				requestedBlockHeaders++
-				go mp.onRequestHeaderHandlerByNonce(shardId, i)
-				continue
+				shardHeader, shardHeaderHash, err = mp.getShardHeaderWithNonce(shardId, i)
```
sasurobert: Is this an edge case when a header is evicted from the nonce pool, but it exists in the header pool ? It looks like an ugly hack, maybe the real error is somewhere else.

process/block/metablock.go
```diff
@@ -1294,3 +1297,28 @@ func (mp *metaProcessor) IsInterfaceNil() bool {
 	}
 	return false
 }
+
+func (mp *metaProcessor) getShardHeaderWithNonce(shardId uint32, nonce uint64) (*block.Header, []byte, error) {
```
AdoAdoAdo: this code is similar to getMetaHeaderWithNonce

it looks like we can move this function to baseProcess and maybe give as parameter the pool while returning a HeaderHandler.
I think we can then do the type assertion after call and have a single function to handle both.

process/block/metablock.go
SebastianMarian: Yes, this is the case because we use the same HeaderNonce pool either for meta headers and for shard headers and because of bootstrap this pool is filled with Shard Headers and Meta Headers are evicted after you bootstrap 1000 shar headers. Instead the MetaBlocks pool will still have them as it is dedicated only for meta blcoks, and in this time while you bootstrap 1000 shard headers you probably have around 20 new meta blocks in.

SebastianMarian: But the real problem I found it today and is because we don't evict shard headers committed from pool because of an old approach which was used to quickly response to requests. Now anyway the resolver is looking first in the storage(cahcer) so we don't nedd to keep them there. So I will replace this implementation today with some few lines of remove old code :).

process/block/metablock.go
SebastianMarian: I found the real problem, and I think I will remove this methods anyway

process/block/shardblock.go
```diff
@@ -1092,6 +1096,9 @@ func (sp *shardProcessor) receivedMetaBlock(metaBlockHash []byte) {
 	if metaBlock.GetRound() <= lastNotarizedHdr.GetRound() {
 		return
 	}
+	if metaBlock.GetNonce() > lastNotarizedHdr.GetNonce()+process.MaxHeadersToRequestInAdvance {
```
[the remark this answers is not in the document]

process/block/shardblock.go
```diff
@@ -1077,6 +1081,11 @@ func (sp *shardProcessor) receivedMetaBlock(metaBlockHash []byte) {
 		return
 	}
 
+	isMetaBlockHigherThanLastNotarized := metaBlock.GetNonce() > lastNotarizedHdr.GetNonce()+process.MaxHeadersToRequestInAdvance
```
AdoAdoAdo: maybe call it metaBlockOutOfRange ? 
isMetaBlockHigherThanLastNotarized is misleading, as if metablock has a higher nonce than last notarized but in the accepted range you will set this variable to false

PR #531, opened 2019-10-11
EN-4317: preferences config file

bogdan-rosianu:
- added a preferences config file which holds the specific preferences for each node, such as the node display name
- the reason is that when updating to other elrond-go version, the prefs.toml file won't be updated
- this file can also be used on more instances of the same owner


cmd/node/main.go
```diff
@@ -1037,6 +1051,15 @@ func loadEconomicsConfig(filepath string, log *logger.Logger) (*config.ConfigEco
 	return cfg, nil
 }
 
+func loadPreferencesConfig(filepath string, log *logger.Logger) (*config.ConfigPreferences, error) {
+	cfg := &config.ConfigPreferences{}
+	err := core.LoadTomlFile(cfg, filepath, log)
+	if err != nil {
+		return nil, err
+	}
+	return cfg, nil
```
sasurobert: add an empty line before.

PR #543, opened 2019-10-13
transaction fee calculation refactor:

iulianpascalau:
transaction fee calculation refactor:
- changed the fee handler so it can compute fees, required gas limits and such: behavior is now consistent between interceptor and transaction processor
- found and fixed the bug that made the cost of a transaction minTxGasLimit * minTxGasPrice regardless the tx data notarized
- found and fixed a bug that made the miniblock max gaslimit calculation based only on minGasLimit instead of the real gas limit of the transaction

### DIFF
--- a/data/rewardTx/rewardTx.go
+++ b/data/rewardTx/rewardTx.go
@@ -18,22 +18,22 @@ type RewardTx struct {
 }
 
 // Save saves the serialized data of a RewardTx into a stream through Capnp protocol
-func (scr *RewardTx) Save(w io.Writer) error {
+func (rtx *RewardTx) Save(w io.Writer) error {
 	seg := capn.NewBuffer(nil)
-	RewardTxGoToCapn(seg, scr)
+	RewardTxGoToCapn(seg, rtx)
 	_, err := seg.WriteTo(w)
 	return err
 }
 
 // Load loads the data from the stream into a RewardTx object through Capnp protocol
-func (scr *RewardTx) Load(r io.Reader) error {
+func (rtx *RewardTx) Load(r io.Reader) error {
 	capMsg, err := capn.ReadFromStream(r, nil)
 	if err != nil {
 		return err
 	}
 
 	z := capnp.ReadRootRewardTxCapn(capMsg)
-	RewardTxCapnToGo(z, scr)
+	RewardTxCapnToGo(z, rtx)
 	return nil
 }
 
--- a/data/rewardTx/rewardTx.go
+++ b/data/rewardTx/rewardTx.go
@@ -76,44 +76,44 @@ func RewardTxGoToCapn(seg *capn.Segment, src *RewardTx) capnp.RewardTxCapn {
 }
 
 // IsInterfaceNil verifies if underlying object is nil
-func (scr *RewardTx) IsInterfaceNil() bool {
-	return scr == nil
+func (rtx *RewardTx) IsInterfaceNil() bool {
+	return rtx == nil
 }
 
 // GetValue returns the value of the reward transaction
-func (scr *RewardTx) GetValue() *big.Int {
-	return scr.Value
+func (rtx *RewardTx) GetValue() *big.Int {
+	return rtx.Value
 }
 
 // GetData returns the data of the reward transaction
-func (scr *RewardTx) GetData() string {
+func (rtx *RewardTx) GetData() string {
 	return ""
 }
 
 // GetRecvAddress returns the receiver address from the reward transaction
-func (scr *RewardTx) GetRecvAddress() []byte {
-	return scr.RcvAddr
+func (rtx *RewardTx) GetRecvAddress() []byte {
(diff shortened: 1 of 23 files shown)


process/economics/economicsData.go
```diff
@@ -123,14 +123,48 @@ func (ed *EconomicsData) BurnPercentage() float64 {
...
+	if requiredGasLimit > tx.GetGasLimit() {
+		return process.ErrInsufficientGasLimitInTx
+	}
+
+	return nil
+}
+
+// ComputeGasLimit returns the gas limit need by the provided transaction in order to be executed
+func (ed *EconomicsData) ComputeGasLimit(tx process.TransactionWithFeeHandler) uint64 {
+	gasLimit := ed.minGasLimit
```
sasurobert: should not this be protected by mutex ?

process/interface.go
```diff
@@ -416,13 +416,23 @@ type RewardsHandler interface {
 	IsInterfaceNil() bool
 }
 
-// FeeHandler will return information about fees
+// FeeHandler is able to perform some economics calculation on a provided transaction
 type FeeHandler interface {
-	MinGasPrice() uint64
-	MinGasLimit() uint64
+	SetMinGasPrice(minGasPrice uint64)
+	SetMinGasLimit(minGasLimit uint64)
```
sasurobert: I think these settings should be done in the constructor. Not separate functions.

process/interface.go
```diff
@@ -416,13 +416,23 @@ type RewardsHandler interface {
...
+	ComputeGasLimit(tx TransactionWithFeeHandler) uint64
+	ComputeFee(tx TransactionWithFeeHandler) *big.Int
+	CheckTxHandler(tx TransactionWithFeeHandler) error
```
sasurobert: might rename the function, CheckTxValuesValidity or something similar ?

process/transaction/process.go
```diff
@@ -128,29 +125,18 @@ func (txProc *txProcessor) processTxFee(tx *transaction.Transaction, acntSnd *st
...
-	minTxFee.Mul(minTxFee, big.NewInt(0).SetUint64(txProc.economicsFee.MinGasPrice()))
-
-	minFee := big.NewInt(0)
-	minFee.Mul(big.NewInt(txDataLen), big.NewInt(0).SetUint64(txProc.economicsFee.MinGasPrice()))
-	minFee.Add(minFee, minTxFee)
-	txProc.mutTxFee.RUnlock()
-
-	if minFee.Cmp(cost) > 0 {
-		return nil, process.ErrNotEnoughFeeInTransactions
+	err := txProc.economicsFee.CheckTxHandler(tx)
```
sasurobert: rename the function name, as it does not say what is does in the background.

process/interface.go
sasurobert: I like simple interfaces. I prefer this without these 2 functions.

iulianpascalau: Done, added test wrapper struct

integrationTests/singleShard/transaction/interceptedResolvedTx_test.go
```diff
@@ -52,13 +52,17 @@ func TestNode_RequestInterceptTransactionWithMessenger(t *testing.T) {
 	valMinting := big.NewInt(100000)
 	integrationTests.CreateMintingForSenders([]*integrationTests.TestProcessorNode{nRequester}, 0, []crypto.PrivateKey{nRequester.OwnAccount.SkTxSign}, valMinting)
 	//Step 1. Generate a signed transaction
+	txData := "tx notarized data"
+	//TODO change here when gas limit will no longer be linear with the tx data length
+	txNotarizedDataGasLimit := uint64(len(txData))
```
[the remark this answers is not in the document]
iulianpascalau: Done

process/economics/economicsData.go
iulianpascalau: Not required, one write on the constructor and multiple reads throughout the struct. If we had multiple write points, race detector would have complained but still it would work as expected for uint64 values (maps complain on concurrent writes the most)

process/interface.go
iulianpascalau: CheckValidityTxValues

process/transaction/process.go
iulianpascalau: renamed to CheckValidityTxValues

data/rewardTx/rewardTx.go
```diff
@@ -76,44 +76,44 @@ func RewardTxGoToCapn(seg *capn.Segment, src *RewardTx) capnp.RewardTxCapn {
 }
 
 // IsInterfaceNil verifies if underlying object is nil
-func (scr *RewardTx) IsInterfaceNil() bool {
-	return scr == nil
+func (rtx *RewardTx) IsInterfaceNil() bool {
```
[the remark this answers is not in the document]
iulianpascalau: done

PR #550, opened 2019-10-15
En 4379-Cross-build

traian-schiau:


cmd/node/main.go
```diff
@@ -238,7 +238,7 @@ VERSION:
 		Value: logger.LogInfo,
 	}
 	// bootstrapRoundIndex defines a flag that specifies the round index from which node should bootstrap from storage
-	bootstrapRoundIndex = cli.UintFlag{
+	bootstrapRoundIndex = cli.Uint64Flag{
```
bogdan-rosianu: you can also change in main.go line 696 -> change `GlobalUint` to `GlobalUint64` and remove the cast

cross_build.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "Building $output_name"
+    if [ $GOOS = "windows" ]; then
+        output_name+='.exe'
+    fi
+
+    pushd cmd/node
+    env GOOS=$GOOS GOARCH=$GOARCH go build -o $output_name -a -i -v -ldflags="-X main.appVersion=$APP_VER"
+    popd
+    if [ $? -ne 0 ]; then
+        echo 'An error has occurred! '
```
bogdan-rosianu: you can also echo the `$?` so we can see the error code

cmd/node/main.go
traian-schiau: ok

cross_build.sh
traian-schiau: Will not help, is just the numeric error code 
+ the compile output is on stdout already

cross_build.sh
```diff
@@ -0,0 +1,28 @@
...
+    GOOS=${platform_split[0]}
+    GOARCH=${platform_split[1]}
+    output_name=$BASEDIR/$GOOS/$GOARCH/$package_name
+    echo "Building $output_name"
+    if [ $GOOS = "windows" ]; then
+        output_name+='.exe'
+    fi
+
+    pushd cmd/node
+    env GOOS=$GOOS GOARCH=$GOARCH go build -o $output_name -a -i -v -ldflags="-X main.appVersion=$APP_VER"
```
iulianpascalau: It seems this renders some "permission denied" errors as it tries to write in /usr/local/go/pkg dir. $GOPATH is somewhere in home.

traian-schiau: That's a go install/setup issue, in my opinion

PR #559, opened 2019-10-17
heartbeat race conditions fixed

raduchis:


node/heartbeat/monitor.go
```diff
@@ -169,21 +170,15 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P) error {
...
 
 func (m *Monitor) addHeartbeatMessageToMap(hb *Heartbeat) {
-	m.mutHeartbeatMessages.Lock()
-	defer m.mutHeartbeatMessages.Unlock()
-
 	pubKeyStr := string(hb.Pubkey)
+	m.mutHeartbeatMessages.Lock()
 	hbmi, ok := m.heartbeatMessages[pubKeyStr]
 	if hbmi == nil || !ok {
 		var err error
```
sasurobert: add m.mutheartbeatmessages.unlock to line 188

node/heartbeat/monitor.go
```diff
@@ -243,31 +245,33 @@ func (m *Monitor) computeShardID(pubkey string) uint32 {
 }
 
 func (m *Monitor) computeAllHeartbeatMessages() {
+	m.mutHeartbeatMessages.Lock()
```
iulianpascalau: I do not like nested mutex operations. They may lead to deadlocks. 
You can use updateMutex inside heartbeatMessageInfo instances (can move the lock/unlock operations there)

sasurobert: +1

node/heartbeat/monitor.go
raduchis: fixed

node/heartbeat/monitor.go
iulianpascalau: heartbeatMessageInfo.updateMutex should be used exclusively inside heartbeatMessageInfo struct. The monitor struct that uses heartbeatMessageInfo instances will need to know too much to protect some operations by a mutex.

raduchis: Fixed

PR #566, opened 2019-10-19
En 4565 fix problem with pool capacity

SebastianMarian:
* Fixed bug with multiple requests for the same finality attesting headers
* Fixed bug in requested/received reward transactions (negative values)
* Fixed bug with bad cast to MiniBlock instead *MiniBlock
* Fixed a problem when too many headers were requested in advance
* Improved the mechanism which requests miniblocks from meta blocks


process/block/baseProcess.go
```diff
@@ -454,6 +465,11 @@ func (bp *baseProcessor) requestHeadersIfMissing(sortedHdrs []data.HeaderHandler
 			return process.ErrNilRequestHeaderHandlerByNonce
 		}
 
+		isHeaderOutOfRange := nonce > lastNotarizedHdrNonce+allowedSize
```
sasurobert: sort missingNonces then you can request only the allowedSize

process/block/metablock.go
```diff
@@ -827,6 +826,11 @@ func (mp *metaProcessor) receivedShardHeader(shardHeaderHash []byte) {
 		mp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 	}
 
+	if mp.isHeaderOutOfRange(shardHeader, shardHeaderPool) {
```
sasurobert: this might result in errors. LIke mp.chRcvAllHdrs gets set to true, then you forcefully deleted a shardHeader.

process/block/shardblock.go
```diff
@@ -1070,6 +1069,11 @@ func (sp *shardProcessor) receivedMetaBlock(metaBlockHash []byte) {
 		sp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 	}
 
+	if sp.isHeaderOutOfRange(metaBlock, metaBlockPool) {
```
sasurobert: this might results in errors further on at processing. same situation as in metablock.

process/block/shardblock.go
```diff
@@ -1107,15 +1113,25 @@ func (sp *shardProcessor) requestMissingFinalityAttestingHeaders() uint32 {
 			sp.dataPool.HeadersNonces())
 
 		if err != nil {
-			requestedBlockHeaders++
-			go sp.onRequestHeaderHandlerByNonce(sharding.MetachainShardId, i)
+			missingFinalityAttestingHeaders++
```
sasurobert: you could create a function at baseProcess for lines 1110-1121 and call from here and metaprocessor as well.

process/block/shardblock.go
```diff
@@ -1107,15 +1113,25 @@ func (sp *shardProcessor) requestMissingFinalityAttestingHeaders() uint32 {
...
+			wasHeaderRequested := sp.wasHeaderRequested(sharding.MetachainShardId, i)
+			if !wasHeaderRequested {
+				requestedBlockHeaders++
+				sp.hdrsForCurrBlock.requestedFinalityAttestingHdrs[sharding.MetachainShardId] = append(sp.hdrsForCurrBlock.requestedFinalityAttestingHdrs[sharding.MetachainShardId], i)
```
AdoAdoAdo: you should protect with mutex

process/block/baseProcess.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/marshal"
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/ElrondNetwork/elrond-go/storage"
 )
```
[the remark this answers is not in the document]

process/block/baseProcess.go
```diff
@@ -617,3 +634,30 @@ func (bp *baseProcessor) sortHeaderHashesForCurrentBlockByNonce(usedInBlock bool
...
+	}
+
+	allowedSize := uint64(float64(cacher.MaxSize()) * process.MaxOccupancyPercentageAllowed)
+	isHeaderOutOfRange := header.GetNonce() > lastNotarizedHdr.GetNonce()+allowedSize
+
+	return isHeaderOutOfRange
+}
+
+func (bp *baseProcessor) wasHeaderRequested(shardId uint32, nonce uint64) bool {
+	requestedNonces, ok := bp.hdrsForCurrBlock.requestedFinalityAttestingHdrs[shardId]
```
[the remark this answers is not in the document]
SebastianMarian: The resource is already protected in upward methods from which this method is called:  metaBlock: requestMissingFinalityAttestingHeaders -> receivedShardHeader / requestShardHeaders
shardBlock: requestMissingFinalityAttestingHeaders -> receivedMetaHeader / requestMetaHeaders

process/block/baseProcess.go
SebastianMarian: They are already sorted and the allowedSize is used to request only the maximum occupancy percentage allowed to be used from the full pool capacity, for the received data which are not needed in the near future.

process/block/metablock.go
SebastianMarian: The header will be deleted only if it is far away from lastNotarized, like 90% from pool size away in the term of nonces. This is the case when you are in bootstrap mode and you are receiving headers from the network with higher nonces than what you are processing now. This mechanism tries to reserve at least 10% from pool for current nonces which have been requested as part of the current block which needs to be processed.

process/block/shardblock.go
SebastianMarian: The header will be deleted only if it is far away from lastNotarized, like 90% from pool size away in the term of nonces. This is the case when you are in bootstrap mode and you are receiving headers from the network with higher nonces than what you are processing now. This mechanism tries to reserve at least 10% from pool for current nonces which have been requested as part of the current block which needs to be processed.

process/block/shardblock.go
SebastianMarian: Extracted/refactored the whole method in baseProcessor

process/block/metablock.go
```diff
@@ -855,16 +861,26 @@ func (mp *metaProcessor) requestMissingFinalityAttestingHeaders() uint32 {
 				mp.dataPool.HeadersNonces())
 
 			if err != nil {
-				requestedBlockHeaders++
-				go mp.onRequestHeaderHandlerByNonce(shardId, i)
+				missingFinalityAttestingHeaders++
+				wasHeaderRequested := mp.wasHeaderRequested(shardId, i)
+				if !wasHeaderRequested {
+					requestedBlockHeaders++
+					mp.hdrsForCurrBlock.requestedFinalityAttestingHdrs[shardId] = append(mp.hdrsForCurrBlock.requestedFinalityAttestingHdrs[shardId], i)
```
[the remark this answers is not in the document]
SebastianMarian: The resource is already protected in upward methods from which this method is called: metaBlock: receivedShardHeader / requestShardHeaders
shardBlock: receivedMetaHeader / requestMetaHeaders

process/block/shardblock.go
SebastianMarian: The resource is already protected in upward methods from which this method is called: metaBlock: receivedShardHeader / requestShardHeaders
shardBlock: receivedMetaHeader / requestMetaHeaders

process/block/shardblock.go
```diff
@@ -93,6 +93,7 @@ func NewShardProcessor(arguments ArgShardProcessor) (*shardProcessor, error) {
 
 	sp.hdrsForCurrBlock.hdrHashAndInfo = make(map[string]*hdrInfo)
```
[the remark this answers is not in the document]

process/block/metablock.go
```diff
@@ -21,8 +21,8 @@ import (
 // metaProcessor implements metaProcessor interface and actually it tries to execute block
```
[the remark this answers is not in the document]

process/block/preprocess/smartContractResults.go
```diff
@@ -346,24 +351,26 @@ func (scr *smartContractResults) processSmartContractResult(
 }
 
 // RequestTransactionsForMiniBlock requests missing smartContractResults for a certain miniblock
-func (scr *smartContractResults) RequestTransactionsForMiniBlock(mb block.MiniBlock) int {
-	missingScrsForMiniBlock := scr.computeMissingScrsForMiniBlock(mb)
-	scr.onRequestSmartContractResult(mb.SenderShardID, missingScrsForMiniBlock)
+func (scr *smartContractResults) RequestTransactionsForMiniBlock(miniBlock *block.MiniBlock) int {
+	missingScrsForMiniBlock := scr.computeMissingScrsForMiniBlock(miniBlock)
```
AdoAdoAdo: can you treat the case miniBlock == nil?

Check the other files as well

process/block/shardblock.go
```diff
@@ -729,6 +743,12 @@ func (sp *shardProcessor) CommitBlock(
 
 	sp.blockSizeThrottler.Succeed(header.Round)
 
+	log.Info(fmt.Sprintf("Pools len: Headers = %d   MetaBlocks = %d   MiniBlocks = %d\n",
```
AdoAdoAdo: no need for \n in log.Info

PR #582, opened 2019-10-23
En 4496 trie snapshot optimization

BeniaminDrasovean:
Create snapshot and delete old snapshots on goroutines.

### DIFF
--- a/data/trie/node_test.go
+++ b/data/trie/node_test.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/data"
@@ -15,6 +16,9 @@ import (
 	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
 	"github.com/stretchr/testify/assert"
 )
+
+var snapshotDelay = time.Millisecond
+var batchDelay = 2 * time.Second
 
 func TestNode_hashChildrenAndNodeBranchNode(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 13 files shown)


data/trie/patriciaMerkleTrie.go
```diff
@@ -451,11 +469,18 @@ func (tr *patriciaMerkleTrie) Snapshot() error {
 	tr.mutOperation.Lock()
 	defer tr.mutOperation.Unlock()
 
-	err := tr.root.commit(false, 0, tr.db, tr.marshalizer, tr.hasher)
+	if tr.snapshotInProgress {
```
sasurobert: mutex protection should be enough. make sure that it stops there, if it is called the 2nd time.

BeniaminDrasovean: After the `Snapshot()` func is done, the mutex is released, but the `snapshot(newTrie *patriciaMerkleTrie, db data.DBWriteCacher)` is still in progress and is not protected by a mutex.

data/trie/node_test.go
```diff
@@ -900,3 +923,77 @@ func rollbackTrieState(t *testing.T, index int, tr data.Trie, rootHashes [][]byt
...
+		_ = tr.Commit()
+
+		previousRootHashIndex := len(rootHashes) - 1
+		currentRootHash := tr.root.getHash()
+
+		_ = tr.Prune(rootHashes[previousRootHashIndex], data.OldRoot)
+		_ = tr.Prune(currentRootHash, data.NewRoot)
+		rootHashes = append(rootHashes, currentRootHash)
+	}
+	assert.Equal(t, 21, len(evictionWaitList.Cache))
```
bogdan-rosianu: 21 ?

PR #590, opened 2019-10-25
En 4606 tx value as string

bogdan-rosianu:


integrationTests/multiShard/smartContract/executingSCCalls_test.go
```diff
@@ -434,7 +435,14 @@ func processAndTestIntermediateResults(t *testing.T, proposerNodeShardSC *testNo
 		txBytes, _ := proposerNodeShardSC.store.Get(dataRetriever.UnsignedTransactionUnit, hash)
 		assert.NotNil(t, txBytes)
 
-		tx := &transaction.Transaction{}
+		//TODO should refactor this, maybe remove the whole integrationTests/multiShard/smartContract package as these tests
+		// are included by integrationTests/multiShard/block package bu should modify the minimum tx gasprice != 0
```
sasurobert: "but"

data/transaction/transaction.go
```diff
@@ -102,7 +97,8 @@ func (tx *Transaction) IsInterfaceNil() bool {
 
 // GetValue returns the value of the transaction
 func (tx *Transaction) GetValue() *big.Int {
-	return tx.Value
+	val, _ := big.NewInt(0).SetString(tx.Value, 10)
+	return val
```
sasurobert: this is suboptimal. Every time value is called to create a new big.int. also setString consumes processing.

iulianpascalau: Totally agree. I'm not happy of this solution, either.

integrationTests/multiShard/smartContract/executingSCCalls_test.go
iulianpascalau: Done

data/transaction/transaction.go
iulianpascalau: Done, changed back so transaction.Value is *big.Int but transaction has custom methods for marshaling and unmarshaling that treats the Value field as string.

integrationTests/frontend/wallet/txInterception_test.go
```diff
@@ -121,16 +101,17 @@ func testInterceptedTxFromFrontendGeneratedParams(
 
 	integrationTests.MintAddress(node.AccntState, sndAddrBytes, valMinting)
 
-	txHexHash, err = node.SendTransaction(&transaction.Transaction{
+	tx := &transaction.Transaction{
```
SebastianMarian: You can put back: Value:     frontendValue and remove tx.SetValue(frontendValue)

data/transaction/transaction.go
```diff
@@ -154,3 +156,63 @@ func (tx *Transaction) SetRecvAddress(addr []byte) {
...
+		SndAddr   []byte `json:"sender"`
+		GasPrice  uint64 `json:"gasPrice,omitempty"`
+		GasLimit  uint64 `json:"gasLimit,omitempty"`
+		Data      string `json:"data,omitempty"`
+		Signature []byte `json:"signature,omitempty"`
+	}{}
+	if err := json.Unmarshal(dataBuff, &aux); err != nil {
+		return err
+	}
+	tx.Nonce = aux.Nonce
```
sasurobert: add an empty line before

PR #596, opened 2019-10-28
EN-4648: refactor rewards tx interceptors

miiu96:
1. Refactor rewards-tx-interceptors
2. Add move unit tests for interceptedRewardTransaction


process/interceptors/factory/shardInterceptedDataFactory.go
```diff
@@ -92,6 +93,8 @@ func (sidf *shardInterceptedDataFactory) Create(buff []byte) (process.Intercepte
 		return sidf.createInterceptedMetaHeader(buff)
 	case InterceptedTxBlockBody:
 		return sidf.createInterceptedTxBlockBody(buff)
+	case InterceptedRewardTx:
```
raduchis: This method is the basic example why Polymorphism works and you don't need a type for different switches. This could be easily refactored into multiple InterceptedSomething that each implement an interface that has just "Create" inside. Open-Closed Principle is not ok here

raduchis: Please refactor also metaInterceptedDataFactory

miiu96: Done.

PR #623, opened 2019-11-01
Skipped some long tests with short

iulianpascalau:


integrationTests/state/stateExecTransaction_test.go
```diff
@@ -173,6 +173,10 @@ func testExecTransactionsMoreTxWithRevert(
 func TestExecTransaction_MoreTransactionsMoreIterationsWithRevertShouldWork(t *testing.T) {
 	t.Parallel()
 
+	if testing.Short() {
```
sasurobert: put before t.Parallel

iulianpascalau: Done

PR #631, opened 2019-11-06
En 4717 termui cpu load

miiu96:
- Fix in termui problem with CPU load
-Update gopsutil library


core/statistics/machine/cpuStatistics.go
```diff
@@ -11,33 +11,36 @@ var durationSecond = time.Second
...
-		cs.setZeroStatsAndWait()
 		return
 	}
 
-	numPhysicalCores, err := cpu.Counts(false)
-	if err != nil {
-		cs.setZeroStatsAndWait()
-		return
-	}
+	cpuUsagePercent := percent / float64(cs.numCpu)
```
SebastianMarian: Is there a chance that cs.numCpu to be 0. Maybe you can check this and if it is 0 you can set cpuUsagePercent to 100

core/statistics/machine/cpuStatistics.go
```diff
@@ -11,33 +11,36 @@ var durationSecond = time.Second
...
 
-	cpuUsagePercent = cpuUsagePercent / float64(numPhysicalCores)
 	atomic.StoreUint64(&cs.cpuPercentUsage, uint64(cpuUsagePercent))
```
SebastianMarian: I would rename cs.cpuPercentUsage in cs.cpuUsagePercent

core/statistics/machine/cpuStatistics.go
```diff
@@ -11,33 +11,36 @@ var durationSecond = time.Second
...
-	time.Sleep(durationSecond)
 }
 
 func (cs *CpuStatistics) setZeroStatsAndWait() {
```
SebastianMarian: Is there this method used anymore? If not, you can remove it

core/statistics/machine/cpuStatistics.go
```diff
@@ -11,33 +11,36 @@ var durationSecond = time.Second
 
 // CpuStatistics can compute the cpu usage percent
 type CpuStatistics struct {
+	numCpu          int
 	cpuPercentUsage uint64
 }
 
+// NewCpuStatistics will create a CpuStatistics object
+func NewCpuStatistics() *CpuStatistics {
+	numCpu, _ := cpu.Counts(true)
```
iulianpascalau: You are hiding error here. Must be careful with these things, as mentioned by Sebastian that this function might return in some situations numCpu = 0 and error. Your code will break in panic on line 41.
The constructor might return (*CpuStatistics, error) tuple.

core/statistics/machine/cpuStatistics.go
iulianpascalau: We need to test this on all archs/oses (linux, mac, windows, raspPi)

core/statistics/machine/cpuStatistics.go
miiu96: Done.

core/statistics/machine/cpuStatistics.go
miiu96: Done. Method removed.

core/statistics/machine/cpuStatistics.go
miiu96: Done.

core/statistics/machine/cpuStatistics.go
miiu96: Done. Works on all that 4 platforms.

iulianpascalau: ok

PR #635, opened 2019-11-06
En 4832 log viewer app

iulianpascalau:
added logviewer binary with built-in test websocket server


cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+	return conn, nil
+}
+
+func listeningOnWebSocket() {
+	for {
+		msgType, message, err := webSocket.ReadMessage()
+		if msgType == websocket.CloseMessage {
+			return
+		}
+		if err != nil {
```
sasurobert: if err == nil continue

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+		os.Exit(1)
+	}
+}
+
+func initCliFlags() {
+	cliApp = cli.NewApp()
+	cli.AppHelpTemplate = nodeHelpTemplate
+	cliApp.Name = "Elrond Node CLI App"
+	cliApp.Version = fmt.Sprintf("%s/%s/%s-%s", "1.0.0", runtime.Version(), runtime.GOOS, runtime.GOARCH)
+	cliApp.Usage = "This is the entry point for starting a new Elrond node - the app will start after the genesis timestamp"
```
sasurobert: These comments should be changed.

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func listeningOnWebSocket() {
+	for {
+		msgType, message, err := webSocket.ReadMessage()
+		if msgType == websocket.CloseMessage {
+			return
+		}
+		if err != nil {
+			isConnectionClosed := strings.Contains(err.Error(), "websocket: close")
```
traian-schiau: if _, ok := err.(websocket.CloseError) { ..... }

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+		outputMessage(message)
+	}
+}
+
+func waitForUserToTerminateApp(conn *websocket.Conn) {
+	stop := make(chan struct{}, 1)
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
```
traian-schiau: why the dedicate routine to just forward the event and not change the first case to receive from sigs?

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+func outputMessage(message []byte) {
+	fmt.Printf(string(message))
+
+	if fileForLogs == nil {
+		return
+	}
+
+	_, _ = fileForLogs.Write(stripAnsiColors(message))
+	err := fileForLogs.Sync()
```
traian-schiau: sync for every log message is overkill

cmd/logviewer/main.go
iulianpascalau: changed, although it's a little bit harder to get it right

cmd/logviewer/main.go
iulianpascalau: changed

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+	logLevelPatterns = cli.StringFlag{
+		Name:  "level",
+		Usage: "This flag specifies the logger levels and patterns",
+		Value: "*:" + logger.LogInfo.String(),
+	}
+
+	//logFile is used when the log output needs to be logged in a file
+	logFile = cli.BoolFlag{
+		Name:  "file",
+		Usage: "will automatically log into a file",
```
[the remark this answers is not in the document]
iulianpascalau: Changed

cmd/logviewer/main.go
iulianpascalau: Right, removed go routine and stop chan

cmd/logviewer/main.go
iulianpascalau: Removed

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+	if ctx.GlobalBool(logFile.Name) {
+		err = prepareLogFile()
+		if err != nil {
+			return err
+		}
+
+		defer func() {
+			_ = fileForLogs.Close()
+		}()
```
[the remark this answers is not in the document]
iulianpascalau: It shows warning as error not handled

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+func startLogViewer(ctx *cli.Context) error {
+	var err error
+	logLevelPattern := ctx.GlobalString(logLevelPatterns.Name)
+	logLevels, _, err := logger.ParseLogLevelAndMatchingString(logLevelPattern)
+	if err != nil {
+		return err
+	}
+
+	if ctx.GlobalBool(testServer.Name) {
+		_ = testing.StartTestWebSocket("127.0.0.1:8080")
```
ccorcoveanu: If we can send the address through a flag, the testserver should also start at that provided address. If it's moved as a separate binary it should have the same configurable option

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func openWebSocket(address string, logLevelPatterns string) (*websocket.Conn, error) {
+	u := url.URL{
+		Scheme: "ws",
```
ccorcoveanu: Is this thought of as running on the same machine as the node? Maybe there should be an option of running `wss` also

cmd/logviewer/main.go
ccorcoveanu: Agree, use the websocket.CloseError

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// logLevelPatterns defines the logger levels and patterns
+	logLevelPatterns = cli.StringFlag{
+		Name:  "level",
+		Usage: "This flag specifies the logger levels and patterns",
+		Value: "*:" + logger.LogInfo.String(),
+	}
+
+	//logFile is used when the log output needs to be logged in a file
+	logFile = cli.BoolFlag{
```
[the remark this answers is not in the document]
iulianpascalau: removed the internal test server. Moved in a separate binary

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+var log = logger.GetOrCreate("logviewer")
+var cliApp *cli.App
+var webSocket *websocket.Conn
+var manualStop chan struct{}
+var fileForLogs *os.File
+
+func main() {
+	initCliFlags()
+	manualStop = make(chan struct{})
+	log.Info("log viewer application started", "version", cliApp.Version)
```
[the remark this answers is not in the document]
iulianpascalau: moved

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+	logFile = cli.BoolFlag{
+		Name:  "file",
+		Usage: "will automatically log into a file",
+	}
+
+	// testServer is used when a test web socket is need for testing purposes
+	testServer = cli.BoolFlag{
+		Name:  "test-server",
+		Usage: "starts a test server that will send log lines continuously. For development and debugging purposes only.",
+	}
```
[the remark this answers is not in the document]
iulianpascalau: Done, split in a new binary

cmd/logviewer/main.go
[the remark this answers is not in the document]
traian-schiau: "it" should be ignored :) in this kind of situation

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+	ansiCleaner = regexp.MustCompile(ansi)
+)
+
+var log = logger.GetOrCreate("logviewer")
+var cliApp *cli.App
+var webSocket *websocket.Conn
+var manualStop chan struct{}
+var fileForLogs *os.File
```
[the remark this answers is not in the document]
iulianpascalau: main func will look something like:
```GO
instanceData.cliApp.Run(os.Args)
```
not a big fan of it. Won't change

cmd/logviewer/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+func startLogViewer(ctx *cli.Context) error {
+	var err error
+	logLevelPattern := ctx.GlobalString(logLevelPatterns.Name)
+	logLevels, _, err := logger.ParseLogLevelAndMatchingString(logLevelPattern)
+	if err != nil {
+		return err
+	}
+
+	if ctx.GlobalBool(testServer.Name) {
```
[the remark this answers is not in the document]
iulianpascalau: done

cmd/logviewer/main.go
iulianpascalau: Used closeError. Can not use like this:
```GO
if _, ok := err.(websocket.CloseError) { ..... }
```

cmd/logviewer/main.go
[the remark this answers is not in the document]
iulianpascalau: Done, added workingdir option

cmd/logviewer/main.go
iulianpascalau: Done when moved to a new binary file

cmd/logviewer/main.go
iulianpascalau: Added option for wss

cmd/logviewer/main.go
```diff
@@ -205,57 +228,54 @@ func listeningOnWebSocket() {
...
+			continue
 		}
 
-		outputMessage(message)
+		_, isConnectionClosed := err.(*websocket.CloseError)
+		if !isConnectionClosed {
+			log.Error("logviewer websocket error", "error", err.Error())
+			manualStop <- struct{}{}
+		}
+		return
```
ccorcoveanu: What can we do here extra, is instead of return, the connection can be set to nil and as long as it is nil we can try and redial. There could be the situation where we stop the node (for an update) and the logview app can reconnect automatically without having to close it and reopen it

cmd/logviewer/main.go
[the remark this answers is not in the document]
iulianpascalau: We decided, at team level, that we should closely follow Goland IDE's warnings and weak warning. So we will keep it like this even though it is a little bit cumbersome.

cmd/logviewer/main.go
```diff
@@ -1,29 +1,39 @@
...
-	"regexp"
 	"runtime"
-	"strings"
 	"syscall"
 	"time"
 
-	"github.com/ElrondNetwork/elrond-go/cmd/logviewer/testing"
 	"github.com/ElrondNetwork/elrond-go/core"
 	"github.com/ElrondNetwork/elrond-go/logger"
+	protobuf "github.com/ElrondNetwork/elrond-go/logger/proto"
```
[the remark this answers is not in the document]
iulianpascalau: The LogLineMessage structure can also be used with json marshalizer. Because it was autogenerated in proto folder I used it as it is (without creating a secondary wrapper). I used protobuf for performance reasons here. Could have used json Marshalizer as well.

cmd/logviewer/main.go
iulianpascalau: Good ideea, implementing...

cmd/logviewer/main.go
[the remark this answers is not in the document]
ccorcoveanu: Or capnp for performance since we already use it in the project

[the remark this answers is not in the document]
iulianpascalau: Added capnp marshaling capabilities. Added tests for that :D. The app still use protobuf.

cmd/logviewer/main.go
traian-schiau: I did missed an ";ok" but 
https://play.golang.org/p/KFZyUZTf_o_d

PR #659, opened 2019-11-11
En 4931 update stats with trie db size

BeniaminDrasovean:


core/statistics/resourceMonitor.go
```diff
@@ -70,18 +76,31 @@ func (rm *ResourceMonitor) GenerateStatistics() string {
 		fileDescriptors,
 		numOpenFiles,
 		numConns,
+		getDirMemSize(accountsTrieDbFilePath),
+		getDirMemSize(evictionWaitingListDbFilePath),
 	)
 }
 
+func getDirMemSize(dir string) string {
+	files, _ := ioutil.ReadDir(dir)
```
bogdan-rosianu: I know it's not important to check for errors here because it's not crucial, but if any of these stuff receive a `nil` the app will crash. I would recommend checking for errors and if there are not nil, then return the message as the return string.

BeniaminDrasovean: The app will not crash, because `files` will be nil, and that means that len(files)==0, so the range instruction will be skipped.

PR #660, opened 2019-11-11
Arwen metering and namespacing

camilbancioiu:
Update the Node to make use of configurable gas costs:
* Gas costs are loaded from `gasCosts.toml`, parsed as a map and passed to the Arwen constructor
* Some changes were manually brought in from the branch `EN-4559-Add-gas-limit-check-in-process-block`, which adds gas limits to Economics (branch is currently in PR review)

Update the Node's `go.mod` to target `arwen-wasm-vm v0.2.6`, which adds configurable gas metering and WASM namespacing.


cmd/node/main.go
```diff
@@ -131,6 +131,12 @@ VERSION:
 		Usage: "The configuration file for servers confidential data",
 		Value: "./config/server.toml",
 	}
+	// gasScheduleConfigurationFile defines a flag for the path to the toml file containing the gas costs used in SmartContract execution
+	gasScheduleConfigurationFile = cli.StringFlag{
+		Name:  "gascostsconfig",
```
sasurobert: gasCostsConfig ?

config/tomlConfig_test.go
```diff
@@ -155,6 +157,7 @@ func TestTomlEconomicsParser(t *testing.T) {
     LeaderPercentage = ` + fmt.Sprintf("%.6f", leaderPercentage) + `
     BurnPercentage = 	` + fmt.Sprintf("%.6f", burnPercentage) + `
 [FeeSettings]
+	MaxGasLimitPerBlock = "` + maxGasLimitPerBlock + `"
```
sasurobert: use spaces.

core/configParser.go
```diff
@@ -26,3 +26,22 @@ func LoadServersPConfig(filepath string) (*config.ServersConfig, error) {
...
+
+// LoadGasScheduleConfig returns a map[string]uint64 of gas costs read from the provided config file
+func LoadGasScheduleConfig(filepath string) (map[string]uint64, error) {
+	gasScheduleConfig, err := LoadTomlFileToMap(filepath, log)
+	if err != nil {
+		return nil, err
+	}
+
+	flattenedGasSchedule := make(map[string]uint64)
+
```
sasurobert: delete empty line

go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	github.com/satori/go.uuid v1.2.0
 	github.com/shirou/gopsutil v0.0.0-20190731134726-d80c43f9c984
 	github.com/sirupsen/logrus v1.4.0
-	github.com/stretchr/testify v1.3.0
+	github.com/stretchr/testify v1.4.0
```
sasurobert: no need for 1.4. if it does not crash any other machine, or breaks compatibility, it is fine.

integrationTests/testInitializer.go
```diff
@@ -1321,3 +1324,89 @@ func StartP2pBootstrapOnProcessorNodes(nodes []*TestProcessorNode) {
 	fmt.Println("Delaying for nodes p2p bootstrap...")
 	time.Sleep(p2pBootstrapStepDelay)
 }
+
+func MakeDummyGasSchedule() map[string]uint64 {
+	gasSchedule := make(map[string]uint64)
+
+	gasSchedule = arwenconfig.FillGasMapWithWASMOpcodeValues(gasSchedule, 1)
+	gasSchedule["StorePerByte"] = 1
```
sasurobert: make a function for these API calls as well in arwen.

integrationTests/testInitializer.go
```diff
@@ -1321,3 +1324,89 @@ func StartP2pBootstrapOnProcessorNodes(nodes []*TestProcessorNode) {
 	fmt.Println("Delaying for nodes p2p bootstrap...")
 	time.Sleep(p2pBootstrapStepDelay)
 }
+
+func MakeDummyGasSchedule() map[string]uint64 {
+	gasSchedule := make(map[string]uint64)
+
+	gasSchedule = arwenconfig.FillGasMapWithWASMOpcodeValues(gasSchedule, 1)
```
sasurobert: arwenConfig

process/factory/shard/vmContainerFactory_test.go
```diff
@@ -66,3 +89,89 @@ func TestVmContainerFactory_Create(t *testing.T) {
 	acc := vmf.VMAccountsDB()
 	assert.NotNil(t, acc)
 }
+
+func makeDummyGasSchedule() map[string]uint64 {
+	gasSchedule := make(map[string]uint64)
+
+	gasSchedule = arwenconfig.FillGasMapWithWASMOpcodeValues(gasSchedule, 1)
+	gasSchedule["StorePerByte"] = 1
```
sasurobert: move api calls gas filler to a public function in arwen.

cmd/node/main.go
camilbancioiu: Changed.

config/tomlConfig_test.go
camilbancioiu: Changed.

core/configParser.go
camilbancioiu: Done.

go.mod
camilbancioiu: Left it to v1.4.0.

integrationTests/testInitializer.go
camilbancioiu: Done.

integrationTests/testInitializer.go
camilbancioiu: Changed.

process/factory/shard/vmContainerFactory_test.go
camilbancioiu: Done.

camilbancioiu:
> Optional: is it possible to add `typedef` for `map[string]uint64` in file `structs.go` and name it `GasSchedule` or something like that?
This would require importing `arwen-wasm-vm/config` into multiple other files of `elrond-go`. Will leave it without a type alias for now.

> Is it possible for the following cast to fail? If so, do a check, and a test for garbage values in config.
> 
> ```
> flattenedGasSchedule[operationName] = uint64(cost.(int64))
> ```
>
It will either fail with a parse error, or a Go panic. But incorrect values in the toml file will not be treated as correct by the Node.

> In `LoadTomlFileToMap`, can `ioutil.ReadAll(file)` be used instead? Less code (no more `fileinfo` and `stat` etc.).
It could, but I wanted to use the utilities that use the `logger`, which can't use `ReadAll`.
 
> In type `FeeSettings`, why using `string` instead of `uint`? Because the values may be `big.Int`? If so, then change the parsing in `economicsData.go`, and use `big.Int` instead of `ParseUint`.
This is part of the branch `EN-4559-Add-gas-limit-check-in-process-block`, manually copied here to get `MaxBlockGasLimit`. This branch also reads toml files without extra validation, relying on the toml parser and Go panics to avoid using incorrect values as correct.

> Can we add a default `gasSchedule.toml` for the node config? Also, if we add such a file with some missing entries (such as `StorePerByte`, `DataCopyPerByte`), and start the node, we receive errors without context: `field StorePerByte has the value 0`. Please make messages more clear that the gas is not configured properly.
This was a mistake indeed. Fixed now.

PR #661, opened 2019-11-11
EN-4932 : reset counter every time when clean is called

miiu96:


process/block/poolsCleaner/txPoolsCleaner.go
```diff
@@ -81,6 +82,8 @@ func (tpc *TxPoolsCleaner) Clean(duration time.Duration) (bool, error) {
 }
 
 func (tpc *TxPoolsCleaner) cleanPools(haveTime func() bool) {
+	atomic.StoreUint64(&tpc.numRemovedTxs, 0)
```
bogdan-rosianu: on line 61, please change the comment to something like:
`Clean will check if transactions with a nonce which is lower that the nonce of the sender's account exist in pool. If yes, they will be removed from pools`

miiu96: Done.

iulianpascalau: Clean removes the transactions with lower nonces than the senders accounts.

PR #664, opened 2019-11-12
En 4767 logger api route

iulianpascalau:
- created logs route on web sockets
- removed some fmt.prints
- the rest api server now starts on the localhost:8080 interface (configurable through the -rest-api-interface option)


cmd/node/main.go
```diff
@@ -52,13 +52,13 @@ import (
...
-	defaultShardString = "Shard"
-	metachainShardName = "metachain"
-	defaultRestApiPort = "off"
+	defaultLogPath         = "logs"
+	defaultStatsPath       = "stats"
+	defaultDBPath          = "db"
+	defaultEpochString     = "Epoch"
+	defaultShardString     = "Shard"
+	metachainShardName     = "metachain"
+	defaultRestApiInerface = "localhost:8080"
```
raduchis: Do we want to enable port 8080? Can the restApi be enabled just for localhost and not the outside?

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/prometheus/client_golang v1.0.0
 	github.com/satori/go.uuid v1.2.0
 	github.com/shirou/gopsutil v0.0.0-20190731134726-d80c43f9c984
-	github.com/sirupsen/logrus v1.4.0
+	github.com/sirupsen/logrus v1.4.0 // indirect
```
raduchis: What is the indirect?

logger/logOutputSubject.go
```diff
@@ -5,7 +5,7 @@ import (
 	"io"
 	"sync"
 
-	protobuf "github.com/ElrondNetwork/elrond-go/logger/proto"
```
raduchis: +1 for wrapping the inside of the marshalizer so that the subject does not know about the internals

cmd/node/main.go
iulianpascalau: Yes, we will re-enable 8080. As found in this PR, one can configure also the interface on which the gin server is allowed to be bound on. localhost:8080 means that the server is able to open the port only on localhost interface (lan and wan connections are not possible).

go.mod
iulianpascalau: removed the whole line, it appears that logrus library is used by some other packages and since it was used also by elrond-go and was present in go.mod, goland did not remove it automatically.

logger/logOutputSubject.go
iulianpascalau: Yeah, finally found the proper way to deal with all 3 marshalizers and having a cleaner, comprehensible implementation.

process/transaction/shardProcess.go
```diff
@@ -121,7 +121,7 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction, round
 
 func (txProc *txProcessor) processTxFee(tx *transaction.Transaction, acntSnd *state.Account) (*big.Int, error) {
 	if acntSnd == nil {
-		return nil, nil
+		return big.NewInt(0), nil
```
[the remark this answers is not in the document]
iulianpascalau: No, it's just a small bug that I found during the system test of this PR. Inefficient to make another PR just for that

logger/logOutputSubject.go
```diff
@@ -44,12 +44,11 @@ func (los *logOutputSubject) convertLogLine(logLine *LogLine) LogLineHandler {
...
-		Message:   logLine.Message,
-		LogLevel:  int32(logLine.LogLevel),
-		Args:      make([]string, len(logLine.Args)),
-		Timestamp: logLine.Timestamp.Unix(),
-	}
+	line := &LogLineWrapper{}
+	line.Message = logLine.Message
+	line.LogLevel = int32(logLine.LogLevel)
+	line.Args = make([]string, len(logLine.Args))
+	line.Timestamp = logLine.Timestamp.Unix()
```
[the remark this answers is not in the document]
iulianpascalau: Refactored in such a way to take the handler pointer before the for loop

cmd/node/main.go
```diff
@@ -189,11 +189,18 @@ VERSION:
...
 
-	// restApiPort defines a flag for port on which the rest API will start on
-	restApiPort = cli.StringFlag{
-		Name:  "rest-api-port",
-		Usage: "The port on which the rest API will start on",
-		Value: defaultRestApiPort,
+	// restApiInterface defines a flag for the interface on which the rest API will try to bind with
+	restApiInterface = cli.StringFlag{
+		Name: "rest-api-interface",
+		Usage: "The interface on which the rest API will try to bind with. Default is 'localhost:8080'. To use it" +
```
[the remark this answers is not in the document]
iulianpascalau: Rephrased. Removed the default message as it is already displayed by default `./node -h`

cmd/node/main.go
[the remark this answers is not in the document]
iulianpascalau: removed from main

facade/elrondNodeFacade.go
```diff
@@ -16,8 +15,8 @@ import (
 	"github.com/ElrondNetwork/elrond-go/ntp"
 )
 
-// DefaultRestPort is the default port the REST API will start on if not specified
-const DefaultRestPort = "8080"
+// DefaultRestInterface is the default interface the rest API will start on if not specified
+const DefaultRestInterface = "localhost:8080"
```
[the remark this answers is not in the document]
iulianpascalau: removed from main

PR #679, opened 2019-11-17
En 4939 network sharding integration

iulianpascalau:
Network sharding component that gathers info from the system (consensus topic and heartbeat topic) about the connections between peer ids and shard ids. It uses 3 cachers for storing link information.
Refactored p2p.networkSharding package
Done integration between elrond-go, sharding/networkSharding and p2p/networkSharding


sharding/networkSharding/peerShardMapper.go
```diff
@@ -0,0 +1,133 @@
...
+		return sharding.UnknownShardId, nil, false
+	}
+
+	pkBuff, ok := pkObj.([]byte)
+	if !ok {
+		return sharding.UnknownShardId, nil, false
+	}
+
+	_, shardId, err := psm.nodesCoordinator.GetValidatorWithPublicKey(pkBuff)
+	if err == nil {
```
sasurobert: if err != nil -> return bad stuff. return the remaining good ones :)

sharding/networkSharding/peerShardMapper.go
```diff
@@ -0,0 +1,133 @@
...
+	shard, ok := shardObj.(uint32)
+	if !ok {
+		return sharding.UnknownShardId, false
+	}
+
+	return shard, true
+}
+
+func (psm *PeerShardMapper) byIDSearchingPidInFallbackCache(pid p2p.PeerID) (shardId uint32) {
+	shardObj, ok := psm.peerIdPk.Get([]byte(pid))
```
sasurobert: peek is faster.

sharding/networkSharding/peerShardMapper.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return sharding.UnknownShardId, pkBuff, false
+}
+
+func (psm *PeerShardMapper) byIDSearchingPkInFallbackCache(pkBuff []byte) (shardId uint32, ok bool) {
+	if len(pkBuff) == 0 {
+		return sharding.UnknownShardId, false
+	}
+
+	shardObj, ok := psm.fallbackPkShard.Get(pkBuff)
```
sasurobert: peek.

sharding/networkSharding/peerShardMapper.go
```diff
@@ -0,0 +1,133 @@
...
+	shardId, ok = psm.byIDSearchingPkInFallbackCache(pk)
+	if ok {
+		return shardId
+	}
+
+	return psm.byIDSearchingPidInFallbackCache(pid)
+}
+
+func (psm *PeerShardMapper) byIDWithNodesCoordinator(pid p2p.PeerID) (shardId uint32, pk []byte, ok bool) {
+	pkObj, ok := psm.peerIdPk.Get([]byte(pid))
```
sasurobert: peek

consensus/spos/worker.go
```diff
@@ -303,6 +316,26 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
...
+	case wrk.checkSigChan <- struct{}{}:
+	default:
+		return
+	}
+
+	defer func() {
+		<-wrk.checkSigChan
+	}()
+
+	sigVerifErr := wrk.checkSignature(cnsMsg)
```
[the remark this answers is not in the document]
iulianpascalau: no, consensus worker is the interceptor for the consensus topic

consensus/spos/worker.go
```diff
@@ -241,6 +252,8 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
 		cnsDta.RoundIndex,
 	))
 
+	go wrk.updateNetworkShardingVals(message, cnsDta)
```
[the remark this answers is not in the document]
iulianpascalau: This is the first implementation as to have something that runs (not optimized yet). The reason behind updating each and every message is that this implementation should keep the mappings as accurate as possible, not keeping this accurate may lead into topic fragmentation.

p2p/libp2p/netMessenger.go
```diff
@@ -542,6 +545,21 @@ func (netMes *networkMessenger) ThresholdMinConnectedPeers() int {
...
 }
 
+// SetPeerShardResolver sets the peer shard resolver component that is able to resolve the link
+// between p2p.PeerID and shardId
+func (netMes *networkMessenger) SetPeerShardResolver(peerShardResolver p2p.PeerShardResolver) error {
+	if check.IfNil(peerShardResolver) {
+		return p2p.ErrNilPeerShardResolver
+	}
+
+	kadSharder, err := networksharding.NewKadSharder(kadSharderPrioBits, peerShardResolver)
```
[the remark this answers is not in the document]
iulianpascalau: Will think about a solution to this. The whole file is swarming with creating new implementations because of the need to create the sub-components that do the job done.

sharding/networkSharding/peerShardMapper.go
iulianpascalau: Will think of a solution

sharding/networkSharding/peerShardMapper.go
iulianpascalau: Correct, but I want to update the pid rating

sharding/networkSharding/peerShardMapper.go
iulianpascalau: Need to update the pkbuff rating inside cacher

sharding/networkSharding/peerShardMapper.go
iulianpascalau: same reason

sharding/networkSharding/peerShardMapper.go
```diff
@@ -0,0 +1,133 @@
...
+		peerIdPk:         peerIdPk,
+		fallbackPkShard:  fallbackPkShard,
+		fallbackPidShard: fallbackPidShard,
+		nodesCoordinator: nodesCoordinator,
+	}, nil
+}
+
+// ByID returns the corresponding shard ID of a given peer ID.
+// If the pid is unknown, it will return sharding.UnknownShardId
+func (psm *PeerShardMapper) ByID(pid p2p.PeerID) (shardId uint32) {
```
traian-schiau: I expect to have more ID requests than nodes distribution in stards update , maybe is better to have the ID to shard mapping in one cache, which is maintained by the update calls

iulianpascalau: Ok, we can do that in a next PR. Right now it should suffice and should work. The reason behind this implementation is that I wanted to use an interface (NodesCoordinator) that should be able to efficiently tell the mapper the shard ID for a provided pk. NodesCoordinator is used each time a new header is intercepted because the header interceptor needs to verify signatures with the help of the NodesCoordinator.
Since connecting/disconnecting to/from a peer is a lengthy process (~hundreds of milliseconds) compared to searching through a map of slices with limited number of items, I think we can leave it like this for the moment and improve the updating with an event-based scheme at a later time.

p2p/libp2p/netMessenger.go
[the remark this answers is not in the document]
iulianpascalau: Refactoring the netMessenger struct is too complicated for this PR. Created EN-5014, please check description. Will remain as it is for the moment

node/interface.go
```diff
@@ -21,3 +21,10 @@ type P2PMessenger interface {
 	IsConnectedToTheNetwork() bool
 	IsInterfaceNil() bool
 }
+
+// NetworkShardingUpdater defines the updating methods used by the network sharding component
+type NetworkShardingUpdater interface {
```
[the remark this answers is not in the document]
iulianpascalau: Collector as it sounds better. Done

sharding/networkSharding/peerShardMapper.go
iulianpascalau: Done

p2p/libp2p/networksharding/noSharder.go
```diff
@@ -7,16 +7,16 @@ import (
 )
 
 // NoSharder default sharder, only uses Kademlia distance in sorting
-type noSharder struct {
+type NoSharder struct {
```
raduchis: Do you need this? Isn't this already implemented?

consensus/interface.go
```diff
@@ -68,3 +69,10 @@ type P2PMessenger interface {
 	Broadcast(topic string, buff []byte)
 	IsInterfaceNil() bool
 }
+
+// NetworkShardingUpdater defines the updating methods used by the network sharding component
+type NetworkShardingUpdater interface {
```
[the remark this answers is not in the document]
iulianpascalau: Done on all interfaces, mocks, files and on the implementation

p2p/libp2p/libp2pConnectionMonitor.go
```diff
@@ -35,6 +39,7 @@ func newLibp2pConnectionMonitor(reconnecter p2p.Reconnecter, thresholdMinConnect
 		thresholdDiscoveryResume:   0,
 		thresholdDiscoveryPause:    math.MaxInt32,
 		thresholdConnTrim:          math.MaxInt32,
+		sharder:                    &networksharding.NoSharder{},
```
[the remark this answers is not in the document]
iulianpascalau: Will be refactored. As stated, the networkMessenger will have a predefined set of options that will be applied before running the ApplyOptions method on the networkMessenger. This will ensure the instance is ready to be used with some basic functionality activated that can be modified at a later time and will need a minimum provisioning of dependencies for that basic usage. The behavior is intensely used elsewhere, for example: https://github.com/libp2p/go-libp2p/blob/master/options.go

p2p/libp2p/networksharding/noSharder.go
iulianpascalau: We might need this in some tests where network sharding is not wanted

cmd/node/config/config.toml
```diff
@@ -214,6 +214,18 @@
     Size = 1000
     Type = "LRU"
 
+[PublicKeyShardId]
+    Size = 1000
```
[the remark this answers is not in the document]
iulianpascalau: Good point :+1: Will change to 15000 
pidPkCacherSize = ~ (32 + 128) * 15000 < 3MB
pkShardCacherSize = ~ (128 + 4) * 15000 < 2MB
pidShardCacherSize = ~ (32 + 4) * 15000 < 1MB

consensus/spos/worker_test.go
```diff
@@ -522,12 +556,51 @@ func TestWorker_NewWorkerSyncTimerNilShouldFail(t *testing.T) {
...
-		nil)
+		nil,
+		createMockNetworkShardingCollector(),
+	)
 
 	assert.Nil(t, wrk)
 	assert.Equal(t, spos.ErrNilSyncTimer, err)
 }
 
+func TestWorker_NewWorkerNilNetworkShardingCollectorShouldFailShouldFail(t *testing.T) {
```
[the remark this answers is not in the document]
iulianpascalau: done

cmd/node/config/config.toml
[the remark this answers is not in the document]
iulianpascalau: done

PR #690, opened 2019-11-20
update multi data interceptor and trie sync

BeniaminDrasovean:


process/interceptors/multiDataInterceptor.go
```diff
@@ -74,20 +75,32 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
...
+
+		nodeData, ok := interceptedData.(*trie.InterceptedTrieNode)
+		if !ok {
+			mdi.throttler.EndProcessing()
+			return
 		}
 
-		err = interceptedData.CheckValidity()
+		nodeData.CreateEndOfProcessingTriggerNode()
+		_ = mdi.processor.Save(nodeData)
```
bogdan-rosianu: treat the error or at least log it

process/interceptors/multiDataInterceptor.go
```diff
@@ -74,20 +75,32 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
 	lastErrEncountered := error(nil)
 	wgProcess := &sync.WaitGroup{}
 	wgProcess.Add(len(multiDataBuff))
+
+	firstDataBuff := multiDataBuff[0]
```
bogdan-rosianu: you can move this to line 75

data/trie/interceptedNode.go
```diff
@@ -79,3 +79,9 @@ func (inTn *InterceptedTrieNode) IsInterfaceNil() bool {
 func (inTn *InterceptedTrieNode) EncodedNode() []byte {
 	return inTn.encNode
 }
+
+// CreateEndOfProcessingTriggerNode changes the hash of the current node by appending the hash to the current hash.
+// This construction will be used to trigger the end of processing for all of the received data
+func (inTn *InterceptedTrieNode) CreateEndOfProcessingTriggerNode() {
+	inTn.hash = append(inTn.hash, inTn.hash...)
```
iulianpascalau: This is a bit messy and racy (calling append concurrently will yield unexpected results). Why not using a chan struct{}? Way safer and cleaner.

process/interceptors/multiDataInterceptor.go
BeniaminDrasovean: Done.

process/interceptors/multiDataInterceptor.go
BeniaminDrasovean: Done.

data/trie/interceptedNode.go
BeniaminDrasovean: The current architecture does not allow for a chan. Added a todo to use a channel at the next refactoring, and added a mutex to protect from race inconsistencies.

iulianpascalau: ok

process/interceptors/multiDataInterceptor.go
```diff
@@ -71,23 +72,39 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
...
+		interceptedData, err := mdi.interceptedData(firstDataBuff)
 		if err != nil {
-			lastErrEncountered = err
-			wgProcess.Done()
-			continue
+			mdi.throttler.EndProcessing()
+			return
+		}
+
+		nodeData, ok := interceptedData.(*trie.InterceptedTrieNode)
```
iulianpascalau: This is used in a wrong manner.

BeniaminDrasovean: Refactored.

PR #691, opened 2019-11-20
[Net shard.] Discovery pause rework

traian-schiau:
- Don't pause discovery unless the upper layers agrees
 - Only pause after trimming
 - Do an automatic (limited) discovery every 20 conn intervals skipped
 - Increase the number of discovery queries

### DIFF
--- a/p2p/libp2p/discovery/kadDhtDiscoverer.go
+++ b/p2p/libp2p/discovery/kadDhtDiscoverer.go
@@ -12,13 +12,12 @@ import (
 )
 
 const (
-	initReconnectMul = 20
+	initReconnectMul     = 20
+	peerDiscoveryTimeout = 5 * time.Second
+	noOfQueries          = 3
+
+	kadDhtName = "kad-dht discovery"
 )
-
-var peerDiscoveryTimeout = 10 * time.Second
-var noOfQueries = 1
-
-const kadDhtName = "kad-dht discovery"
 
 var log = logger.GetOrCreate("p2p/libp2p/kaddht")
 
--- a/p2p/libp2p/discovery/kadDhtDiscoverer.go
+++ b/p2p/libp2p/discovery/kadDhtDiscoverer.go
@@ -1,14 +1,17 @@
 package discovery
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/ElrondNetwork/elrond-go/logger"
 	"github.com/ElrondNetwork/elrond-go/p2p"
 	"github.com/ElrondNetwork/elrond-go/p2p/libp2p"
+	protocol "github.com/libp2p/go-libp2p-core/protocol"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
-	"github.com/libp2p/go-libp2p-kbucket"
+	opts "github.com/libp2p/go-libp2p-kad-dht/opts"
+	kbucket "github.com/libp2p/go-libp2p-kbucket"
 )
 
 const (
@@ -73,11 +76,16 @@ func (kdd *KadDhtDiscoverer) Bootstrap() error {
 	ctx := kdd.contextProvider.Context()
 	h := kdd.contextProvider.Host()
 
+	protos := []protocol.ID{
+		protocol.ID(fmt.Sprintf("/ipfs/kad/1.0.0/erd_%s", kdd.randezVous)),
+		protocol.ID("/ipfs/kad/1.0.0/erd"),
+	}
+
 	// Start a DHT, for use in peer discovery. We can't just make a new DHT
 	// client because we want each peer to maintain its own local copy of the
 	// DHT, so that the bootstrapping node of the DHT can go down without
 	// inhibiting future peer discovery.
-	kademliaDHT, err := dht.New(ctx, h)
+	kademliaDHT, err := dht.New(ctx, h, opts.Protocols(protos...))
 	if err != nil {
 		return err
 	}
--- a/p2p/libp2p/discovery/kadDhtDiscoverer.go
+++ b/p2p/libp2p/discovery/kadDhtDiscoverer.go
@@ -32,6 +32,7 @@ type KadDhtDiscoverer struct {
 	randezVous       string
 	initialPeersList []string
 	initConns        bool // Initiate new connections
+	watchdogKick     chan struct{}
 }
 
 // NewKadDhtPeerDiscoverer creates a new kad-dht discovery type implementation
@@ -227,4 +228,34 @@ func (kdd *KadDhtDiscoverer) IsInterfaceNil() bool {
 		return true
 	}
 	return false
+}
+
+// StartWatchdog start the watchdog, and return the kick channel
+func (kdd *KadDhtDiscoverer) StartWatchdog(timeout time.Duration) chan struct{} {
+	kdd.mutKadDht.Lock()
+	defer kdd.mutKadDht.Unlock()
+
+	if kdd.contextProvider == nil {
+		return nil
+	}
+
+	if kdd.watchdogKick != nil {
+		return kdd.watchdogKick
+	}
+
+	kdd.watchdogKick = make(chan struct{})
+	ctx := kdd.contextProvider.Context()
+	go func() {
+		for {
+			select {
+			case <-time.After(timeout):
+				kdd.Resume()
+			case <-ctx.Done():
+				return
+			case <-kdd.watchdogKick:
+			}
+		}
+	}()
+
+	return kdd.watchdogKick
 }
--- a/p2p/libp2p/discovery/kadDhtDiscoverer.go
+++ b/p2p/libp2p/discovery/kadDhtDiscoverer.go
@@ -295,7 +295,7 @@ func (kdd *KadDhtDiscoverer) StopWatchdog() error {
 	return nil
 }
 
-// KickWatchdog stops the discovery watchdog
+// KickWatchdog extends the discovery resume timeout
 func (kdd *KadDhtDiscoverer) KickWatchdog() error {
 	kdd.mutKadDht.Lock()
 	defer kdd.mutKadDht.Unlock()
--- a/p2p/libp2p/discovery/kadDhtDiscoverer.go
+++ b/p2p/libp2p/discovery/kadDhtDiscoverer.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"time"
@@ -20,6 +21,8 @@ const (
 	noOfQueries          = 3
 
 	kadDhtName = "kad-dht discovery"
+
+	minWatchdogTimeout = time.Second
 )
 
 var log = logger.GetOrCreate("p2p/libp2p/kaddht")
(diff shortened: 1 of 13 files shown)


p2p/libp2p/networksharding/kadSharder.go
```diff
@@ -15,6 +15,7 @@ const (
 // KadPeerShardResolver peer to shard mapping interface
 type KadPeerShardResolver interface {
 	ByID(p2p.PeerID) uint32 //ByID get the shard id of the given peer.ID
+	IsBalanced() bool       //IsBalanced checks if the sharding is balanced and connections initiation can be paused
```
iulianpascalau: This can not be computed by the Mapper implementation. It simply does not have the required info to decide nor it should care how the lower level component manage its connections. I think this should be resolved by Sharded implementations. For example, on the bit masking kadSharder implementation, while iterating the sortingList and calling resolver.ByID, you could also have some counters and then decide if the the list is balanced or not.

traian-schiau: done, but I'll need the total number of shards

iulianpascalau: That can be provided by the KadPeerShardResolver as it a high level info. Just add a function like:
```GO
NumberOfShards() uint32
```

p2p/libp2p/discovery/kadDhtDiscoverer.go
```diff
@@ -33,6 +35,9 @@ type KadDhtDiscoverer struct {
 	randezVous       string
 	initialPeersList []string
 	initConns        bool // Initiate new connections
+	watchdogKick     chan struct{}
+
+	localShard uint32
```
golangcibot: `localShard` is unused (from `structcheck`)

p2p/libp2p/networksharding/kadSharder.go
```diff
@@ -15,6 +15,7 @@ const (
 // KadPeerShardResolver peer to shard mapping interface
 type KadPeerShardResolver interface {
 	ByID(p2p.PeerID) uint32 //ByID get the shard id of the given peer.ID
+	NumShards() int         //NumShards get the number of shards
```
iulianpascalau: This is not used. Do you plan to use it in the near future?

p2p/libp2p/networksharding/sortingList.go
```diff
@@ -33,3 +34,24 @@ func (sl *sortingList) Less(i int, j int) bool {
...
+	sort.Sort(sl)
+	ret := make([]peer.ID, len(sl.peers))
+
+	for i, id := range sl.peers {
+		ret[i] = id.id
+	}
+	return ret
+}
+
+func (sl *sortingList) InShardCount() int {
```
iulianpascalau: This sorting list knows how to count in shard peers. I think this should be move inside kadSharder and be unexported because it is not the sortingList's job.

p2p/libp2p/discovery/kadDhtDiscoverer.go
```diff
@@ -229,3 +237,33 @@ func (kdd *KadDhtDiscoverer) IsInterfaceNil() bool {
...
+func (kdd *KadDhtDiscoverer) StartWatchdog(timeout time.Duration) chan struct{} {
+	kdd.mutKadDht.Lock()
+	defer kdd.mutKadDht.Unlock()
+
+	if kdd.contextProvider == nil {
+		return nil
+	}
+
+	if kdd.watchdogKick != nil {
+		return kdd.watchdogKick
```
iulianpascalau: This should yield an error as the watchdog process is already started. Otherwise, the function does at least 2 things: start a watchdog and, if already started, returns the same chan.

p2p/libp2p/networksharding/sharder.go
```diff
@@ -40,15 +39,15 @@ type Sharder interface {
 	// GetDistance get the distance between a and b
 	GetDistance(a, b sortingID) *big.Int
```
iulianpascalau: Trying to reuse the Sharder interface & implementations elsewhere and found out that we might change the interface and implementations in a following manner:
```GO
type Sharder interface {
     CanConnectToPeer(newPeer peer.ID, currentList []peer.ID) bool
     CanClosePeer(newPeer peer.ID, currentList []peer.ID) bool
     CanCallBootstrap() bool
}
```

In this way, the clients of this interface (KadDhtDiscoverer, libp2pConnectionMonitor and possible KadDhtHost) will only need this component to tell them what to do: resume the discovery process, closing unwanted peers and if it's worth the connection to be really made. The logic with all the required parameters, watchdog, kademlia implementation, sorting, bit masking and so on will be encapsulated in the kad sharder implementation(s). 
Sorting is just too general as we need the target peers to be inserted just about everywhere ending up having different portions of duplicated code.

p2p/libp2p/networksharding/sortingList.go
traian-schiau: Done

p2p/libp2p/discovery/kadDhtDiscoverer.go
traian-schiau: Done, (changed the interface)

p2p/libp2p/networksharding/sharder.go
traian-schiau: OK, but in a future PR

p2p/libp2p/networksharding/kadSharder.go
```diff
@@ -69,11 +69,24 @@ func (ks *kadSharder) GetDistance(a, b sortingID) *big.Int {
 }
 
 // SortList sort the provided peers list
-func (ks *kadSharder) SortList(peers []peer.ID, ref peer.ID) []peer.ID {
-	return sortList(ks, peers, ref)
+func (ks *kadSharder) SortList(peers []peer.ID, ref peer.ID) ([]peer.ID, bool) {
+	sl := getSortingList(ks, peers, ref)
+	// for balance we just try to keep at least 1 connection ouside of shard
+	balanced := len(peers) > inShardCount(sl)
```
AdoAdoAdo: I think at least one connection would not be enough, why don't we have this as percentage?
float(len(peers)-inShardCount)/float(len(peers))> percentageOutside

where percentageOutside is subunitar e.g 0.2 meaning 20%
same for intrashard

p2p/libp2p/networksharding/sharder.go
iulianpascalau: Ok, will be done in a future PR.

p2p/libp2p/libp2pConnectionMonitor.go
```diff
@@ -7,11 +7,15 @@ import (
...
+	ns "github.com/ElrondNetwork/elrond-go/p2p/libp2p/networksharding"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/multiformats/go-multiaddr"
 )
 
+const (
+	watchdogTimeout = 5 * time.Minute
+)
+
 // DurationBetweenReconnectAttempts is used as to not call reconnecter.ReconnectToNetwork() to often
```
AdoAdoAdo: typos:
to often -> too often
succeed -> succeeded/succeeds

p2p/libp2p/discovery/kadDhtDiscoverer.go
```diff
@@ -230,32 +242,71 @@ func (kdd *KadDhtDiscoverer) IsInterfaceNil() bool {
...
 
-	return kdd.watchdogKick
+	if kdd.watchdogCancel == nil {
+		return p2p.ErrWatchdogNotStarted
+	}
+
+	kdd.watchdogCancel()
+	kdd.watchdogCancel = nil
+
+	close(kdd.watchdogKick)
```
iulianpascalau: warning: this might trigger instruction on line 272

p2p/libp2p/networksharding/kadSharder.go
raduchis: Don't we want to also have some connections outside of the shard in the "balanced". Could it be that all peers are in my shard and the kadSharder says it is balanced?

traian-schiau: Done, we try to keep between 1 connection and 20%

p2p/libp2p/libp2pConnectionMonitor.go
traian-schiau: done

p2p/libp2p/discovery/kadDhtDiscoverer.go
traian-schiau: unlikely, but even if so, the nest select will hit 270 and return

p2p/libp2p/discovery/kadDhtDiscoverer.go
```diff
@@ -229,3 +241,72 @@ func (kdd *KadDhtDiscoverer) IsInterfaceNil() bool {
...
+	kdd.watchdogKick = nil
+	return nil
+}
+
+// KickWatchdog stops the discovery watchdog
```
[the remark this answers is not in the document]
traian-schiau: done

p2p/libp2p/networksharding/kadSharder.go
traian-schiau: Done, we try to keep between 1 connection and 20%

p2p/libp2p/discovery/kadDhtDiscoverer_test.go
```diff
@@ -273,3 +273,44 @@ func TestKadDhtPeerDiscoverer_ApplyContextShouldWork(t *testing.T) {
...
+
+	s1 := kdd.StartWatchdog(interval)
+	s2 := kdd.StartWatchdog(interval)
+
+	assert.Nil(t, s1)
+	assert.Equal(t, s2, p2p.ErrWatchdogAlreadyStarted)
+
+	if !testing.Short() {
+		//kick
+		kdd.KickWatchdog()
```
golangcibot: Error return value of `kdd.KickWatchdog` is not checked (from `errcheck`)

p2p/libp2p/discovery/kadDhtDiscoverer_test.go
```diff
@@ -273,3 +273,44 @@ func TestKadDhtPeerDiscoverer_ApplyContextShouldWork(t *testing.T) {
...
+		kdd.Pause()
+		assert.True(t, kdd.IsDiscoveryPaused())
+		time.Sleep(interval / 2)
+		assert.True(t, kdd.IsDiscoveryPaused())
+		time.Sleep(interval)
+		assert.False(t, kdd.IsDiscoveryPaused())
+	}
+
+	kdd.StopWatchdog()
```
golangcibot: Error return value of `kdd.StopWatchdog` is not checked (from `errcheck`)

p2p/libp2p/networksharding/kadSharder.go
```diff
@@ -69,11 +69,31 @@ func (ks *kadSharder) GetDistance(a, b sortingID) *big.Int {
...
-func (ks *kadSharder) SortList(peers []peer.ID, ref peer.ID) []peer.ID {
-	return sortList(ks, peers, ref)
+func (ks *kadSharder) SortList(peers []peer.ID, ref peer.ID) ([]peer.ID, bool) {
+	sl := getSortingList(ks, peers, ref)
+	// for balance we should have between 1 and 20% connections outside of shard
+	peerCnt := len(peers)
+	inShardCnt := inShardCount(sl)
+	balanced := peerCnt > inShardCnt
+
+	if balanced && peerCnt > 5 {
```
AdoAdoAdo: I do not understand your magic numbers
why multiply by 4 and divide by 5... to get 20 %?
the percentages should be configurable, worst case have them in constants on top of the file.

PR #694, opened 2019-11-21
Add testnet and txgen scripts

camilbancioiu:
Scripts for running a testnet locally (i.e. the "Ireland" testnet in a vial)

What works:
* configuring the network layout
* automatically starting nodes according to configured network layout
* automatically starting the proxy connected to the Shard observers
* automatically starting the txgen, connected to the proxy
* verification against missing components (`proxy` and `txgen` are private repositories, but community members must still be able to use these scripts without them)
* tested on Ubuntu 18.04 and Fedora 30


scripts/testnet/include/config.sh
```diff
@@ -0,0 +1,152 @@
...
+  echo "Updated configuration for Nodes."
+}
+
+copyProxyConfig() {
+  cd $TESTNETDIR
+
+  cp $PROXYDIR/config/config.toml ./proxy/config/
+
+  cp ./node/config/economics.toml ./proxy/config/
+  cp ./node/config/initialBalancesSk.pem ./proxy/config
```
traian-schiau: why not symliks? (for most of the configs)

scripts/testnet/include/config.sh
```diff
@@ -0,0 +1,152 @@
...
+  rm p2p_edit.toml
+
+  echo "Updated configuration for Nodes."
+}
+
+copyProxyConfig() {
+  cd $TESTNETDIR
+
+  cp $PROXYDIR/config/config.toml ./proxy/config/
+
```
traian-schiau: maybe do updateProxyConfig at this point

scripts/testnet/start.sh
```diff
@@ -0,0 +1,61 @@
...
+startObservers
+startValidators
+showTerminalSession "elrond-nodes"
+echo "Waiting for the Nodes to start ($NODE_DELAY s)..."
+sleep $NODE_DELAY
+
+# Phase 5: start the Proxy
+startProxy
+echo "Waiting for the Proxy to start ($PROXY_DELAY s)..."
+sleep $PROXY_DELAY
```
traian-schiau: maybe move the waiting in startProxy, same for the seednode

scripts/testnet/include/build.sh
```diff
@@ -0,0 +1,67 @@
...
+
+
+buildNode() {
+  echo "Building Node executable..."
+  cd $NODEDIR
+  go build .
+
+  cd $TESTNETDIR
+  cp $NODE ./node/
+  echo "Node executable built."
```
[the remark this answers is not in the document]
camilbancioiu: The functions in `build.sh` were intended to be ran as independently as possible from another, and they can't depend on the directory list being present. But `pushd` and `popd` (and `dirs`) are excellent, I didn't know they exist, thank you!

scripts/testnet/include/config.sh
camilbancioiu: These files will be edited in-place with `sed`, so I wanted to have independent copies of each. Symlinks would have worked for some, but not all config files. This was the case with `nodesSetup.json`, which doesn't need editing on this branch (symlinking possible), but will need editing on a new branch soon to be merged in development (symlinking not possible anymore). So I used `cp` for all of them, in case some might need more editing.

scripts/testnet/include/config.sh
camilbancioiu: Fair enough - I'll keep `copyProxyConfig` and `updateProxyConfig` separate for now, so it will be visible in `start.sh` that the config files are in fact being edited. Same applies for the other config functions.

scripts/testnet/start.sh
camilbancioiu: Waiting for a command to finish should not be part of the command, and it would make you dig for the `sleep` command into `include/tools.sh`. This way, `start.sh` remains a list of atomic actions (at least this was the intention).

PR #702, opened 2019-11-22
EN-4808/fix bad config for end of epoch

sasurobert:


process/block/shardblock.go
```diff
@@ -1222,7 +1247,12 @@ func (sp *shardProcessor) receivedMetaBlock(metaBlockHash []byte) {
 		return
 	}
 
-	go sp.epochStartTrigger.ReceivedHeader(metaBlock)
+	sp.epochStartTrigger.ReceivedHeader(metaBlock)
+	if sp.epochStartTrigger.IsEpochStart() {
+		sp.mutEpochStart.Lock()
+		sp.mutEpochStart.Unlock()
```
golangcibot: SA2001: empty critical section (from `staticcheck`)

process/block/shardblock.go
```diff
@@ -279,36 +283,50 @@ func (sp *shardProcessor) ProcessBlock(
 	return nil
 }
 
+func (sp *shardProcessor) requestEpochStartInfo(header *block.Header, waitTime time.Duration) error {
+	sp.mutEpochStart.Lock()
+	_ = process.EmptyChannel(sp.chRcvEpochStart)
+	haveMissingMetaHeaders := header.IsStartOfEpochBlock() && sp.epochStartTrigger.IsEpochStart() == false
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!sp.epochStartTrigger.IsEpochStart()` (from `gosimple`)

sasurobert: done

process/block/shardblock.go
sasurobert: done

process/block/shardblock.go
```diff
@@ -279,36 +283,50 @@ func (sp *shardProcessor) ProcessBlock(
...
+	isEpochIncorrect := header.GetEpoch() != currentBlockHeader.GetEpoch() &&
 		sp.epochStartTrigger.Epoch() == currentBlockHeader.GetEpoch()
 	if isEpochIncorrect {
 		return process.ErrEpochDoesNotMatch
 	}
 
 	isOldEpochAndShouldBeNew := sp.epochStartTrigger.IsEpochStart() &&
-		headerHandler.GetRound() > sp.epochStartTrigger.EpochStartRound()+process.MetaBlockFinality &&
-		headerHandler.GetEpoch() != sp.epochStartTrigger.Epoch()
+		header.GetRound() > sp.epochStartTrigger.EpochStartRound()+process.MetaBlockFinality+1 &&
```
AdoAdoAdo: what is the +1 for?

cmd/node/factory/structs.go
```diff
@@ -709,8 +708,8 @@ func newEpochStartTrigger(
 			Storage:         args.data.Store,
 			RequestHandler:  requestHandler,
 			Epoch:           args.startEpochNum,
-			Validity:        0,
-			Finality:        0,
+			Validity:        process.MetaBlockFinality,
```
AdoAdoAdo: what is this field for and why is it initialized with MetaBlockFinality?

epochStart/shardchain/trigger.go
```diff
@@ -214,7 +214,7 @@ func (t *trigger) updateTriggerFromMeta(metaHdr *block.MetaBlock, hdrHash []byte
 
 	for hash, meta := range t.mapEpochStartHdrs {
 		canActivateEpochStart := t.checkIfTriggerCanBeActivated(hash, meta)
-		if canActivateEpochStart && t.epoch != meta.Epoch {
+		if canActivateEpochStart && t.epoch+1 == meta.Epoch {
```
AdoAdoAdo: why was this changed to t.epoch+1 == meta.Epoch, was it possible that shard would trigger when it was behind meta with more than one epoch? or ahead? 
otherwise t.epoch != meta.Epoch should have been ok as well.

config/config.go
```diff
@@ -56,13 +56,14 @@ type NTPConfig struct {
 
 // EpochStartConfig will hold the configuration of EpochStart settings
 type EpochStartConfig struct {
-	MinRoundsBetweenEpochs int64
-	RoundsPerEpoch         int64
+	MinRoundsBetweenEpochs int
```
raduchis: this will be int32 on 32 bit and int64 on 64 bit

epochStart/interface.go
```diff
@@ -9,14 +9,14 @@ import (
...
+	ForceEpochStart(round uint64) error
 	IsEpochStart() bool
 	Epoch() uint32
 	ReceivedHeader(header data.HeaderHandler)
-	Update(round int64)
+	Update(round uint64)
 	EpochStartRound() uint64
 	EpochStartMetaHdrHash() []byte
-	Processed()
+	Processed(header data.HeaderHandler)
```
raduchis: The "Processed" name is not very suggestive.

cmd/node/factory/structs.go
raduchis: is it used as the number of blocks that give the finality?

process/block/shardblock.go
sasurobert: metachain accepts blocks from shard even in epoch change if their round is smaller than epoch change round + finality + 1. The same is here. Maybe the leader did not get the metaheaders, so he doesn't no yet there was an epoch change.

cmd/node/factory/structs.go
sasurobert: added verification that metablock is valid as final as well. you must have at least validity num of blocks before the epoch announcing metablock, and finality number of blocks after.

epochStart/shardchain/trigger.go
sasurobert: in tests, created scenarios when shard is with 2-3 epoch behind and wants to catch up.

config/config.go
sasurobert: done

epochStart/interface.go
sasurobert: done

epochStart/metachain/trigger.go
```diff
@@ -91,42 +90,53 @@ func (t *trigger) ForceEpochStart(round int64) error {
...
-		t.epochStartTime = t.rounder.TimeStamp()
 		t.isEpochStart = true
-		t.epochStartRound = t.currentRound
+		t.currEpochStartRound = t.currentRound
 	}
 }
 
 // Processed sets start of epoch to false and cleans underlying structure
-func (t *trigger) Processed() {
+func (t *trigger) Processed(header data.HeaderHandler) {
```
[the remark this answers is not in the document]
sasurobert: done

cmd/node/config/config.toml
```diff
@@ -231,6 +242,10 @@
 [Marshalizer]
    Type = "json"
 
+[EpochStartConfig]
+    MinRoundsBetweenEpochs = 30
+    RoundsPerEpoch         = 50
```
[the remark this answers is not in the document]
sasurobert: still left it like this. It is easier for Luci to test directly from this config.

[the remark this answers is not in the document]
AdoAdoAdo: then add a todo to change this to 20k or how many rounds we want an epoch to have

process/block/shardblock.go
AdoAdoAdo: let's name this instead and take it either as from config if this is dependent on other config parameters and may change or have a constant.

epochStart/shardchain/trigger.go
AdoAdoAdo: add a variable/config param instead of a magic number so we know what it is.

process/block/shardblock.go
sasurobert: it is named. Added constant in process.

epochStart/shardchain/trigger.go
sasurobert: it is not in config. as you want to go from epoch to epoch, so the change has to be plus 1.

cmd/node/factory/structs.go
raduchis: Yes but they are set o the same value MetaBlockFinality. Is it ok?

sasurobert: done

PR #704, opened 2019-11-22
fixed bug that cause seldom application freeze on startup

iulianpascalau:
fixed bug that cause seldom application freeze on startup
added log.trace prints on main file for easier future debugging sessions
optimized signal waiting go routine on main
removed with-ui flag
removed log folder creation
removed unused config.toml setting

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -13,10 +13,8 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
-	"runtime/debug"
 	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
@@ -55,7 +53,6 @@ import (
 )
 
 const (
-	defaultLogPath     = "logs"
 	defaultStatsPath   = "stats"
 	defaultDBPath      = "db"
 	defaultEpochString = "Epoch"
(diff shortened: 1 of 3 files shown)


cmd/node/main.go
```diff
@@ -818,27 +821,21 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	ef.SetTpsBenchmark(tpsBenchmark)
 	ef.SetConfig(efConfig)
 
-	wg := sync.WaitGroup{}
-	go ef.StartBackgroundServices(&wg)
-	wg.Wait()
+	log.Trace("starting background services")
+	ef.StartBackgroundServices()
```
raduchis: The problem was with the starting of backgroundservices on a goroutine?

iulianpascalau: Not quite, the ef.StartBackgroundServices that started on a go routing had and wg.Add on the first line. Starting the rest API web server is blocking so wg.Done will never be called. Depending on how fast the go routine was started, the first instruction after that would either have been wg.Add or wg.Wait. If wg.Wait would have triggered first, since there was no prior wg.Add call, it would have returned immediately, otherwise it would wait indefinitely.

PR #707, opened 2019-11-25
[Net. Sharding] Make the number of priority bits configurable

traian-schiau:
This pr should be merged after en-4996.. change the destination after 4996 is merged


traian-schiau:
redon

PR #712, opened 2019-11-27
trie checkpoint

BeniaminDrasovean:
Implemented trie checkpoint.
Recreate trie also checks snapshots for root hash.
Fix golangcibot comments from feat/storage-trie-pruning.

### DIFF
--- a/data/trie/branchNode_test.go
+++ b/data/trie/branchNode_test.go
@@ -51,6 +51,8 @@ func newEmptyTrie(marsh marshal.Marshalizer, hsh hashing.Hasher) *patriciaMerkle
 	evictionWaitListSize := 100
 	evictionWaitList, _ := mock.NewEvictionWaitingList(evictionWaitListSize, mock.NewMemDbMock(), marsh)
 
+	// TODO change this initialization of the persister  (and everywhere in this package)
+	// by using a persister factory
 	tempDir, _ := ioutil.TempDir("", "leveldb_temp")
 	cfg := config.DBConfig{
 		FilePath:          tempDir,
--- a/data/trie/branchNode_test.go
+++ b/data/trie/branchNode_test.go
@@ -9,14 +9,13 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ElrondNetwork/elrond-go/storage"
-
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/data"
 	"github.com/ElrondNetwork/elrond-go/data/mock"
-	protobuf "github.com/ElrondNetwork/elrond-go/data/trie/proto"
+	"github.com/ElrondNetwork/elrond-go/data/trie/proto"
 	"github.com/ElrondNetwork/elrond-go/hashing"
 	"github.com/ElrondNetwork/elrond-go/marshal"
+	"github.com/ElrondNetwork/elrond-go/storage"
 	"github.com/ElrondNetwork/elrond-go/storage/lrucache"
 	"github.com/ElrondNetwork/elrond-go/storage/memorydb"
 	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
--- a/data/trie/branchNode_test.go
+++ b/data/trie/branchNode_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ElrondNetwork/elrond-go/storage"
+
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/data"
 	"github.com/ElrondNetwork/elrond-go/data/mock"
@@ -64,7 +66,7 @@ func newEmptyTrie(marsh marshal.Marshalizer, hsh hashing.Hasher) *patriciaMerkle
 
 	return &patriciaMerkleTrie{
 		db:                    db,
-		snapshots:             make([]data.DBWriteCacher, 0),
+		snapshots:             make([]storage.Persister, 0),
 		snapshotDbCfg:         cfg,
 		dbEvictionWaitingList: evictionWaitList,
 		marshalizer:           marsh,
--- a/data/trie/branchNode_test.go
+++ b/data/trie/branchNode_test.go
@@ -9,9 +9,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ElrondNetwork/elrond-go/storage/memorydb"
-	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
-
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/data"
 	"github.com/ElrondNetwork/elrond-go/data/mock"
@@ -19,6 +16,8 @@ import (
 	"github.com/ElrondNetwork/elrond-go/hashing"
 	"github.com/ElrondNetwork/elrond-go/marshal"
 	"github.com/ElrondNetwork/elrond-go/storage/lrucache"
+	"github.com/ElrondNetwork/elrond-go/storage/memorydb"
+	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
 	"github.com/stretchr/testify/assert"
 )
 
--- a/data/trie/branchNode_test.go
+++ b/data/trie/branchNode_test.go
@@ -3,10 +3,14 @@ package trie
 import (
 	"encoding/hex"
 	"fmt"
+	"io/ioutil"
 	"reflect"
 	"strconv"
 	"testing"
 	"time"
+
+	"github.com/ElrondNetwork/elrond-go/storage/memorydb"
+	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
 
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/data"
@@ -44,9 +48,27 @@ func getBnAndCollapsedBn(marshalizer marshal.Marshalizer, hasher hashing.Hasher)
 }
 
 func newEmptyTrie(marsh marshal.Marshalizer, hsh hashing.Hasher) *patriciaMerkleTrie {
-	cacheSize := 100
-	tr, _ := NewTrie(mock.NewMemDbMock(), marsh, hsh, mock.NewMemDbMock(), cacheSize, config.DBConfig{})
-	return tr
+	db := memorydb.New()
+	evictionWaitListSize := 100
+	evictionWaitList, _ := mock.NewEvictionWaitingList(evictionWaitListSize, mock.NewMemDbMock(), marsh)
+
+	tempDir, _ := ioutil.TempDir("", "leveldb_temp")
+	cfg := config.DBConfig{
+		FilePath:          tempDir,
+		Type:              string(storageUnit.LvlDbSerial),
+		BatchDelaySeconds: 1,
+		MaxBatchSize:      1,
+		MaxOpenFiles:      10,
+	}
+
+	return &patriciaMerkleTrie{
+		db:                    db,
+		snapshots:             make([]data.DBWriteCacher, 0),
+		snapshotDbCfg:         cfg,
+		dbEvictionWaitingList: evictionWaitList,
+		marshalizer:           marsh,
+		hasher:                hsh,
+	}
 }
 
 func initTrie() *patriciaMerkleTrie {
--- a/data/trie/errors.go
+++ b/data/trie/errors.go
@@ -51,3 +51,9 @@ var ErrInvalidHash = errors.New("the received hash is invalid")
 
 // ErrTimeIsOut signals that time is out
 var ErrTimeIsOut = errors.New("time is out")
+
+// ErrTrieNotCommited signals that the trie is dirty
+var ErrTrieNotCommited = errors.New("trie is not commited")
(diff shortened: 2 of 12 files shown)


data/trie/patriciaMerkleTrie.go
```diff
@@ -648,16 +711,17 @@ func (tr *patriciaMerkleTrie) GetSerializedNodes(rootHash []byte, maxBuffToSend
...
+	_, err := tr.db.Get(rootHash)
+
 	hashPresent := err == nil
 	if hashPresent {
 		return tr.db
 	}
 
 	for i := range tr.snapshots {
-		encNode, err = tr.snapshots[i].Get(rootHash)
+		_, err := tr.snapshots[i].Get(rootHash)
```
AdoAdoAdo: why :=?
variable already declared

data/trie/patriciaMerkleTrie.go
```diff
@@ -445,37 +449,71 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
-	db, err := storageUnit.NewDB(
-		storageUnit.DBType(tr.snapshotDbCfg.Type),
-		path.Join(tr.snapshotDbCfg.FilePath, strconv.Itoa(tr.snapshotId)),
-		tr.snapshotDbCfg.BatchDelaySeconds,
-		tr.snapshotDbCfg.MaxBatchSize,
-		tr.snapshotDbCfg.MaxOpenFiles,
-	)
+	db, err := tr.newSnapshotDb()
 	if err != nil {
+		tr.snapshotInProgress = false
```
AdoAdoAdo: can you instead defer the setting of snapshot in progress to false ? and remove this from all returns?

data/trie/patriciaMerkleTrie.go
```diff
@@ -445,37 +449,71 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
+func (tr *patriciaMerkleTrie) Snapshot() error {
+	tr.mutOperation.Lock()
+	defer tr.mutOperation.Unlock()
+
+	if tr.root.isDirty() {
+		return ErrTrieNotCommited
+	}
+
+	if tr.snapshotInProgress {
+		return ErrSnapshotInProgress
```
iulianpascalau: Since all tr.snapshotInProgress calls are protected by tr.mutOperation these if branches that return ErrSnapshotInProgress are technically unreachable so we might drop tr.snapshotInProgress. The only use case remain in snapshot function L593. 
You should take into account that there is an edge case that might cause repetitive Snapshots triggers even though the previous snapshots are not done (caused by the L593 in conjunction with L596-L603 and L498-L554). Is this the intended behavior?

data/trie/patriciaMerkleTrie.go
```diff
@@ -515,6 +556,24 @@ func (tr *patriciaMerkleTrie) Snapshot() error {
 	return nil
 }
 
+func (tr *patriciaMerkleTrie) newSnapshotDb() (data.DBWriteCacher, error) {
+	db, err := storageUnit.NewDB(
+		storageUnit.DBType(tr.snapshotDbCfg.Type),
+		path.Join(tr.snapshotDbCfg.FilePath, strconv.Itoa(tr.snapshotId)),
+		tr.snapshotDbCfg.BatchDelaySeconds,
+		tr.snapshotDbCfg.MaxBatchSize,
+		tr.snapshotDbCfg.MaxOpenFiles,
```
iulianpascalau: Warning: we can reach the 1024 opened files limit by using this mechanism if checkpoints are done too often.

data/trie/patriciaMerkleTrie.go
```diff
@@ -445,37 +449,71 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
 	if err != nil {
+		tr.snapshotInProgress = false
 		return err
 	}
-	tr.snapshotId++
-
-	tr.snapshots = append(tr.snapshots, db)
 
 	if len(tr.snapshots) > maxSnapshots {
 		dbUniqueId := strconv.Itoa(tr.snapshotId - len(tr.snapshots))
```
iulianpascalau: L522-523 you remove an old snapshot. Why not calling Close before erasing physical folder?

data/trie/patriciaMerkleTrie.go
```diff
@@ -515,6 +556,24 @@ func (tr *patriciaMerkleTrie) Snapshot() error {
...
+		path.Join(tr.snapshotDbCfg.FilePath, strconv.Itoa(tr.snapshotId)),
+		tr.snapshotDbCfg.BatchDelaySeconds,
+		tr.snapshotDbCfg.MaxBatchSize,
+		tr.snapshotDbCfg.MaxOpenFiles,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	tr.snapshotId++
```
iulianpascalau: tr.snapshot is always 0 when restarting the node. Is this the intended behavior? It might overwrite previously saved checkpoint data and be unable to serve other peers with checkpoint trie nodes.

data/trie/errors.go
```diff
@@ -51,3 +51,9 @@ var ErrInvalidHash = errors.New("the received hash is invalid")
 
 // ErrTimeIsOut signals that time is out
 var ErrTimeIsOut = errors.New("time is out")
+
+// ErrTrieNotCommited signals that the trie is dirty
```
SebastianMarian: committed

data/trie/node_test.go
```diff
@@ -653,61 +653,27 @@ func TestTrieAddHashesToOldHashes(t *testing.T) {
 func TestRecreateTrieFromSnapshotDb(t *testing.T) {
```
SebastianMarian: Remove redundant alias protobuf from imports

data/trie/patriciaMerkleTrie.go
```diff
@@ -445,37 +449,71 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Snapshot creates a new database in which the current state of the trie is saved.
+// If the maximum number of snapshots has been reached, the oldest snapshot is removed.
+func (tr *patriciaMerkleTrie) Snapshot() error {
```
iulianpascalau: This is not called at the current moment. An integration PR will follow?

data/trie/patriciaMerkleTrie.go
SebastianMarian: +1. Also in Checkpoint method

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Removed.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: I can not use defer, because of the goroutine on line 554. I don't want to set snapshotInProgress to false when this function has finished execution, but when the snapshot() function that runs on a goroutine on line 554 will be done.

BeniaminDrasovean: In the checkpoint method I can and will use defer.

BeniaminDrasovean: Used defer on the Checkpoint method, but I left the `tr.snapshotInProgress = false` from L484, because the snapshoting process is done before executing L486-L491.

data/trie/branchNode_test.go
```diff
@@ -44,9 +47,27 @@ func getBnAndCollapsedBn(marshalizer marshal.Marshalizer, hasher hashing.Hasher)
...
 
 func newEmptyTrie(marsh marshal.Marshalizer, hsh hashing.Hasher) *patriciaMerkleTrie {
-	cacheSize := 100
-	tr, _ := NewTrie(mock.NewMemDbMock(), marsh, hsh, mock.NewMemDbMock(), cacheSize, config.DBConfig{})
-	return tr
+	db := memorydb.New()
+	evictionWaitListSize := 100
+	evictionWaitList, _ := mock.NewEvictionWaitingList(evictionWaitListSize, mock.NewMemDbMock(), marsh)
+
+	tempDir, _ := ioutil.TempDir("", "leveldb_temp")
```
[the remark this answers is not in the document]
BeniaminDrasovean: Added TODO.

data/trie/errors.go
```diff
@@ -51,3 +51,9 @@ var ErrInvalidHash = errors.New("the received hash is invalid")
...
+var ErrTrieNotCommited = errors.New("trie is not commited")
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: There are cases where the if statements are reached, and used. For example, if I call `Snapshot()`, `tr.snapshotInProgress = true` will be set, but the mutex will be released at the end of this function, while  `tr.snapshotInProgress = false` will be set in the `go tr.snapshot(newTrie, db)` function call on line 555. If, before the setting of  `tr.snapshotInProgress = false` from the function that runs on a goroutine, we call `Snapshot()` again, we will acquire the released mutex, but when we reach the if from L506, the snapshotInProgress was not set to false, so we can not continue with the process of snapshoting.

BeniaminDrasovean: Technically, the snapshoting process has finished on L594. The pruning was buffered while snapshoting, but now (L594), thet the snapshot was finished, I can begin pruning from the bufferd values (L596-L603). If another snapshot is called before finishing the operations from L596-L603, the new snapshot will not interfere with these operations.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: The only case in which the Checkpoint creates a new db, is if there is no snapshot db (see L469). This is the case if the Checkpoint is called, and no Snapshot was called ever before. Only `Snapshot()` creates new db, so only one new db will be created each epoch.

BeniaminDrasovean: Changed the code a bit, so if Checkpoint is called and no Snapshot was called before, Checkpoint will call Snapshot.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Changed, so now before creating a db, we check if the path exists.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Yes. Snapshot will be called at the beginning of a new epoch, so the endOfEpoch is needed in order to call this function. Also for the checkpoint, an integration PR will follow.

data/trie/errors.go
BeniaminDrasovean: Changed.

data/trie/node_test.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Changed the interface that the snapshots implement (from DBWriteCacher to Persister), and used the Close() method. The interface change was needed because the trie db also uses DBWriteCacher, but the trie db is instantiated as a storage unit, and the snapshots as Persisters. The storage unit interface has DestroyUnit() whereas persister has Close(), so I couldn't add a new method to the DBWriteCacer interface and also expect to implement bothPersister and StorageUnit interfaces.

iulianpascalau: Great :+1:

data/trie/patriciaMerkleTrie.go
```diff
@@ -517,6 +516,12 @@ func (tr *patriciaMerkleTrie) Snapshot() error {
...
 		dbUniqueId := strconv.Itoa(tr.snapshotId - len(tr.snapshots))
+
+		err = tr.snapshots[0].Close()
+		if err != nil {
+			tr.snapshotInProgress = false
+			return err
+		}
 		tr.snapshots = tr.snapshots[1:]
 
 		removePath := path.Join(tr.snapshotDbCfg.FilePath, dbUniqueId)
```
AdoAdoAdo: do we have a test to validate that we are removing the right db?
otherwise, instead of computing the id based on size and current snapshot id I would store it (or the dbpath) in the snapshot list and just take it from there

BeniaminDrasovean: Yes, there is a test that verifies that the right db was removed. (TestDeleteOldSnapshots() from `node_test.go`)

data/trie/patriciaMerkleTrie.go
```diff
@@ -516,6 +516,12 @@ func (tr *patriciaMerkleTrie) Snapshot() error {
 
 	if len(tr.snapshots) > maxSnapshots {
 		dbUniqueId := strconv.Itoa(tr.snapshotId - len(tr.snapshots))
+
+		err = tr.snapshots[0].Close()
+		if err != nil {
+			tr.snapshotInProgress = false
```
SebastianMarian: You can use var err error and a defer at line 508 which set tr.snapshotInProgress to false if err != nil to avoid setting it on false for each return in this method

BeniaminDrasovean: Never thought of this. Changed.

PR #713, opened 2019-11-27
Add connected peers classification metric

iulianpascalau:
added metric for peer connections type
added prints in logger
modified an error in topicResolver for a better debugging session
added dedicated route for p2p status
p2pstatus route can return the connected peers (full addresses) split on classification


statusHandler/statusMetricsProvider.go
```diff
@@ -95,12 +96,44 @@ func (nd *statusMetrics) Close() {
...
 	})
 
-	return statusMetricsMap, nil
+	return statusMetricsMap
+}
+
+// StatusMetricsMapWithoutP2P will return the non-p2p metrics in a map
+func (nd *statusMetrics) StatusMetricsMapWithoutP2P() map[string]interface{} {
+	statusMetricsMap := make(map[string]interface{})
+	nd.nodeMetrics.Range(func(key, value interface{}) bool {
```
camilbancioiu: `sync.Map.Range()` stops iterating over the key-value pairs when the `func` provided as argument will return `false`, but the one provided here never seems to return `false`. Shouldn't the line `if string.Contains() ... ` be written as a stop-condition?

statusHandler/statusMetricsProvider.go
```diff
@@ -95,12 +96,44 @@ func (nd *statusMetrics) Close() {
...
+		return true
+	})
+
+	return statusMetricsMap
+}
+
+// StatusP2pMetricsMap will return the p2p metrics in a map
+func (nd *statusMetrics) StatusP2pMetricsMap() map[string]interface{} {
+	statusMetricsMap := make(map[string]interface{})
+	nd.nodeMetrics.Range(func(key, value interface{}) bool {
```
camilbancioiu: `sync.Map.Range()` stops iterating over the key-value pairs when the `func` provided as argument will return `false`, but the one provided here never seems to return `false`. Shouldn't the line `if string.Contains() ... ` be written as a stop-condition?

statusHandler/statusMetricsProvider.go
iulianpascalau: Returning true each time will cause the range function iterate over all values stored and then stop (by itself). There were no need to returning false in any case.

statusHandler/statusMetricsProvider.go
iulianpascalau: Answered above

iulianpascalau: Answered above

statusHandler/statusMetricsProvider.go
```diff
@@ -95,12 +96,44 @@ func (nd *statusMetrics) Close() {
 }
 
 // StatusMetricsMap will return all metrics in a map
-func (nd *statusMetrics) StatusMetricsMap() (map[string]interface{}, error) {
+func (nd *statusMetrics) StatusMetricsMap() map[string]interface{} {
```
bogdan-rosianu: This is not used anymore. I assume that you left it in case we need it for later uses where we would like all the metrics in one place

statusHandler/statusMetricsProvider.go
```diff
@@ -95,12 +96,44 @@ func (nd *statusMetrics) Close() {
...
 		statusMetricsMap[key.(string)] = value
 		return true
 	})
 
-	return statusMetricsMap, nil
+	return statusMetricsMap
+}
+
+// StatusMetricsMapWithoutP2P will return the non-p2p metrics in a map
+func (nd *statusMetrics) StatusMetricsMapWithoutP2P() map[string]interface{} {
```
bogdan-rosianu: rename the receiver to `sm`

api/node/routes_test.go
```diff
@@ -280,6 +283,38 @@ func TestStatusMetrics_ShouldDisplayMetrics(t *testing.T) {
 
 	keyAndValueFoundInResponse := strings.Contains(respStr, key) && strings.Contains(respStr, value)
 	assert.True(t, keyAndValueFoundInResponse)
+
```
bogdan-rosianu: remove empty line

api/node/routes.go
```diff
@@ -102,20 +103,29 @@ func Statistics(c *gin.Context) {
...
 
-	details, err := ef.StatusMetrics().StatusMetricsMap()
-	if err != nil {
-		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+	details := ef.StatusMetrics().StatusMetricsMapWithoutP2P()
+
+	c.JSON(http.StatusOK, gin.H{"details": details})
+}
+
+// P2pStatusMetrics returns the node's p2p statistics exported by an StatusMetricsHandler
```
bogdan-rosianu: by a *

statusHandler/statusMetricsProvider.go
```diff
@@ -95,12 +96,44 @@ func (nd *statusMetrics) Close() {
...
-func (nd *statusMetrics) StatusMetricsMap() (map[string]interface{}, error) {
+func (nd *statusMetrics) StatusMetricsMap() map[string]interface{} {
 	statusMetricsMap := make(map[string]interface{})
 	nd.nodeMetrics.Range(func(key, value interface{}) bool {
 		statusMetricsMap[key.(string)] = value
 		return true
 	})
 
-	return statusMetricsMap, nil
+	return statusMetricsMap
```
bogdan-rosianu: Fix tests on `elrond-go/statusHandler/statusMetricsProvider_test.go` . the teamcity build for tests failed

statusHandler/statusMetricsProvider.go
iulianpascalau: Removed

statusHandler/statusMetricsProvider.go
iulianpascalau: Renamed. All of them.

api/node/routes_test.go
iulianpascalau: removed

api/node/routes.go
iulianpascalau: changed

statusHandler/statusMetricsProvider.go
iulianpascalau: :+1: done

PR #726, opened 2019-11-29
updated to new arwen.

sasurobert:


integrationTests/vm/iele/vmDeploy_test.go
```diff
@@ -37,8 +37,7 @@ func TestVMInvalidSmartContractCodeShouldNotGenerateAccount(t *testing.T) {
 	}
 
 	// tx is not processed due to the invalid sc code
-	err := txProc.ProcessTransaction(tx, round)
-	assert.NotNil(t, err)
+	_ = txProc.ProcessTransaction(tx, round)
```
raduchis: Why the remove for the error test?

process/smartContract/process.go
```diff
@@ -153,48 +187,93 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
...
 }
 
-func (sc *scProcessor) prepareSmartContractCall(tx *transaction.Transaction, acntSnd state.AccountHandler) error {
-	err := sc.argsParser.ParseData(tx.Data)
+func (sc *scProcessor) processIfError(
+	acntSnd state.AccountHandler,
+	tx data.TransactionHandler,
+	returnCode vmcommon.ReturnCode,
+) error {
+	consumedFee := big.NewInt(0).SetUint64(tx.GetGasLimit() * tx.GetGasPrice())
```
raduchis: Do we really want to "eat" the whole txFee if an error happens?

process/smartContract/process.go
```diff
@@ -350,6 +443,9 @@ func (sc *scProcessor) createVMInput(tx *transaction.Transaction) (*vmcommon.VMI
 	vmInput.GasPrice = tx.GasPrice
 	moveBalanceGasConsume := sc.economicsFee.ComputeGasLimit(tx)
 	vmInput.GasProvided = tx.GasLimit - moveBalanceGasConsume
+	if tx.GetGasLimit() < moveBalanceGasConsume {
```
raduchis: Please create an unit test for this

process/smartContract/process.go
```diff
@@ -456,16 +542,16 @@ func (sc *scProcessor) processVMOutput(
...
 
 	if vmOutput.GasRefund.Uint64() > 0 {
 		log.Debug("total gas refunded", "value", vmOutput.GasRefund.Uint64(), "hash", txHash)
 	}
 
 	totalGasRefund := big.NewInt(0)
-	totalGasRefund = totalGasRefund.Add(vmOutput.GasRefund, vmOutput.GasRemaining)
-	sc.gasHandler.SetGasRefunded(totalGasRefund.Uint64(), txHash)
+	totalGasRefund = totalGasRefund.Add(vmOutput.GasRefund, big.NewInt(0).SetUint64(vmOutput.GasRemaining))
+	sc.gasHandler.SetGasRefunded(vmOutput.GasRemaining, txHash)
```
raduchis: There is a computation above for the totalGasRefund, but you are setting just a part of it. Is it ok?

process/smartContract/process.go
```diff
@@ -649,6 +735,10 @@ func (sc *scProcessor) processSCOutputAccounts(outputAccounts []*vmcommon.Output
 		// update the values according to SC output
 		updatedBalance := big.NewInt(0)
 		updatedBalance = updatedBalance.Add(stAcc.Balance, outAcc.BalanceDelta)
+		if updatedBalance.Cmp(big.NewInt(0)) < 0 {
```
raduchis: Don't we have a problem in the vm if the balanceDelta is negative and smaller than the balance? Why would the vm process something like this?

process/smartContract/process.go
```diff
@@ -790,7 +880,7 @@ func (sc *scProcessor) ProcessSmartContractResult(scr *smartContractResult.Smart
 		return process.ErrWrongTypeAssertion
 	}
 
-	storageUpdates, err := sc.argsParser.GetStorageUpdates(scr.Data)
+	storageUpdates, _ := sc.argsParser.GetStorageUpdates(scr.Data)
```
raduchis: No test for the error is needed? If error is not nil, don't we have to do a revert?

integrationTests/vm/iele/vmDeploy_test.go
sasurobert: because it does not return error anymore.

process/smartContract/process.go
sasurobert: gasHandler wants to know only about the actual consumed gas - the gas consumed for processing. The value gasRefund is only for freeing up memory.

process/smartContract/process.go
sasurobert: balanceDelta will be negative for some accounts, in the VM the APIs are protecting from bad behaviour. In case of lottery, when the SC announces the winner, the balanceDelta of the SC will be negative, and the others will be positive.

process/smartContract/process.go
sasurobert: There is no need to test this, as most of the SCR's will not have valid Data for storageUpdates - they will keep other types of data inside.

process/smartContract/process.go
sasurobert: This is how bad processing in handled in eth, it is the most simple way. Otherwise, we have to treat this depending on errors and to add specific code inside Arwen.

process/smartContract/process.go
sasurobert: done.

process/smartContract/process.go
raduchis: ok

process/smartContract/process.go
raduchis: ok

process/smartContract/process.go
raduchis: ok

process/smartContract/process.go
raduchis: ok

process/smartContract/process.go
raduchis: ok

PR #736, opened 2019-12-02
EN-5231 : fix logger problem

miiu96:
fix problem with logger


api/logs/logSender.go
```diff
@@ -103,7 +105,8 @@ func (ls *logSender) waitForPatternMessage() error {
 func (ls *logSender) monitorConnection() {
 	for {
 		mt, _, err := ls.conn.ReadMessage()
-		if mt == websocket.CloseMessage {
+		ls.log.Info("message type", "value", mt)
```
iulianpascalau: Maybe make this logger message of type Trace

api/logs/logSender.go
```diff
@@ -103,7 +105,8 @@ func (ls *logSender) waitForPatternMessage() error {
...
+		if mt == websocket.CloseMessage || mt == disconnectMessage {
 			_ = ls.writer.Close()
```
iulianpascalau: As stated, I think it is better to close the writer if any error is received. Maybe add a defer func on monitorConnection?

api/logs/logSender.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+const disconnectMessage = -1
```
iulianpascalau: instead of this const might check if there is a const in websocket package (like websocket.CloseMessage)

bogdan-rosianu: the constant you mention is unexported so it can't be used

iulianpascalau: ok

api/logs/logSender.go
miiu96: Done.

api/logs/logSender.go
miiu96: Done.

PR #740, opened 2019-12-03
En 5237 check if nil reflect

traian-schiau:
IfNilReflect

### DIFF
--- a/core/check/ifNil.go
+++ b/core/check/ifNil.go
@@ -1,9 +1,24 @@
 package check
 
+import (
+	"reflect"
+)
+
 // IfNil tests if the provided interface pointer or underlying object is nil
 func IfNil(checker NilInterfaceChecker) bool {
 	if checker == nil {
 		return true
 	}
 	return checker.IsInterfaceNil()
+}
+
+// IfNilReflect tests if the provided interface pointer or underlying pointer receiver is nil
+func IfNilReflect(i interface{}) bool {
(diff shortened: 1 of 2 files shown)


core/check/ifNil.go
```diff
@@ -1,9 +1,25 @@
...
+
 // IfNil tests if the provided interface pointer or underlying object is nil
 func IfNil(checker NilInterfaceChecker) bool {
 	if checker == nil {
 		return true
 	}
 	return checker.IsInterfaceNil()
 }
+
+// IfNil tests if the provided interface pointer or underlying pointer receiver is nil
```
miiu96: IfNilReflect tests...

core/check/ifNil.go
```diff
@@ -1,9 +1,25 @@
...
+func IfNilReflect(i interface{}) bool {
+	if v := reflect.ValueOf(i); v.IsValid() {
+		if v.Kind() == reflect.Ptr && v.IsNil() {
+			return true
+		}
+		return false
+	} else {
```
miiu96: maybe you can remove **else**

sasurobert: no need for else. return true

core/check/ifNil.go
sasurobert: +1

traian-schiau: done

core/check/ifNil.go
traian-schiau: done

PR #744, opened 2019-12-03
EN-5153 : extract processed mini blocks map to a struct

miiu96:
-Refactor complex structure(_map[string]map[string]struct{}_)  from shardProcessor in a new type that can be used easier


consensus/mock/blockProcessorMock.go
```diff
@@ -1,6 +1,7 @@
 package mock
 
 import (
+	"github.com/ElrondNetwork/elrond-go/core/complexStructures"
```
bogdan-rosianu: go imports

core/complexStructures/processedMiniBlocks.go
```diff
@@ -0,0 +1,132 @@
+package complexStructures
```
bogdan-rosianu: please rename the directory and the package name.

node/mock/blockProcessorStub.go
```diff
@@ -1,6 +1,7 @@
 package mock
 
 import (
+	"github.com/ElrondNetwork/elrond-go/core/complexStructures"
```
bogdan-rosianu: go imports

process/block/shardblock.go
```diff
@@ -534,12 +533,12 @@ func (sp *shardProcessor) restoreMetaBlockIntoPool(mapMiniBlockHashes map[string
 
 	for metaBlockHash, miniBlockHashes := range mapMetaHashMiniBlockHashes {
 		for _, miniBlockHash := range miniBlockHashes {
-			sp.addProcessedMiniBlock([]byte(metaBlockHash), miniBlockHash)
+			sp.processedMiniBlocks.AddMiniBlockHash([]byte(metaBlockHash), miniBlockHash)
```
bogdan-rosianu: why are you casting to byte array and in the structure you cast back to string ?

core/complexStructures/processedMiniBlocks.go
bogdan-rosianu: also I think you can move this file to the `process/block` directory because it belongs there and this way you will get rid of this bad naming.

core/complexStructures/processedMiniBlocks.go
```diff
@@ -0,0 +1,132 @@
...
+	"sync"
+
+	"github.com/ElrondNetwork/elrond-go/logger"
+	"github.com/ElrondNetwork/elrond-go/process/block/bootstrapStorage"
+)
+
+var log = logger.GetOrCreate("core/complexStructures")
+
+// ProcessedMiniBlocks is used to store all processed min blocks hashes for e meta block hash
+type ProcessedMiniBlocks struct {
```
bogdan-rosianu: Please rename it to something like `ProcessedMiniBlockTracker`

core/complexStructures/processedMiniBlocks.go
```diff
@@ -0,0 +1,132 @@
...
+func (pmb *ProcessedMiniBlocks) GetProcessedMiniBlocksHashes(metaBlockHash []byte) map[string]struct{} {
+	pmb.mutProcessedMiniBlocks.RLock()
+	processedMiniBlocksHashes := pmb.processedMiniBlocks[string(metaBlockHash)]
+	pmb.mutProcessedMiniBlocks.RUnlock()
+
+	return processedMiniBlocksHashes
+}
+
+// IsMiniBlockProcessed will return true if a mini block is processed
+func (pmb *ProcessedMiniBlocks) IsMiniBlockProcessed(metaBlockHash []byte, miniBlockHash []byte) bool {
```
bogdan-rosianu: use `string` as parameters' type because where you use these methods you cast the parameters as `[]byte` and in this structure you cast them back to `string`

consensus/mock/blockProcessorMock.go
miiu96: done.

core/complexStructures/processedMiniBlocks.go
miiu96: done

node/mock/blockProcessorStub.go
miiu96: done.

process/block/shardblock.go
miiu96: done. now string is used.

core/complexStructures/processedMiniBlocks.go
miiu96: done.

core/complexStructures/processedMiniBlocks.go
miiu96: done.

process/block/processedMb/processedMiniBlocks.go
```diff
@@ -7,52 +7,52 @@ import (
 	"github.com/ElrondNetwork/elrond-go/process/block/bootstrapStorage"
 )
 
-var log = logger.GetOrCreate("core/complexStructures")
+var log = logger.GetOrCreate("core/processedMb")
```
bogdan-rosianu: `core` ?

process/block/processedMb/processedMiniBlocks.go
```diff
@@ -7,52 +7,52 @@ import (
...
 
-// ProcessedMiniBlocks is used to store all processed min blocks hashes for e meta block hash
-type ProcessedMiniBlocks struct {
+// ProcessedMiniBlockTracker is used to store all processed min blocks hashes for e meta block hash
```
bogdan-rosianu: mini * 
for **a** meta *

process/block/processedMb/processedMiniBlocks_test.go
```diff
@@ -1,4 +1,4 @@
-package complexStructures
+package processedMb
```
bogdan-rosianu: make the package `processedMb_test` so we can test from "outside"

process/mock/blockProcessorMock.go
```diff
@@ -1,7 +1,7 @@
 package mock
 
 import (
-	"github.com/ElrondNetwork/elrond-go/core/complexStructures"
+	"github.com/ElrondNetwork/elrond-go/process/block/processedMb"
```
bogdan-rosianu: go imports

process/interface.go
```diff
@@ -14,8 +13,9 @@ import (
 	"github.com/ElrondNetwork/elrond-go/data/transaction"
 	"github.com/ElrondNetwork/elrond-go/p2p"
 	"github.com/ElrondNetwork/elrond-go/process/block/bootstrapStorage"
+	"github.com/ElrondNetwork/elrond-go/process/block/processedMb"
 	"github.com/ElrondNetwork/elrond-go/storage"
-	"github.com/ElrondNetwork/elrond-vm-common"
+	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
```
bogdan-rosianu: remove this `vmcommon` alias for the package. it is always inserted by go imports but it is not needed

process/block/processedMb/processedMiniBlocks.go
miiu96: done.

process/block/processedMb/processedMiniBlocks.go
miiu96: done.

process/block/processedMb/processedMiniBlocks_test.go
miiu96: done.

process/mock/blockProcessorMock.go
miiu96: done.

process/interface.go
miiu96: done.

consensus/mock/blockProcessorMock.go
```diff
@@ -29,7 +30,7 @@ func (blProcMock *BlockProcessorMock) SetNumProcessedObj(numObj uint64) {
 
 }
 
-func (blProcMock *BlockProcessorMock) ApplyProcessedMiniBlocks(miniBlocks map[string]map[string]struct{}) {
+func (blProcMock *BlockProcessorMock) ApplyProcessedMiniBlocks(miniBlocks *processedMb.ProcessedMiniBlockTracker) {
```
ccorcoveanu: The `ProcessedMiniBlockTracker` should implement an interface and pass that around, so when we will add more tests in the block/metablock processors we can mock this and have a simpler setup for unit tests.

process/block/processedMb/processedMiniBlocks.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// AddMiniBlockHash will add a miniblock hash
+func (pmb *ProcessedMiniBlockTracker) AddMiniBlockHash(metaBlockHash string, miniBlockHash string) {
+	pmb.mutProcessedMiniBlocks.Lock()
+	defer pmb.mutProcessedMiniBlocks.Unlock()
+
+	miniBlocksProcessed, ok := pmb.processedMiniBlocks[metaBlockHash]
+	if !ok {
+		miniBlocksProcessed := make(MiniBlockHashes)
```
ccorcoveanu: `miniBlocksProcessed = make(MiniBlockHashes)` -> without ':'

process/block/processedMb/processedMiniBlocks.go
```diff
@@ -0,0 +1,135 @@
...
+	delete(pmb.processedMiniBlocks, metaBlockHash)
+	pmb.mutProcessedMiniBlocks.Unlock()
+}
+
+// RemoveMiniBlockHash will remove a mini block hash
+func (pmb *ProcessedMiniBlockTracker) RemoveMiniBlockHash(miniBlockHash string) {
+	pmb.mutProcessedMiniBlocks.Lock()
+	for metaHash, miniBlocksProcessed := range pmb.processedMiniBlocks {
+		_, isProcessed := miniBlocksProcessed[miniBlockHash]
+		if isProcessed {
```
ccorcoveanu: it's safe to delete directly, without checking if it exists before

process/block/processedMb/processedMiniBlocks.go
```diff
@@ -0,0 +1,135 @@
...
+	if !ok {
+		return false
+	}
+
+	_, isProcessed := miniBlocksProcessed[miniBlockHash]
+	return isProcessed
+}
+
+// ConvertProcessedMiniBlocksMapToSlice will convert a map[string]map[string]struct{} in a slice of MiniBlocksInMeta
+func (pmb *ProcessedMiniBlockTracker) ConvertProcessedMiniBlocksMapToSlice() []bootstrapStorage.MiniBlocksInMeta {
```
ccorcoveanu: I don't think this data structure should know about a boot storer. I think this conversion should be done by the boot storer which is the one who knows how it needs the data formatted

process/block/processedMb/processedMiniBlocks.go
miiu96: done.

process/block/processedMb/processedMiniBlocks.go
miiu96: done.

process/block/processedMb/processedMiniBlocks.go
miiu96: i will let it as it is because if I move this method in another file I will need a mutex to protect  ProcessedMiniBlockTracker from concurrency accesses.

consensus/mock/blockProcessorMock.go
miiu96: i will let it as it is because if i will create a interface code will became more complicated

PR #748, opened 2019-12-04
En 4504 db pruning

bogdan-rosianu:
Implemented DB pruning with the following changes:
Added a new storer which will handle:
- multiple persisters
- directories creation and deletion based on an epoch change notification

Added configuration:
- full archive node : no database will be deleted
- num of active persisters : how many persister are kept open at a time
- num of epoch to keep: if node is not full archive, after this number is exceeded, old dbs will be removed

Applied the storer to almost all units, excluding trie related ones and heartbeat

Changes in termui. some reordering in order to make room for 'epoch number' information

Database factory


storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+type persisterData struct {
+	persister storage.Persister
+	path      string
+}
+
+// PruningStorer represents a storer which creates a new persister for each epoch and removes older persisters
+type PruningStorer struct {
+	lock                  sync.RWMutex
+	fullArchive           bool
+	batcher               storage.Batcher
```
golangcibot: `batcher` is unused (from `structcheck`)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+
+	v, ok := ps.cacher.Get(key)
+	var err error
+
+	if !ok {
+		// not found in cache
+		// search it in second persistence medium
+		found := false
+		for idx := uint32(0); (idx < ps.numOfActivePersisters) && (idx < uint32(len(ps.persisters))); idx++ {
+			if ps.bloomFilter == nil || ps.bloomFilter.MayContain(key) == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `ps.bloomFilter.MayContain(key)` (from `gosimple`)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+func (ps *PruningStorer) Has(key []byte) error {
+	ps.lock.RLock()
+	defer ps.lock.RUnlock()
+
+	has := ps.cacher.Has(key)
+	if has {
+		return nil
+	}
+
+	if ps.bloomFilter == nil || ps.bloomFilter.MayContain(key) == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `ps.bloomFilter.MayContain(key)` (from `gosimple`)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+			res, err := ps.getFromClosedPersisters(key)
+			if err != nil {
+				log.Debug("get from closed persisters", "error", err.Error())
+			} else {
+				ps.cacher.Put(key, res)
+				return res, nil
+			}
+		}
+		if !found {
+			return nil, errors.New(fmt.Sprintf("key %s not found in %s",
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+		if err != nil {
+			log.Debug("pruning db: destroy",
+				"error", err.Error())
+			continue
+		}
+		numOfPersistersRemoved++
+	}
+
+	if numOfPersistersRemoved != totalNumOfPersisters {
+		return errors.New(fmt.Sprintf("couldn't destroy all persisters. %d/%d destroyed",
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+	}
+}
+
+// getNewFilePath will return the file path for the new epoch. It uses regex to change the default path
+func (ps *PruningStorer) getNewFilePath(epoch uint32) string {
+	// TODO: the path will be provided by a path naming component
+	// using a regex to match the epoch directory name as placeholder followed by at least one digit
+	// in a string which contains Epoch_X it will replace X with the given epoch number
+	rg := regexp.MustCompile("Epoch_\\d+")
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

storage/pruning/pruningStorer_test.go
```diff
@@ -0,0 +1,286 @@
...
+	expectedRes := "db/Epoch_7/Shard_2"
+	replacementEpoch := "Epoch_7"
+
+	var testPaths []string
+	testPaths = append(testPaths, "db/Epoch_22282493984354/Shard_2")
+	testPaths = append(testPaths, "db/Epoch_0/Shard_2")
+	testPaths = append(testPaths, "db/Epoch_02/Shard_2")
+	testPaths = append(testPaths, "db/Epoch_99999999999999999999999999999999999999999999/Shard_2")
+
+	rg := regexp.MustCompile("Epoch_\\d+")
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	defer func() {
+		err := f.Close()
+		if err != nil {
+			log.Debug("pruning db - file close", "error", err.Error())
+		}
+	}()
+
+	_, err = f.Readdirnames(1) // Or f.Readdir(1)
+	if err == io.EOF {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cmd/node/factory/structs.go
```diff
@@ -981,93 +991,176 @@ func createShardDataStoreFromConfig(
 		}
 	}()
 
-	txUnit, err = storageUnit.NewStorageUnitFromConf(
-		getCacherFromConfig(config.TxStorage.Cache),
-		getDBFromConfig(config.TxStorage.DB, uniqueID),
-		getBloomFromConfig(config.TxStorage.Bloom))
+	fullArchiveMode := config.StoragePruning.FullArchive
```
sasurobert: this function might be put inside a factory - and deleted from structs.go

cmd/node/factory/structs.go
```diff
@@ -981,93 +991,176 @@ func createShardDataStoreFromConfig(
...
+	numOfEpochsToKeep := uint32(config.StoragePruning.NumOfEpochsToKeep)
+	numOfActivePersisters := uint32(config.StoragePruning.NumOfActivePersisters)
+
+	txUnitStorerArgs := &pruning.PruningStorerArgs{
```
sasurobert: you could use only one PruningStorerArgs, and change only those things which are different.

cmd/node/main.go
```diff
@@ -540,6 +560,14 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	}
 
 	log.Trace("creating nodes coordinator")
+	if ctx.IsSet(isNodefullArchive.Name) {
+		generalConfig.StoragePruning.FullArchive = ctx.GlobalBool(isNodefullArchive.Name)
+	}
+	if ctx.IsSet(numEpochsToSave.Name) {
```
sasurobert: you have to add verification that these configs are correct, if the node is a validator than epochsToSave must be higher than 2.

config/config.go
```diff
@@ -118,6 +119,13 @@ type NodeConfig struct {
 	TargetPeerCount int
 }
 
+// StoragePruningConfig will hold settings relates to storage pruning
+type StoragePruningConfig struct {
+	FullArchive           bool
+	NumOfEpochsToKeep     uint64
```
sasurobert: "Of" is not needed.

process/block/shardblock.go
```diff
@@ -819,7 +819,7 @@ func (sp *shardProcessor) CommitBlock(
 		sp.specialAddressHandler.IsCurrentNodeInConsensus(),
 		display.DisplayByteSlice(headerHash),
 		highestFinalBlockNonce,
-		headerMeta.GetNonce(),
+		headerMeta,
```
sasurobert: metaHeader ?

storage/badgerdb/badgerdb.go
```diff
@@ -208,6 +208,14 @@ func (s *DB) Destroy() error {
 	return err
 }
 
+// DestroyClosed removes the already closed storage medium stored data
+func (s *DB) DestroyClosed() error {
+	s.dbClosed <- struct{}{}
+	err := os.RemoveAll(s.path)
```
sasurobert: you can return directly os.RemoveAll(s.path)

storage/factory/persisterFactory.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(path) < 0 {
+		return nil, errors.New("invalid file path")
+	}
+
+	var db storage.Persister
+	var err error
+
+	switch storageUnit.DBType(df.dbType) {
+	case storageUnit.LvlDB:
+		db, err = leveldb.NewDB(path, df.batchDelaySeconds, df.maxBatchSize, df.maxOpenFiles)
```
sasurobert: you can return directly at every case.

storage/leveldb/leveldb.go
```diff
@@ -206,6 +206,20 @@ func (s *DB) Destroy() error {
...
+// DestroyClosed removes the already closed storage medium stored data
+func (s *DB) DestroyClosed() error {
+	s.mutBatch.Lock()
+	s.batch.Reset()
+	s.sizeBatch = 0
+	s.mutBatch.Unlock()
+
+	s.dbClosed <- struct{}{}
+
+	err := os.RemoveAll(s.path)
```
sasurobert: you can return here.

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"io"
+	"os"
+	"strings"
+)
+
+func removeDirectoryIfEmpty(path string) {
+	elementsSplitBySeparator := strings.Split(path, string(os.PathSeparator))
+	// the structure is this way :
```
sasurobert: move all commentary in top of function.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+	db, err = args.PersisterFactory.Create(filePath)
+
+	var persisters []*persisterData
+	persisters = append(persisters, &persisterData{
+		persister: db,
+		path:      filePath,
+	})
+
+	if reflect.DeepEqual(args.BloomFilterConf, storageUnit.BloomConfig{}) {
```
sasurobert: verify only one parameter from the config to decide whether you want with bloom config or not.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,414 @@
...
+		return pdb, nil
+	}
+
+	bf, err = storageUnit.NewBloomFilter(args.BloomFilterConf)
+	if err != nil {
+		return nil, err
+	}
+
+	pdb := &PruningStorer{
+		identifier:            args.Identifier,
```
sasurobert: duplicated code - make the PruningStorer without bloomFilter first. Test if bloom filter is needed and create then, add it to pruningStorer.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,414 @@
...
+		if err != nil {
+			log.Debug("pruning db: destroy",
+				"error", err.Error())
+			continue
+		}
+		numOfPersistersRemoved++
+	}
+
+	if numOfPersistersRemoved != totalNumOfPersisters {
+		return fmt.Errorf("couldn't destroy all persisters. %d/%d destroyed",
```
sasurobert: log.Error then return constant error ?

cmd/node/factory/structs.go
bogdan-rosianu: done. created a factory for shard and meta storage services

cmd/node/factory/structs.go
bogdan-rosianu: didn't change this. half of the params are unique so I'd have to change them always. Also I think it is safer that we have a new arguments instance for all the storers

cmd/node/main.go
bogdan-rosianu: done. also changes some conditions. numActivePersisters should be set in all the cases

config/config.go
bogdan-rosianu: done

process/block/shardblock.go
bogdan-rosianu: done

storage/badgerdb/badgerdb.go
bogdan-rosianu: done

storage/factory/persisterFactory.go
bogdan-rosianu: done

storage/leveldb/leveldb.go
bogdan-rosianu: done

storage/pruning/fileHandling.go
bogdan-rosianu: done

storage/pruning/pruningStorer.go
bogdan-rosianu: done. checked for bloom's size. if the struct is empty (so no config for bloom filter), the default value will be 0

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,49 @@
...
+func removeDirectoryIfEmpty(path string) {
+	elementsSplitBySeparator := strings.Split(path, string(os.PathSeparator))
+
+	epochDirectory := ""
+	for idx := 0; idx < len(elementsSplitBySeparator)-2; idx++ {
+		epochDirectory += elementsSplitBySeparator[idx] + string(os.PathSeparator)
+	}
+
+	shardDirectory := epochDirectory + elementsSplitBySeparator[len(elementsSplitBySeparator)-2]
+	if isDirectoryEmpty(shardDirectory) {
```
miiu96: you can refactor this part of code !isDirectoryEmpy -> return and after that continue with your code

bogdan-rosianu: I don't think this is needed because the function doesn't return anything so the code between the brackets will only be executed if the directory is empty. otherwise, no other action is needed

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+		return nil
+	}
+
+	if ps.bloomFilter == nil || ps.bloomFilter.MayContain(key) {
+		persister, ok := ps.persistersMapByEpoch[epoch]
+		if !ok {
+			return storage.ErrKeyNotFound
+		}
+
+		return persister.persister.Has(key)
```
AdoAdoAdo: how do you know the persister is opened?

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+	return storage.ErrKeyNotFound
+}
+
+// Remove removes the data associated to the given key from both cache and persistence medium
+func (ps *PruningStorer) Remove(key []byte) error {
+	ps.lock.Lock()
+	defer ps.lock.Unlock()
+
+	ps.cacher.Remove(key)
+	return ps.activePersisters[0].persister.Remove(key)
```
AdoAdoAdo: how do you know the key is in the first active persister?

it should be removed from any of the active persisters iterating from most recent to older and break on found.
I could think of the situation where you are right after the start of epoch event and you would need to rollback your chain to previous epoch blocks.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+	if ps.bloomFilter != nil {
+		ps.bloomFilter.Clear()
+	}
+
+	ps.cacher.Clear()
+
+	var err error
+	numOfPersistersRemoved := 0
+	totalNumOfPersisters := len(ps.activePersisters)
+	for _, persister := range ps.activePersisters {
```
AdoAdoAdo: what about the non active persisters?
do you destroy them

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+		if err != nil {
+			log.Debug("pruning db: destroy",
+				"error", err.Error())
+			continue
+		}
+		numOfPersistersRemoved++
+	}
+
+	if numOfPersistersRemoved != totalNumOfPersisters {
+		log.Debug("error destroying pruning db",
```
AdoAdoAdo: this will always return error if the active persisters < num total persisters

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+	// recent activePersisters have to he closed for both scenarios: full archive or not
+	if ps.numOfActivePersisters < uint32(len(ps.activePersisters)) {
+		persisterToClose := ps.activePersisters[ps.numOfActivePersisters]
+		err := persisterToClose.persister.Close()
+		if err != nil {
+			log.Error("error closing persister", "error", err.Error(), "id", ps.identifier)
+			return err
+		}
+		// remove it from the active persisters slice
+		ps.activePersisters = append(ps.activePersisters[:ps.numOfActivePersisters], ps.activePersisters[ps.numOfActivePersisters+1:]...)
```
AdoAdoAdo: why do you need to append older persisters? [numOfActivePersisters+1:]...

statusHandler/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -222,27 +222,27 @@ func (wr *WidgetsRender) prepareChainInfo() {
 		rows[3] = []string{fmt.Sprintf("Number of transactions processed: %d", numTxProcessed)}
 	}
 
+	epoch := wr.presenter.GetEpochNumber()
+	rows[4] = []string{fmt.Sprintf("Current epoch: %d", epoch)}
```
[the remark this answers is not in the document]
bogdan-rosianu: I don't think this is necessary. We have 3 groups of these things. The index only represents the order in that group

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,564 @@
...
+			}
+		}
+	}()
+
+	fullArchiveMode := psf.generalConfig.StoragePruning.FullArchive
+	numOfEpochsToKeep := uint32(psf.generalConfig.StoragePruning.NumEpochsToKeep)
+	numOfActivePersisters := uint32(psf.generalConfig.StoragePruning.NumActivePersisters)
+
+	txUnitStorerArgs := &pruning.PruningStorerArgs{
+		Identifier:            "txUnit",
```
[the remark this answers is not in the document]
bogdan-rosianu: I placed them only for easier debugging / testing. We already have constants for these but they are UnitType which is uint8 so we don't have their full name as string, only the constant name. Do you think it would make sense to extract all these names to constants ?

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,564 @@
...
+				_ = metaBlockUnit.DestroyUnit()
+			}
+			if headerUnit != nil {
+				_ = headerUnit.DestroyUnit()
+			}
+			if metaHdrHashNonceUnit != nil {
+				_ = metaHdrHashNonceUnit.DestroyUnit()
+			}
+			if shardHdrHashNonceUnits != nil {
+				for i := uint32(0); i < psf.shardCoordinator.NumberOfShards(); i++ {
```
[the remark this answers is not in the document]
bogdan-rosianu: you're right. added a nil check

storage/leveldb/leveldb.go
```diff
@@ -206,6 +206,18 @@ func (s *DB) Destroy() error {
 	return err
 }
 
+// DestroyClosed removes the already closed storage medium stored data
+func (s *DB) DestroyClosed() error {
+	s.mutBatch.Lock()
```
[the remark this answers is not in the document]
bogdan-rosianu: you're right. Changed

storage/leveldb/leveldb.go
```diff
@@ -206,6 +206,18 @@ func (s *DB) Destroy() error {
...
+	s.batch.Reset()
+	s.sizeBatch = 0
+	s.mutBatch.Unlock()
+
+	s.dbClosed <- struct{}{}
```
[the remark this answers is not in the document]
bogdan-rosianu: removed

storage/leveldb/leveldbSerial.go
```diff
@@ -206,7 +206,11 @@ func (s *SerialDB) Close() error {
 	s.closed = true
 	s.mutClosed.Unlock()
 
-	_ = s.putBatch()
+	err := s.putBatch()
```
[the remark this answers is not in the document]
bogdan-rosianu: Ok then, changed back

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"io"
+	"os"
+	"strings"
+)
+
+// removeDirectoryIfEmpty will clean the directories after all persisters for one epoch were destroyed
+// the structure is this way :
+// workspace/db/Epoch_X/Shard_Y/DbName
+// we need to remove the last 3 if everything is empty
```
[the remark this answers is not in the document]
bogdan-rosianu: the comment was bad. I meant that we remove the last 3 directories from that path. changed

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,49 @@
...
+func removeDirectoryIfEmpty(path string) {
+	elementsSplitBySeparator := strings.Split(path, string(os.PathSeparator))
+
+	epochDirectory := ""
+	for idx := 0; idx < len(elementsSplitBySeparator)-2; idx++ {
```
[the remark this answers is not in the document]
bogdan-rosianu: if the complete path is `workspace/db/Epoch_X/Shard_Y/DbName` then iterating until the len() - 2 will result in `workspace/db/Epoch_X/` so the epoch directory

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+	"github.com/ElrondNetwork/elrond-go/storage"
+	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
+)
+
+var log = logger.GetOrCreate("storage/pruning")
+
+// DefaultEpochDirectoryName represents the naming pattern for epoch directories
+const DefaultEpochDirectoryName = "Epoch"
+
+// persisterData structure is used so the persister and it's path can be kept in the same place
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+
+	var persisters []*persisterData
+	persisters = append(persisters, &persisterData{
+		persister: db,
+		path:      filePath,
+		isClosed:  false,
+	})
+
+	persistersMapByEpoch := make(map[uint32]*persisterData)
+	persistersMapByEpoch[0] = persisters[0]
```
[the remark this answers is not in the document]
bogdan-rosianu: You're right. In parallel I am working at a directory path naming component and I will add the epoch number as parameter to PruningStorer in that PR.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+	if err != nil {
+		ps.cacher.Remove(key)
+		return err
+	}
+
+	if ps.bloomFilter != nil {
+		ps.bloomFilter.Add(key)
+	}
+
+	return err
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+
+	persisterData, exists := ps.persistersMapByEpoch[epoch]
+	if !exists {
+		return nil, fmt.Errorf("key %s not found in %s",
+			base64.StdEncoding.EncodeToString(key), ps.identifier)
+	}
+
+	if !persisterData.isClosed {
+		return persisterData.persister.Get(key)
+	} else {
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+		persister, err := ps.persisterFactory.Create(persisterData.path)
+		if err != nil {
+			log.Debug("open old persister", "error", err.Error())
+			return nil, err
+		}
+
+		err = persister.Init()
```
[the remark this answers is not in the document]
bogdan-rosianu: added a defer and removed the Close() few lines after

storage/pruning/pruningStorer.go
bogdan-rosianu: very good catch. implemented the same logic from GetFromEpoch now

storage/pruning/pruningStorer.go
bogdan-rosianu: Changed. Now I iterate over all the active persisters and if i get a nil error then it means that the key-value pair was removed. otherwise I try in the next persisters and if no one is successful then I return the last error

storage/pruning/pruningStorer.go
bogdan-rosianu: missed that.. now I iterate over the map of persisters by epoch because it contains all the persisters

storage/pruning/pruningStorer.go
bogdan-rosianu: fixed now

storage/pruning/pruningStorer.go
bogdan-rosianu: I don't understand what you mean. This is a way to remove an element at a given position from a slice (by putting together 2 slices: one with elements before the position and one with elements after the position)

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,444 @@
...
+
+	if ps.bloomFilter != nil {
+		ps.bloomFilter.Add(key)
+	}
+
+	return err
+}
+
+// Get searches the key in the cache. In case it is not found, it searches
+// for the key in bloom filter first and if found
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,564 @@
...
+	var miniBlockHeadersUnit *pruning.PruningStorer
+	var shardHdrHashNonceUnits []*pruning.PruningStorer
+	var bootstrapUnit *pruning.PruningStorer
+	var err error
+
+	defer func() {
+		// cleanup
+		if err != nil {
+			log.Error("create meta store", "error", err.Error())
+			if peerDataUnit != nil {
```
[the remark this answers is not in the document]
bogdan-rosianu: done.

storage/factory/common.go
```diff
@@ -0,0 +1,44 @@
...
+		BatchDelaySeconds: cfg.BatchDelaySeconds,
+		MaxOpenFiles:      cfg.MaxOpenFiles,
+	}
+}
+
+// GetBloomFromConfig will return the bloom config needed for storage unit from a config came from the toml file
+func GetBloomFromConfig(cfg config.BloomFilterConfig) storageUnit.BloomConfig {
+	var hashFuncs []storageUnit.HasherType
+	if cfg.HashFunc != nil {
+		hashFuncs = make([]storageUnit.HasherType, 0)
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/badgerdb/badgerdb.go
```diff
@@ -208,6 +208,12 @@ func (s *DB) Destroy() error {
 	return err
 }
 
+// DestroyClosed removes the already closed storage medium stored data
+func (s *DB) DestroyClosed() error {
+	s.dbClosed <- struct{}{}
```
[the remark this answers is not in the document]
bogdan-rosianu: done

storage/pruning/pruningStorer.go
AdoAdoAdo: what I meant was: aren't you supposed to always remove last? then why do you need to append anything after it?

bogdan-rosianu: I understood now. fixed

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,537 @@
...
+	defer func() {
+		// cleanup
+		if err != nil {
+			log.Error("create meta store", "error", err.Error())
+			for _, storer := range successfullyCreatedStorers {
+				_ = storer.DestroyUnit()
+			}
+			if shardHdrHashNonceUnits != nil {
+				for i := uint32(0); i < psf.shardCoordinator.NumberOfShards(); i++ {
+					if shardHdrHashNonceUnits[i] != nil {
```
AdoAdoAdo: I would append these storers to the successfullyCreatedStorers as well, then there would be only one for loop over its range.

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,537 @@
...
+				}
+			}
+		}
+	}()
+
+	fullArchiveMode := psf.generalConfig.StoragePruning.FullArchive
+	numOfEpochsToKeep := uint32(psf.generalConfig.StoragePruning.NumEpochsToKeep)
+	numOfActivePersisters := uint32(psf.generalConfig.StoragePruning.NumActivePersisters)
+
+	metaBlockUnitArgs := &pruning.PruningStorerArgs{
```
AdoAdoAdo: can you extract the pruning storer arguments creation in a method? as I see you only need as parameter the identifier and the storer you want to create arguments for, as the other params you have in the receiver.

this would remove a lot of duplicated initialization code.

storage/leveldb/leveldbSerial.go
```diff
@@ -258,6 +259,26 @@ func (s *SerialDB) Destroy() error {
 	return err
 }
 
+// DestroyClosed removes the already closed storage medium stored data
+func (s *SerialDB) DestroyClosed() error {
+	s.mutBatch.Lock()
```
AdoAdoAdo: I think here as well you just need to remove all, as the close should have already been done.

storage/factory/pruningStorerFactory.go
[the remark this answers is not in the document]
AdoAdoAdo: do we need to use the identifiers in other places? if we don't and it's only for debugging then no need to extract.

storage/pruning/fileHandling.go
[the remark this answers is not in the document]
AdoAdoAdo: ok, didn't read the comment previously

storage/pruning/fileHandling.go
```diff
@@ -0,0 +1,51 @@
...
+		epochDirectory += elementsSplitBySeparator[idx] + string(os.PathSeparator)
+	}
+
+	if len(elementsSplitBySeparator) > 2 { // if length is less than 2, the path is incorrect
+		shardDirectory := epochDirectory + elementsSplitBySeparator[len(elementsSplitBySeparator)-2]
+		if isDirectoryEmpty(shardDirectory) {
+			err := os.RemoveAll(shardDirectory)
```
AdoAdoAdo: I see you are removing the epoch directory if the shard directory is empty, is this correct? isn't it possible we have some other shard data inside? e.g multiple observers (different shards) running on the same machine.

If we anyway remove shard and epoch (which is parent) why do we have to do them separately? wouldn't removing epoch directory remove also all children ?

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+func (ps *PruningStorer) Get(key []byte) ([]byte, error) {
+	ps.lock.Lock()
+	defer ps.lock.Unlock()
+
+	v, ok := ps.cacher.Get(key)
+	var err error
+
+	if !ok {
+		// not found in cache
+		// search it in second persistence medium
```
AdoAdoAdo: why second persistence medium? 
search it in active persisters

statusHandler/view/termuic/termuiRenders/widgetsRender.go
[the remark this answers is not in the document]
AdoAdoAdo: can you then use append to the rows instead? I just don't like explicitly using index values for for multiple data.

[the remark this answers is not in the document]
bogdan-rosianu: can we do this in another PR ?

storage/factory/pruningStorerFactory.go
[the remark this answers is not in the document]
bogdan-rosianu: no, we use them just for better log messages

storage/factory/pruningStorerFactory.go
bogdan-rosianu: done

storage/pruning/fileHandling.go
miiu96: you can use return without any value just to go out from function

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/leveldb/leveldbSerial.go
bogdan-rosianu: done

storage/factory/pruningStorerFactory.go
bogdan-rosianu: done

storage/pruning/fileHandling.go
bogdan-rosianu: Added a check for empty `Epoch_X` directory and only if this is empty I remove it.

cmd/node/main.go
```diff
@@ -540,6 +560,23 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
...
+		generalConfig.StoragePruning.NumEpochsToKeep = ctx.GlobalUint64(numEpochsToSave.Name)
+	}
+	if ctx.IsSet(numActivePersisters.Name) {
+		generalConfig.StoragePruning.NumActivePersisters = ctx.GlobalUint64(numActivePersisters.Name)
+	}
+
+	if generalConfig.StoragePruning.NumEpochsToKeep < uint64(2) && !generalConfig.StoragePruning.FullArchive {
+		return errors.New("invalid number of epochs to save")
+	}
+	if generalConfig.StoragePruning.NumActivePersisters < uint64(1) {
```
sasurobert: use constant instead of magic number

cmd/node/main.go
```diff
@@ -540,6 +560,23 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
...
+		generalConfig.StoragePruning.FullArchive = ctx.GlobalBool(isNodefullArchive.Name)
+	}
+	if ctx.IsSet(numEpochsToSave.Name) {
+		generalConfig.StoragePruning.NumEpochsToKeep = ctx.GlobalUint64(numEpochsToSave.Name)
+	}
+	if ctx.IsSet(numActivePersisters.Name) {
+		generalConfig.StoragePruning.NumActivePersisters = ctx.GlobalUint64(numActivePersisters.Name)
+	}
+
+	if generalConfig.StoragePruning.NumEpochsToKeep < uint64(2) && !generalConfig.StoragePruning.FullArchive {
```
sasurobert: use constant instead of magic number

storage/factory/pruningStorerFactory.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// NewStorageServiceFactory will return a new instance of StorageServiceFactory
+func NewStorageServiceFactory(
+	config *config.Config,
+	shardCoordinator sharding.Coordinator,
+	uniqueID string,
+	epochStartNotifier storage.EpochStartNotifier,
+) (*StorageServiceFactory, error) {
+	if config == nil {
```
sasurobert: the validity of the config can be inside the factory. not in main.go. like checking epochtokeep and numpersisters.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+		return nil, storage.ErrInvalidNumberOfPersisters
+	}
+	if check.IfNil(args.Notifier) {
+		return nil, storage.ErrNilEpochStartNotifier
+	}
+	if check.IfNil(args.PersisterFactory) {
+		return nil, storage.ErrNilPersisterFactory
+	}
+
+	cache, err = storageUnit.NewCache(args.CacheConf.Type, args.CacheConf.Size, args.CacheConf.Shards)
```
sasurobert: storageService.GetCacheFromConfig ?

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+		persistersMapByEpoch:  persistersMapByEpoch,
+		cacher:                cache,
+		bloomFilter:           nil,
+		dbPath:                filePath,
+		numOfEpochsToKeep:     args.NumOfEpochsToKeep,
+		numOfActivePersisters: args.NumOfActivePersisters,
+	}
+
+	if args.BloomFilterConf.Size != 0 { // if size is 0, that means an empty config was used so bloom filter will be nil
+		bf, err = storageUnit.NewBloomFilter(args.BloomFilterConf)
```
sasurobert: storageService.GetBloomFilterFromConfig  ?

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+	var err error
+
+	if !ok {
+		// not found in cache
+		// search it in active persisters
+		found := false
+		for idx := uint32(0); (idx < ps.numOfActivePersisters) && (idx < uint32(len(ps.activePersisters))); idx++ {
+			if ps.bloomFilter == nil || ps.bloomFilter.MayContain(key) {
+				v, err = ps.activePersisters[idx].persister.Get(key)
+
```
sasurobert: delete empty line.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+				if err != nil {
+					continue
+				}
+
+				found = true
+				// if found in persistence unit, add it in cache
```
sasurobert: add it to the cache

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+	}()
+
+	err = persister.Init()
+	if err != nil {
+		log.Debug("init old persister", "error", err.Error())
+		return nil, err
+	}
+
+	res, err := persister.Get(key)
+
```
sasurobert: delete empty line.

storage/pruning/pruningStorer.go
```diff
@@ -0,0 +1,484 @@
...
+}
+
+// Put adds data to both cache and persistence medium and updates the bloom filter
+func (ps *PruningStorer) Put(key, data []byte) error {
+	ps.lock.Lock()
+	defer ps.lock.Unlock()
+
+	ps.cacher.Put(key, data)
+
+	err := ps.activePersisters[0].persister.Put(key, data)
```
sasurobert: shouldn't you add according to epoch ?

storage/pruning/pruningStorer.go
bogdan-rosianu: no, it only returns the config which is expected for this constructor `NewCache`

storage/pruning/pruningStorer.go
bogdan-rosianu: same as above

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/pruning/pruningStorer.go
bogdan-rosianu: done

storage/pruning/pruningStorer.go
bogdan-rosianu: Would it make sense to add data to a previous epoch ? for `get` and `has` it definitely makes sense, but is there any use case you would need to put in a certain epoch ?

storage/factory/pruningStorerFactory.go
bogdan-rosianu: done

cmd/node/main.go

cmd/node/main.go

storage/factory/pruningStorerFactory.go
AdoAdoAdo: looks better, can be refactored more but it can go in as it is.

PR #749, opened 2019-12-04
Correct sync on metachain to resolve block body

sasurobert:


process/sync/metablock.go
```diff
@@ -333,7 +366,63 @@ func (boot *MetaBootstrap) getMetaHeaderFromPool(headerHash []byte) (data.Header
...
+	if err != nil {
+		return nil, err
+	}
+
+	blockBody := block.Body(miniBlockSlice)
+
+	return blockBody, nil
+}
+
+func (boot *MetaBootstrap) requestMiniBlocksFromHeaderWithNonceIfMissing(_ uint32, nonce uint64) {
```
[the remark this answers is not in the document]
sasurobert: done.

cmd/node/config/economics.toml
```diff
@@ -17,4 +17,4 @@
 
 [ValidatorSettings]
     StakeValue = "500000000000000000000000"
-    UnBoundPeriod = "100000"
+    UnBoundPeriod = "10"
```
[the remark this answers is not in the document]
sasurobert: only for testing.

process/sync/metablock.go
```diff
@@ -405,6 +405,8 @@ func (boot *MetaBootstrap) requestMiniBlocksFromHeaderWithNonceIfMissing(_ uint3
 		return
 	}
 
+	boot.requestedItemsHandler.Sweep()
+
 	hashes := make([][]byte, len(header.MiniBlockHeaders))
```
SebastianMarian: You forgot to call Has from requestedItemsHandler in next for. You should replace lines between 410-413 with these:

	hashes := make([][]byte, 0)
	for i := 0; i < len(header.MiniBlockHeaders); i++ {
		if boot.requestedItemsHandler.Has(string(header.MiniBlockHeaders[i].Hash)) {
			continue
		}

		hashes = append(hashes, header.MiniBlockHeaders[i].Hash)
	}

sasurobert: done.

integrationTests/sync/edgeCases_test.go
```diff
@@ -20,6 +21,8 @@ func TestSyncMetaNodeIsSyncingReceivedHigherRoundBlockFromShard(t *testing.T) {
 		t.Skip("this is not a short test")
 	}
 
+	_ = logger.SetLogLevel("*:DEBUG")
```
iulianpascalau: might remove this as running all tests in debug mode will output a plethora of messages.

process/peer/process.go
```diff
@@ -122,13 +123,98 @@ func (p *validatorStatistics) IsNodeValid(node *sharding.InitialNode) bool {
...
+	zero := big.NewInt(0)
+	if peerChange.ValueChange.Cmp(zero) != 0 {
+		actualValue := zero.Add(account.Stake, peerChange.ValueChange)
+		err := account.SetStakeWithJournal(actualValue)
+		if err != nil {
+			return err
+		}
+	}
+
+	if peerChange.Action == block.PeerRegistrantion && peerChange.TimeStamp != account.Nonce {
```
iulianpascalau: We use the timestamp as peer account's nonce? So counter intuitive.

sasurobert: yeah, timeStamp is equal with nonce, maybe it should be round.

integrationTests/sync/edgeCases_test.go
sasurobert: done.

cmd/node/config/economics.toml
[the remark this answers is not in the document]
sasurobert: done.

process/peer/process.go
iulianpascalau: It remains counter intuitive. Why didn't we have a specialized field called Timestamp or Round (just like Balance in the plain Account)?

SebastianMarian: Can you put at least a TODO there to refactor this

PR #751, opened 2019-12-04
Fix validator statistics root hash

SebastianMarian:
* Added ValidatorStatsRootHash in displayed block
* Fixed situation of roll back to genesis when set of current hash/header/body was not done correctly to nil
* Set ValidatorStatsRootHash to genesis block before committed it to storage


cmd/node/factory/structs.go
```diff
@@ -550,6 +550,14 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
 		return nil, err
 	}
 
+	validatorStatisticsProcessor, err := newValidatorStatisticsProcessor(args)
+	if err != nil {
+		return nil, err
+	}
+
+	validatorStatsRootHash, _ := validatorStatisticsProcessor.RootHash()
```
sasurobert: treat the error.

SebastianMarian: done

PR #753, opened 2019-12-04
EN-5239 changed address converter for validator statistics trie

ccorcoveanu:


integrationTests/vm/systemVM/systemVm_test.go
```diff
@@ -62,9 +63,9 @@ func TestStakingUnstakingAndUnboundingOnMultiShardEnvironment(t *testing.T) {
 
 	///////////------- send stake tx and check sender's balance
 	var txData string
-	for _, node := range nodes {
-		pubKey, _ := node.NodeKeys.Pk.ToByteArray()
-		txData = "stake" + "@" + hex.EncodeToString(pubKey)
+	for index, node := range nodes {
```
sasurobert: could you add the peerProcessor for an integration test which is equal with this ?

PR #754, opened 2019-12-04
En 4011 new tx cache (optimization)

andreibancioiu:
The new transactions cache implementation uses an inserted sort technique when collecting transactions. Also, it yields transactions (to be included in the next mini-block) in a reasonably fair manner. Furthermore, it has an improved eviction mechanism, implemented in 3 passes.


storage/txcache/stopwatch.go
```diff
@@ -0,0 +1,24 @@
...
+	milliseconds int64
+	checkpoint   time.Time
+}
+
+func (watch *stopWatch) start() {
+	watch.checkpoint = time.Now()
+}
+
+func (watch *stopWatch) pause() {
+	sinceCheckpoint := time.Now().Sub(watch.checkpoint)
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,91 @@
...
+	batchSize := list.CopyBatchSize
+	availableLength := len(destination)
+
+	if element == nil {
+		return 0
+	}
+
+	// todo rewrite loop, make it more readable
+	copied := 0
+	for true {
```
golangcibot: S1006: should use for {} instead of for true {} (from `gosimple`)

storage/txcache/txCache.go
```diff
@@ -0,0 +1,176 @@
...
+	cache.sendersCount.Decrement()
+}
+
+func (cache *TxCache) removeSenders(senders []string) {
+	for _, senderKey := range senders {
+		cache.removeSender(senderKey)
+	}
+}
+
+func (cache *TxCache) selfCheck() {
```
golangcibot: func `(*TxCache).selfCheck` is unused (from `unused`)

storage/txcache/txCache_test.go
```diff
@@ -0,0 +1,164 @@
...
+		sender := string(tx.GetSndAddress())
+		previousNonce := nonces[sender]
+
+		assert.LessOrEqual(t, previousNonce, nonce)
+		nonces[sender] = nonce
+	}
+}
+
+func Test_AddManyTransactionsToCacheWithEviction_UniformDistribution(t *testing.T) {
+	logger.SetLogLevel("txcache/eviction:DEBUG")
```
golangcibot: Error return value of `logger.SetLogLevel` is not checked (from `errcheck`)

storage/txcache/txCache_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+
+	return b
+}
+
+func printMemUsage(message string) {
```
golangcibot: `printMemUsage` is unused (from `deadcode`)

storage/txcache/txCache_test.go
```diff
@@ -0,0 +1,164 @@
...
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+	// For info on each, see: https://golang.org/pkg/runtime/#MemStats
+	fmt.Printf("Alloc = %v MiB", bytesToMegabytes(m.Alloc))
+	fmt.Printf("\tTotalAlloc = %v MiB", bytesToMegabytes(m.TotalAlloc))
+	fmt.Printf("\tSys = %v MiB", bytesToMegabytes(m.Sys))
+	fmt.Printf("\tNumGC = %v\n", m.NumGC)
+}
+
+func bytesToMegabytes(b uint64) uint64 {
```
golangcibot: func `bytesToMegabytes` is unused (from `unused`)

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
+package txcache
+
+import (
+	"sync"
+)
+
+// ConcurrentMap is a "thread" safe map of type string:Anything.
+// To avoid lock bottlenecks this map is dived to several (ShardCount) map shards.
```
sasurobert: divided

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+type ConcurrentMap struct {
+	shardCount int
```
sasurobert: i think we might use "chunks" here, as it is pretty confusing with shards from the architecture. only a suggestion.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	shards     []*ConcurrentMapShard
+}
+
+// ConcurrentMapShard is a "thread" safe string to anything map.
+type ConcurrentMapShard struct {
+	maxSize      int
+	items        map[string]interface{}
+	sync.RWMutex // Read Write mutex, guards access to internal map.
```
sasurobert: no need for comment.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// ConcurrentMap is a "thread" safe map of type string:Anything.
+// To avoid lock bottlenecks this map is dived to several (ShardCount) map shards.
+type ConcurrentMap struct {
+	shardCount int
+	shards     []*ConcurrentMapShard
+}
+
+// ConcurrentMapShard is a "thread" safe string to anything map.
```
sasurobert: no need for ""

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	if shardSize == 0 {
+		shardSize = 1
+	}
+	if maxSize%shardCount != 0 {
+		shardSize++
+	}
+	for i := 0; i < shardCount; i++ {
+		m.shards[i] = &ConcurrentMapShard{
+			maxSize: shardSize,
+
```
sasurobert: delete empty line.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	shardSize := maxSize / shardCount
+	if shardSize == 0 {
+		shardSize = 1
+	}
+	if maxSize%shardCount != 0 {
+		shardSize++
+	}
+	for i := 0; i < shardCount; i++ {
```
sasurobert: add empty line berfore

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+
+			items: make(map[string]interface{}),
+		}
+	}
+	return &m
+}
+
+// GetShard returns shard under given key.
+func (m *ConcurrentMap) GetShard(key string) *ConcurrentMapShard {
+	return m.shards[uint(fnv32(key))%uint(m.shardCount)]
```
sasurobert: use uint32, int32 instead if uint and int.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Set sets the given value under the specified key.
+func (m *ConcurrentMap) Set(key string, value interface{}) {
+	// Get map shard.
```
sasurobert: no need for comment

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	shard.items[key] = value
+	shard.Unlock()
+}
+
+// Get retrieves an element from map under given key.
+func (m *ConcurrentMap) Get(key string) (interface{}, bool) {
+	// Get shard
+	shard := m.GetShard(key)
+	shard.RLock()
+	// Get item from shard.
```
sasurobert: no need for comments.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	val, ok := shard.items[key]
```
sasurobert: if !ok return nil

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+	shard.RLock()
+	// See if element is within shard.
+	_, ok := shard.items[key]
+	shard.RUnlock()
+	return ok
+}
+
+// Remove removes an element from the map.
+func (m *ConcurrentMap) Remove(key string) {
+	// Try to get shard.
```
sasurobert: no need for comment inside function.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,134 @@
...
+
+// ConcurrentMapShard is a "thread" safe string to anything map.
+type ConcurrentMapShard struct {
+	maxSize      int
+	items        map[string]interface{}
+	sync.RWMutex // Read Write mutex, guards access to internal map.
+}
+
+// NewConcurrentMap creates a new concurrent map.
+func NewConcurrentMap(maxSize int, shardCount int) *ConcurrentMap {
```
sasurobert: do we really need a new concurrent map ? can't we reuse the existing setup ?

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,191 @@
+package txcache
+
+import (
+	linkedList "container/list"
```
sasurobert: delete linkedList

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/data/transaction"
+	"github.com/ElrondNetwork/elrond-go/logger"
+)
+
+var log = logger.GetOrCreate("txcache/eviction")
+
+// EvictionStrategyConfig is a cache eviction model
+type EvictionStrategyConfig struct {
+	CountThreshold                 int
```
sasurobert: use int32

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+type EvictionStrategy struct {
+	Cache  *TxCache
+	Config EvictionStrategyConfig
+	mutex  sync.Mutex
+}
+
+// NewEvictionStrategy creates a new EvictionModel
+func NewEvictionStrategy(cache *TxCache, config EvictionStrategyConfig) *EvictionStrategy {
+	model := &EvictionStrategy{
+		Cache:  cache,
```
sasurobert: you need to check if cache is nil or config is wrong and return error in those cases.

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+
+// EvictionStrategy is a cache eviction model
+type EvictionStrategy struct {
+	Cache  *TxCache
+	Config EvictionStrategyConfig
+	mutex  sync.Mutex
+}
+
+// NewEvictionStrategy creates a new EvictionModel
+func NewEvictionStrategy(cache *TxCache, config EvictionStrategyConfig) *EvictionStrategy {
```
sasurobert: instead of a type, you should input an interface.

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+func (model *EvictionStrategy) DoEviction(incomingTx *transaction.Transaction) {
+	if !model.areThereTooManyTxs() {
+		return
+	}
+
+	// We do not allow more evictions to start concurrently
+	model.mutex.Lock()
+
+	// First pass
+	// Senders capacity is close to be reached first (before txs capacity) when there are a lot of senders with little or one transaction
```
sasurobert: no need for comments inside method

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+	if model.areThereTooManyTxs() && !model.areThereJustAFewSenders() {
+		steps, countTxs, countSenders := model.EvictSendersWhileTooManyTxs()
+		log.Debug("DoEviction, 3rd pass:", "steps", steps, "countTxs", countTxs, "countSenders", countSenders)
+	}
+
+	model.mutex.Unlock()
+}
+
+func (model *EvictionStrategy) areThereTooManySenders() bool {
+	noSenders := model.Cache.txListBySender.Counter.Get()
```
sasurobert: this is too long. too many structs inside structs.

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+	CountJustAFewSenders           int
+	NoOldestSendersToEvict         int
+	ALotOfTransactionsForASender   int
+	NoTxsToEvictForASenderWithALot int
+}
+
+// EvictionStrategy is a cache eviction model
+type EvictionStrategy struct {
+	Cache  *TxCache
+	Config EvictionStrategyConfig
```
sasurobert: too many public fields

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (model *EvictionStrategy) areThereTooManyTxs() bool {
+	noTxs := model.Cache.txByHash.Counter.Get()
+	tooManyTxs := noTxs > int64(model.Config.CountThreshold)
+	return tooManyTxs
+}
+
+// EvictOldestSenders removes transactions from the cache
+func (model *EvictionStrategy) EvictOldestSenders() (int, int) {
```
sasurobert: this method should be private.

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+func (model *EvictionStrategy) doEvictItems(txsToEvict [][]byte, sendersToEvict []string) (countTxs int, countSenders int) {
+	countTxs = model.Cache.txByHash.RemoveTxsBulk(txsToEvict)
+	countSenders = model.Cache.txListBySender.RemoveSendersBulk(sendersToEvict)
+	return
+}
+
+// EvictHighNonceTransactions removes transactions from the cache
+// For senders with many transactions (> "ALotOfTransactionsForASender"), evict "NoTxsToEvictForASenderWithALot" transactions
+// Also makes sure that there's no sender with 0 transactions
+func (model *EvictionStrategy) EvictHighNonceTransactions() (int, int) {
```
sasurobert: make it private

storage/txcache/evictionStrategy.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	})
+
+	return model.doEvictItems(txsToEvict, sendersToEvict)
+}
+
+// EvictSendersWhileTooManyTxs removes transactions
+// Eviction happens in ((transaction count) - CountThreshold) / NoOldestSendersToEvict + 1 steps
+// One batch of senders is removed in each step
+func (model *EvictionStrategy) EvictSendersWhileTooManyTxs() (step int, countTxs int, countSenders int) {
```
sasurobert: make it private

storage/txcache/txByHashMap.go
```diff
@@ -0,0 +1,66 @@
+package txcache
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/data/transaction"
+)
+
+// TxByHashMap is
```
sasurobert: explain the structure

storage/txcache/txByHashMap.go
```diff
@@ -0,0 +1,66 @@
...
+type TxByHashMap struct {
+	Map     *ConcurrentMap
+	Counter core.AtomicCounter
+}
+
+// NewTxByHashMap creates a new map-like structure for holding and accessing transactions by txHash
+func NewTxByHashMap(size int, shardsHint int) TxByHashMap {
```
sasurobert: this should get the concurrentmap as an input parameter.

storage/txcache/txCache.go
```diff
@@ -0,0 +1,99 @@
...
+type TxCache struct {
+	txListBySender   TxListBySenderMap
+	txByHash         TxByHashMap
+	EvictionStrategy *EvictionStrategy
+}
+
+// NewTxCache creates a new transaction cache
+// "size" dictates the maximum number of transactions to hold in this cache at a given time
+// "shardsHint" is used to configure the internal concurrent maps on which the implementation relies
+func NewTxCache(size int, shardsHint int) *TxCache {
```
sasurobert: this should get txByHash and txListBySender as an input parameter. And use a factory to set up all the txCache.

storage/txcache/txCache.go
```diff
@@ -0,0 +1,99 @@
...
+
+// GetTransactions gets a reasonably fair list of transactions to be included in the next miniblock
+// It returns at most "noRequested" transactions
+// Each sender gets the chance to give at least "batchSizePerSender" transactions, unless "noRequested" limit is reached before iterating over all senders
+func (cache *TxCache) GetTransactions(noRequested int, batchSizePerSender int) []*transaction.Transaction {
+	result := make([]*transaction.Transaction, noRequested)
+	resultFillIndex := 0
+	resultIsFull := false
+
+	for pass := 0; ; pass++ {
```
sasurobert: add !resultIsFull as condition.

storage/txcache/txListBySenderMap.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return listForSender
+}
+
+// RemoveTx removes a transaction from the map
+func (txMap *TxListBySenderMap) RemoveTx(tx *transaction.Transaction) {
+	sender := string(tx.SndAddr)
+	listForSender, ok := txMap.getListForSender(sender)
+	if !ok {
+		// This should never happen (eviction should never cause this kind of inconsistency between the two internal maps)
```
sasurobert: log.Error.

storage/txcache/txListBySenderMap.go
```diff
@@ -0,0 +1,128 @@
...
+type TxListBySenderMap struct {
+	Map             *ConcurrentMap
+	Counter         core.AtomicCounter
+	nextOrderNumber core.AtomicCounter
+}
+
+// NewTxListBySenderMap creates a new map-like structure for holding and accessing transactions by sender
+func NewTxListBySenderMap(size int, shardsHint int) TxListBySenderMap {
+	// We'll hold at most "size" lists of at least 1 transaction
+	backingMap := NewConcurrentMap(size, shardsHint)
```
sasurobert: same as above.

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,191 @@
...
+	copyBatchIndex *linkedList.Element
+	copyBatchSize  int
+	orderNumber    int64
+	sender         string
+}
+
+// TxListForSenderNode is a node of the linked list
+type TxListForSenderNode struct {
+	TxHash []byte
+	Tx     *transaction.Transaction
```
sasurobert: you could use data.TransactionHandler interface instead.

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,191 @@
...
+		if value.Tx.Nonce > nonce {
+			return element
+		}
+	}
+
+	return nil
+}
+
+// RemoveTx removes a transaction from the sender's list
+func (list *TxListForSender) RemoveTx(tx *transaction.Transaction) {
```
sasurobert: use transactionHandler interface instead of defined struct. Add check.ifNil for input parameters for public methods.

storage/txcache/concurrentMap.go
andreibancioiu: `val` is already `<nil>` when key is missing (a bit less clutter with forwarding (returning) `val`, `ok` in my opinion).

storage/txcache/evictionStrategy.go
andreibancioiu: The `EvictionStrategy` is actually very coupled to `TxCache`'s implementation, because it knows a lot about its internal structures: `txByHash` and `txListBySender`. I will try to find a better design.

storage/txcache/evictionStrategy.go
andreibancioiu: I plead guilty for breaking the Law of Demeter.

storage/txcache/concurrentMap.go
andreibancioiu: https://github.com/ElrondNetwork/concurrent-map/blob/master/concurrent_map.go also implements eviction, and we wanted to explicitly trigger eviction (and have full control over it).

storage/txcache/txByHashMap.go
andreibancioiu: Actually, `TxByHashMap` wants to have full control over the map (and its associated counter). It wants to construct it by itself. But I have made it private and renamed it to "backingMap". Is it ok?

storage/txcache/txListBySenderMap.go
andreibancioiu: Actually, `TxListBySenderMap` wants to have full control over the map (and its associated counter). It wants to construct it by itself. But I have made it private and renamed it to `backingMap`. Is it ok?

storage/txcache/evictionStrategy.go
andreibancioiu: Fixed for `areThereTooManySenders`, `areThereJustAFewSenders`, `areThereTooManyTxs`.

storage/txcache/txCache_test.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func Test_RemoveByTxHash(t *testing.T) {
+	cache := NewTxCache(250000, 16)
+
+	txHash := []byte("hash-1")
+	tx := createTx("alice", 1)
+
+	cache.AddTx(txHash, tx)
+	cache.RemoveTxByHash(txHash)
```
golangcibot: Error return value of `cache.RemoveTxByHash` is not checked (from `errcheck`)

storage/txcache/txCache.go
andreibancioiu: `TxCache` owns `txByHash` and `txListBySender`. It is composition, not aggregation. Therefore, `TxCache` wants to explicitly instantiate and manage the two owned objects.

storage/txcache/evictionStrategy.go
andreibancioiu: Added `NewTxCacheWithEviction`, and it receives the config struct now (no nil).

storage/txcache/evictionStrategy.go
andreibancioiu: Redesigned coupling of cache and eviction strategy - eviction methods are now methods of the cache, but in a separate file - `eviction.go`.

@sasurobert, is it better?

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+
+	listForSender.mutex.Unlock()
+
+	return removedTxHashes
+}
+
+func (listForSender *txListForSender) findTx(txToFind data.TransactionHandler) *list.Element {
+	for element := listForSender.items.Front(); element != nil; element = element.Next() {
+		value := element.Value.(txListForSenderNode)
+		if value.tx == txToFind {
```
AdoAdoAdo: I think this will fail if the interface pointers are different even if the values are the same.

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+
+// IsEmpty checks whether the list is empty
+func (listForSender *txListForSender) IsEmpty() bool {
+	return listForSender.items.Len() == 0
+}
+
+// startBatchCopying resets the internal state used for copy operations
+func (listForSender *txListForSender) startBatchCopying(batchSize int) {
+	// We cannot copy or start copy from multiple goroutines at the same time
+	listForSender.mutex.Lock()
```
AdoAdoAdo: because between startBatchCopying call and copyBatchTo call, a removal could be triggered on a go routine (e.g RemoveTx is waiting on the listForSender.mutex because it was taken first by startBatchCopying, and then after startBatchCopying releases the mutex it is immediately taken by RemoveTx) there is a chance that you remove the front item. 
Not sure what inconsistency this could cause afterwards, so I think startBatchCopying should be under the same critical section as copyBatchTo, with no release of mutex in the middle.

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+	listForSender.mutex.Unlock()
+	return copied
+}
+
+// GetTxHashes returns the hashes of transactions in the list
+func (listForSender *txListForSender) GetTxHashes() [][]byte {
+	result := make([][]byte, listForSender.items.Len())
+
+	index := 0
+	for element := listForSender.items.Front(); element != nil; element = element.Next() {
```
AdoAdoAdo: don't we need any rlocking here?

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+		listForSender.items.PushBack(newNode)
+	} else {
+		listForSender.items.InsertBefore(newNode, mark)
+	}
+
+	listForSender.mutex.Unlock()
+}
+
+func (listForSender *txListForSender) findTxWithLargerNonce(nonce uint64) *list.Element {
+	for element := listForSender.items.Front(); element != nil; element = element.Next() {
```
AdoAdoAdo: can you add a comment for this function to only be used under listForSender.mutex.Lock ?

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,128 @@
...
+	m := ConcurrentMap{
+		noChunks: noChunks,
+		chunks:   make([]*concurrentMapChunk, noChunks),
+	}
+
+	chunkSize := maxSize / noChunks
+	if chunkSize == 0 {
+		chunkSize = 1
+	}
+	if maxSize%noChunks != 0 {
```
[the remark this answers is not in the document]
andreibancioiu: Indeed, `NewConcurrentMap(41, 42)` would have `chunkSize=2`, which is not ok. It won't happen in practice, but we have to fix this nonetheless.

(I took this implementation: https://github.com/ElrondNetwork/concurrent-map/blob/master/concurrent_map.go)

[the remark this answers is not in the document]
andreibancioiu: I removed the `maxSize` parameter, since it wasn't really necessary (it is for the `concurrent_map` with FIFO behavior). Capacity is given by `EvictionConfig`.

storage/txcache/txListBySenderMap.go
```diff
@@ -0,0 +1,131 @@
...
+	noRemoved := oldCount - newCount
+	return noRemoved
+}
+
+// GetListsSortedByOrderNumber gets the list of sender addreses, sorted by the global order number
+func (txMap *txListBySenderMap) GetListsSortedByOrderNumber() []*txListForSender {
+	lists := make([]*txListForSender, txMap.counter.Get())
+
+	index := 0
+	txMap.backingMap.IterCb(func(key string, item interface{}) {
```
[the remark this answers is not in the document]
andreibancioiu: In this implementation, eviction of senders is done in bulk according to `EvictionConfig.NoOldestSendersToEvict`. Which means that the sorting isn't executed extremely often. Though a circular buffer is better, indeed.

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+		index++
+	}
+
+	listForSender.mutex.Unlock()
+
+	return removedTxHashes
+}
+
+func (listForSender *txListForSender) findTx(txToFind data.TransactionHandler) *list.Element {
+	for element := listForSender.items.Front(); element != nil; element = element.Next() {
```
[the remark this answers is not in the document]
andreibancioiu: Right :)

And this also makes us think about same-sender & same-nonce & different gas price transactions. At this moment, the insertion of a new transaction in the sorted list is done like this: find the transaction with a higher nonce, insert the new transaction before this one. Though, this is not ok for same-nonce (resent) transactions. What should we do? **Perhaps, at insertion, replace the same-nonce transaction from the list if the gas price is higher?**

@AdoAdoAdo

storage/txcache/txListForSender.go
andreibancioiu: It seems that it's ok if the interface pointers point to the same *transaction in memory (which is our case as far as I know, since we don't make copies).

Example: https://play.golang.org/p/t20Y9ZmIFpC

Anyway, now we use the nonce in `findTx()`.

storage/txcache/txListForSender.go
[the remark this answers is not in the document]
AdoAdoAdo: I would replace but we need to agree on the replacement criteria (if new tx has greater gas price?)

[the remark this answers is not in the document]
AdoAdoAdo: I think I would search for the exact transaction in pool and stop with iteration if we exceed the nonce. 

The reason is the following:
1. Block with a specific set of TxHashes is agreed upon in consensus.
2. In the meantime some other TxHash came for the same nonce with a different gas/receiver/etc which was considered better for that nonce by the node so a. it replaced the old one, b. it keeps both old and new.
3. Nodes in the shard receive the signed block and have to process it.
4. The node needs to select the old TxHash, so it searches in the map txHash->txPointer:
a) The txHash was removed when replaced by a better one, so the node cannot find it and even if it requests it he cannot add it in the pool as the replacement criteria would refuse it. This case I think would invalidate the replace approach in step 2. so maybe we should not replace but keep all versions?
b) if it is not removed, then we still need to find the tx by iterating, but it is important to get the right tx (the one given as parameter), it is not sufficient to match the nonce. We can still stop iterating when finding the tx or reaching a higher nonce.

storage/txcache/txListForSender.go
AdoAdoAdo: Right, this is just used internally by the txcache so calling findTx should be done only after getting the tx pointer from the hash to tx pointer map, which will hold the original pointer, not a copy. 

So I take back my comment

storage/txcache/txListForSender.go
```diff
@@ -0,0 +1,193 @@
...
+		value := element.Value.(txListForSenderNode)
+		result[index] = value.txHash
+		index++
+	}
+
+	return result
+}
+
+func (listForSender *txListForSender) getHighestNonceTx() data.TransactionHandler {
+	back := listForSender.items.Back()
```
camilbancioiu: Mutex lock / unlock around the body, until `return`?

storage/txcache/concurrentMap.go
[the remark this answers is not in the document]
AdoAdoAdo: just had a discussion in the office and probably we should measure/maintain the capacity in bytes not number of items. this can come in another pull request.

storage/txcache/txListForSender.go
andreibancioiu: Fixed, and `startBatchCopying` and `copyBatchTo` are now a single function.

storage/txcache/txListForSender.go
andreibancioiu: Added the same locking as for copy and mutate (for simplicity). `getTxHashes` is only called at eviction time. I made it private, too.

storage/txcache/concurrentMap.go
```diff
@@ -0,0 +1,119 @@
...
+		for key, value := range chunk.items {
+			fn(key, value)
+		}
+		chunk.RUnlock()
+	}
+}
+
+func fnv32(key string) uint32 {
+	hash := uint32(2166136261)
+	const prime32 = uint32(16777619)
```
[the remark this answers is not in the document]
andreibancioiu: Added comment (link to Wikipedia) - they are "magic numbers" for `fvn32`.

storage/txcache/eviction.go
```diff
@@ -0,0 +1,143 @@
...
+		batch := batchesSource[batchStart:batchEnd]
+
+		stepCountTxs, stepCountSenders := cache.evictSendersAndTheirTxs(batch)
+
+		countTxs += stepCountTxs
+		countSenders += stepCountSenders
+		batchStart += batchSize
+
+		// Infinite loop otherwise
+		if stepCountTxs == 0 {
```
sasurobert: make a variable - shoudlBreak := stepCountTxs == 0 || stepCountSenders < batchSzie

storage/txcache/txByHashMap.go
```diff
@@ -0,0 +1,66 @@
...
+
+// txByHashMap is a new map-like structure for holding and accessing transactions by txHash
+type txByHashMap struct {
+	backingMap *ConcurrentMap
+	counter    core.AtomicCounter
+}
+
+// newTxByHashMap creates a new TxByHashMap instance
+func newTxByHashMap(nChunksHint uint32) txByHashMap {
+	// We'll hold at most "size" transactions
```
sasurobert: delete comment.

storage/txcache/txCache.go
```diff
@@ -0,0 +1,125 @@
...
+	"sync"
+
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/logger"
+)
+
+var log = logger.GetOrCreate("txcache")
+
+// TxCache is
```
sasurobert: add comment.

storage/txcache/eviction.go
```diff
@@ -0,0 +1,143 @@
...
+	CountThreshold                  uint32
+	ThresholdEvictSenders           uint32
+	NumOldestSendersToEvict         uint32
+	ALotOfTransactionsForASender    uint32
+	NumTxsToEvictForASenderWithALot uint32
+}
+
+// doEviction does cache eviction
+// We do not allow more evictions to start concurrently
+func (cache *TxCache) doEviction(incomingTx data.TransactionHandler) {
```
sasurobert: incommingTx not used.

PR #755, opened 2019-12-05
Fix journalization on data tries

BeniaminDrasovean:


data/state/accountsDB_test.go
```diff
@@ -668,9 +663,9 @@ func TestAccountsDB_LoadDataWithSomeValuesShouldWork(t *testing.T) {
 func TestAccountsDB_CommitShouldCallCommitFromTrie(t *testing.T) {
 	t.Parallel()
 
-	account := generateAccount()
 	commitCalled := 0
-
+	marsh := &mock.MarshalizerMock{}
+	serializedAccount, err := marsh.Marshal(mock.AccountWrapMock{})
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

data/state/dataTriesHolder.go
```diff
@@ -0,0 +1,54 @@
...
+	return dth.tries[string(key)]
+}
+
+// GetAll returns all trie pointers from the map
+func (dth *dataTriesHolder) GetAll() []data.Trie {
+	dth.mutex.Lock()
+	defer dth.mutex.Unlock()
+
+	tries := make([]data.Trie, 0)
+	for key, _ := range dth.tries {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for key := range ...` (from `gosimple`)

data/state/accountsDB.go
```diff
@@ -112,83 +115,61 @@ func (adb *AccountsDB) loadDataTrie(accountHandler AccountHandler) error {
...
 
-	var newDataTrie data.Trie
-	var err error
-	if accountHandler.DataTrie() == nil || accountHandler.DataTrie().IsInterfaceNil() {
-		newDataTrie, err = adb.mainTrie.Recreate(make([]byte, 0))
-		if err != nil {
-			return err
-		}
-	} else {
-		newDataTrie, err = accountHandler.DataTrie().DeepClone()
```
iulianpascalau: I think this was the main reason it ended up eating a lot of RAM

data/state/journalEntries.go
```diff
@@ -38,3 +38,51 @@ func (jeb *JournalEntryBalance) IsInterfaceNil() bool {
...
+	return &JournalEntryDataTrieUpdates{
+		trieUpdates: trieUpdates,
+		account:     account,
+	}, nil
+}
+
+// Revert applies undo operation
+func (jedtu *JournalEntryDataTrieUpdates) Revert() (AccountHandler, error) {
+	for key := range jedtu.trieUpdates {
+		err := jedtu.account.DataTrie().Update([]byte(key), jedtu.trieUpdates[key])
```
iulianpascalau: Nice one! (this also works when the value is nil causing deletion of the trie node) :+1:

PR #756, opened 2019-12-05
Edgecase when nothing proposed in the epoch

sasurobert:


process/errors.go
```diff
@@ -589,3 +586,6 @@ var ErrNilMediator = errors.New("nil mediator")
 
 // ErrMissingShardDataInStorage signals that some ShardData information is missing from storage
 var ErrMissingShardDataInStorage = errors.New("missing shard data in storage")
+
+// ErrEpochStartDataDoesNotMatch signals that EpochStartData is not the same as the leader created
+var ErrEpochStartDataDoesNotMatch = errors.New("epooch start data does not match")
```
SebastianMarian: Typo: epooch

process/block/metablock.go
```diff
@@ -1628,6 +1633,33 @@ func (mp *metaProcessor) ApplyBodyToHeader(hdr data.HeaderHandler, bodyHandler d
...
+	if err != nil {
+		return err
+	}
+
+	receivedEpochStartHash, err := core.CalculateHash(mp.marshalizer, mp.hasher, metaBlock.EpochStart)
+	if err != nil {
+		return err
+	}
+
+	createEpochStartHash, err := core.CalculateHash(mp.marshalizer, mp.hasher, *epochStart)
```
SebastianMarian: created

process/errors.go
sasurobert: done

process/block/metablock.go
sasurobert: done

process/block/metablock.go
```diff
@@ -1628,6 +1633,33 @@ func (mp *metaProcessor) ApplyBodyToHeader(hdr data.HeaderHandler, bodyHandler d
 	return nil
 }
 
+func (mp *metaProcessor) verifyEpochStartDataForMetablock(metaBlock *block.MetaBlock) error {
+	if !metaBlock.IsStartOfEpochBlock() {
+		return nil
+	}
+
+	epochStart, err := mp.createEpochStartForMetablock(metaBlock)
```
raduchis: metablock is not used in method createEpochStartForMetablock, should this not take into consideration metaBlock because you are comparing the receivedMetaBlock "metaBlock" with what you are creating?

process/block/metablock.go
```diff
@@ -1740,7 +1772,8 @@ func (mp *metaProcessor) getLastFinalizedMetaHashForShard(shardHdr *block.Header
 		currentHdr = prevShardHdr
 	}
 
-	return nil, nil, process.ErrLastFinalizedMetaHashForShardNotFound
+	//TODO: get header hash from last epoch start metablock
+	return nil, nil, nil
```
raduchis: Is it ok to return all nils? Will the method that uses these results not set some values that should not be nil to nil?

process/block/metablock_test.go
```diff
@@ -2690,7 +2690,7 @@ func TestShardProcessor_getLastFinalizedMetaHashForShardMetaHashNotFoundShouldEr
 	last, lastFinal, err := mp.GetLastFinalizedMetaHashForShard(shardHdr)
 	assert.Nil(t, last)
 	assert.Nil(t, lastFinal)
-	assert.Equal(t, process.ErrLastFinalizedMetaHashForShardNotFound, err)
+	assert.Equal(t, nil, err)
```
raduchis: Should the GetlastFinalizedMetaHashForShard return 3 nils ever?

process/block/metablock.go
```diff
@@ -1628,6 +1633,33 @@ func (mp *metaProcessor) ApplyBodyToHeader(hdr data.HeaderHandler, bodyHandler d
 	return nil
 }
 
+func (mp *metaProcessor) verifyEpochStartDataForMetablock(metaBlock *block.MetaBlock) error {
+	if !metaBlock.IsStartOfEpochBlock() {
```
raduchis: This method checks if the EpochStart.LastFinalizedHeaders slice contains anything and this struct has been added to all metaBlocks. Couldn't it work to extend the metaBlock struct with the new struct and have a EndOfEpochMetaBlock?

process/block/metablock.go
sasurobert: done.

process/block/metablock_test.go
raduchis: talked about it. Will be done when the start epoch is implemented

process/block/metablock.go
raduchis: will be done when the start epoch is implemented

process/block/metablock.go
raduchis: could be a refactoring in the future

PR #757, opened 2019-12-06
En 5276 added genesis time check in fork detector

SebastianMarian:
* Added prints info
* Added request for missing overwritten header when a finality check is done
* Added check for genesis time missmatch in fork detector

### DIFF
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -1133,7 +1133,12 @@ func (mp *metaProcessor) checkShardHeadersFinality(highestNonceHdrs map[uint32]d
 		}
 
 		if nextBlocksVerified < mp.shardBlockFinality {
-			go mp.onRequestHeaderHandlerByNonce(lastVerifiedHdr.GetShardID(), lastVerifiedHdr.GetNonce()+1)
+			for nonce := lastVerifiedHdr.GetNonce(); nonce <= lastVerifiedHdr.GetNonce()+1; nonce++ {
+				go func(requestedNonce uint64) {
+					mp.onRequestHeaderHandlerByNonce(lastVerifiedHdr.GetShardID(), requestedNonce)
+				}(nonce)
+			}
+
 			errFinal = process.ErrHeaderNotFinal
 		}
 	}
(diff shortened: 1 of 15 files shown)


process/block/metablock.go
```diff
@@ -1133,7 +1133,12 @@ func (mp *metaProcessor) checkShardHeadersFinality(highestNonceHdrs map[uint32]d
 		}
 
 		if nextBlocksVerified < mp.shardBlockFinality {
-			go mp.onRequestHeaderHandlerByNonce(lastVerifiedHdr.GetShardID(), lastVerifiedHdr.GetNonce()+1)
+			for nonce := lastVerifiedHdr.GetNonce(); nonce <= lastVerifiedHdr.GetNonce()+1; nonce++ {
```
sasurobert: this is a for for 2 elements, would it be better to have two calls of go mp.onRequestHeaderHandler

process/block/shardblock.go
```diff
@@ -377,7 +377,12 @@ func (sp *shardProcessor) checkMetaHdrFinality(header data.HeaderHandler) error
 	}
 
 	if nextBlocksVerified < sp.metaBlockFinality {
-		go sp.onRequestHeaderHandlerByNonce(lastVerifiedHdr.GetShardID(), lastVerifiedHdr.GetNonce()+1)
+		for nonce := lastVerifiedHdr.GetNonce(); nonce <= lastVerifiedHdr.GetNonce()+1; nonce++ {
+			go func(requestedNonce uint64) {
```
sasurobert: same as for metachain. 2 calls are easier to understand than this for.

process/sync/baseForkDetector.go
```diff
@@ -61,21 +62,17 @@ func (bfd *baseForkDetector) checkBlockBasicValidity(
...
-			log.Trace("add requested item with error",
-				"error", err.Error(),
-				"key", headerHash)
-
-		}
-
+		process.AddHeaderToBlackList(bfd.blackListHandler, headerHash)
 		return process.ErrHeaderIsBlackListed
 	}
+	if genesisTimeFromHeader != bfd.genesisTime {
```
sasurobert: I think this check is better in the interceptor.

process/block/metablock.go
SebastianMarian: done

process/block/shardblock.go
SebastianMarian: done

process/sync/baseForkDetector.go
SebastianMarian: added TODO to remove this check when implementation on interceptors side will be done

PR #758, opened 2019-12-06
En 5266  unmarshalizing size check

traian-schiau:


process/factory/metachain/interceptorsContainerFactory.go
```diff
@@ -75,6 +76,9 @@ func NewInterceptorsContainerFactory(
 	if check.IfNil(store) {
 		return nil, process.ErrNilStore
 	}
+	if sizeCheckDelta > 0 {
```
AdoAdoAdo: does it mean that even if correct messages have been marshalised, if verified after unmarshal/marshal they will have some slight difference in size?

traian-schiau: That depends on the actual marshaller used, for instance json with and without indentation

marshal/sizeCheckUnmarshalizer.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Unmarshal tries to unmarshal input buffer values into output object, and checks
+// for additional unused data
+func (scu *sizeCheckUnmarshalizer) Unmarshal(obj interface{}, buff []byte) error {
+	err := scu.Marshalizer.Unmarshal(obj, buff)
+	if err != nil {
+		return err
+	}
+
+	out, erro := scu.Marshal(obj)
```
iulianpascalau: This **might** prove to be overkill, needs testing. For all received messages we first unmarshal the bytes and then re-marshal them back. 
Had something similar for a transaction packer and prove to be a bad solution as CPU went most of the time to 100%, go routines started to clutter up and memory consumption increased dramatically. Had to drop that solution and implemented one that was not so precise but it had a lower resource demand. 
If `out of memory` panics will rise in tests, this point is the first to check upon and have a new implementation.
For that, we might make all structures we use for transferring data between nodes implement an interface called `SizeChecker` that is able to roughly compute the memory consumed for the instance as the sum of all contained fields.

marshal/sizeCheckUnmarshalizer.go
```diff
@@ -0,0 +1,46 @@
...
+	if erro != nil {
+		return err
+	}
+
+	maxSize := len(out) + len(out)*int(scu.acceptedDelta)/100
```
iulianpascalau: Bad naming. This is not delta. It resembles more to a threshold or percentage.

config/config.go
```diff
@@ -93,6 +93,7 @@ type Config struct {
 	Hasher         TypeConfig
 	MultisigHasher TypeConfig
 	Marshalizer    TypeConfig
+	SizeCheckDelta uint32
```
iulianpascalau: Should also add the setting in config.toml as to not forget about it.

marshal/sizeCheckUnmarshalizer.go
traian-schiau: It is a delta since it measures a variation

marshal/sizeCheckUnmarshalizer.go
iulianpascalau: Ok, let's see how it runs on Irlanda with high transaction load.

PR #762, opened 2019-12-06
En 5272 fix computation of fee txs when gas limit is reached

SebastianMarian:


process/block/preprocess/transactions.go
```diff
@@ -746,6 +736,19 @@ func (txs *transactions) ProcessMiniBlock(
 		if err != nil {
 			return err
 		}
+
+		processedTxHashes = append(processedTxHashes, miniBlockTxHashes[index])
+	}
+
+	for index := range miniBlockTxs {
```
sasurobert: i don't think we need a new for. It could stay in the first for cycle. the code would be easier as well.

process/block/preprocess/transactions.go
```diff
@@ -580,6 +580,27 @@ func (txs *transactions) CreateAndProcessMiniBlock(
 		}
 
 		snapshot := txs.accounts.JournalLen()
+		oldGasConsumedByMiniBlockInSenderShard := gasConsumedByMiniBlockInSenderShard
```
sasurobert: can we add an integration test ? Creating a block which reaches the gasLimit and processing it should reach the same results.

process/block/preprocess/transactions.go
SebastianMarian: If we move this inside the next for there could be an edge case. Actually this method is called only for crossshard miniblocks from processCompleteMiniBlock method, when a node is proposer. If let say that this miniblock is send by a malicious consensus group from another shard and it would have gas limit consumption over the max limit, but this limit would be detected only after the node would processed some txs from this miniblock, the return from this method it would let the accumulatedFees dirty. From this point even if this proposer would skip this bad miniblock and it would take another one the accumulatedFees would be wrongly calculated on its side compared with what would be calculated on its consensus group side.

sasurobert: ok. can remain as this.

process/block/preprocess/transactions.go
SebastianMarian: Added

PR #760, opened 2019-12-06
En 5269 chain id in headers

iulianpascalau:
- Added chain ID identifier on headers
- Added integration tests that proves that invalid headers can not pass through
- Changed the logic in worker.go such that, when an invalid header is received it will immediately return error as to minimize spam

PR #766, opened 2019-12-07
Rc bon fix probable highest nonce

SebastianMarian:
* Fixed problem when nodes are entered in sync mode because of a probable highest nonce set by a proposed block

### DIFF
--- a/consensus/spos/worker.go
+++ b/consensus/spos/worker.go
@@ -267,21 +267,24 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
 
 		//TODO: Block validity should be checked here and also on interceptors side, taking into consideration the following:
 		//(previous random seed, round, shard id and current random seed to verify if the block has been sent by the right proposer)
-		isHeaderValid := !check.IfNil(header) && headerHash != nil
-		if isHeaderValid {
-			errNotCritical := wrk.forkDetector.AddHeader(header, headerHash, process.BHProposed, nil, nil, false)
-			if errNotCritical != nil {
-				log.Debug("add header in forkdetector", "error", errNotCritical.Error())
-			}
-
-			log.Debug("received proposed block",
-				"from", core.GetTrimmedPk(core.ToHex(cnsDta.PubKey)),
-				"header hash", cnsDta.BlockHeaderHash,
-				"round", header.GetRound(),
-				"nonce", header.GetNonce(),
-				"prev hash", header.GetPrevHash(),
-			)
+		isHeaderInvalid := check.IfNil(header) || headerHash == nil
+		if isHeaderInvalid {
+			return ErrInvalidHeader
 		}
+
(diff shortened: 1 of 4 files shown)


consensus/spos/worker.go
```diff
@@ -267,21 +267,24 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
 
 		//TODO: Block validity should be checked here and also on interceptors side, taking into consideration the following:
 		//(previous random seed, round, shard id and current random seed to verify if the block has been sent by the right proposer)
-		isHeaderValid := !check.IfNil(header) && headerHash != nil
-		if isHeaderValid {
-			errNotCritical := wrk.forkDetector.AddHeader(header, headerHash, process.BHProposed, nil, nil, false)
-			if errNotCritical != nil {
-				log.Debug("add header in forkdetector", "error", errNotCritical.Error())
-			}
+		isHeaderInvalid := check.IfNil(header) || headerHash == nil
```
iulianpascalau: This is also done in PR #760 :D

consensus/spos/worker.go
```diff
@@ -267,21 +267,24 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
...
+		}
 
-			log.Debug("received proposed block",
-				"from", core.GetTrimmedPk(core.ToHex(cnsDta.PubKey)),
-				"header hash", cnsDta.BlockHeaderHash,
-				"round", header.GetRound(),
-				"nonce", header.GetNonce(),
-				"prev hash", header.GetPrevHash(),
-			)
+		err = wrk.forkDetector.AddHeader(header, headerHash, process.BHProposed, nil, nil, false)
```
iulianpascalau: Why do we still add the proposed block if it is not been taken into account? (you removed lines from 168:171) from baseForkDetector? Might also remove process.BHProposed constant

process/sync/baseForkDetector.go
```diff
@@ -165,10 +164,6 @@ func (bfd *baseForkDetector) removeCheckpointsBehindNonce(nonce uint64) {
 // computeProbableHighestNonce computes the probable highest nonce from the valid received/processed headers
 func (bfd *baseForkDetector) computeProbableHighestNonce() uint64 {
 	probableHighestNonce := bfd.finalCheckpoint().nonce
-	lastProposedBlockNonce := bfd.lastProposedBlockNonce()
```
iulianpascalau: Closing security gaps. :+1:

consensus/spos/worker.go
SebastianMarian: It is still used as a trigger for forced fork when consensus is stuck and cannot advance after 20 proposed blocks

consensus/spos/worker.go
SebastianMarian: Ok. It would be merged then

consensus/spos/worker.go
iulianpascalau: ok. great

PR #774, opened 2019-12-09
Dev to end of epoch dec9

sasurobert:


process/smartContract/hooks/vmCrypto.go
```diff
@@ -1,12 +1,9 @@
 package hooks
 
 import (
-	"encoding/hex"
-	"math/big"
-
 	"github.com/ElrondNetwork/elrond-go/hashing/keccak"
 	"github.com/ElrondNetwork/elrond-go/hashing/sha256"
-	"github.com/ElrondNetwork/elrond-vm-common"
+	"golang.org/x/crypto/ripemd160"
```
golangcibot: SA1019: Package golang.org/x/crypto/ripemd160 is deprecated: RIPEMD-160 is a legacy hash and should not be used for new applications. Also, this package does not and will not provide an optimized implementation. Instead, use a modern hash like SHA-256 (from crypto/sha256).  (from `staticcheck`)

PR #789, opened 2019-12-11
rc-fix-edge-case-in-bootstrap-from-storage-and-sync-state

SebastianMarian:
* Fixed an edge case in bootstrap from storage when no block could be applied
* Fixed an edge case in ShouldSync method when connected peers could fall below minimum accepted threshold, while node participate in consensus (the node has been synced at the beginning of the round) 
* Fixed an edge case when RevertAccountState would be executed while node processes block


process/block/shardblock.go
```diff
@@ -259,6 +259,12 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	//TODO: This part of code should be removed, it has been used only for debugging
```
iulianpascalau: Should remove all these TODO's when testing is over

PR #794, opened 2019-12-12
En 5313 refactor trie snapshoting

BeniaminDrasovean:
Create a new component (trieStorageManager) that manages all storage operations (snapshot, checkpoint, pruning) on the trie.


data/trie/snapshotsQueue_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	sq := newSnapshotsQueue()
+	numSnapshots := 2
+
+	for i := 0; i < numSnapshots; i++ {
+		sq.add([]byte(strconv.Itoa(i)), true)
+	}
+
+	assert.False(t, sq.removeFirst())
+	assert.True(t, sq.removeFirst())
```
bogdan-rosianu: as mentioned in the above file, it is not intuitive that the method will return true if the slice will be empty after removing first element.
Also if you called one more time that method, the test will fail with a runtime error.

data/trie/snapshotsQueue.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return sq.queue[0]
+}
+
+func (sq *snapshotsQueue) removeFirst() bool {
+	sq.mut.Lock()
+	defer sq.mut.Unlock()
+
+	sq.queue = sq.queue[1:]
+	return len(sq.queue) == 0
```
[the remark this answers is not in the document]
bogdan-rosianu: you can use the `len()` method in the places you check if the inner slice is empty.

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+	snapshotId      int
+	snapshotDbCfg   config.DBConfig
+	snapshotsBuffer snapshotsBuffer
+
+	dbEvictionWaitingList data.DBRemoveCacher
+	storageOperationMutex sync.RWMutex
+}
+
+// NewTrieStorageManager creates a new instance of trieStorageManager
+func NewTrieStorageManager(db data.DBWriteCacher, snapshotDbCfg config.DBConfig, ewl data.DBRemoveCacher) (*trieStorageManager, error) {
```
bogdan-rosianu: send the `config.DBConfig` as a pointer and check if it is nil

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+// SetDatabase sets the provided database as the main database
+func (tsm *trieStorageManager) SetDatabase(db data.DBWriteCacher) {
+	tsm.storageOperationMutex.Lock()
+	tsm.db = db
+	tsm.storageOperationMutex.Unlock()
+}
+
+// Clone returns a new instance of trieStorageManager
+func (tsm *trieStorageManager) Clone() data.StorageManager {
+	tsm.storageOperationMutex.Lock()
```
bogdan-rosianu: I think you can replace with something like:
``` GOLANG
tsmCopy := *tsm
tsm.storageOperationMutex.Lock()
tsmCopy.snasphotsBuffer = tsm.snapshotsBuffer.clone()
tsm.storageOperationMutex.Unlock()

return &tsmCopy
```

check if it will work this way

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (tsm *trieStorageManager) snapshot(msh marshal.Marshalizer, hsh hashing.Hasher) {
+	var keys [][]byte
+	for {
+		tsm.storageOperationMutex.Lock()
+
+		snapshot := tsm.snapshotsBuffer.getFirst()
+		tr, err := newSnapshotTrie(tsm.db, msh, hsh, snapshot.rootHash)
+		if err != nil {
+			log.Error(err.Error())
```
bogdan-rosianu: `log.Error("trie storage manager: newSnapshotTrie", "error", err.Error())`

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+			return
+		}
+		db := tsm.getSnapshotDb(snapshot.newDb)
+
+		tsm.storageOperationMutex.Unlock()
+
+		err = tr.root.commit(true, 0, tsm.db, db)
+		if err != nil {
+			log.Error(err.Error())
```
bogdan-rosianu: `log.Error("trie storage manager: commit", "error", err.Error())`

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+			return
+		}
+
+		tsm.storageOperationMutex.Lock()
+		isEmpty := tsm.snapshotsBuffer.removeFirst()
```
bogdan-rosianu: here check for len() as mentioned before

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+	if tsm.snapshotsBuffer.len() > 1 {
+		return
+	}
+
+	go tsm.snapshot(msh, hsh)
+}
+
+func (tsm *trieStorageManager) snapshot(msh marshal.Marshalizer, hsh hashing.Hasher) {
+	var keys [][]byte
+	for {
```
bogdan-rosianu: it would be easier to understand if you add a condition to this `for`. for example:
``` GOLANG
isSnapshotsBufferEmpty := false
for !isSnapshotsBufferEmpty{
....
  isSnapshotsBufferEmpty = tsm.snapshotsBuffer.len() == 0
....
}
```

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+
+	tsm.removeKeysFromDb(keys)
+}
+
+func (tsm *trieStorageManager) removeKeysFromDb(keys [][]byte) {
+	for i := range keys {
+		tsm.storageOperationMutex.Lock()
+		err := tsm.removeFromDb(keys[i])
+		if err != nil {
+			log.Error(err.Error())
```
bogdan-rosianu: change log the way mentioned before

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (tsm *trieStorageManager) getSnapshotDb(newDb bool) data.DBWriteCacher {
+	createNewDb := newDb || len(tsm.snapshots) == 0
+	if !createNewDb {
+		return tsm.snapshots[len(tsm.snapshots)-1]
+	}
+
+	db, err := tsm.newSnapshotDb()
+	if err != nil {
+		log.Error(err.Error())
```
bogdan-rosianu: change log the way mentioned before

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+
+	return db
+}
+
+func (tsm *trieStorageManager) removeSnapshot() {
+	dbUniqueId := strconv.Itoa(tsm.snapshotId - len(tsm.snapshots))
+
+	err := tsm.snapshots[0].Close()
+	if err != nil {
+		log.Error(err.Error())
```
bogdan-rosianu: change log the way mentioned before

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+	tsm.snapshots = tsm.snapshots[1:]
+
+	removePath := path.Join(tsm.snapshotDbCfg.FilePath, dbUniqueId)
+	go removeDirectory(removePath)
+}
+
+func removeDirectory(path string) {
+	err := os.RemoveAll(path)
+	if err != nil {
+		log.Error(err.Error())
```
bogdan-rosianu: change log the way mentioned before

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,326 @@
...
+		trieStorage: trieStorage,
+		marshalizer: msh,
+		hasher:      hsh,
+	}, nil
+}
+
+func (tsm *trieStorageManager) newSnapshotDb() (storage.Persister, error) {
+	snapshotPath := path.Join(tsm.snapshotDbCfg.FilePath, strconv.Itoa(tsm.snapshotId))
+	_, err := os.Stat(snapshotPath)
+	for err == nil {
```
bogdan-rosianu: `if _, err := os.Stat(snapshotPath); !os.IsNotExist(err) {`

data/trie/trieStorageManager_test.go
```diff
@@ -0,0 +1,379 @@
...
+		_ = tr.Update(testVal.key, testVal.value)
+		_ = tr.Commit()
+		_ = tr.Snapshot()
+		for trieStorage.snapshotsBuffer.len() != 0 {
+			time.Sleep(snapshotDelay)
+		}
+
+		snapshotId := strconv.Itoa(trieStorage.snapshotId - 1)
+		snapshotPath := path.Join(trieStorage.snapshotDbCfg.FilePath, snapshotId)
+		f, _ := os.Stat(snapshotPath)
```
bogdan-rosianu: `if f, err := os.Stat(snapshotPath); !os.IsNotExist(err) {`

data/trie/patriciaMerkleTrie.go
```diff
@@ -32,62 +24,39 @@ const maxSnapshots = 2
...
+	trieStorage data.StorageManager,
 	msh marshal.Marshalizer,
 	hsh hashing.Hasher,
-	evictionDb storage.Persister,
-	evictionWaitListSize int,
-	snapshotDbCfg config.DBConfig,
 ) (*patriciaMerkleTrie, error) {
-	if db == nil || db.IsInterfaceNil() {
-		return nil, ErrNilDatabase
+	if trieStorage == nil || trieStorage.IsInterfaceNil() {
```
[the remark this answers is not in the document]

data/interface.go
```diff
@@ -114,6 +117,7 @@ type Trie interface {
 	CancelPrune(rootHash []byte, identifier TriePruningIdentifier)
 	Prune(rootHash []byte, identifier TriePruningIdentifier) error
 	Snapshot() error
+	Checkpoint() error
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
```diff
@@ -340,32 +309,37 @@ func (tr *patriciaMerkleTrie) Recreate(root []byte) (data.Trie, error) {
...
-	if db == nil {
-		return tr.recreateFromDb(root, tr.db)
+	if emptyTrie(root) {
+		clonedTrieStorage := tr.trieStorage.Clone()
+		newTr, err := NewTrie(
+			clonedTrieStorage,
+			tr.marshalizer,
+			tr.hasher,
+		)
+		if err != nil {
```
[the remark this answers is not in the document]

data/trie/snapshotsQueue.go
```diff
@@ -0,0 +1,68 @@
...
+func (sq *snapshotsQueue) len() int {
+	sq.mut.Lock()
+	defer sq.mut.Unlock()
+
+	return len(sq.queue)
+}
+
+func (sq *snapshotsQueue) clone() snapshotsBuffer {
+	sq.mut.Lock()
+	defer sq.mut.Unlock()
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/snapshotsQueue.go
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/snapshotsQueue_test.go
```diff
@@ -0,0 +1,105 @@
...
+	sq := newSnapshotsQueue()
+	sq.add([]byte("root hash"), true)
+
+	assert.Equal(t, 1, len(sq.queue))
+	assert.Equal(t, []byte("root hash"), sq.queue[0].rootHash)
+	assert.True(t, sq.queue[0].newDb)
+}
+
+func TestSnapshotsQueue_AddConcurrently(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/snapshotsQueue_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func TestSnapshotsQueue_Clone(t *testing.T) {
+	t.Parallel()
+
+	sq := newSnapshotsQueue()
+	sq.add([]byte("root hash"), true)
+
+	newSq := sq.clone()
+	assert.Equal(t, sq.len(), newSq.len())
+
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: I think the way it is right now is a bit cleaner.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Initially, there was a race problem that did not allow me to write it like this, but now it is solved, so changed.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager_test.go
BeniaminDrasovean: I don't check the err here.

data/trie/errors.go
```diff
@@ -60,3 +60,6 @@ var ErrNilTrieStorage = errors.New("nil trie storage provided")
 
 // ErrNilEvictionWaitingList is raised when a nil eviction waiting list is provided
 var ErrNilEvictionWaitingList = errors.New("nil eviction waiting list provided")
+
+// ErrNilSnapshotDbConfig is raised when a nil snapshot db config is provided
+var ErrNilSnapshotDbConfig = errors.New("nil nil snapshot db config provided")
```
bogdan-rosianu: delete one 'nil'

data/trie/trieStorageManager.go
```diff
@@ -0,0 +1,329 @@
...
+		trieStorage: trieStorage,
+		marshalizer: msh,
+		hasher:      hsh,
+	}, nil
+}
+
+func (tsm *trieStorageManager) newSnapshotDb() (storage.Persister, error) {
+	snapshotPath := path.Join(tsm.snapshotDbCfg.FilePath, strconv.Itoa(tsm.snapshotId))
+	_, err := os.Stat(snapshotPath)
+	for !os.IsNotExist(err) {
```
ccorcoveanu: Please extract into a function or variable (`!os.IsNotExist(err)`)- it's hard to make sense of every aspect of what is treated here.

data/trie/errors.go
BeniaminDrasovean: Done.

cmd/node/factory/structs.go
```diff
@@ -758,13 +759,20 @@ func getTrie(
 
 	snapshotDbCfg.FilePath = filepath.Join(uniqueID, cfg.DB.FilePath, snapshotDbCfg.FilePath)
 
+	ewl, err := evictionWaitingList.NewEvictionWaitingList(int(evictionWaitingListCfg.Size), evictionDb, marshalizer)
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/trie/interface.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"io"
+	"sync"
+
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/hashing"
+	"github.com/ElrondNetwork/elrond-go/marshal"
+)
+
+type node interface {
```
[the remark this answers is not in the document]
BeniaminDrasovean: I agree, there are a lot of methods to this interface, but I can't find another solution.The idea is that the trie holds a pointer to the root, which can be either branchNode, extensionNode or leafNode. Instead of using a lot of switch - case statements, I chose to use polymorphism, so each node type implements the node interface.

data/trie/patriciaMerkleTrie.go
```diff
@@ -413,258 +383,87 @@ func (tr *patriciaMerkleTrie) Prune(rootHash []byte, identifier data.TriePruning
...
+	return tr.trieStorage.Prune(rootHash)
 }
 
 // CancelPrune invalidates the hashes that correspond to the given root hash from the eviction waiting list
 func (tr *patriciaMerkleTrie) CancelPrune(rootHash []byte, identifier data.TriePruningIdentifier) {
 	tr.mutOperation.Lock()
 	rootHash = append(rootHash, byte(identifier))
-	_, _ = tr.dbEvictionWaitingList.Evict(rootHash)
+	tr.trieStorage.CancelPrune(rootHash)
 	tr.mutOperation.Unlock()
```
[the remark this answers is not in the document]

data/trie/patriciaMerkleTrie_test.go
```diff
@@ -16,7 +16,6 @@ import (
 	"github.com/ElrondNetwork/elrond-go/hashing"
```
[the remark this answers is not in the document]
BeniaminDrasovean: There are test for `ResetOldHashes` and `SetCheckpoint`: 
- TestTrieResetOldHashes
- TestTrieCheckpoint
- TestTrieCheckpointWithNoSnapshotCreatesSnapshot
- TestTrieSnapshottingAndCheckpointConcurrently

I chose to put them in different test files, because this is the trie_test package, and I don't have access to the unexported variables. If I were to create an export_test file, there would be a lot of functions and members that would have to be exported. There are some test that check that the trie operations work concurrently, but I will add more.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

cmd/node/factory/structs.go
```diff
@@ -2324,7 +2324,7 @@ func generateInMemoryAccountsAdapter(
 	hasher hashing.Hasher,
 	marshalizer marshal.Marshalizer,
 ) (state.AccountsAdapter, error) {
-	cacheSize := 100
+	cacheSize := uint(100)
```
ccorcoveanu: Should this be hardcoded here?

BeniaminDrasovean: This function is called only when creating genesis header. After that, the real accounts adapter will be used.

cmd/node/factory/structs.go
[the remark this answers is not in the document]
bogdan-rosianu: why not `uint32` ? 
`uint is an unsigned integer type that is at least 32 bits in size. It is a distinct type, however, and not an alias for, say, uint32. `

data/trie/patriciaMerkleTrie.go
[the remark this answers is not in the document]
bogdan-rosianu: I disagree. We should use `defer unlock` only in cases where we have multiple "exit points" inside a function. Otherwise, we can unlock after the usage of the protected resource. Defer will involve a little bit more computation and if it is not necessary we can avoid it.

cmd/node/factory/structs.go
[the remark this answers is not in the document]
BeniaminDrasovean: What would be the advantage of using `uint32`?

data/trie/patriciaMerkleTrie.go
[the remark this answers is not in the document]
BeniaminDrasovean: For me it's ok either way. @ccorcoveanu ?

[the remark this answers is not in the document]
ccorcoveanu: @bogdan-rosianu Yes, that's the general recommendation using mutexes. But in this special case it seems the use case is to lock the mutex through the whole function execution. It's not as performant but it's a lot easier to read. 
Consider that somebody will edit the function later in time adding some functionality. He will have to go through the whole thought process to figure out if the new code should or shouldn't be protected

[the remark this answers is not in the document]
ccorcoveanu: If you still don't agree we can leave it as it was :) I don't have a strong opinion, I just prefer code that is easy to read :)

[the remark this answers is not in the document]
bogdan-rosianu: I understand what you say, but in this case the functions were pretty simple. In case of larger functions, we can use defers. Also, this is inside the trie so we can use each small improvement. So @BeniaminDrasovean , please put it as it was before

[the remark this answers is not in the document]
ccorcoveanu: Exactly, I was talking about those simple things.
```go
func (tr *patriciaMerkleTrie) AppendToOldHashes(hashes [][]byte) {
	tr.mutOperation.RLock()
	tr.oldHashes = append(tr.oldHashes, hashes...)
        tr.mutOperation.RUnlock()

        // ^ This is a classic lock - And now, what if I do:
       tr.trieStorage.TakeSnapshot(rootHash, tr.marshalizer, tr.hasher)
}
```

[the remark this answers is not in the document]
BeniaminDrasovean: Changed with `Lock()` and `Unlock()` instead of `RLock()` and `RUnlock()`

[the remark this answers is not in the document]
bogdan-rosianu: @iulianpascalau please leave an opinion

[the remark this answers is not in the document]
iulianpascalau: My opinion is to use mut.Lock and mut.Unlock (without defer) if the function is concise with no early exits. In this way we have a slight faster execution times. Moreover, since we are dealing with tries, I suggest to keep the code as optimum as we can.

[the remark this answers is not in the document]
ccorcoveanu: I don't like it, but ok :)

PR #802, opened 2019-12-12
Move ArgumentParser into vmcommon

camilbancioiu:
* Move `atArgumentParser.go` into `elrond-vm-common`
* Move errors related to the ArgumentParser into `elrond-vm-common`
* Update all references to ArgumentParser and its errors to use `vmcommon`
* Keep the ArgumentParser interface in `process`


process/smartContract/atArgumentParser_test.go
```diff
@@ -4,35 +4,34 @@ import (
 	"encoding/hex"
 	"testing"
 
-	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-vm-common"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestNewAtArgumentParser(t *testing.T) {
```
iulianpascalau: Please move this file in elrond-vm-common

camilbancioiu: We don't run tests in `elrond-vm-common` very often. Can we leave the file here, to ensure continuous testing?

PR #817, opened 2019-12-13
more linter fixes

iulianpascalau:
shown by running `go vet ./...`


iulianpascalau:
not the correct solution

PR #819, opened 2019-12-14
Fix mem p2p

iulianpascalau:
fixed memory p2p messenger to work in integration tests


iulianpascalau:
Not the correct solution

p2p/memp2p/messenger.go
```diff
@@ -264,100 +281,101 @@ func (messenger *Messenger) BroadcastOnChannelBlocking(channel string, topic str
...
+		}
+
+		// numReceived gets incremented because the message arrived on a registered topic
+		atomic.AddUint64(&messenger.numReceived, 1)
+		validator, _ := messenger.topicValidators[topic]
+		if check.IfNil(validator) {
+			messenger.topicsMutex.Unlock()
+			continue
+		}
+		messenger.topicsMutex.Unlock()
```
camilbancioiu: Move this before the `if` and only do it once.

integrationTests/testInitializer.go
```diff
@@ -816,6 +817,35 @@ func CreateNodes(
...
+	for shardId := uint32(0); shardId < uint32(numOfShards); shardId++ {
+		for j := 0; j < nodesPerShard; j++ {
+			n := NewTestProcessorNodeWithMemP2P(uint32(numOfShards), shardId, shardId, network)
+
+			nodes[idx] = n
+			idx++
+		}
+	}
+
+	for i := 0; i < numMetaChainNodes; i++ {
```
camilbancioiu: This would be clearer:
```
numValidators := len(nodes)
for i := i < numMetaChainNodes; i++ {
    metaNode := ...
    nodes[i + numValidators] = metaNode
}
```

integrationTests/testInitializer.go
```diff
@@ -816,6 +817,35 @@ func CreateNodes(
...
 }
 
+// CreateNodesWithMemP2P creates multiple nodes in different shards
+func CreateNodesWithMemP2P(
+	numOfShards int,
+	nodesPerShard int,
+	numMetaChainNodes int,
+	network *memp2p.Network,
+) []*TestProcessorNode {
+	//first node generated will have is pk belonging to firstSkShardId
```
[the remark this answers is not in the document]
iulianpascalau: Removed useless, rotten comment. From this function and the originating function

p2p/memp2p/messenger.go
```diff
@@ -264,100 +281,101 @@ func (messenger *Messenger) BroadcastOnChannelBlocking(channel string, topic str
...
+
+func (messenger *Messenger) processFromQueue() {
+	for {
+		message := <-messenger.processQueue
+		if check.IfNil(message) {
+			continue
 		}
+
+		identifier := string(message.Peer()) + string(message.SeqNo())
+		messenger.seenMessages.Sweep()
```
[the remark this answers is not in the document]
iulianpascalau: removed

p2p/memp2p/messenger.go
iulianpascalau: this is not possible as I need the same mutex for the second map that holds topicValidators. I want the check to be done atomically.

integrationTests/testInitializer.go
iulianpascalau: Something is not right with the for loop condition

PR #846, opened 2019-12-18
EN-5525 - Block tracker component

SebastianMarian:


process/track/blockDisplayer.go
```diff
@@ -0,0 +1,67 @@
+package track
+
+import (
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+func (bbt *baseBlockTrack) displayHeaders() {
```
golangcibot: func `(*baseBlockTrack).displayHeaders` is unused (from `unused`)

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,175 @@
...
+
+	for _, startHeader := range startHeaders {
+		shardID := startHeader.GetShardID()
+		bbt.crossNotarizedHeaders[shardID] = append(bbt.crossNotarizedHeaders[shardID], startHeader)
+	}
+
+	return nil
+}
+
+func (bbt *baseBlockTrack) setCrossNotarizedHeadersForShard(
```
golangcibot: func `(*baseBlockTrack).setCrossNotarizedHeadersForShard` is unused (from `unused`)

process/track/blockDisplayer.go
SebastianMarian: done

process/track/blockNotarizer.go
SebastianMarian: done

process/track/blockProcessor.go
```diff
@@ -0,0 +1,249 @@
...
+	return isHeaderOutOfRange
+}
+
+func (bbt *baseBlockTrack) addHeaders(
+	header data.HeaderHandler,
+	headerHash []byte,
+	lastNotarizedHeaderNonce uint64,
+	getHeaderFromPoolWithNonce func(nonce uint64, shardID uint32) (data.HeaderHandler, []byte, error),
+) {
+	nextHeaderNonce := lastNotarizedHeaderNonce + bbt.blockFinality + 1
```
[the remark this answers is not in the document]
SebastianMarian: refactored

process/track/blockNotifier.go
```diff
@@ -0,0 +1,45 @@
+package track
+
+import (
+	"github.com/ElrondNetwork/elrond-go/data"
+)
+
+// RegisterSelfNotarizedHeadersHandler registers a new handler to be called when self notarized header is changed
+func (bbt *baseBlockTrack) RegisterSelfNotarizedHeadersHandler(handler func(headers []data.HeaderHandler, headersHashes [][]byte)) {
+	if handler == nil {
+		log.Debug("attempt to register a nil handler to a tracker object")
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+func (bbt *baseBlockTrack) setCrossNotarizedHeaders(startHeaders map[uint32]data.HeaderHandler) error {
+	bbt.mutCrossNotarizedHeaders.Lock()
+	defer bbt.mutCrossNotarizedHeaders.Unlock()
+
+	if startHeaders == nil {
+		return process.ErrCrossNotarizedHdrsSliceIsNil
+	}
+
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,175 @@
...
+func (bbt *baseBlockTrack) setSelfNotarizedHeaders(startHeaders map[uint32]data.HeaderHandler) error {
+	bbt.mutSelfNotarizedHeaders.Lock()
+	defer bbt.mutSelfNotarizedHeaders.Unlock()
+
+	if startHeaders == nil {
+		return process.ErrSelfNotarizedHdrsSliceIsNil
+	}
+
+	bbt.selfNotarizedHeaders = make(map[uint32][]data.HeaderHandler)
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/shardBlockTrack.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type shardBlockTrack struct {
+	*baseBlockTrack
+	store dataRetriever.StorageService
+}
+
+// NewShardBlockTrack creates an object for tracking the received shard blocks
+func NewShardBlockTrack(
+	hasher hashing.Hasher,
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/track/shardBlockTrack.go
```diff
@@ -0,0 +1,122 @@
...
+		hasher:            hasher,
+		marshalizer:       marshalizer,
+		rounder:           rounder,
+		shardCoordinator:  shardCoordinator,
+		shardHeadersPool:  poolsHolder.Headers(),
+		metaBlocksPool:    poolsHolder.MetaBlocks(),
+		headersNoncesPool: poolsHolder.HeadersNonces(),
+	}
+
+	err = bbt.setCrossNotarizedHeaders(startHeaders)
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/track/blockProcessor.go
```diff
@@ -0,0 +1,249 @@
...
+import (
+	"bytes"
+	"sort"
+
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+func (bbt *baseBlockTrack) receivedShardHeader(shardHeaderHash []byte) {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/blockNotifier.go
```diff
@@ -0,0 +1,45 @@
+package track
+
+import (
+	"github.com/ElrondNetwork/elrond-go/data"
+)
+
+// RegisterSelfNotarizedHeadersHandler registers a new handler to be called when self notarized header is changed
+func (bbt *baseBlockTrack) RegisterSelfNotarizedHeadersHandler(handler func(headers []data.HeaderHandler, headersHashes [][]byte)) {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,175 @@
+package track
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+func (bbt *baseBlockTrack) setCrossNotarizedHeaders(startHeaders map[uint32]data.HeaderHandler) error {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,431 @@
...
+type baseBlockTrack struct {
+	hasher            hashing.Hasher
+	marshalizer       marshal.Marshalizer
+	rounder           consensus.Rounder
+	shardCoordinator  sharding.Coordinator
+	metaBlocksPool    storage.Cacher
+	shardHeadersPool  storage.Cacher
+	headersNoncesPool dataRetriever.Uint64SyncMapCacher
+
+	mutHeaders sync.RWMutex
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/blockProcessor.go
```diff
@@ -0,0 +1,249 @@
...
+		return
+	}
+
+	shardID := header.GetShardID()
+	nonce := header.GetNonce()
+
+	bbt.mutHeaders.Lock()
+	defer bbt.mutHeaders.Unlock()
+
+	headersForShard, ok := bbt.headers[shardID]
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,175 @@
...
+	bbt.mutCrossNotarizedHeaders.Unlock()
+}
+
+func (bbt *baseBlockTrack) addCrossNotarizedHeaders(shardID uint32, crossNotarizedHeaders []data.HeaderHandler) {
+	bbt.mutCrossNotarizedHeaders.Lock()
+	bbt.crossNotarizedHeaders[shardID] = append(bbt.crossNotarizedHeaders[shardID], crossNotarizedHeaders...)
+	bbt.mutCrossNotarizedHeaders.Unlock()
+}
+
+func (bbt *baseBlockTrack) getLastCrossNotarizedHeaderNonce(shardID uint32) uint64 {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,431 @@
...
+	blockTracker blockTracker
+}
+
+// AddTrackedHeader adds header to the tracker lists
+func (bbt *baseBlockTrack) AddTrackedHeader(header data.HeaderHandler, hash []byte) {
+	bbt.addHeader(header, hash)
+}
+
+// AddCrossNotarizedHeaders adds cross notarized header to the tracker lists
+func (bbt *baseBlockTrack) AddCrossNotarizedHeader(shardID uint32, crossNotarizedHeader data.HeaderHandler) {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,431 @@
...
+	bbt.addCrossNotarizedHeaders(crossNotarizedHeader.GetShardID(), []data.HeaderHandler{crossNotarizedHeader})
+}
+
+// AddSelfNotarizedHeader adds self notarized header to the tracker lists
+func (bbt *baseBlockTrack) AddSelfNotarizedHeader(shardID uint32, selfNotarizedHeader data.HeaderHandler) {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/track/blockProcessor.go
```diff
@@ -0,0 +1,249 @@
...
+	"bytes"
+	"sort"
+
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+func (bbt *baseBlockTrack) receivedShardHeader(shardHeaderHash []byte) {
+	shardHeader, err := process.GetShardHeaderFromPool(shardHeaderHash, bbt.shardHeadersPool)
```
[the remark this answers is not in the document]
SebastianMarian: Yep

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/ElrondNetwork/elrond-go/hashing"
+	"github.com/ElrondNetwork/elrond-go/marshal"
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+type blockNotarizer struct {
+	hasher      hashing.Hasher
+	marshalizer marshal.Marshalizer
+
+	startHeaders map[uint32]data.HeaderHandler
```
golangcibot: `startHeaders` is unused (from `structcheck`)

SebastianMarian: Removed

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,425 @@
...
+// GetLastHeader returns the last header received (highest round) for the given shard
+func (bbt *baseBlockTrack) GetLastHeader(shardID uint32) data.HeaderHandler {
+	bbt.mutHeaders.RLock()
+	defer bbt.mutHeaders.RUnlock()
+
+	var lastHeaderForShard data.HeaderHandler
+
+	headersForShard, ok := bbt.headers[shardID]
+	if !ok {
+		return lastHeaderForShard
```
sasurobert: this return is pretty problematic. as it in only a pointer, without the underlying structure. Those who use this function has to be pretty careful.

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,425 @@
...
+
+	headers := make([]data.HeaderHandler, 0)
+	headersHashes := make([][]byte, 0)
+
+	for _, headerInfo := range sortedHeadersInfo {
+		headers = append(headers, headerInfo.header)
+		headersHashes = append(headersHashes, headerInfo.hash)
+	}
+
+	return headers, headersHashes
```
sasurobert: these results could be cached, and redo the sorting when changes were done.

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+// IsShardStuck returns true if the given shard is stuck
+func (bbt *baseBlockTrack) IsShardStuck(shardId uint32) bool {
+	header := bbt.GetLastHeader(shardId)
+	if check.IfNil(header) {
+		return false
```
sasurobert: if there is no last header means the shard is ok ? I think it should return true

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,228 @@
...
+			continue
+		}
+
+		headersInfo = append(headersInfo, headerInfo)
+	}
+
+	bn.notarizedHeaders[shardID] = headersInfo
+}
+
+func (bn *blockNotarizer) displayNotarizedHeaders(shardID uint32, message string) {
```
sasurobert: this should be class independet. you could have a display.go file and have all the functions there. Here the displayNotarizedHeaders should get the list / sorted notarized headers.

process/track/metaBlockTrack.go
```diff
@@ -0,0 +1,130 @@
...
+		selfMetaBlocksInfo = append(selfMetaBlocksInfo, &headerInfo{hash: metaBlockHash, header: metaBlock})
+	}
+
+	return selfMetaBlocksInfo
+}
+
+func (mbt *metaBlockTrack) computeLongestSelfChain() (data.HeaderHandler, []byte, []data.HeaderHandler, [][]byte) {
+	lastSelfNotarizedHeader, lastSelfNotarizedHeaderHash, err := mbt.selfNotarizer.getLastNotarizedHeader(mbt.shardCoordinator.SelfId())
+	if err != nil {
+		return nil, nil, nil, nil
```
sasurobert: masked error. i think it should be treated.

process/track/shardBlockTrack.go
```diff
@@ -0,0 +1,135 @@
...
+		log.Debug("getSelfHeaders", process.ErrWrongTypeAssertion)
+		return selfHeadersInfo
+	}
+
+	for _, shardInfo := range metaBlock.ShardInfo {
+		if shardInfo.ShardID != sbt.shardCoordinator.SelfId() {
+			continue
+		}
+
+		header, err := process.GetShardHeader(shardInfo.HeaderHash, sbt.shardHeadersPool, sbt.marshalizer, sbt.store)
```
sasurobert: there is no way for this header to be in the poo. It is only in the store.

process/track/shardBlockTrack.go
```diff
@@ -0,0 +1,135 @@
...
+		selfHeadersInfo = append(selfHeadersInfo, &headerInfo{hash: shardInfo.HeaderHash, header: header})
+	}
+
+	return selfHeadersInfo
+}
+
+func (sbt *shardBlockTrack) computeLongestSelfChain() (data.HeaderHandler, []byte, []data.HeaderHandler, [][]byte) {
+	lastSelfNotarizedHeader, lastSelfNotarizedHeaderHash, err := sbt.selfNotarizer.getLastNotarizedHeader(sharding.MetachainShardId)
+	if err != nil {
+		return nil, nil, nil, nil
```
sasurobert: treat the error.

sasurobert: it is critical if that returns error. should not ever happen. In that case there should be a node restart.

process/track/blockProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+	prevHeader := header
+	numFinalityAttestingHeaders := uint64(0)
+
+	for i := index; i < len(sortedHeaders); i++ {
+		currHeader := sortedHeaders[i]
+		if numFinalityAttestingHeaders >= bp.blockFinality || currHeader.GetNonce() > prevHeader.GetNonce()+1 {
+			break
+		}
+
+		if currHeader.GetNonce() == prevHeader.GetNonce()+1 {
```
sasurobert: the same check is inside IsHeaderConstructionValid - so this is not needed.

process/track/blockProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		return
+	}
+
+	for i := index; i < len(sortedHeaders); i++ {
+		currHeader := sortedHeaders[i]
+		if currHeader.GetNonce() > prevHeader.GetNonce()+1 {
+			break
+		}
+
+		if currHeader.GetNonce() == prevHeader.GetNonce()+1 {
```
sasurobert: this check exists inside IsHeaderConstructionValid so it is not needed here.

process/track/blockProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+				continue
+			}
+
+			err = bp.checkHeaderFinality(currHeader, sortedHeaders, i+1)
+			if err != nil {
+				continue
+			}
+
+			headersIndexes = append(headersIndexes, i)
+			bp.getNextHeader(longestChainHeadersIndexes, headersIndexes, currHeader, sortedHeaders, i+1)
```
sasurobert: i think it would be better to do it iteratively: https://stackoverflow.com/questions/15013956/height-of-binary-search-tree-iteratively/15015738

process/track/blockNotarizer.go
iulianpascalau: Still exists.

process/track/blockNotarizer.go
iulianpascalau: Disagree, it depends on this structure as it operates on the same mutex with the rest of the functions. It is better to keep all functionality regarding the same data structures close together (in the same struct) [encapsulation] rather than have separate "structs" or files that groups functions based on behavior. Polymorphism also relies on the fact that multiple structs (classes) have the same function signatures.

process/track/baseBlockTrack.go
```diff
@@ -0,0 +1,425 @@
...
+				maxRound = headerInfo.header.GetRound()
+				lastHeaderForShard = headerInfo.header
+			}
+		}
+	}
+
+	return lastHeaderForShard
+}
+
+// GetTrackedHeadersForShard returns tracked headers for a given shard
```
iulianpascalau: Wrong comment

process/track/blockNotarizer.go
```diff
@@ -0,0 +1,228 @@
...
+	if bn.notarizedHeaders == nil {
+		return nil, nil, process.ErrNotarizedHeadersSliceIsNil
+	}
+
+	headersInfo := bn.notarizedHeaders[shardID]
+	if headersInfo == nil {
+		return nil, nil, process.ErrNotarizedHeadersSliceForShardIsNil
+	}
+
+	headerInfo := headersInfo[0]
```
iulianpascalau: this code will panic if, for some reason (bug, for example), there is an empty slice on `bn.notarizedHeaders[shardID]` (maybe replace the condition on L171 with `if len(headersInfo) == 0 {`
Also, the function returns the first notarized header id offset is out of bound (returns something else as the name suggests)

process/track/interface.go
```diff
@@ -0,0 +1,33 @@
+package track
+
+import (
+	"github.com/ElrondNetwork/elrond-go/data"
+)
+
+type blockNotarizerHandler interface {
```
iulianpascalau: Suggestion to keep the interface names unexported but the functions exported in order to be able to stub these interfaces in a separate package (mock). Can be left of the next PR when unit tests will be written.

process/track/metaBlockTrack.go
iulianpascalau: ...or at least logged.

process/track/blockNotarizer.go
SebastianMarian: Not pushed yet

process/track/baseBlockTrack.go
SebastianMarian: Actully this method is used only in the exported method from this file -> IsShardSuck (which returns true or false). So I made it unexported

process/track/baseBlockTrack.go
SebastianMarian: This method is will be called only one time per round in BlockProcessor and it does only (1-2 appends for each call) as it returns headers which are stored per shard / per nonce. It will have two records when there are forks(network split) so maximum 2 records for the same nonce.

process/track/baseBlockTrack.go
SebastianMarian: This method is used in fork detector and could be used anywhere to determine if one cross chain is stuck or not. Actually GetLastHeader could not be nil as it is initialized with startHeaders(genesis header for each shard) on New.. method. But even if it would be nil for some reason, you can not say that a shard is stuck just because you have no info about it. This is why it returns false in this case, but again this would not be possible with the current implementation.

process/track/blockNotarizer.go
SebastianMarian: Actually display.go has been deleted and its functionality has been moved here and with this approach the duplicate code has been also removed.

process/track/metaBlockTrack.go
SebastianMarian: I have added a log. Actually getLastNotarizedHeader could not return err as it is initialized with startHeaders(genesis header for each shard) on New.. method.This method returns the new finalized headers constructed above the last finalized one in self chain. If the chain didn't advance and the node is sync or the last notarized header could not be determined, this method has nothing to return.

process/track/shardBlockTrack.go
SebastianMarian: If one node constructed on the fork in shard and now it received a meta block which contains the wright shard headers which have been notarized in meta, these shard headers would be only in pool as they were not been removed because node didn't use them as it constructed on the wrong chain.

process/track/metaBlockTrack.go
SebastianMarian: These shard headers contained in the received meta block are used to notify all the registered handlers (for now only fork detector) that there are new self headers notarized in meta, so the node can now check sooner if it constructed on the notarized chain or not (not need anymore to wait until it will include these meta headers in its processed shard headers to determine this)

process/track/shardBlockTrack.go
SebastianMarian: These shard headers contained in the received meta block are used to notify all the registered handlers (for now only fork detector) that there are new self headers notarized in meta, so the node can now check sooner if it constructed on the notarized chain or not (not need anymore to wait until it will include these meta headers in its processed shard headers to determine this)

process/track/shardBlockTrack.go
SebastianMarian: I have added a warn log. Actually getLastNotarizedHeader could not return err as notarizedHeaders map is initialized with startHeaders(genesis header for each shard) on New.. method.This method returns the new finalized headers constructed above the last finalized one in self chain. If the chain didn't advance and the node is sync or the last notarized header could not be determined, this method has nothing to return. Its return is used only to notify the registered handlers, if they will exist, that a new finalized header(s) in self chain has/have appeared.

process/track/blockProcessor.go
SebastianMarian: I just wanted to avoid the consumed time with the other checks which are made before the nonce check in headerValidator. Now I checked again the IsHeaderConstructionValid method and it checks only the round before the nonce so I removed this check here.

process/track/blockProcessor.go
SebastianMarian: Done. I just wanted to avoid the consumed time with the other checks which are made before the nonce check in headerValidator, as here in checkHeaderFinality the headers are already sorted. Now I have checked again the IsHeaderConstructionValid method and it checks only the round before the nonce so I removed this check here.

process/track/blockProcessor.go
SebastianMarian: This method from the link is good if you want to determine only the height of the trie. In my case I need to returns the real value from each node in the longest chain (slice of headers) so it would be much complicated as I need to use additional lists. More than this, for our case the method from the link would consumes much time as it would use a lot of deletes from the lists. In our case when there could not be more than one chain with many children and the rest of the nodes from this trie would have no children (because there could be hardly more than one fork) would not make sense to change this approach. Its for sure that it would be calculated near than instantly as this computation is done only upward from the last final and all the records behind final would be deleted.

process/track/baseBlockTrack.go
SebastianMarian: Changed

process/track/interface.go
SebastianMarian: ok

process/track/blockNotarizer.go
SebastianMarian: Refactored

PR #867, opened 2019-12-19
fixed a bug that caused delayed recovery to a shard when a forkdetect…

iulianpascalau:
fixed a bug that caused delayed recovery to a shard when a forkdetector error actually prevented a message to be broadcast on the network


consensus/spos/worker.go
```diff
@@ -252,6 +252,8 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, _ func(buffToS
 			hex.EncodeToString(cnsDta.ChainID),
 			hex.EncodeToString(wrk.chainID),
 		)
+		log.Debug("consensus data chain ID mismatch",
```
sasurobert: log.Warn

PR #900, opened 2019-12-23
fix pruning on rollback

BeniaminDrasovean:
Bug 1: if root hash stays the same in some consecutive blocks, and then one block does a rollback, rootHash+newRootIdentifier will be pruned from db, thus removing values that are needed. 

Bug 2: When bootstrapping from storage, if there were some trie snapshots, they were not taken in consideration.

### DIFF
--- a/data/trie/patriciaMerkleTrie.go
+++ b/data/trie/patriciaMerkleTrie.go
@@ -74,7 +74,7 @@ func (tr *patriciaMerkleTrie) Get(key []byte) ([]byte, error) {
 
 	val, err := tr.root.tryGet(hexKey, tr.trieStorage.Database())
 	if err != nil {
-		log.Debug("trie get", "error", key)
+		log.Trace("trie get", "error", key)
 	}
 
 	return val, err
--- a/data/trie/patriciaMerkleTrie.go
+++ b/data/trie/patriciaMerkleTrie.go
@@ -72,7 +72,12 @@ func (tr *patriciaMerkleTrie) Get(key []byte) ([]byte, error) {
 	}
 	hexKey := keyBytesToHex(key)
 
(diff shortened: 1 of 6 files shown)


integrationTests/state/stateTrie_test.go
```diff
@@ -1282,6 +1282,67 @@ func TestRollbackBlockAndCheckThatPruningIsCancelled(t *testing.T) {
...
+	round = integrationTests.IncrementAndPrintRound(round)
+	nonce++
+	round, nonce = integrationTests.ProposeAndSyncOneBlock(t, nodes, idxProposers, round, nonce)
+
+	rootHashOfFirstBlock, _ := shardNode.AccntState.RootHash()
+
+	assert.Equal(t, uint64(1), nodes[0].BlockChain.GetCurrentBlockHeader().GetNonce())
+	assert.Equal(t, uint64(1), nodes[1].BlockChain.GetCurrentBlockHeader().GetNonce())
+
+	round, _ = integrationTests.ProposeAndSyncOneBlock(t, nodes, idxProposers, round, nonce)
```
golangcibot: ineffectual assignment to `round` (from `ineffassign`)

data/trie/patriciaMerkleTrie.go
```diff
@@ -72,7 +72,12 @@ func (tr *patriciaMerkleTrie) Get(key []byte) ([]byte, error) {
 	}
 	hexKey := keyBytesToHex(key)
 
-	return tr.root.tryGet(hexKey, tr.trieStorage.Database())
+	val, err := tr.root.tryGet(hexKey, tr.trieStorage.Database())
+	if err != nil {
```
iulianpascalau: We should not follow the pattern in which you both log and return the error as you might get double prints
Maybe switch this to Trace?

PR #901, opened 2019-12-23
EN-4436: bitmap check number of signers

bogdan-rosianu:


integrationTests/singleShard/block/consensusNotAchieved_test.go
```diff
@@ -0,0 +1,143 @@
...
+	nodesMap[0][0].BroadcastBlock(body, hdr)
+	time.Sleep(stepDelay)
+
+	// the block should have not pass the interceptor
+	assert.Equal(t, int32(0), nodesMap[0][1].CounterHdrRecv)
+
+	round = integrationTests.IncrementAndPrintRound(round)
+	nonce++
+
+	bitMapEnough := []byte{1, 0, 1, 0, 1}
```
AdoAdoAdo: I don't think this is ok, you have 5 nodes per shard and your bitmap is selecting the nodes with indexes 0, 16, 32.

integrationTests/singleShard/block/consensusNotAchieved_test.go
```diff
@@ -0,0 +1,143 @@
...
+	body, hdr, _ = proposeBlock(nodesMap[0][0], round, nonce, bitMapEnough)
+	assert.NotNil(t, body)
+	assert.NotNil(t, hdr)
+
+	nodesMap[0][0].BroadcastBlock(body, hdr)
+	time.Sleep(stepDelay)
+
+	// this block should have passed the interceptor
+	assert.Equal(t, int32(1), nodesMap[0][1].CounterHdrRecv)
```
AdoAdoAdo: If this block passed, then this is suspicious as the bitmap was not correctly constructed.

integrationTests/testProcessorNodeWithMultisigner.go
```diff
@@ -58,6 +55,26 @@ func NewTestProcessorNodeWithCustomNodesCoordinator(
...
 	)
+	singleSigner := &mock.SignerMock{
+		VerifyStub: func(public crypto.PublicKey, msg []byte, sig []byte) error {
+			return nil
+		},
+		SignStub: func(private crypto.PrivateKey, msg []byte) ([]byte, error) {
+			return nil, nil
+		},
+	}
+	keyGen := &mock.KeyGenMock{}
```
AdoAdoAdo: I would prefer the signing and verification was the real one not mocked, if we need these mocked we already have the other test processor nodes, this was supposed to be, as the name of the file implies, with multisigner(real)

process/headerCheck/headerSignatureVerify.go
```diff
@@ -120,6 +127,26 @@ func (hsv *HeaderSigVerifier) VerifySignature(header data.HeaderHandler) error {
...
+	for index := range bitmap {
+		numOfOnesInBitmap += bits.OnesCount8(bitmap[index])
+	}
+
+	minNumRequiredSignatures := consensusSize*2/3 + 1
+	if numOfOnesInBitmap >= minNumRequiredSignatures {
+		return nil
+	}
+
+	log.Warn("not enough signatures",
```
iulianpascalau: Might leave this as Debug, otherwise someone could "log flood" the other peers. 
Added task to create unit tests for this file.

integrationTests/singleShard/block/consensusNotAchieved_test.go
bogdan-rosianu: changed

integrationTests/singleShard/block/consensusNotAchieved_test.go
bogdan-rosianu: it passed because the size of the bitmap was not checked. and the function only counted the number of ones in the bitmap.
Now I added a check for the bitmap size. Also, changed the logic for calculating the size when the bitmap is created.

integrationTests/testProcessorNodeWithMultisigner.go
bogdan-rosianu: created a `CreateNodesWithNodesCoordinatorAndHeaderSigVerifier`

process/headerCheck/headerSignatureVerify.go
bogdan-rosianu: done

PR #914, opened 2019-12-26
added 2 WS middleware limiters

iulianpascalau:
added 2 middle-ware web server limiters


api/middleware/globalThrottler.go
```diff
@@ -0,0 +1,34 @@
...
+
+// globalThrottler is a middleware global limiter used to limit total number of simultaneous requests
+type globalThrottler struct {
+	queue chan struct{}
+}
+
+// NewGlobalThrottler creates a new instance of a globalThrottler
+func NewGlobalThrottler(maxConnections uint32) *globalThrottler {
+	return &globalThrottler{
+		queue: make(chan struct{}, maxConnections),
```
sasurobert: I think maxConnections should be at least 1.

facade/elrondNodeFacade.go
```diff
@@ -140,16 +163,52 @@ func (ef *ElrondNodeFacade) startRest() {
...
+		return nil, err
+	}
+	go ef.sourceLimiterReset(sourceLimiter)
+
+	globalLimiter := middleware.NewGlobalThrottler(ef.wsAntifloodConfig.SimultaneousRequests)
+
+	return []api.MiddlewareLimiter{sourceLimiter, globalLimiter}, nil
+}
+
+func (ef *ElrondNodeFacade) sourceLimiterReset(reset resetter) {
```
sasurobert: resetHandler ?

api/middleware/sourceThrottler.go
```diff
@@ -0,0 +1,60 @@
...
+		c.Next()
+	}
+}
+
+// Reset resets all accumulated counters
+func (st *sourceThrottler) Reset() {
+	st.mutRequests.Lock()
+	st.sourceRequests = make(map[string]uint32)
+	st.mutRequests.Unlock()
+}
```
sasurobert: this satisfies an interface, you could add isinterfacenil check as well.

api/middleware/globalThrottler.go
iulianpascalau: right, this might lead to a deadlock. Good catch! :+1:

iulianpascalau: done

facade/elrondNodeFacade.go
iulianpascalau: done

api/middleware/sourceThrottler.go
iulianpascalau: done

api/middleware/sourceThrottler.go
```diff
@@ -0,0 +1,65 @@
...
+			c.AbortWithStatus(http.StatusTooManyRequests)
+			return
+		}
+
+		c.Next()
+	}
+}
+
+// Reset resets all accumulated counters
+func (st *sourceThrottler) Reset() {
```
ccorcoveanu: If we are using this approach we should know there is an edge - we can actually hit with double requests/interval than set up in config - we just have to catch the reseting point between request batches.
Isn't there a middleware package from a third party that would have bees useful?

api/middleware/globalThrottler.go
```diff
@@ -0,0 +1,43 @@
...
+	}, nil
+}
+
+// Limit returns the handler func used by the gin server to limit simultaneous requests
+func (gt *globalThrottler) Limit() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		select {
+		case gt.queue <- struct{}{}:
+		default:
+			c.AbortWithStatus(http.StatusTooManyRequests)
```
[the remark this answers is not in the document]
iulianpascalau: Seemed a good idea but then, taking a look at how it was implemented, found out the following:
```GO
// AbortWithError calls `AbortWithStatus()` and `Error()` internally.
// This method stops the chain, writes the status code and pushes the specified error to `c.Errors`.
// See Context.Error() for more details.
func (c *Context) AbortWithError(code int, err error) *Error {
	c.AbortWithStatus(code)
	return c.Error(err)
}
```
Which calls `c.AbortWithStatus(code)` but also appends the error to the current context and causes more processing on the server side. Suggestion to leave it like it was?

api/api.go
```diff
@@ -65,6 +71,13 @@ func Start(elrondFacade MainApiHandler) error {
 	}
 	ws = gin.Default()
 	ws.Use(cors.Default())
+	for _, limiter := range limiters {
```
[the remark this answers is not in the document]
iulianpascalau: Done

api/middleware/sourceThrottler.go
iulianpascalau: Agree, this will still plague the p2p limiters also as their implementations follow the same principles. This burst can be achieved once on two rounds thus. Tried to keep the implementation as lightweight as it could even with the above mentioned problem. If we spot an attack, we can lower the configurable thresholds. 

I have searched for some already implemented middleware limiters for the gin web server but found simple solutions like the one implemented here. Decided to go with our own variant to match the implementation code with the rest of the code base.

api/middleware/globalThrottler.go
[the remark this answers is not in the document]
ccorcoveanu: Yes, I saw the type assertion too now, ok 👍

PR #915, opened 2019-12-26
updated libp2p's libraries to latest releases

iulianpascalau:
- updated libp2p's libraries to latest releases
- refactored pubsub's topic usage
- removed skipping the failing p2p tests with race because of https://github.com/koron/go-ssdp dependency


p2p/libp2p/netMessenger.go
```diff
@@ -387,7 +416,7 @@ func (netMes *networkMessenger) HasTopic(name string) bool {
 // HasTopicValidator returns true if the topic has a validator set
 func (netMes *networkMessenger) HasTopicValidator(name string) bool {
 	netMes.mutTopics.RLock()
-	validator, _ := netMes.topics[name]
+	validator, _ := netMes.processors[name]
```
golangcibot: S1005: should write `validator := netMes.processors[name]` instead of `validator, _ := netMes.processors[name]` (from `gosimple`)

iulianpascalau: Good catch, thanks! :))

p2p/libp2p/netMessenger.go
```diff
@@ -437,16 +466,18 @@ func (netMes *networkMessenger) Broadcast(topic string, buff []byte) {
...
 	netMes.mutTopics.Lock()
 	defer netMes.mutTopics.Unlock()
-	validator, found := netMes.topics[topic]
+
+	_, found := netMes.topics[topic]
 	if !found {
 		return p2p.ErrNilTopic
 	}
+	validator := netMes.processors[topic]
 	if validator != nil {
```
SebastianMarian: Is this condition correct? It not should be validator == nil ?

p2p/libp2p/netMessenger.go
```diff
@@ -467,20 +498,20 @@ func (netMes *networkMessenger) RegisterMessageProcessor(topic string, handler p
...
 	netMes.mutTopics.Lock()
 	defer netMes.mutTopics.Unlock()
-	validator, found := netMes.topics[topic]
 
+	_, found := netMes.topics[topic]
 	if !found {
 		return p2p.ErrNilTopic
 	}
-
+	validator := netMes.processors[topic]
```
sasurobert: validator, ok := ... if !ok

iulianpascalau: actually, this variant covers the case in witch (due to a programming error) we inserted value nil with a topic key. In that case ok would have returned true but the value would have been nil

p2p/libp2p/netMessenger.go
iulianpascalau: The condition is correct, you can not register a topic validator (messageProcessor) on an already set-up topic. (we are not allowing setting more than 1 topic validators on a specific topic)

iulianpascalau:
> go mod tidy to clean up go.sum.

already called `go mod tidy` to clean up go.sum

p2p/libp2p/netMessenger.go
SebastianMarian: ok

PR #950, opened 2020-01-07
EN-5579: add close method for all store units

miiu96:


storage/storageUnit/storageunit.go
```diff
@@ -112,6 +115,14 @@ func (s *Unit) Put(key, data []byte) error {
 	return err
 }
 
+// Close will close unit
+func (s *Unit) Close() error {
+	err := s.persister.Close()
```
AdoAdoAdo: The Close should return an error but you log the error and always return nil, you should return the error instead.

storage/pruning/pruningStorer.go
```diff
@@ -202,6 +202,16 @@ func (ps *PruningStorer) Get(key []byte) ([]byte, error) {
 	return v.([]byte), nil
 }
 
+// Close will close PruningStorer
+func (ps *PruningStorer) Close() error {
+	for _, persister := range ps.activePersisters {
+		err := persister.persister.Close()
+		log.LogIfError(err)
```
AdoAdoAdo: the function never returns an error, it will always return nil

storage/storageUnit/storageunit.go
miiu96: done.

data/trie/patriciaMerkleTrie.go
```diff
@@ -286,6 +286,11 @@ func (tr *patriciaMerkleTrie) String() string {
 	return writer.String()
 }
 
+// ClosePersister will close trie persister
+func (tr *patriciaMerkleTrie) ClosePersister() {
+	_ = tr.db.Close()
```
bogdan-rosianu: this method returns an error. In the interface, the `ClosePersister()` is void type because in AccountsDB there is no returned error. So please change the interface's method to return an error and return it here, or ar least log here if there is an error

bogdan-rosianu: better log the error here and leave the interface as it is

node/mock/accountsStub.go
```diff
@@ -21,6 +21,10 @@ type AccountsStub struct {
 	RecreateTrieCalled          func(rootHash []byte) error
 }
 
+func (aam *AccountsStub) ClosePersister() {
```
bogdan-rosianu: change the receiver to `as`

process/mock/accountsStub.go
```diff
@@ -43,6 +43,10 @@ func (aam *AccountsStub) Commit() ([]byte, error) {
 	return nil, errNotImplemented
 }
 
+func (aam *AccountsStub) ClosePersister() {
```
bogdan-rosianu: change the receiver to `as`

process/mock/chainStorerMock.go
```diff
@@ -15,6 +15,13 @@ type ChainStorerMock struct {
 	PutCalled       func(unitType dataRetriever.UnitType, key []byte, value []byte) error
 	GetAllCalled    func(unitType dataRetriever.UnitType, keys [][]byte) (map[string][]byte, error)
 	DestroyCalled   func() error
+	CloseAllCalled  func()
+}
+
+func (bc *ChainStorerMock) CloseAll() {
```
bogdan-rosianu: change the receiver to `csm`

storage/pruning/pruningStorer.go
bogdan-rosianu: agree. maybe define a base error something like `ErrClosingPersisters` and save all errors and return them using `fmt.Errorf`.

storage/storageUnit/storageunit.go
bogdan-rosianu: :+1:

storage/pruning/pruningStorer.go
miiu96: method return error now if a persister cannot be closed successfully

process/mock/accountsStub.go
miiu96: done./

data/trie/patriciaMerkleTrie.go
miiu96: error is now returned

node/mock/accountsStub.go
miiu96: done.

process/mock/chainStorerMock.go
miiu96: done.

process/mock/accountsStub.go
```diff
@@ -25,123 +25,123 @@ type AccountsStub struct {
...
 	}
 
 	return errNotImplemented
 }
 
 // IsInterfaceNil returns true if there is no value under the interface
-func (aam *AccountsStub) IsInterfaceNil() bool {
-	if aam == nil {
+func (as *AccountsStub) IsInterfaceNil() bool {
+	if as == nil {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

process/mock/chainStorerMock.go
```diff
@@ -15,74 +15,83 @@ type ChainStorerMock struct {
...
+		return csm.DestroyCalled()
 	}
 	return nil
 }
 
 // IsInterfaceNil returns true if there is no value under the interface
-func (bc *ChainStorerMock) IsInterfaceNil() bool {
-	if bc == nil {
+func (csm *ChainStorerMock) IsInterfaceNil() bool {
+	if csm == nil {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

node/mock/accountsStub.go
```diff
@@ -21,65 +21,69 @@ type AccountsStub struct {
...
+
+func (as *AccountsStub) RecreateTrie(rootHash []byte) error {
+	return as.RecreateTrieCalled(rootHash)
 }
 
 // IsInterfaceNil returns true if there is no value under the interface
-func (aam *AccountsStub) IsInterfaceNil() bool {
-	if aam == nil {
+func (as *AccountsStub) IsInterfaceNil() bool {
+	if as == nil {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

storage/errors.go
```diff
@@ -78,3 +78,6 @@ var ErrInvalidNumberOfEpochsToSave = errors.New("invalid number of epochs to sav
 
 // ErrInvalidNumberOfActivePersisters signals that an invalid number of active persisters has been provided
 var ErrInvalidNumberOfActivePersisters = errors.New("invalid number of active persisters")
+
+//ErrClosingPersisters signals that not all persisters was closed
```
bogdan-rosianu: were closed *
Also, I don't think it is enough. We won't know why any persister failed to close. So, at least log the errors when a persister.Close() fails.

cmd/node/main.go
```diff
@@ -869,9 +869,9 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	log.Info("terminating at user's signal...")
 
 	log.Info("closing all store units....")
-	dataComponents.Store.CloseAll()
-	coreComponents.Trie.ClosePersister()
-	stateComponents.PeerAccounts.ClosePersister()
+	_ = dataComponents.Store.CloseAll()
+	_ = coreComponents.Trie.ClosePersister()
```
bogdan-rosianu: why do you ignore the errors ? you added them with no purpose then

miiu96: error is logged now

data/state/accountsDB.go
```diff
@@ -98,6 +102,32 @@ func (adb *AccountsDB) addCodeToTrieIfMissing(codeHash []byte, code []byte) erro
 	return nil
 }
 
+// ClosePersister will close trie persister
+func (adb *AccountsDB) ClosePersister() error {
+	closedSuccessfully := true
+
+	err := adb.mainTrie.ClosePersister()
+	log.LogIfError(err)
```
AdoAdoAdo: you are anyway checking for error != nil, so I wouldn't do logIfError as that would mean you are doing the check twice.
you can directly log it inside if err!= nil{}

storage/storageUnit/storageunit.go
```diff
@@ -112,6 +115,14 @@ func (s *Unit) Put(key, data []byte) error {
...
+	log.LogIfError(err)
```
AdoAdoAdo: unless there is something different you are logging compared to what you are returning, I wouldn't do both.

data/state/accountsDB.go
miiu96: done.

storage/storageUnit/storageunit.go
miiu96: fixed

PR #958, opened 2020-01-09
Api middleware

bogdan-rosianu:


cmd/node/config/config.toml
```diff
@@ -270,14 +270,14 @@
...
         PeerBanDurationInSeconds = 300
     [Antiflood.WebServer]
-        # SimultaneousRequests represents the number of councurrent requests accepted by the web server
+        # SimultaneousRequests represents the number of concurrent requests accepted by the web server
         # this is a global throttler that acts on all http connections regardless of the originating source
         SimultaneousRequests = 100
         # SameSourceRequests defines how many requests are allowed from the same source in the specified
         # time frame (SameSourceResetIntervalInSec)
-        SameSourceRequests = 10
+        SameSourceRequests = 10000
```
ccorcoveanu: This number could prove quite large but combined with 100 concurrent requests limit can be mitigated. It will be interesting how it behaves under some load

PR #974, opened 2020-01-13
EN-5599: topic antiflood

bogdan-rosianu:
Implemented topic antiflood with calls from worker, heartbeat and header resolver


process/throttle/antiflood/topicFloodPreventer.go
```diff
@@ -0,0 +1,114 @@
...
+	_, ok = tfp.counterMap[topic][identifier]
+	if !ok {
+		tfp.counterMap[topic][identifier] = 1
+		return true
+	}
+
+	// if this was already in the map, just increment it
+	tfp.counterMap[topic][identifier]++
+
+	if tfp.counterMap[topic][identifier] > tfp.maxMessagesForTopic(topic) {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cmd/node/config/config.toml
```diff
@@ -264,6 +264,10 @@
     [Antiflood.Cache]
         Size = 5000
         Type = "LRU"
+    [Antiflood.Topic]
+        DefaultMaxRequestsPerSec = 75
```
iulianpascalau: Might change the namings as the keyword Requests will be used in antiflooding in a next PR

cmd/node/factory/structs.go
```diff
@@ -595,6 +601,17 @@ func startResetingFloodPreventer(floodPreventer p2p.FloodPreventer) {
 	}()
 }
 
+func handleTopicFloodPreventer(topicFloodPreventer p2p.TopicFloodPreventer, config *config.TopicAntifloodConfig) {
+	topicFloodPreventer.SetMaxMessagesForTopic("heartbeat", config.HeartbeatMaxMessagesPerSec)
+	go func() {
```
iulianpascalau: this can be merged with the startResettingFloodPreventer, init can be taken out of this function

p2p/antiflood/p2pAntiflood.go
```diff
@@ -50,6 +54,24 @@ func (af *p2pAntiflood) CanProcessMessage(message p2p.MessageP2P, fromConnectedP
 	return nil
 }
 
+// CanProcessMessageOnTopic signals if a p2p message can be processed or not for a given topic
+func (af *p2pAntiflood) CanProcessMessageOnTopic(message p2p.MessageP2P, fromConnectedPeer p2p.PeerID, topic string) error {
```
iulianpascalau: I would just leave the function signature as 
```GO
func (af *p2pAntiflood) CanProcessMessageOnTopic(peer p2p.PeerID, topic string) error {
```
as the message payload size is computed on the other flood preventer. It would be clearer how to use it

p2p/antiflood/p2pAntiflood.go
```diff
@@ -50,6 +54,24 @@ func (af *p2pAntiflood) CanProcessMessage(message p2p.MessageP2P, fromConnectedP
...
+	topicFloodPreventer := af.TopicFloodPreventer
+	if check.IfNil(topicFloodPreventer) {
+		return p2p.ErrNilTopicFloodPreventer
```
iulianpascalau: this is already checked on the constructor

p2p/antiflood/p2pAntiflood.go
```diff
@@ -50,6 +54,24 @@ func (af *p2pAntiflood) CanProcessMessage(message p2p.MessageP2P, fromConnectedP
...
+	}
+	if message == nil {
+		return p2p.ErrNilMessage
+	}
+
+	ok := topicFloodPreventer.Accumulate(message.Peer().Pretty(), topic)
+	if !ok {
+		return fmt.Errorf("%w in p2pAntiflood for originator", p2p.ErrSystemBusy)
```
iulianpascalau: might output the pid.Pretty also in the error

test_test.go
```diff
@@ -0,0 +1,16 @@
+package elrond_go
+
+import (
+	"encoding/base64"
+	"encoding/binary"
+	"fmt"
+	"testing"
+)
+
+func Test1(t *testing.T) {
```
iulianpascalau: Should remove this file

dataRetriever/resolvers/headerResolver.go
```diff
@@ -88,6 +96,18 @@ func (hdrRes *HeaderResolver) SetEpochHandler(epochHandler dataRetriever.EpochHa
 	return nil
 }
 
+func (hdrRes *HeaderResolver) topicNameForAntiflood() string {
+	prefix := factory.ShardBlocksTopic
+	if hdrRes.shardCoordinator.SelfId() > hdrRes.shardCoordinator.NumberOfShards() {
```
iulianpascalau: Please add a TODO here as this won't work on metachain nodes and intra shard header resolvers.
This should have been done on the header interceptor rather than header resolver.

p2p/antiflood/p2pAntiflood.go
bogdan-rosianu: right now I use the message to get the peer ID. Should I use the fromConnected to ge the peer and send it to this function ?

p2p/antiflood/p2pAntiflood.go
bogdan-rosianu: also in `CanProcessMessage` there is a nil check. Should I remove that check also ?

cmd/node/config/config.toml
bogdan-rosianu: done

cmd/node/factory/structs.go
bogdan-rosianu: done

p2p/antiflood/p2pAntiflood.go
bogdan-rosianu: done

test_test.go
bogdan-rosianu: done

dataRetriever/resolvers/headerResolver.go
bogdan-rosianu: done

p2p/antiflood/p2pAntiflood.go
iulianpascalau: Depends, if used inside worker and heartbeat, you should take the peer ID from the message. Inside the header interceptor, you should use the fromConnected peer.

p2p/antiflood/p2pAntiflood.go
iulianpascalau: Right, since we are checking on the underlying composing pointers. Should remain like this. Also, might add a supplemental check on IsInterfaceNil function

p2p/antiflood/p2pAntiflood.go
bogdan-rosianu: done

p2p/antiflood/p2pAntiflood.go
bogdan-rosianu: done

consensus/spos/bls/blsWorker.go
```diff
@@ -5,6 +5,8 @@ import (
 	"github.com/ElrondNetwork/elrond-go/consensus/spos"
 )
 
+const peerMaxMessagesPerSec = uint32(2)
```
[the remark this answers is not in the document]
bogdan-rosianu: @iulianpascalau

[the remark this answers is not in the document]
iulianpascalau: It will work as expected because we are taking into account the message.Peer() here (not the connected peers that send the message to self). So we will process at max (2 * num consensus peers) messages in a consensus round.
For each consensus round we need at max one message originated from a consensus peer. The value 2 took into account the possibility to receive in the same round the messages from 2 adjacent rounds.

api/transaction/routes_test.go
```diff
@@ -277,6 +277,88 @@ func TestSendTransaction_ReturnsSuccessfully(t *testing.T) {
...
+	t.Parallel()
+
+	ws := startNodeServerWrongFacade()
+	req, _ := http.NewRequest("POST", "/transaction/send-multiple", nil)
+	resp := httptest.NewRecorder()
+	ws.ServeHTTP(resp, req)
+
+	transactionResponse := TransactionResponse{}
+	loadResponse(resp.Body, &transactionResponse)
+	assert.Equal(t, resp.Code, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
bogdan-rosianu: these tests arrived by mistake here.  however, done

p2p/antiflood/p2pAntiflood.go
```diff
@@ -50,7 +54,22 @@ func (af *p2pAntiflood) CanProcessMessage(message p2p.MessageP2P, fromConnectedP
...
+		return fmt.Errorf("%w in p2pAntiflood for originator. peer id = %s", p2p.ErrSystemBusy, peer.Pretty())
+	}
+
+	return nil
+}
+
 // IsInterfaceNil returns true if there is no value under the interface
 func (af *p2pAntiflood) IsInterfaceNil() bool {
-	return af == nil || check.IfNil(af.FloodPreventer)
+	return af == nil || check.IfNil(af.FloodPreventer) || check.IfNil(af.TopicFloodPreventer)
```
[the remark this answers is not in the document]
bogdan-rosianu: I know what you mean but it won't work if these 2 interfaces who reside here would be nil. So it makes sense that `IsInterfaceNil` checks them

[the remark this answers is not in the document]
iulianpascalau: This is a pattern we started using when a struct composes over one or more interfaces (or struct pointers) (@traian-schiau started using this on feat/network-sharding branches)

process/throttle/antiflood/topicFloodPreventer.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+const topicMinMessages = 1
+const wildcardCharacter = "*"
+
+var log = logger.GetOrCreate("process/throttle/antiflood")
+
+// topicFloodPreventer represents a cache of quotas per peer used in antiflooding mechanism
+type topicFloodPreventer struct {
+	mutOperation              *sync.RWMutex
```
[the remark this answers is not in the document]
bogdan-rosianu: renamed to `mutTopicMaxMessages`

p2p/antiflood/p2pAntiflood.go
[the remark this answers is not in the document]
AdoAdoAdo: This can be checked on the constructor instead of this function.
IsInterfaceNil should only check if the interface is nil.

PR #980, opened 2020-01-14
[Network sharding] - Rework min out of shard connection limit

traian-schiau:


p2p/libp2p/networksharding/kadSharder.go
```diff
@@ -82,7 +83,7 @@ func (ks *kadSharder) SortList(peers []peer.ID, ref peer.ID) ([]peer.ID, bool) {
 	// for balance we should have between 1 and 20% connections outside of shard
 	peerCnt := len(peers)
 	inShardCnt := inShardCount(sl)
-	balanced := peerCnt > inShardCnt
+	balanced := getMinOOS(ks.prioBits, peerCnt) <= (peerCnt - inShardCnt)
```
iulianpascalau: Thought of this solution but I'm afraid is not enough. Setting the balance to false will not solve the issue found on BoN testing described below.
After a long time (around one day) all peers had almost only intra shard peers. This happened because over time, the discoverer found more and more intra-shard peers and thus, their relative distance promote them in the sorted list. By having this balance return false if the number of connected cross shard peers drop under a threshold will just keep the discoverer running but not preventing the immediately closing of a new connection to the cross shard peer.
That is the main reason I came up with the list based implementation. After one day of running, the peers have the desired levels set (initially one half for intra shard and one half for cross shard)

traian-schiau: a similar level (50 50) can be achieved by setting the priobits value to 1

iulianpascalau: not guaranteed, anyway, can go as it is, we might test this soon

PR #984, opened 2020-01-15
En 5614 kad dht parameters

iulianpascalau:
added parameters for kad dht bucket size and routing table refresh interval


p2p/libp2p/factory/peerDiscovererFactory.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/ElrondNetwork/elrond-go/p2p/libp2p/discovery"
+)
+
+type peerDiscovererFactory struct {
+	p2pConfig config.P2PConfig
+}
+
+// NewPeerDiscovererFactory creates a new instance of peer discovery factory
+func NewPeerDiscovererFactory(pConfig config.P2PConfig) *peerDiscovererFactory {
+	return &peerDiscovererFactory{
```
bogdan-rosianu: maybe check the config for nil and also return an error ?

p2p/libp2p/factory/peerDiscovererFactory.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	return discovery.NewNullDiscoverer(), nil
+}
+
+func (pdf *peerDiscovererFactory) createKadDhtPeerDiscoverer() (p2p.PeerDiscoverer, error) {
+	if pdf.p2pConfig.KadDhtPeerDiscovery.RefreshIntervalInSec < 1 {
+		return nil, fmt.Errorf("%w for RefreshIntervalInSec, expected value > 0", p2p.ErrInvalidValue)
+	}
+	if pdf.p2pConfig.KadDhtPeerDiscovery.RoutingTableRefreshIntervalInSec < 1 {
```
bogdan-rosianu: I don't like that you check 2 times the same config values. Can you skip the checking in one of them ?
Right now, you check the config and then send them to the NewKadDhtPeerDiscoverer after converting to time.Duration so right now there is no way that the values would arrive broken there

cmd/seednode/config/p2p.toml
```diff
@@ -43,4 +43,10 @@
...
     #phase but will accept connections and will do the network discovery if another peer connects to it
-   InitialPeerList = []
\ No newline at end of file
+   InitialPeerList = []
+
+   #kademlia's routing table bucket size
+   BucketSize = 100
+
+   #RoutingTableRefreshIntervalInSec defines how many seconds should past between 2 kad routing table auto refresh calls
+   RoutingTableRefreshIntervalInSec = 300
```
sasurobert: add empty line at the end of file.

p2p/libp2p/factory/peerDiscovererFactory.go
iulianpascalau: Agree

p2p/libp2p/factory/peerDiscovererFactory.go
sasurobert: config could be verified at NewPeerDiscovererFactory time.

p2p/libp2p/factory/peerDiscovererFactory.go
sasurobert: +1 check in a single time.

cmd/node/config/p2p.toml
```diff
@@ -48,3 +48,10 @@
     #If the initial peers list is left empty, the node will not try to connect to other peers during initial bootstrap
     #phase but will accept connections and will do the network discovery if another peer connects to it
     InitialPeerList = ["/ip4/127.0.0.1/tcp/10000/p2p/16Uiu2HAmAzokH1ozUF52Vy3RKqRfCMr9ZdNDkUQFEkXRs9DqvmKf"]
+
+    #kademlia's routing table bucket size
+    BucketSize = 100
+
+    #RoutingTableRefreshIntervalInSec defines how many seconds should past between 2 kad routing table auto refresh calls
```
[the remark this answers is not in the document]
iulianpascalau: right

p2p/libp2p/factory/peerDiscovererFactory.go
iulianpascalau: pConfig can not be nil here as it has been pass as value not pointer

p2p/libp2p/factory/peerDiscovererFactory.go
iulianpascalau: There are some subtleties regarding factories. The pattern should be uniformized. Although it works in this use case (create method is called just after the factory has been instantiated, it might "lazy fail" in some other use cases (such as intercepted transaction factory). Will try to find out a good solution to this problem. Long story, short, for now, will remove the checking in this create method.

p2p/libp2p/factory/peerDiscovererFactory.go
iulianpascalau: > config could be verified at NewPeerDiscovererFactory time.

The checking logic will be duplicated in each constructor. I would avoid duplicate code in this case

cmd/seednode/config/p2p.toml
iulianpascalau: done

p2p/libp2p/factory/peerDiscovererFactory.go
iulianpascalau: Left a TODO

PR #1004, opened 2020-01-20
Merge dev end of epoch p1 20.01.20

raduchis:


integrationTests/state/stateTrieSync_test.go
```diff
@@ -2,14 +2,14 @@ package state
 
 import (
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/core"
```
sasurobert: goimports

PR #1007, opened 2020-01-21
fixed sorting for tx preprocessors - added unit test for rewards to b…

sasurobert:
fixed sorting for tx preprocessors - added unit test for rewards to be always the last one. Added invalid transactions tx fee to the fee accumulator

### DIFF
--- a/data/block/block.go
+++ b/data/block/block.go
@@ -29,19 +29,20 @@ type MiniBlockSlice []*MiniBlock
 
 const (
 	// TxBlock identifies a miniblock holding transactions
-	TxBlock Type = 0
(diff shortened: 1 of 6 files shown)


data/block/block.go
```diff
@@ -18,7 +18,7 @@ import (
 // The body can be transmitted even before having built the heder and go through a prevalidation of each transaction
 
 // Type identifies the type of the block
-type Type uint8
+type Type uint32
```
SebastianMarian: Why do we need uint32? Do you think that we can have more than 256 miniblock types?

PR #1021, opened 2020-01-23
En 5649 reject headers with nonce under final

iulianpascalau:
added check on the header interceptors not to process old, under finality nonce block headers. This will prevent spam.


cmd/node/factory/interface.go
```diff
@@ -19,4 +20,8 @@ type HeaderSigVerifierHandler interface {
 	VerifyRandSeedAndLeaderSignature(header data.HeaderHandler) error
 	VerifySignature(header data.HeaderHandler) error
 	IsInterfaceNil() bool
-}
\ No newline at end of file
+}
+
+type interceptedDataFactoryWithFinalAttesterSetter interface {
```
sasurobert: add comment to interface.

cmd/node/factory/structs.go
```diff
@@ -780,6 +785,26 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
...
+func applyblockTrackerOnInterceptors(
+	finalityAttester process.FinalityAttester,
+	interceptorsContainer process.InterceptorsContainer,
+) error {
+
+	var foundErr error
+	interceptorsContainer.Iterate(func(key string, interceptor process.Interceptor) bool {
+		dataFactory := interceptor.InterceptedDataFactory()
+		setter, ok := dataFactory.(interceptedDataFactoryWithFinalAttesterSetter)
+		if ok {
```
sasurobert: if !ok return true else setter.set and return foundErr == nil. If you have a set of errors, you only catch the last one. Should save founderr into an other variable.

integrationTests/testProcessorNode.go
```diff
@@ -1320,3 +1321,16 @@ func (tpn *TestProcessorNode) initHeaderValidator() {
 
 	tpn.HeaderValidator, _ = block.NewHeaderValidator(argsHeaderValidator)
 }
+
+func (tpn *TestProcessorNode) applyBlockTrackerOnInterceptors() {
+	tpn.InterceptorsContainer.Iterate(func(key string, interceptor process.Interceptor) bool {
+		dataFactory := interceptor.InterceptedDataFactory()
+		setter, ok := dataFactory.(interceptedDataFactoryWithFinalAttesterSetter)
+		if ok {
+			foundErr := setter.SetFinalityAttester(tpn.BlockTracker)
```
sasurobert: same comments as in structs.go

process/interceptors/factory/nilFinalityAttester.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/data/block"
+	"github.com/ElrondNetwork/elrond-go/logger"
+)
+
+var log = logger.GetOrCreate("process/interceptors/factory")
+
+// nilFinalityAttester is a nil implementation that will cause all received headers
```
sasurobert: that returns default value, without checking actual finality.

process/mock/finalityAttesterStub.go
```diff
@@ -0,0 +1,25 @@
...
+type FinalityAttesterStub struct {
+	GetFinalHeaderCalled func(shardID uint32) (data.HeaderHandler, []byte, error)
+}
+
+func (fas *FinalityAttesterStub) GetFinalHeader(shardID uint32) (data.HeaderHandler, []byte, error) {
+	if fas.GetFinalHeaderCalled != nil {
+		return fas.GetFinalHeaderCalled(shardID)
+	}
+
+	return &block.Header{
```
sasurobert: define header separately like in nilFinalityAttester

process/track/baseBlockTrack.go
```diff
@@ -78,6 +92,20 @@ func (bbt *baseBlockTrack) receivedMetaBlock(headerHandler data.HeaderHandler, m
 		return
 	}
 
+	err := bbt.checkBlockBasicValidity(metaBlock)
```
sasurobert: no need for this as implementation is nil. log.Trace will be never called, neither return.

process/track/baseBlockTrack.go
```diff
@@ -89,6 +117,10 @@ func (bbt *baseBlockTrack) receivedMetaBlock(headerHandler data.HeaderHandler, m
 	bbt.blockProcessor.processReceivedHeader(metaBlock)
 }
 
+func (bbt *baseBlockTrack) checkBlockBasicValidity(_ data.HeaderHandler) error {
```
sasurobert: rename checkBasicBlockValidity. However I do not think this is needed. Interceptors are doing the basic validity.

process/interface.go
```diff
@@ -637,3 +640,9 @@ type EpochStartDataCreator interface {
 	VerifyEpochStartDataForMetablock(metaBlock *block.MetaBlock) error
 	IsInterfaceNil() bool
 }
+
+// FinalityAttester is able to manage the final blocks
+type FinalityAttester interface {
```
sasurobert: Maybe rename FinalityAttester to something nice :).

process/interceptors/factory/interceptedMetaHeaderDataFactory_test.go
```diff
@@ -141,3 +141,27 @@ func TestNewInterceptedMetaHeaderDataFactory_ShouldWorkAndCreate(t *testing.T) {
...
+	t.Parallel()
+
+	arg := createMockArgument()
+	imh, _ := NewInterceptedMetaHeaderDataFactory(arg)
+	newAttester := &nilFinalityAttester{}
+
+	err := imh.SetFinalityAttester(newAttester)
+
+	assert.Nil(t, err)
+	assert.True(t, imh.finalityAttester == newAttester)
```
AdoAdoAdo: why not assert equal?

process/interceptors/factory/interceptedShardHeaderDataFactory_test.go
```diff
@@ -80,3 +80,27 @@ func TestInterceptedShardHeaderDataFactory_ShouldWorkAndCreate(t *testing.T) {
...
+	t.Parallel()
+
+	arg := createMockArgument()
+	ish, _ := NewInterceptedShardHeaderDataFactory(arg)
+	newAttester := &nilFinalityAttester{}
+
+	err := ish.SetFinalityAttester(newAttester)
+
+	assert.Nil(t, err)
+	assert.True(t, ish.finalityAttester == newAttester)
```
AdoAdoAdo: why not assert.Equal

process/interceptors/factory/nilFinalityAttester.go
```diff
@@ -0,0 +1,31 @@
...
+//not to be actually checked
+type nilFinalityAttester struct {
```
AdoAdoAdo: I don't like the this mock in production code, this is only used in tests right?

I would move this to the mock folder instead
You also don't need any unit tests over a mock implementation

process/interceptors/singleDataInterceptor_test.go
```diff
@@ -82,14 +83,16 @@ func TestNewSingleDataInterceptor_NilInterceptorThrottlerShouldErr(t *testing.T)
...
+		factory,
 		&mock.InterceptorProcessorStub{},
 		&mock.InterceptorThrottlerStub{},
 	)
 
-	assert.NotNil(t, sdi)
-	assert.Nil(t, err)
+	require.False(t, check.IfNil(sdi))
+	require.Nil(t, err)
+	assert.True(t, sdi.InterceptedDataFactory() == factory)
```
AdoAdoAdo: why not assert.equal?

process/track/baseBlockTrack.go
AdoAdoAdo: why is this needed? does it need to implement any interface that requires this? otherwise I would remove it.

cmd/node/factory/interface.go
iulianpascalau: removed, no need for this at it will be injected

cmd/node/factory/structs.go
iulianpascalau: function removed, no longer necessary

integrationTests/testProcessorNode.go
iulianpascalau: function removed

process/interceptors/factory/nilFinalityAttester.go
iulianpascalau: implementation removed, no longer necessary

process/interface.go
iulianpascalau: Do you have a suggestion? FinalityAttester describes best what this interface should suppose to do: attest finality (in case of block headers)

process/block/interceptedBlocks/common.go
```diff
@@ -77,6 +82,24 @@ func checkHeaderHandler(hdr data.HeaderHandler) error {
...
 
+func checkHeaderHandlerFinality(hdr data.HeaderHandler, attester process.FinalityAttester) error {
+	attestedHdr, _, err := attester.GetFinalHeader(hdr.GetShardID())
+	if err != nil {
+		return err
+	}
+
+	isHeaderReceivedUnderFinal := hdr.GetNonce() < attestedHdr.GetNonce()
+	if isHeaderReceivedUnderFinal {
+		return fmt.Errorf("%w while checking attested hdr, got nonce: %d, final hdr nonce: %d",
```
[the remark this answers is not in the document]
iulianpascalau: ok, renamed to `checkHeaderHandlerHasLowerNonceThanFinalHeader`

process/interceptors/factory/interceptedMetaHeaderDataFactory_test.go
iulianpascalau: Test removed. The assert.True tested that the same pointer has been set (set worked correctly)

process/interceptors/factory/interceptedShardHeaderDataFactory_test.go
iulianpascalau: Test removed. The assert.True tested that the same pointer has been set (set worked correctly)

process/interceptors/factory/nilFinalityAttester.go
iulianpascalau: Nil implementation removed, no longer required as we inject the implementation on the constructor

process/interceptors/singleDataInterceptor_test.go
iulianpascalau: Test removed. The assert.True tested that the same pointer has been set (set worked correctly)

process/track/baseBlockTrack.go
iulianpascalau: function removed

iulianpascalau: function removed

process/track/baseBlockTrack.go
iulianpascalau: function removed

process/mock/finalityAttesterStub.go
iulianpascalau: done

process/interface.go
iulianpascalau: HeaderFinalityAttester? @AdoAdoAdo

process/track/baseBlockTrack.go
```diff
@@ -275,6 +276,64 @@ func (bbt *baseBlockTrack) GetCrossNotarizedHeader(shardID uint32, offset uint64
 	return bbt.crossNotarizer.getNotarizedHeader(shardID, offset)
 }
 
+// CheckBlockBasicValidity checks if the given header is valid related to some basic checks
+func (bbt *baseBlockTrack) CheckBlockBasicValidity(headerHandler data.HeaderHandler) error {
+	if check.IfNil(headerHandler) {
+		return fmt.Errorf("nil header handler")
+	}
+
+	finalHeader, _, err := bbt.GetFinalHeader(headerHandler.GetShardID())
```
AdoAdoAdo: this error is misleading, this function should only check if the block is valid or invalid.
This error is unrelated to this specific block, so it shouldn't be here.

iulianpascalau: done, split the function in 2

PR #1026, opened 2020-01-27
En-5654 : remove unused storage databases

miiu96:


storage/storageUnit/storageunit.go
```diff
@@ -345,10 +344,6 @@ func NewDB(dbType DBType, path string, batchDelaySeconds int, maxBatchSize int,
 		db, err = leveldb.NewDB(path, batchDelaySeconds, maxBatchSize, maxOpenFiles)
 	case LvlDbSerial:
 		db, err = leveldb.NewSerialDB(path, batchDelaySeconds, maxBatchSize, maxOpenFiles)
-	case BadgerDB:
```
SebastianMarian: You should also delete these constants declaration in the beginning of this file, and then you could also remove two more unit tests which are using them

miiu96: done.

PR #1031, opened 2020-01-28
En 5664 replace resolver finder with request handler

SebastianMarian:
* Replaced ResolverFinder with RequestHandler in sync package
* Added arguments to bootstrapper


process/sync/baseSync.go
```diff
@@ -316,104 +313,72 @@ func (boot *baseBootstrap) cleanCachesAndStorageOnRollback(header data.HeaderHan
...
-			err := hdrRes.RequestDataFromNonce(currentNonce, boot.epochHandler.Epoch())
-			if err != nil {
-				log.Debug("RequestDataFromNonce", "error", err.Error())
-				continue
-			}
-
-			err = boot.requestedItemsHandler.Add(key)
-			if err != nil {
-				log.Trace("add requested item with error", "error", err.Error())
+			if boot.shardCoordinator.SelfId() == sharding.MetachainShardId {
```
miiu96: maybe you can change logic here to reduce complexity if haveHeader { continue}...

miiu96: and also add  one more method in _blockBootstrapper_ interface to can requestHeaderByNonce(nonde uint64) and also if instruction where is checked shard id can be removed.

process/sync/shardblock.go
```diff
@@ -108,42 +66,13 @@ func NewShardBootstrap(
...
 	base.forkInfo = process.NewForkInfo()
 
-	miniBlocksRes, ok := miniBlocksResolver.(dataRetriever.MiniBlocksResolver)
-	if !ok {
-		return nil, process.ErrWrongTypeAssertion
-	}
-
-	boot.miniBlocksResolver = miniBlocksRes
-
 	boot.chRcvHdrNonce = make(chan bool)
```
miiu96: maybe you can create a method in baseSync that init common code
boot.chRcvHdrNonce = make(chan bool)
	boot.chRcvHdrHash = make(chan bool)
	boot.chRcvMiniBlocks = make(chan bool)
boot.chStopSync = make(chan bool)

	boot.setRequestedHeaderNonce(nil)
	boot.setRequestedHeaderHash(nil)
	boot.setRequestedMiniBlocks(nil)

process/sync/baseSync.go

process/sync/shardblock.go
SebastianMarian: Done

PR #1036, opened 2020-01-30
Merge dev into feat antiflood

bogdan-rosianu:


consensus/spos/worker_test.go
```diff
@@ -818,82 +818,6 @@ func TestWorker_ProcessReceivedMessageShouldErrIfFloodIsDetectedOnTopic(t *testi
 	assert.Equal(t, expectedErr, err)
 }
 
-func TestWorker_ProcessReceivedMessageWrongHeaderShouldErr(t *testing.T) {
```
iulianpascalau: Why did you remove this test? It appears both in development and feat/antiflood?

bogdan-rosianu: it is already in the file. it just moved to the bottom of it

iulianpascalau: right

PR #1049, opened 2020-02-04
EN-5703: ineffassign fixes

bogdan-rosianu:


p2p/libp2p/netMessenger.go
```diff
@@ -461,10 +461,14 @@ func (netMes *networkMessenger) RegisterMessageProcessor(topic string, handler p
 		netMes.Broadcast(topic, buffToSend)
 	}
 
-	err := netMes.pb.RegisterTopicValidator(topic, func(ctx context.Context, pid peer.ID, message *pubsub.Message) bool {
-		err := handler.ProcessReceivedMessage(NewMessage(message), broadcastHandler)
+	err := netMes.pb.RegisterTopicValidator(topic, func(ctx context.Context, pid peer.ID, pubsubMessage *pubsub.Message) bool {
+		message, err := NewMessage(pubsubMessage)
 		if err != nil {
-			log.Trace("p2p validator", "error", err.Error(), "topics", message.TopicIDs)
+			return false
```
iulianpascalau: Maybe add this log line?
```GO
log.Trace("p2p validator - new message", "error", err.Error(), "topics", pubsubMessage.TopicIDs)
```

p2p/libp2p/message.go
```diff
@@ -31,11 +31,12 @@ func NewMessage(message *pubsub.Message) *Message {
 	id, err := peer.IDFromBytes(msg.from)
 	if err != nil {
 		log.Trace("creating new p2p message", "error", err.Error())
+		return nil, err
 	} else {
```
SebastianMarian: You can drop the whole else now, and add in line 38 this line: msg.peer = p2p.PeerID(id)

p2p/libp2p/netMessenger.go
SebastianMarian: +1

bogdan-rosianu: done

p2p/libp2p/message.go
bogdan-rosianu: done

p2p/libp2p/netMessenger.go
```diff
@@ -461,10 +461,15 @@ func (netMes *networkMessenger) RegisterMessageProcessor(topic string, handler p
...
+	err := netMes.pb.RegisterTopicValidator(topic, func(ctx context.Context, pid peer.ID, pubsubMessage *pubsub.Message) bool {
+		message, err := NewMessage(pubsubMessage)
 		if err != nil {
-			log.Trace("p2p validator", "error", err.Error(), "topics", message.TopicIDs)
+			log.Trace("p2p validator - new message", "error", err.Error(), "topics", pubsubMessage.TopicIDs)
+			return false
+		}
+		err = handler.ProcessReceivedMessage(message, broadcastHandler)
+		if err != nil {
+			log.Trace("p2p validator", "error", err.Error(), "topics", pubsubMessage.TopicIDs)
```
SebastianMarian: I would say that is better to return directly false here and at the end of the method to simply return true, because if somebody will extend this method after this point, could forgot to add a return false here in the case of an error

bogdan-rosianu: done

PR #1050, opened 2020-02-04
En 5702 fix trie bugs

BeniaminDrasovean:
- When canceling prune, also remove the hash from pruning buffer.
- When updating the trie with an existent key-value pair, nothing should be modified.

### DIFF
--- a/data/trie/trieStorageManager.go
+++ b/data/trie/trieStorageManager.go
@@ -1,6 +1,7 @@
 package trie
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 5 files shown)


data/trie/trieStorageManager.go
```diff
@@ -168,6 +169,15 @@ func (tsm *trieStorageManager) CancelPrune(rootHash []byte) {
 
 	log.Trace("trie storage manager cancel prune", "root", rootHash)
 	_, _ = tsm.dbEvictionWaitingList.Evict(rootHash)
+	tsm.removeHashFromPruningBuffer(rootHash)
+}
+
+func (tsm *trieStorageManager) removeHashFromPruningBuffer(rootHash []byte) {
+	for i := range tsm.pruningBuffer {
+		if bytes.Equal(tsm.pruningBuffer[i], rootHash) {
+			tsm.pruningBuffer = append(tsm.pruningBuffer[:i], tsm.pruningBuffer[i+1:]...)
```
SebastianMarian: Could be found twice or more the value of rootHash in the tsm.pruningBuffer? If not, you can do a break after this line, just to avoid the next computations of bytes.Equal for the entire range

BeniaminDrasovean: Added break.

PR #1053, opened 2020-02-04
EN-5705: subround end round code coverage improvements

bogdan-rosianu:


consensus/spos/bls/subroundEndRound_test.go
```diff
@@ -610,6 +629,152 @@ func TestSubroundEndRound_CheckSignaturesValidityShouldRetunNil(t *testing.T) {
 	assert.Equal(t, nil, err)
 }
 
+func TestSubroundEndRound_DoEndRoundJobByParticipant_RoundCanceledShouldReturnFalse(t *testing.T) {
```
miiu96: test name is to long maybe you can remove from test name SubroundEndRound because already the name of the file is tell you that you are testing SubroundEndRound

consensus/spos/bls/subroundEndRound_test.go
```diff
@@ -610,6 +629,152 @@ func TestSubroundEndRound_CheckSignaturesValidityShouldRetunNil(t *testing.T) {
...
+	}}
+	container.SetRounder(&rounder)
+	sr := *initSubroundEndRoundWithContainer(container)
+
+	sr.RoundTimeStamp = time.Now().AddDate(0, 0, -1)
+
+	res := sr.IsOutOfTime()
+	assert.True(t, res)
+}
+
```
SebastianMarian: Can you check after merge with development, the code coverage again, as there are 2-3 more methods which have entered with the last consensus improvement, which also could need some unit tests. 

receivedBlockHeaderFinalInfo
haveConsensusHeaderWithFullInfo
createAndBroadcastHeaderFinalInfo

consensus/spos/bls/subroundEndRound_test.go
bogdan-rosianu: No. I will leave it like this. Imagine the test fails in TeamCity. You won't be able to know where does this test belong. I will keep with the standard

consensus/spos/bls/subroundEndRound_test.go
bogdan-rosianu: done

PR #1058, opened 2020-02-05
Epoch change processing fixes

sasurobert:


process/block/shardblock_test.go
```diff
@@ -4281,3 +4281,173 @@ func TestShardProcessor_updateStateStorage(t *testing.T) {
...
+
+	blockChain := &mock.BlockChainMock{GetCurrentBlockHeaderCalled: func() data.HeaderHandler {
+		return &block.Header{Epoch: epochStartTrigger.Epoch() - 1, Round: epochStartTrigger.EpochFinalityAttestingRound() + process.EpochChangeGracePeriod + 1}
+	}}
+
+	err := sp.CheckEpochCorrectnessCrossChain(blockChain)
+	assert.Equal(t, process.ErrMissingHeader, err)
+}
+
+func TestShardProcessor_checkEpochCorrectnessCrossChainInCorrectEpochRollback1(t *testing.T) {
```
SebastianMarian: Rollback1Block

process/block/shardblock.go
```diff
@@ -680,6 +721,12 @@ func (sp *shardProcessor) CommitBlock(
 		return err
 	}
 
+	err = sp.checkEpochCorrectnessCrossChain(chainHandler)
```
SebastianMarian: This code should be executed now without the initial condition if header.IsStartOfEpochBlock() {  ?

process/block/shardblock.go
```diff
@@ -680,6 +721,12 @@ func (sp *shardProcessor) CommitBlock(
...
+	sp.epochStartTrigger.SetProcessed(header)
```
SebastianMarian: It is ok to set processed this header even if there is an err return by checkEpochCorrectnessCrossChain ?

process/block/shardblock.go
```diff
@@ -258,6 +258,14 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	if header.IsStartOfEpochBlock() {
+		err = sp.checkEpochCorrectnessCrossChain(chainHandler)
+		sp.epochStartTrigger.SetProcessed(header)
```
SebastianMarian: It is ok to set processed this header even if there is an err return by checkEpochCorrectnessCrossChain ?

process/block/shardblock.go
```diff
@@ -258,6 +258,14 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	if header.IsStartOfEpochBlock() {
```
SebastianMarian: Do you need this part of code also in commit block again?

process/block/shardblock.go
```diff
@@ -347,28 +355,62 @@ func (sp *shardProcessor) checkEpochCorrectness(
...
+		}
+
+		metaBlockHash, err := core.CalculateHash(sp.marshalizer, sp.hasher, metaBlock)
+		if err != nil {
+			return fmt.Errorf("%w could not calculate hash for epoch start metablock for epoch %d",
+				err, header.GetEpoch())
+		}
+
+		if !bytes.Equal(header.EpochStartMetaHash, metaBlockHash) {
+			log.Warn("epoch start meta hash missmatch", "proposed", header.EpochStartMetaHash, "calculated", sp.epochStartTrigger.EpochStartMetaHdrHash())
```
SebastianMarian: metaBlockHash instead sp.epochStartTrigger.EpochStartMetaHdrHash()

process/block/shardblock_test.go
sasurobert: done.

process/block/shardblock.go
sasurobert: initial condition is not more needed. It is better to check every time.

process/block/shardblock.go
sasurobert: put set only if not error.

process/block/shardblock.go
sasurobert: changed. put the verification only if isstartofepochblock and setprocessed only if epochcorrectess does not return error.

process/block/shardblock.go
sasurobert: yes. As the leader cannot make revert in create or applybodytoheader.

process/block/shardblock.go
sasurobert: done.

process/block/shardblock.go
SebastianMarian: ok

process/block/shardblock.go
```diff
@@ -262,6 +262,13 @@ func (sp *shardProcessor) ProcessBlock(
 		return err
 	}
 
+	if header.IsStartOfEpochBlock() {
+		err = sp.checkEpochCorrectnessCrossChain(chainHandler)
+		if err != nil {
+			return err
+		}
+	}
```
SebastianMarian: Do you forgot to add this line after checking the err? sp.epochStartTrigger.SetProcessed(header)

PR #1065, opened 2020-02-06
added the possibility to log to file directly from the node

iulianpascalau:
- added the possibility to log to file directly from the node
- adapted scripts not to use the tee log file method


cmd/node/main.go
```diff
@@ -383,8 +391,30 @@ func getSuite(config *config.Config) (crypto.Suite, error) {
...
+		defer func() {
+			_ = fileForLogs.Close()
+		}()
+
+		err = logger.AddLogObserver(fileForLogs, &logger.PlainFormatter{})
+		if err != nil {
+			return fmt.Errorf("%w adding file log observer", err)
+		}
+	}
+
```
camilbancioiu: Might want to put this in a function called `redirectNodeOutputToLogfile()`. The function `startNode()` is already quite long.

cmd/node/main.go
```diff
@@ -874,6 +892,44 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	return nil
 }
 
+func getWorkingDir(ctx *cli.Context, log logger.Logger) string {
+	var workingDir = ""
```
golangcibot: ineffectual assignment to `workingDir` (from `ineffassign`)

cmd/node/main.go
```diff
@@ -383,8 +391,30 @@ func getSuite(config *config.Config) (crypto.Suite, error) {
...
+		fileForLogs, err = createLogFile(workingDir)
+		if err != nil {
+			return fmt.Errorf("%w creating a log file", err)
+		}
+
+		defer func() {
+			_ = fileForLogs.Close()
+		}()
+
+		err = logger.AddLogObserver(fileForLogs, &logger.PlainFormatter{})
```
camilbancioiu: Might want to attach the `logger` to `fileForLogs` in the same place where `stderr` is redirected to `fileForLogs`. They're currently in separate functions (the other being `createLogFile()`).

cmd/node/main.go
```diff
@@ -874,6 +892,44 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
...
+func createLogFile(workingDir string) (*os.File, error) {
+	logDirectory := filepath.Join(workingDir, defaultLogsPath)
+	f, err := core.CreateFile("elrond-go", logDirectory, "log")
+	if err != nil {
+		return nil, err
+	}
+
+	//we need this function as to close file.Close() when the code panics and the defer func associated
+	//with the file pointer in the main func will never be reached
+	runtime.SetFinalizer(f, func(f *os.File) {
```
camilbancioiu: Using `f` for both variables (outside the closure + inside the closure) might work, but it's not very clean. Might want to use different names.

iulianpascalau: right, will be fixed

cmd/node/main.go

cmd/node/main.go
iulianpascalau: this is not possible as I need to create the defer function inside startNode (as to be able to be executed when startNode ends) not when the redirectNodeOutputToLogfile ends.

cmd/node/main.go

cmd/node/main.go
```diff
@@ -239,6 +241,11 @@ VERSION:
 		Usage: "This flag specifies the logger level",
 		Value: "*:" + logger.LogInfo.String(),
 	}
+	//logFile is used when the log output needs to be logged in a file
+	logFile = cli.BoolFlag{
+		Name:  "logFile",
```
bogdan-rosianu: please rename this.  this does not sound like a bool flag. maybe `save log` and don't use camel case in flags. The rest of the flags in the node should be refactored also. so maybe `save-log` for flag name and `saveLog` for the variable

iulianpascalau: done, both log-related flags. The rest should be done in a separate PR as there might be implications regarding the scripts

PR #1073, opened 2020-02-07
En 5722 tsc fix blake empty value hash size

traian-schiau:


integrationTests/consensus/testInitializer.go
```diff
@@ -5,14 +5,15 @@ import (
...
 	"fmt"
-	"github.com/ElrondNetwork/elrond-go/storage/lrucache"
 	"io/ioutil"
 	"math/big"
 	"math/rand"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ElrondNetwork/elrond-go/storage/lrucache"
```
sasurobert: goimports.

hashing/blake2b/blake2b.go
```diff
@@ -6,38 +6,49 @@ import (
 	"golang.org/x/crypto/blake2b"
 )
 
-var b2bEmptyHash []byte
-
 // Blake2b is a blake2b implementation of the hasher interface.
 type Blake2b struct {
-	HashSize int
+	HashSize  int
+	emptyHash []byte
```
sasurobert: I think we could correct this implementation, like making a constructor for Blake2B with a given hashSize, not accepting hashSizes of 0 - as it does not have any value. GetHasher function implementation can be put in constructor, as well as the call for setEmpty.

traian-schiau: Indeed, using a constructor could be a better approach, however since the full hashing package needs some refactoring (use []byte as input data), fix some logic in all the other hashers, I think is better to do it then

integrationTests/consensus/testInitializer.go
traian-schiau: done

hashing/blake2b/blake2b.go
```diff
@@ -6,38 +6,49 @@ import (
...
 }
 
-// Compute takes a string, and returns the blake2b hash of that string
-func (b2b Blake2b) Compute(s string) []byte {
-	if len(s) == 0 && len(b2bEmptyHash) != 0 {
-		return b2b.EmptyHash()
-	}
-	var h hash.Hash
+func (b2b *Blake2b) getHeser() hash.Hash {
```
AdoAdoAdo: can you fix typo in function name heser -> hasher

PR #1075, opened 2020-02-07
Fix commit with longer time

SebastianMarian:


process/block/baseProcess.go
```diff
@@ -905,3 +905,55 @@ func (bp *baseProcessor) DecodeBlockBodyAndHeader(dta []byte) (data.BodyHandler,
...
 }
+
+func (bp *baseProcessor) saveBody(body block.Body) {
+	errNotCritical := bp.txCoordinator.SaveBlockDataToStorage(body)
+	if errNotCritical != nil {
+		log.Warn("saveBody.SaveBlockDataToStorage", "error", errNotCritical.Error())
+	}
+
+	for i := 0; i < len(body); i++ {
+		marshalizedBody, errNotCritical := bp.marshalizer.Marshal(body[i])
```
iulianpascalau: *marshalizedMiniblock

process/block/shardblock_test.go
```diff
@@ -1683,127 +1690,11 @@ func TestShardProcessor_CommitBlockStorageFailsForBodyShouldWork(t *testing.T) {
 	})
 
 	err = sp.CommitBlock(blkc, hdr, body)
-
+	wg.Wait()
 	assert.Nil(t, err)
 	assert.True(t, wasCalled)
 }
 
-func TestShardProcessor_CommitBlockNoTxInPoolShouldErr(t *testing.T) {
```
AdoAdoAdo: why was this test removed?

SebastianMarian: Because previously there was the below listed code in method CommitBlock from shard/meta processor and its return was treated wrongly as an critical error. At some point we have decided that all the errors received after a put action should be not critical and only logged with WARN.

Previous code from CommitBlock method:

err = mp.txCoordinator.SaveBlockDataToStorage(body)
if err != nil {
  return err
}

Now, with this refactor, I also corrected this part, which seems that remained as the only one treated differently. 

func (bp *baseProcessor) saveBody(body block.Body) {
	errNotCritical := bp.txCoordinator.SaveBlockDataToStorage(body)
	if errNotCritical != nil {
		log.Warn("saveBody.SaveBlockDataToStorage", "error", errNotCritical.Error())
	}
...
...
...
}

process/block/shardblock_test.go
```diff
@@ -3212,10 +3212,13 @@ func TestShardProcessor_RemoveAndSaveLastNotarizedMetaHdrNoDstMB(t *testing.T) {
 
 	wg := sync.WaitGroup{}
 	wg.Add(4)
+	mutPut := sync.RWMutex{}
```
AdoAdoAdo: can you use atomic instead

process/block/shardblock_test.go
```diff
@@ -1560,12 +1560,15 @@ func TestShardProcessor_CommitBlockStorageFailsForHeaderShouldErr(t *testing.T)
...
 	}
 	body := make(block.Body, 0)
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	hdrUnit := &mock.StorerStub{
 		GetCalled: func(key []byte) (i []byte, e error) {
 			return marshalizer.Marshal(&block.Header{})
 		},
 		PutCalled: func(key, data []byte) error {
 			wasCalled = true
```
AdoAdoAdo: please use atomic as well for was called (put called)

process/block/shardblock_test.go
```diff
@@ -1538,7 +1538,7 @@ func TestShardProcessor_CommitBlockStorageFailsForHeaderShouldErr(t *testing.T)
 
 	tdp := initDataPool([]byte("tx_hash1"))
 	errPersister := errors.New("failure")
-	wasCalled := false
+	putCalledNr := uint32(0)
```
iulianpascalau: for future naming: Nr->Num

storage/pruning/pruningStorer.go
```diff
@@ -88,6 +88,9 @@ func initPruningStorer(
 	if check.IfNil(args.PathManager) {
 		return nil, storage.ErrNilPathManager
 	}
+	if args.MaxBatchSize > int(args.CacheConf.Size) {
```
iulianpascalau: why not moving this on the persister constructors (SerialDB/DB)?

SebastianMarian: Done

process/block/shardblock_test.go
SebastianMarian: Done

process/block/shardblock_test.go
SebastianMarian: Ok

PR #1082, opened 2020-02-10
Config to enable/disable staking and auction by nonce.

sasurobert:


process/economics/economicsData.go
```diff
@@ -291,14 +304,14 @@ func (ed *EconomicsData) NumNodes() uint32 {
 	return ed.numNodes
 }
 
-// AuctionEnabled returns whether full auction process is enabled
-func (ed *EconomicsData) AuctionEnabled() bool {
-	return ed.auctionEnabled
+// AuctionEnableNonce returns whether full auction process is enabled
```
bogdan-rosianu: maybe `// AuctionEnableNonce returns the nonce that will trigger auction beginning`

process/economics/economicsData.go
```diff
@@ -291,14 +304,14 @@ func (ed *EconomicsData) NumNodes() uint32 {
...
+func (ed *EconomicsData) AuctionEnableNonce() uint64 {
+	return ed.auctionEnableNonce
 }
 
-// SetAuctionEnabled sets whether the auction is enabled
-func (ed *EconomicsData) SetAuctionEnabled(auctionEnabled bool) {
-	ed.auctionEnabled = auctionEnabled
+// StakeEnableNonce returns whether full auction process is enabled
```
bogdan-rosianu: `// StakeEnableNonce returns the nonce that will trigger staking beginning`

vm/systemSmartContracts/auction.go
```diff
@@ -308,7 +310,25 @@ func (s *stakingAuctionSC) getVerifiedBLSKeysFromArgs(txPubKey []byte, args [][]
 	return blsKeys
 }
 
+func (s *stakingAuctionSC) isFunctionEnabled(nonce uint64) bool {
+	currentNonce := s.eei.BlockChainHook().CurrentNonce()
+	if currentNonce == 0 {
+		return true
+	}
+
+	if currentNonce < nonce {
```
bogdan-rosianu: can simply use `return currentNonce >= nonce`

vm/systemSmartContracts/auction.go
```diff
@@ -117,11 +117,106 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
+	if len(args.Arguments[0]) != len(args.CallerAddr) {
+		return vmcommon.UserError
+	}
+
+	registrationData, err := s.getOrCreateRegistrationData(args.CallerAddr)
+	if err != nil {
+		return vmcommon.UserError
+	}
+
+	if len(registrationData.RewardAddress) == 0 {
```
bogdan-rosianu: please use only one way for checking the length. the first check in the method checks that the length is less than one and here you check that the length is equal to 0.

vm/systemSmartContracts/auction.go
```diff
@@ -117,11 +117,106 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
+	err = s.saveRegistrationData(args.CallerAddr, registrationData)
+	if err != nil {
+		return vmcommon.UserError
+	}
+
+	return vmcommon.Ok
+}
+
+func (s *stakingAuctionSC) changeValidatorKeys(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	if len(args.Arguments) < 4 {
```
bogdan-rosianu: why 4 ? at least define a constant

vm/systemSmartContracts/auction.go
```diff
@@ -117,11 +117,106 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
+
+	return vmcommon.Ok
+}
+
+func (s *stakingAuctionSC) replaceBLSKey(registrationData *AuctionData, oldBlsKey []byte, newBlsKey []byte) error {
+	foundOldKey := false
+	for i, registeredKey := range registrationData.BlsPubKeys {
+		if bytes.Equal(registeredKey, oldBlsKey) {
+			foundOldKey = true
+			registrationData.BlsPubKeys[i] = newBlsKey
```
bogdan-rosianu: you can refactor this by returning `nil` when you found the key. And return that error if the code reaches out of the `for` loop

vm/systemSmartContracts/staking.go
```diff
@@ -84,11 +84,23 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+		return r.jail(args)
+	case "unJail":
+		return r.unJail(args)
 	}
 
 	return vmcommon.UserError
 }
 
+func (r *stakingSC) unJail(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	return vmcommon.Ok
```
iulianpascalau: Same as above, valid also for the next function

vm/interface.go
```diff
@@ -68,7 +68,8 @@ type ValidatorSettingsHandler interface {
 	MinStepValue() *big.Int
 	TotalSupply() *big.Int
 	NumNodes() uint32
-	AuctionEnabled() bool
+	AuctionEnableNonce() uint64
+	StakeEnableNonce() uint64
 	IsInterfaceNil() bool
 }
```
iulianpascalau: auction.go and staking.go have only ~63% code coverage. I would like a percent close to 100 for these files. Should add a new task to increase the code coverage

sasurobert: agreed. will not merge to dev till close to 100% is not reached.

vm/systemSmartContracts/staking.go
sasurobert: implemented

vm/systemSmartContracts/auction.go
```diff
@@ -115,11 +117,106 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
+		oldBlsKey := args.Arguments[i]
+		newBlsKey := args.Arguments[i+1]
+		signedWithNewKey := args.Arguments[i+2]
+
+		err := s.sigVerifier.Verify(args.CallerAddr, signedWithNewKey, newBlsKey)
+		if err != nil {
+			return vmcommon.UserError
+		}
+
+		err = s.replaceBLSKey(registrationData, oldBlsKey, newBlsKey)
```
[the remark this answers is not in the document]
sasurobert: yes. otherwise it would break a lot of things.

vm/systemSmartContracts/auction.go
```diff
@@ -115,11 +117,106 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
+		return s.changeValidatorKeys(args)
+	case "unJail":
+		return s.unJail(args)
 	}
 
 	return vmcommon.UserError
 }
 
+func (s *stakingAuctionSC) unJail(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	return vmcommon.Ok
```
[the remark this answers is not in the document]
sasurobert: implemented

vm/systemSmartContracts/auction.go
sasurobert: no, as it is calling another function with the results.

vm/systemSmartContracts/auction.go
sasurobert: done.

vm/systemSmartContracts/auction.go
sasurobert: checked.

vm/systemSmartContracts/auction.go
sasurobert: implemented completely the function.

process/economics/economicsData.go
sasurobert: done.

process/economics/economicsData.go
sasurobert: done.

vm/systemSmartContracts/auction.go
```diff
@@ -199,6 +296,7 @@ func (s *stakingAuctionSC) getConfig(epoch uint32) AuctionConfig {
 		baseConfigData, err := json.Marshal(s.baseConfig)
 		if err != nil {
 			log.Debug("marshal error on getConfig function")
```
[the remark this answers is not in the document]
sasurobert: log.Warn

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+	case "changeRewardAddress":
+		return r.changeRewardAddress(args)
+	case "changeValidatorKeys":
+		return r.changeValidatorKey(args)
 	}
 
 	return vmcommon.UserError
 }
 
+func getPercentageOfValue(value *big.Int, percentage float64) *big.Int {
```
iulianpascalau: Do we really need this in this exact form? I would avoid percentage as float64. Can we have percentage as int and all operations done on ints? The main reason is that this code is treated as a "VM" code.

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+	if startRound > endRound {
+		return stake
+	}
+	if endRound-startRound < r.numRoundsWithoutBleed {
+		return stake
+	}
+
+	totalRoundsToBleed := endRound - startRound - r.numRoundsWithoutBleed
+	totalPercentageToBleed := float64(totalRoundsToBleed) * r.bleedPercentagePerRound
+	totalPercentageToBleed = math.Max(totalPercentageToBleed, r.maximumPercentageToBleed)
```
iulianpascalau: Shouldn't this be Min?

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+
+	bleedValue := big.NewInt(0).Neg(getPercentageOfValue(stake, totalPercentageToBleed))
```
iulianpascalau: This is a bit awkward. What should I understand from a negative bleedValue? Is a reward? :D
I would have the bleed value positive here and subtract it from stake on the next line

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+	for _, argument := range args.Arguments {
+		stakedData, err := r.getOrCreateRegisteredData(argument)
+		if err != nil {
+			return vmcommon.UserError
+		}
+		if len(stakedData.RewardAddress) == 0 {
+			return vmcommon.UserError
+		}
+
+		stakedData.StakeValue = r.calculateStakeAfterBleed(
```
iulianpascalau: Question: unjailing will take effect in the next epoch?

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+		)
+		return err
+	}
+
+	r.eei.SetStorage(key, data)
+	return nil
+}
+
 func (r *stakingSC) jail(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	if !bytes.Equal(args.CallerAddr, r.jailAccessAddr) {
```
iulianpascalau: I do not think this condition is required: all meta nodes will automatically call this function when a peer will need to be moved in jail. It is useless to try to call this function with the jailing address for the sake of this condition.

vm/systemSmartContracts/staking.go
```diff
@@ -88,16 +113,206 @@ func (r *stakingSC) Execute(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+	for _, argument := range args.Arguments {
+		stakedData, err := r.getOrCreateRegisteredData(argument)
+		if err != nil {
+			return vmcommon.UserError
+		}
+		if len(stakedData.RewardAddress) == 0 {
+			return vmcommon.UserError
+		}
+
+		stakedData.JailedRound = r.eei.BlockChainHook().CurrentRound()
```
iulianpascalau: I presume this will come in effect (the peer will no longer be a validator/proposer) also at eoe

vm/systemSmartContracts/staking.go
iulianpascalau: What happens if the remaining staked value is below the threshold minimum stake value?

vm/systemSmartContracts/auction.go
```diff
@@ -144,22 +158,35 @@ func (s *stakingAuctionSC) changeRewardAddress(args *vmcommon.ContractCallInput)
 	if err != nil {
 		return vmcommon.UserError
 	}
-
 	if len(registrationData.RewardAddress) == 0 {
 		return vmcommon.UserError
 	}
+	if bytes.Equal(registrationData.RewardAddress, args.Arguments[0]) {
```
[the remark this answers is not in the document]
sasurobert: If error is returned, all the fee is taken.

vm/systemSmartContracts/staking.go
sasurobert: Any change of validators take effect at end of epoch.

vm/systemSmartContracts/staking.go
sasurobert: Somebody might create a transaction to Jail everybody. This is a protection for it.

vm/systemSmartContracts/staking.go
sasurobert: EOE.

vm/systemSmartContracts/staking.go
sasurobert: will not be selected in auction.

vm/systemSmartContracts/auction.go
```diff
@@ -129,6 +132,17 @@ func (s *stakingAuctionSC) Execute(args *vmcommon.ContractCallInput) vmcommon.Re
...
 func (s *stakingAuctionSC) unJail(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	config := s.getConfig(s.eei.BlockChainHook().CurrentEpoch())
+	totalUnJailPrice := big.NewInt(0).Mul(config.UnJailPrice, big.NewInt(int64(len(args.Arguments))))
+
+	if totalUnJailPrice.Cmp(args.CallValue) < 0 {
+		return vmcommon.UserError
+	}
+
+	for _, argument := range args.Arguments {
+		_, _ = s.executeOnStakingSC([]byte("unJail@" + hex.EncodeToString(argument)))
```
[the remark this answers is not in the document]
sasurobert: because you try to unjail all those who you can.

process/economics/economicsData.go
```diff
@@ -11,25 +11,29 @@ import (
...
-	burnAddress          string
-	stakeValue           *big.Int
-	unBondPeriod         uint64
-	ratingsData          *RatingsData
-	minStep              *big.Int
-	totalSupply          *big.Int
-	numNodes             uint32
-	auctionEnableNonce   uint64
-	stakeEnableNonce     uint64
+	rewardsValue             *big.Int
```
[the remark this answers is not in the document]
sasurobert: it holds only some values.

vm/systemSmartContracts/staking.go
sasurobert: min to be.

vm/systemSmartContracts/staking.go
sasurobert: forgot about Sub.

vm/systemSmartContracts/staking.go
```diff
@@ -31,30 +36,50 @@ type StakedData struct {
 	UnStakedEpoch uint32   `json:"UnStakedEpoch"`
 	RewardAddress []byte   `json:"RewardAddress"`
 	StakeValue    *big.Int `json:"StakeValue"`
+	JailedRound   uint64   `json:"JailedRound"`
+}
+
+// ArgsNewStakingSmartContract
```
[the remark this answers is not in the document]
sasurobert: done.

vm/systemSmartContracts/staking.go
sasurobert: we have some percentages here and there in the code as float64. We have to make a denomination if we go with INT. Open for change, but has to be done on the whole protocol.

vm/systemSmartContracts/auction.go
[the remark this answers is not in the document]
iulianpascalau: Yes, but I still think it's not an error. :D

vm/systemSmartContracts/staking.go
iulianpascalau: good point!

PR #1089, opened 2020-02-12
Optimised validator statistics - bulk trie write for missed blocks

ccorcoveanu:


data/state/peerAccount.go
```diff
@@ -321,6 +321,20 @@ func (pa *PeerAccount) IncreaseValidatorSuccessRateWithJournal(value uint32) err
 	return pa.accountTracker.SaveAccount(pa)
 }
 
+// UpdateValidatorSuccessRateWithJournal increases the account's number of successful signing with the provided amount,
+// saving the old state before changing
+func (pa *PeerAccount) UpdateValidatorSuccessRateWithJournal(amount uint32) error {
```
raduchis: the new methods are exactly the same as the changed methods with value. Is it really needed?

process/peer/export_test.go
```diff
@@ -24,8 +24,6 @@ func (vs *validatorStatistics) GetMatchingPrevShardData(currentShardData block.S
 	return vs.getMatchingPrevShardData(currentShardData, shardInfo)
 }
 
-func (vs *validatorStatistics) PrevShardInfo() map[string]block.ShardData {
-	vs.mutPrevShardInfo.RLock()
-	defer vs.mutPrevShardInfo.RUnlock()
-	return vs.prevShardInfo
+func (vs *validatorStatistics) GetLeaderDecreaseCount(key []byte) uint32 {
+	return vs.missedBlocksCounters.get(key).leaderDecreaseCount
```
raduchis: is not mutex needed?

data/state/peerAccount.go
ccorcoveanu: removed

process/peer/export_test.go
ccorcoveanu: Will add a mutex here and a concurrent unit test 👍

sasurobert: this in export_test.go so only if you make the test concurrent it is needed. This is not in the production code.

process/peer/validatorCounters.go
```diff
@@ -0,0 +1,38 @@
...
+func (vrc validatorRoundCounters) increaseLeader(key []byte) {
+	vrc.get(key).leaderIncreaseCount++
+}
+
+func (vrc validatorRoundCounters) decreaseLeader(key []byte) {
+	vrc.get(key).leaderDecreaseCount++
+}
+
+func (vrc validatorRoundCounters) get(key []byte) *validatorCounters {
+	if vrc[string(key)] == nil {
```
sasurobert: this can be made a little bit faster, with vdCounter, ok := vrc[string(key)] if !ok {vrc[string(key] =&validatorCounters{} return vrc[string(key)]} return vdCounter

ccorcoveanu: done

PR #1090, opened 2020-02-12
Fix recreate trie from snapshot bug

BeniaminDrasovean:


cmd/node/config/config.toml
```diff
@@ -183,10 +183,11 @@
         MaxBatchSize = 100
         MaxOpenFiles = 10
 
+# TODO: add cacher to trie snapshot
```
sasurobert: I think adding a cacher can be done in this PR.

BeniaminDrasovean: After a talk with Sebi, realised the cacher is no longer needed.

PR #1097, opened 2020-02-14
En 5736 index peer account

bogdan-rosianu:


data/state/peerAccount.go
```diff
@@ -368,6 +371,12 @@ func (pa *PeerAccount) GetRating() uint32 {
 	return pa.Rating
 }
 
+// SetListAndIndex will update the peer's list (eligible, waiting) and the index inside it
+func (pa *PeerAccount) SetListAndIndex(list string, index int) {
+	pa.List = list
```
AdoAdoAdo: I think you should save with journal the changes

process/rating/nilListIndexUpdater.go
```diff
@@ -0,0 +1,15 @@
+package rating
+
+// NilListIndexUpdater is a nil implementation for an list index updater
+type NilListIndexUpdater struct {
```
AdoAdoAdo: what are we using the nilListindexUpdater for right now?
If this is not meant for production code but for testing, I would prefer if this was a mock instead and in mock folder.

sharding/interface.go
```diff
@@ -24,9 +24,11 @@ type Validator interface {
 // NodesCoordinator defines the behaviour of a struct able to do validator group selection
 type NodesCoordinator interface {
 	PublicKeysSelector
-	SetNodesPerShards(eligible map[uint32][]Validator, waiting map[uint32][]Validator, epoch uint32) error
+	SetNodesPerShards(eligible map[uint32][]Validator, waiting map[uint32][]Validator, epoch uint32, updatePeers bool) error
```
AdoAdoAdo: do we really need two interfaces ? can't we use the other interface for this purpose in a loop?

node/heartbeat/monitor.go
```diff
@@ -64,6 +73,8 @@ func NewMonitor(
 	mon := &Monitor{
 		marshalizer:                 marshalizer,
 		heartbeatMessages:           make(map[string]*heartbeatMessageInfo),
+		eligibleListProvider:        eligibleListProvider,
```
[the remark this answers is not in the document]
bogdan-rosianu: I can tell if a peer is in waiting list if it is not in eligible list

node/heartbeat/monitor.go
```diff
@@ -281,6 +293,22 @@ func (m *Monitor) computeShardID(pubkey string) uint32 {
 	return m.heartbeatMessages[pubkey].computedShardID
 }
 
+func (m *Monitor) isIsEligibleList(pubkey []byte, shardID uint32) bool {
```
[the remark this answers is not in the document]
bogdan-rosianu: done

node/heartbeat/sender.go
```diff
@@ -104,3 +128,27 @@ func (s *Sender) SendHeartbeat() error {
 
 	return nil
 }
+
+func (s *Sender) updateMetrics(hb *Heartbeat) {
+	result := "in waiting list"
```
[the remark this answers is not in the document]
bogdan-rosianu: in heartbeat / node status there is also the information if a node is validator or observer. so this should be enough

node/heartbeat/sender.go
```diff
@@ -104,3 +128,27 @@ func (s *Sender) SendHeartbeat() error {
...
+	if s.isIsEligibleList(hb.Pubkey, hb.ShardID) {
+		result = "in eligible list"
+	}
+	s.statusHandler.SetStringValue(core.MetricValidatorType, result)
+}
+
+func (s *Sender) isIsEligibleList(pubkey []byte, shardID uint32) bool {
```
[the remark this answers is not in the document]
bogdan-rosianu: renamed

data/state/peerAccount.go
bogdan-rosianu: done

process/rating/nilListIndexUpdater.go
bogdan-rosianu: it is not meant for testing. It's the usage of `null object design pattern`. It follows the same direction for rating reader because these 2 components are initialized "lazy" inside validator statistics constructor. Will rename both of them with `disabled` instead of `nil`

data/state/interface.go
```diff
@@ -73,6 +73,8 @@ type PeerAccountHandler interface {
 	DecreaseLeaderSuccessRateWithJournal(value uint32) error
 	IncreaseValidatorSuccessRateWithJournal(value uint32) error
 	DecreaseValidatorSuccessRateWithJournal(value uint32) error
+	SetListAndIndex(list string, index int)
```
raduchis: Do we need the method with no Journal?

integrationTests/mock/peerProcessorMock.go
```diff
@@ -65,10 +64,12 @@ func (vsp *ValidatorStatisticsProcessorMock) GetPeerAccount(address []byte) (sta
 	return nil, nil
 }
 
+// updatePeersListAndIndex -
+func (vsp *ValidatorStatisticsProcessorMock) UpdatePeerAccounts() error {
```
raduchis: name and comment do not match

node/heartbeat/heartbeat.go
```diff
@@ -6,27 +6,29 @@ import (
...
+	HexPublicKey     string    `json:"hexPublicKey"`
+	VersionNumber    string    `json:"versionNumber"`
+	NodeDisplayName  string    `json:"nodeDisplayName"`
+	TotalUpTime      int       `json:"totalUpTimeSec"`
+	TotalDownTime    int       `json:"totalDownTimeSec"`
+	MaxInactiveTime  Duration  `json:"maxInactiveTime"`
+	ReceivedShardID  uint32    `json:"receivedShardID"`
+	ComputedShardID  uint32    `json:"computedShardID"`
+	IsValidator      bool      `json:"isValidator"`
+	IsInEligibleList bool      `json:"isInEligibleList"`
```
raduchis: Could we send the actual list name better? How do you know who is in which list?

node/heartbeat/monitor.go
```diff
@@ -64,6 +73,8 @@ func NewMonitor(
 	mon := &Monitor{
 		marshalizer:                 marshalizer,
 		heartbeatMessages:           make(map[string]*heartbeatMessageInfo),
+		eligibleListProvider:        eligibleListProvider,
+		epochHandler:                epochHandler,
```
raduchis: epochHandler is not testes for nil. Is it ok?

node/heartbeat/monitor.go
```diff
@@ -281,6 +293,22 @@ func (m *Monitor) computeShardID(pubkey string) uint32 {
 	return m.heartbeatMessages[pubkey].computedShardID
 }
 
+func (m *Monitor) isInEligibleList(pubkey []byte, shardID uint32) bool {
+	nodesMap, err := m.eligibleListProvider.GetNodesPerShard(m.epochHandler.Epoch())
+	if err != nil {
+		return false
```
raduchis: I would also add a log.warn(err) here, if something went wrong with the eligibleListProvier

node/heartbeat/monitor.go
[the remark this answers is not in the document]
raduchis: What about if it is the "leaving" nodes list? Could you differentiate between eligible, waiting and leaving? And joining?

node/heartbeat/sender.go
```diff
@@ -104,3 +128,27 @@ func (s *Sender) SendHeartbeat() error {
...
+
+func (s *Sender) updateMetrics(hb *Heartbeat) {
+	result := "in waiting list"
+	if s.isInEligibleList(hb.Pubkey, hb.ShardID) {
+		result = "in eligible list"
+	}
+	s.statusHandler.SetStringValue(core.MetricValidatorType, result)
+}
+
+func (s *Sender) isInEligibleList(pubkey []byte, shardID uint32) bool {
```
raduchis: this method is dupplicate with the one in monitor. Could the eligibleListProvider compute this directly?

node/mock/peerAccountHandlerMock.go
```diff
@@ -25,12 +25,31 @@ type PeerAccountHandlerMock struct {
...
+	SetListAndIndexWithJournalCalled func(list string, index int) error
+
 	IncreaseLeaderSuccessRateWithJournalCalled    func(value uint32) error
 	DecreaseLeaderSuccessRateWithJournalCalled    func(value uint32) error
 	IncreaseValidatorSuccessRateWithJournalCalled func(value uint32) error
 	DecreaseValidatorSuccessRateWithJournalCalled func(value uint32) error
 }
 
+// SetListAndIndex -
+func (pahm *PeerAccountHandlerMock) SetListAndIndex(list string, index int) {
```
raduchis: maybe not needed any more

process/mock/peerAccountHandlerMock.go
```diff
@@ -171,10 +174,23 @@ func (pahm *PeerAccountHandlerMock) SetTempRatingWithJournal(rating uint32) erro
 	return nil
 }
 
+// SetListAndIndex -
+func (pahm *PeerAccountHandlerMock) SetListAndIndex(list string, index int) {
```
raduchis: maybe not needed any more

process/peer/process.go
```diff
@@ -122,9 +122,26 @@ func NewValidatorStatisticsProcessor(arguments ArgValidatorStatisticsProcessor)
 
 	ratingReaderSetter.SetRatingReader(rr)
 
+	listIndexUpdaterSetter, ok := rater.(sharding.ListIndexUpdaterSetter)
+	if !ok {
+		return nil, process.ErrNilRatingReader
```
raduchis: It was my bad when creating the ratingReader and returning ErrNilRatingReader. Could you please change line 115 and 127 with 2 other errors? ErrRaterNotRatingReaderSetter ErrRaterNotListIndexUpdaterSetter

sharding/indexHashedNodesCoordinator.go
```diff
@@ -179,6 +185,13 @@ func (ihgs *indexHashedNodesCoordinator) SetNodesPerShards(
 	nodesConfig.shardId = ihgs.computeShardForPublicKey(nodesConfig)
 	ihgs.nodesConfig[epoch] = nodesConfig
 
+	if updatePeersListAndIndex {
```
raduchis: is there any time where we are updating the PeerList but we do not want to UpdatePeersListAndIndex?

data/state/interface.go
bogdan-rosianu: removed

integrationTests/mock/peerProcessorMock.go
bogdan-rosianu: fixed

node/heartbeat/heartbeat.go
bogdan-rosianu: you're right. done

node/heartbeat/monitor.go
bogdan-rosianu: I will add a check for it but we should rethink these things. In this situation let's say it is ok because epochhandler comes from a `with` option. And for example if you forget that with in an integration test, it should fail. But there are cases where we forward a component many layers down and check it for nil every time.

node/heartbeat/monitor.go
bogdan-rosianu: done

node/heartbeat/sender.go
bogdan-rosianu: I would not create another structure which is only a wrapper over the nodes coordinator so I can move this method there. Another way of solving the duplicated code issue is to move this function to a common file so it can be used by both of  them, but in this case the function will have 2 additional parameters (eligible list provider and epoch handler). Which way do you prefer ?

node/mock/peerAccountHandlerMock.go
bogdan-rosianu: removed

process/mock/peerAccountHandlerMock.go
bogdan-rosianu: removed

process/peer/process.go
bogdan-rosianu: done

sharding/indexHashedNodesCoordinator.go
bogdan-rosianu: yes. As discussed, when we first create the nodes coordinator we don't have the list index updater with ok values, but with the disabled one.

node/heartbeat/sender.go
bogdan-rosianu: regarding to "What about if it is the "leaving" nodes list? Could you differentiate between eligible, waiting and leaving? And joining?":
maybe when we will use those lists also, the nodes coordinator should be able to tell us. And we will just add another method in the interface.

data/state/interface.go
raduchis: ok

integrationTests/mock/peerProcessorMock.go
raduchis: ok

raduchis: ok

node/heartbeat/monitor.go
raduchis: fixed on line 71 now

node/heartbeat/heartbeatMessageInfo_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const waitingList = "waiting list"
```
raduchis: maybe better eligible list, because that will always be populated.

sharding/indexHashedNodesCoordinator.go
```diff
@@ -591,6 +628,44 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
 	return shardEligible, nil
 }
 
+// UpdatePeersListAndIndex will update the list and the index for all peers
+func (ihgs *indexHashedNodesCoordinator) UpdatePeersListAndIndex() error {
+	nodesConfig, ok := ihgs.nodesConfig[ihgs.currentEpoch]
```
AdoAdoAdo: you are missing lock/unlock for relevant mutexes

sharding/peerTypeProvider.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return &PeerTypeProvider{
+		nodesCoordinator: nodesCoordinator,
+		epochHandler:     epochHandler,
+	}, nil
+}
+
+// ComputeForPubKey returns the peer type for a given public key and shard id
+func (ptp *PeerTypeProvider) ComputeForPubKey(pubKey []byte, shardID uint32) (core.ValidatorList, error) {
```
AdoAdoAdo: this seems too compute intensive to execute for each node every 30 seconds, so I suggest you add a cache for the type with the public key as key, and populate this cache on a new epoch trigger.

all requests that are not found in the cache return "observer" and there's no need to add observers to the cache as it may allow for out of memory vectors.

sharding/indexHashedNodesCoordinator.go
raduchis: :+1:

sharding/indexHashedNodesCoordinator.go
raduchis: I'm still not happy with the bool value, but it can be refactored in the future.

node/heartbeat/heartbeatMessageInfo.go
```diff
@@ -192,6 +195,5 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
 func (hbmi *heartbeatMessageInfo) GetIsValidator() bool {
 	hbmi.updateMutex.Lock()
 	defer hbmi.updateMutex.Unlock()
-	isValidator := hbmi.isValidator
-	return isValidator
+	return hbmi.peerType == string(core.EligibleList)
```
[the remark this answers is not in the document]
bogdan-rosianu: No. This function is used when computing the number of validators which will be passed to status handler

sharding/peerTypeProvider.go
bogdan-rosianu: added a cache

sharding/indexHashedNodesCoordinator.go
bogdan-rosianu: added

node/heartbeat/heartbeatMessageInfo.go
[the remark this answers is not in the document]
raduchis: ok

sharding/peerTypeProvider.go
```diff
@@ -0,0 +1,145 @@
...
+		cache:            make(map[string]core.PeerType),
+		mutCache:         sync.RWMutex{},
+	}
+
+	err := ptp.populateCache(epochHandler.Epoch())
+	if err != nil {
+		return nil, err
+	}
+
+	ptp.registerHandler(subscriber)
```
AdoAdoAdo: this should have been:
*startOfEpochNotifier*.RegisterHandler(function_to_be_called_when_starting_new_epoch)

and define below the function_to_be_called_when_starting_new_epoch instead of a registerHandler function.

sharding/indexHashedNodesCoordinator.go
```diff
@@ -591,6 +627,56 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
...
+
+func (ihgs *indexHashedNodesCoordinator) updatePeerAccountsForGivenMap(
+	peers map[uint32][]Validator,
+	list core.PeerType,
+) error {
+	for _, accountsPerShard := range peers {
+		for index, account := range accountsPerShard {
+			err := ihgs.listIndexUpdater.UpdateListAndIndex(string(account.PubKey()), string(list), index)
+			if err != nil {
+				log.Debug("error while updating list and index for peer",
```
AdoAdoAdo: please add this log with Warn

sharding/indexHashedNodesCoordinator.go
```diff
@@ -591,6 +627,56 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
...
+	return nil
+}
+
+func (ihgs *indexHashedNodesCoordinator) updatePeerAccountsForGivenMap(
+	peers map[uint32][]Validator,
+	list core.PeerType,
+) error {
+	for _, accountsPerShard := range peers {
+		for index, account := range accountsPerShard {
+			err := ihgs.listIndexUpdater.UpdateListAndIndex(string(account.PubKey()), string(list), index)
```
AdoAdoAdo: I think we should add also the shard (update the shard as well)

node/node_test.go
```diff
@@ -1324,6 +1338,9 @@ func TestNode_StartHeartbeatShouldWorkAndCallSendHeartbeat(t *testing.T) {
 				return mock.NewStorerMock()
 			},
 		}),
+		node.WithNodesCoordinator(&mock.NodesCoordinatorMock{}),
+		node.WithEpochStartTrigger(&mock.EpochStartTriggerStub{}),
+		node.WithEpochStartSubscriber(&mock.EpochStartNotifierStub{}),
```
AdoAdoAdo: please rename the option, as this is not a subscriber but a notifier.

sharding/indexHashedNodesCoordinator.go
```diff
@@ -591,6 +627,56 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
 	return shardEligible, nil
 }
 
+// UpdatePeersListAndIndex will update the list and the index for all peers
+func (ihgs *indexHashedNodesCoordinator) UpdatePeersListAndIndex() error {
+	ihgs.mutNodesConfig.RLock()
+	defer ihgs.mutNodesConfig.RUnlock()
+
+	nodesConfig, ok := ihgs.nodesConfig[ihgs.currentEpoch]
```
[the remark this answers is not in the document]
bogdan-rosianu: I don't think it would be ok because in the inner called functions we use the nodesConfig variable which has the same pointer as the element in the map. I would leave it like this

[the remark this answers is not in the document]
bogdan-rosianu: I don't think it would be ok because in the inner called functions we use the nodesConfig variable which has the same pointer as the element in the map. I would leave it like this

sharding/peerTypeProvider.go
```diff
@@ -0,0 +1,145 @@
...
+	epochHandler     EpochHandler
+	cache            map[string]core.PeerType
+	mutCache         sync.RWMutex
+}
+
+// NewPeerTypeProvider will return a new instance of PeerTypeProvider
+func NewPeerTypeProvider(
+	nodesCoordinator NodesCoordinator,
+	epochHandler EpochHandler,
+	subscriber EpochStartSubscriber,
```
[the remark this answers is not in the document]
bogdan-rosianu: done

sharding/peerTypeProvider.go
bogdan-rosianu: done

sharding/indexHashedNodesCoordinator.go
bogdan-rosianu: done

sharding/indexHashedNodesCoordinator.go
bogdan-rosianu: done

node/node_test.go
bogdan-rosianu: done

sharding/indexHashedNodesCoordinator.go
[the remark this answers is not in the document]
AdoAdoAdo: That is not an issue, the pointer can change inside the original map, but you already have a copy of the pointer so it is fine.

Inside the nodesCofig you now hold, when accessing the eligible and waiting maps you have other mutexes.

sharding/peerTypeProvider.go
```diff
@@ -0,0 +1,146 @@
...
+			ptp.cache[string(val)] = core.WaitingList
+		}
+	}
+
+	return nil
+}
+
+// ComputeForPubKey returns the peer type for a given public key and shard id
+func (ptp *PeerTypeProvider) ComputeForPubKey(pubKey []byte, shardID uint32) (core.PeerType, error) {
+	ptp.mutCache.Lock()
```
AdoAdoAdo: this can be RLock/RUnlock

node/options.go
```diff
@@ -370,7 +370,8 @@ func WithEpochStartTrigger(epochStartTrigger epochStart.TriggerHandler) Option {
 	}
 }
 
-func WithEpochStartSubscriber(epochStartSubscriber epochStart.EpochStartSubscriber) Option {
+// WithEpochStartEventNotifier sets up the notifier for the epoch start event
+func WithEpochStartEventNotifier(epochStartSubscriber epochStart.EpochStartSubscriber) Option {
```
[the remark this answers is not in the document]
bogdan-rosianu: done

sharding/peerTypeProvider.go
bogdan-rosianu: done

PR #1106, opened 2020-02-17
configuration and flags refactoring

bogdan-rosianu:


cmd/node/config/external.toml
```diff
@@ -0,0 +1,7 @@
+[Explorer]
```
iulianpascalau: Should merge these 2 sections: move the "Explorer" configs inside "ElasticSearchConnector"

scripts/testnet/include/config.sh
```diff
@@ -56,7 +56,6 @@ copyNodeConfig() {
   cp $NODEDIR/config/config.toml ./node/config
   cp $NODEDIR/config/economics.toml ./node/config
   cp $NODEDIR/config/prefs.toml ./node/config
-  cp $NODEDIR/config/server.toml ./node/config
```
iulianpascalau: maybe change to copy external.toml

cmd/node/config/external.toml
bogdan-rosianu: done

scripts/testnet/include/config.sh
bogdan-rosianu: done

cmd/node/main.go
```diff
@@ -932,7 +938,7 @@ func copySingleFile(folder string, configFile string) {
 		return
 	}
 	defer func() {
-		err := destination.Close()
+		err = destination.Close()
```
miiu96: go at line 949 and change print message

cmd/node/main.go
```diff
@@ -1034,8 +1040,18 @@ func loadEconomicsConfig(filepath string) (*config.ConfigEconomics, error) {
 	return cfg, nil
 }
 
-func loadPreferencesConfig(filepath string) (*config.ConfigPreferences, error) {
-	cfg := &config.ConfigPreferences{}
+func loadPreferencesConfig(filepath string) (*config.Preferences, error) {
```
miiu96: you should move all that functions from  main.go in a new file maybe a common file or something like that

config/tomlConfig_test.go
```diff
@@ -168,22 +168,54 @@ func TestTomlEconomicsParser(t *testing.T) {
...
-	cfg := ConfigPreferences{}
+	cfg := Preferences{}
 
 	err := toml.Unmarshal([]byte(testString), &cfg)
 
 	assert.Nil(t, err)
 	assert.Equal(t, cfgPreferencesExpected, cfg)
 }
+
+func TestTomlExternalParser(t *testing.T) {
```
miiu96: :+1:

cmd/node/main.go

cmd/node/main.go
bogdan-rosianu: will leave it like this for now. maybe in a future PR

config/tomlConfig_test.go
bogdan-rosianu: :+1:

cmd/node/main.go
miiu96: why you cant do this now because you have just to copy a part of the code from main.go in a new file ?

PR #1122, opened 2020-02-18
changed refresh topic peers interval values

iulianpascalau:
changed refresh topic peers interval values
updated nodes.sh


p2p/libp2p/peersOnChannel.go
```diff
@@ -74,7 +74,7 @@ func (poc *peersOnChannel) ConnectedPeersOnChannel(topic string) []p2p.PeerID {
 func (poc *peersOnChannel) updateConnectedPeersOnTopic(topic string, connectedPeers []p2p.PeerID) {
 	poc.mutPeers.Lock()
 	poc.peers[topic] = connectedPeers
-	poc.lastUpdated[topic] = poc.clockTime()
+	poc.lastUpdated[topic] = poc.getTimeHandler()
```
SebastianMarian: I think that clockTime method could be removed now, as it is unused

PR #1128, opened 2020-02-19
En 5800 added harmonic mean in ntp synchronization

SebastianMarian:


ntp/syncTime.go
```diff
@@ -73,89 +88,127 @@ type syncTime struct {
...
 		s.sync()
-		time.Sleep(s.syncPeriod)
+		time.Sleep(s.getSleepTime())
 	}
 }
 
-// sync method does the time synchronization and sets the current offset difference between local time
-// and server time with which it has done the synchronization
+func (s *syncTime) getSleepTime() time.Duration {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
miiu96: you should use "crypto/rand" instead of "math/rand" because will get linter warnings

ntp/syncTime.go
```diff
@@ -2,42 +2,58 @@ package ntp
...
 }
 
-// NewNTPOptions creates a new NTPOptions object.
-func NewNTPOptions(ntpConfig config.NTPConfig, hostIndex int) NTPOptions {
-	if ntpConfig.TimeoutMilliseconds <= 0 {
-		ntpConfig.TimeoutMilliseconds = 100
-	}
+// NewNTPOptions creates a new NTPOptions object
+func NewNTPOptions(ntpConfig config.NTPConfig) NTPOptions {
+	ntpConfig.TimeoutMilliseconds = core.MinInt(100, ntpConfig.TimeoutMilliseconds)
```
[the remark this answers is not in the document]
SebastianMarian: Done

ntp/syncTime.go
SebastianMarian: Done

PR #1139, opened 2020-02-24
fix-epoch-start edge case take min finality attesting round

sasurobert:


epochStart/shardchain/trigger.go
```diff
@@ -226,20 +229,35 @@ func (t *trigger) RequestEpochStartIfNeeded(interceptedHeader data.HeaderHandler
 	}
 }
 
+func (t *trigger) checkIfEpochFinalityAttestingWithLowerRound(metaHdr *block.MetaBlock) {
```
SebastianMarian: setLowerEpochFinalityAttestingRoundIfNeeded

epochStart/shardchain/trigger.go
```diff
@@ -42,11 +43,13 @@ type ArgsShardEpochStartTrigger struct {
 type trigger struct {
 	currentRoundIndex           int64
 	epochStartRound             uint64
+	epochStartNonce             uint64
```
SebastianMarian: This could be removed as you have this info in epochStartMeta

epochStart/shardchain/trigger.go
```diff
@@ -226,20 +229,35 @@ func (t *trigger) RequestEpochStartIfNeeded(interceptedHeader data.HeaderHandler
 	}
 }
 
+func (t *trigger) checkIfEpochFinalityAttestingWithLowerRound(metaHdr *block.MetaBlock) {
+	if metaHdr.GetNonce() != t.epochStartNonce+1 {
```
SebastianMarian: This check is done below in IsHeaderConstructionValid method so I think it could be removed from here

epochStart/shardchain/trigger.go
```diff
@@ -284,6 +302,8 @@ func (t *trigger) updateTriggerFromMeta(metaHdr *block.MetaBlock, hdrHash []byte
 			t.epochStartRound = meta.Round
 			t.epochFinalityAttestingRound = finalityAttestingRound
 			t.epochMetaBlockHash = []byte(hash)
+			t.epochStartNonce = meta.Nonce
```
SebastianMarian: Not needed

epochStart/shardchain/trigger.go
```diff
@@ -443,14 +463,30 @@ func (t *trigger) getHeaderWithNonceAndHash(nonce uint64, neededHash []byte) (*b
...
+		}
+
+		if lowestRound > hdrWithNonce.GetRound() {
+			lowestRound = hdrWithNonce.GetRound()
+			chosenMeta = hdrWithNonce
 		}
 	}
-	return nil
+
+	if chosenMeta.GetNonce() != nonce {
```
SebastianMarian: You can check: if lowestRound == uint64(math.MaxUint64) instead this one, as there could be an edge case when you call this method with parameter nonce = 0 and you don't find anything

epochStart/shardchain/trigger.go
```diff
@@ -470,12 +508,20 @@ func (t *trigger) getHeaderWithNonceAndPrevHashFromCache(nonce uint64, prevHash
...
+			chosenMeta = hdrWithNonce
+		}
+
 		t.mapHashHdr[string(hashes[i])] = hdrWithNonce
 		t.mapNonceHashes[hdrWithNonce.Nonce] = append(t.mapNonceHashes[hdrWithNonce.Nonce], string(hashes[i]))
-		return hdrWithNonce
 	}
 
-	return nil
+	if chosenMeta.GetNonce() != nonce {
```
SebastianMarian: The same comment here, you can use  if lowestRound == uint64(math.MaxUint64) instead this one

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go
sasurobert: no way to call this with nonce 0

SebastianMarian: I know, but with this check you will cover any situation

epochStart/shardchain/trigger.go
SebastianMarian: I know, but with this check you will cover any situation

PR #1152, opened 2020-02-26
Fix economics v2

sasurobert:


data/state/peerAccount.go
```diff
@@ -353,15 +353,17 @@ func (pa *PeerAccount) DecreaseValidatorSuccessRateWithJournal(value uint32) err
 
 // IncreaseLeaderAccumulatedFees increases the account's accumulated fees
 func (pa *PeerAccount) AddToAccumulatedFees(value *big.Int) error {
-	newAccumulatedFees := big.NewInt(0).Add(pa.AccumulatedFees, value)
+	if value.Cmp(big.NewInt(0)) == 0 {
```
camilbancioiu: `len(value.Bytes()) == 0` is equivalent and may be faster.

PR #1154, opened 2020-02-26
Fixed an edge case when node not requests next cross header

SebastianMarian:
Fixed an edge case when node not requests next cross header if nothing new would be received because of an epoch change

### DIFF
--- a/process/track/blockProcessor.go
+++ b/process/track/blockProcessor.go
@@ -3,6 +3,7 @@ package track
 import (
 	"sort"
 
+	"github.com/ElrondNetwork/elrond-go/consensus"
 	"github.com/ElrondNetwork/elrond-go/core"
 
 	"github.com/ElrondNetwork/elrond-go/core/check"
@@ -20,6 +21,7 @@ type blockProcessor struct {
 	crossNotarizer                blockNotarizerHandler
 	crossNotarizedHeadersNotifier blockNotifierHandler
 	selfNotarizedHeadersNotifier  blockNotifierHandler
+	rounder                       consensus.Rounder
 
 	blockFinality uint64
 }
@@ -39,6 +41,7 @@ func NewBlockProcessor(arguments ArgBlockProcessor) (*blockProcessor, error) {
 		crossNotarizer:                arguments.CrossNotarizer,
 		crossNotarizedHeadersNotifier: arguments.CrossNotarizedHeadersNotifier,
 		selfNotarizedHeadersNotifier:  arguments.SelfNotarizedHeadersNotifier,
+		rounder:                       arguments.Rounder,
 	}
 
 	bp.blockFinality = process.BlockFinality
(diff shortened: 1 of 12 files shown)


process/track/blockProcessor.go
```diff
@@ -233,31 +241,83 @@ func (bp *blockProcessor) requestHeadersIfNeeded(
...
-	nbLongestChainHeaders := len(longestChainHeaders)
-	if nbLongestChainHeaders > 0 {
-		highestNonceInLongestChain = longestChainHeaders[nbLongestChainHeaders-1].GetNonce()
+	numLongestChainHeaders := len(longestChainHeaders)
+	if numLongestChainHeaders > 0 {
+		highestNonceInLongestChain = longestChainHeaders[numLongestChainHeaders-1].GetNonce()
 	}
 
-	shouldRequestHeaders := highestNonceReceived > highestNonceInLongestChain+bp.blockFinality && nbLongestChainHeaders == 0
+	shouldRequestHeaders = highestNonceReceived > highestNonceInLongestChain+bp.blockFinality && numLongestChainHeaders == 0
```
iulianpascalau: why aren't we requesting if we have a not empty longestChainHeaders?

process/track/blockProcessor.go
```diff
@@ -233,31 +241,83 @@ func (bp *blockProcessor) requestHeadersIfNeeded(
...
+	longestChainHeaders []data.HeaderHandler,
+) {
+	if check.IfNil(lastNotarizedHeader) {
+		return
+	}
+
+	header := lastNotarizedHeader
+	numLongestChainHeaders := len(longestChainHeaders)
+	if numLongestChainHeaders > 0 {
+		header = longestChainHeaders[numLongestChainHeaders-1]
```
iulianpascalau: this can never be reached as how the code is written now

process/track/blockProcessor.go
```diff
@@ -233,31 +241,83 @@ func (bp *blockProcessor) requestHeadersIfNeeded(
...
+	}
+
+	highestRoundReceived := header.GetRound()
+	numSortedHeaders := len(sortedHeaders)
+	if numSortedHeaders > 0 {
+		if sortedHeaders[numSortedHeaders-1].GetRound() > highestRoundReceived {
```
iulianpascalau:
```GO
highestRoundReceived = core.MaxUint64(highestRoundReceived, sortedHeaders[numSortedHeaders-1].GetRound())
```

SebastianMarian: Done

process/track/blockProcessor.go
SebastianMarian: Could be, but anyway I have refactored this part

process/track/blockProcessor.go
SebastianMarian: This is done to avoid many requests when node is in sync mode and it receives headers with higher nonces than it's last processed header nonce. As long as it can advance, so numLongestChainHeaders > 0 than, it should not request for headers in advance

process/track/blockProcessor.go
```diff
@@ -20,6 +21,7 @@ type blockProcessor struct {
 	crossNotarizer                blockNotarizerHandler
 	crossNotarizedHeadersNotifier blockNotifierHandler
 	selfNotarizedHeadersNotifier  blockNotifierHandler
+	rounder                       consensus.Rounder
```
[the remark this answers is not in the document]
SebastianMarian: Done

PR #1158, opened 2020-02-27
Fix sync problem on epoch change

SebastianMarian:


process/sync/baseSync.go
```diff
@@ -298,25 +294,46 @@ func (boot *baseBootstrap) ShouldSync() bool {
...
 
 	roundDiff := uint64(boot.rounder.Index()) - currHeader.GetRound()
-	if roundDiff < tryRequestHeaderDelta {
+	if roundDiff <= process.MaxRoundsWithoutNewBlockReceived {
 		return
 	}
 
-	shouldTrySync := roundDiff%tryRequestHeaderDelta == 0
+	numRequestedHdrs := 0
+	nonce := boot.getNonceForNextBlock()
```
iulianpascalau: is this concurrent safe?

SebastianMarian: If this is not concurrent safe that means we have a problem on each call from chainHandler. This is the method listed:

// getNonceForNextBlock will get the nonce for the next block we should request
func (boot *baseBootstrap) getNonceForNextBlock() uint64 {
	nonce := uint64(1) // first block nonce after genesis block
	if boot.chainHandler != nil && boot.chainHandler.GetCurrentBlockHeader() != nil {
		nonce = boot.chainHandler.GetCurrentBlockHeader().GetNonce() + 1
	}
	return nonce
}

process/sync/baseSync.go
```diff
@@ -298,25 +294,46 @@ func (boot *baseBootstrap) ShouldSync() bool {
 	}
 	boot.statusHandler.SetUInt64Value(core.MetricIsSyncing, result)
 
+	go boot.requestHeadersIfSyncIsStuck()
```
sasurobert: no need for go routine.

process/sync/baseSync.go
sasurobert: check.ifnil

process/sync/baseSync.go

PR #1180, opened 2020-03-02
Fix sync requests on startup

sasurobert:


dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -242,7 +242,7 @@ func (rrh *resolverRequestHandler) RequestShardHeader(shardID uint32, hash []byt
 		return
 	}
 
-	log.Trace("requesting shard header from network",
+	log.Debug("requesting shard header from network",
```
raduchis: Is the Debug needed?

node/node.go
```diff
@@ -252,6 +252,12 @@ func (n *Node) StartConsensus(epoch uint32) error {
 	}
 
 	bootstrapper.StartSync()
+	epoch := uint32(0)
+	crtBlockHeader := n.blkc.GetCurrentBlockHeader()
+	if !check.IfNil(crtBlockHeader) {
+		epoch = crtBlockHeader.GetEpoch()
+		log.Debug("starting consensus", "epoch", epoch)
```
raduchis: I would put log.Info here, because it happens just once when you start the node

process/track/blockProcessor.go
```diff
@@ -311,14 +311,19 @@ func (bp *blockProcessor) getHighestRoundInReceivedHeaders(
 }
 
 func (bp *blockProcessor) requestHeadersIfNothingNewIsReceived(
+	lastNotarizedHeader data.HeaderHandler,
```
raduchis: lastNotarized header is not needed here, just the lastNotarizedHeader.GetNonce(), just the nonce could be passed as param.

dataRetriever/requestHandlers/requestHandler.go
iulianpascalau: yes, for the moment, we want to see the effective requests

node/node.go
iulianpascalau: done

process/track/blockProcessor.go
iulianpascalau: done

node/node.go
```diff
@@ -252,6 +252,12 @@ func (n *Node) StartConsensus(epoch uint32) error {
 	}
 
 	bootstrapper.StartSync()
+	epoch := uint32(0)
+	crtBlockHeader := n.blkc.GetCurrentBlockHeader()
+	if !check.IfNil(crtBlockHeader) {
+		epoch = crtBlockHeader.GetEpoch()
+		log.Info("starting consensus", "epoch", epoch)
```
bogdan-rosianu: shouldn't this log message be displayed in both cases ?

node/node_test.go
```diff
@@ -2406,9 +2406,9 @@ func TestNode_SendBulkTransactionsMultiShardTxsShouldBeMappedCorrectly(t *testin
 			}
 			for _, txBuff := range txsBuff {
 				tx := transaction.Transaction{}
-				err := marshalizer.Unmarshal(&tx, txBuff)
-				if err != nil {
-					assert.Fail(t, err.Error())
+				errMarshal := marshalizer.Unmarshal(&tx, txBuff)
+				if errMarshal != nil {
```
bogdan-rosianu: `assert.Nil(t, errMarshal)` or even `require`

node/node.go
iulianpascalau: :1st_place_medal: done

node/node_test.go
iulianpascalau: require

dataRetriever/requestHandlers/requestHandler.go
raduchis: ok

node/node.go
raduchis: ok

process/track/blockProcessor.go
raduchis: ok

PR #1185, opened 2020-03-04
Fix update peer on meta

iulianpascalau:
- fixed the edge case that appeared on meta nodes when the header cache was full and the peer processor could not find shard headers to process them.
- minor refactorization on block processors


epochStart/metachain/economics.go
```diff
@@ -93,7 +93,7 @@ func (e *economics) ComputeEndOfEpochEconomics(
 	}
 
 	roundsPassedInEpoch := metaBlock.GetRound() - prevEpochStart.GetRound()
```
raduchis: can these 2 be equal? how?

epochStart/metachain/trigger.go
```diff
@@ -238,11 +238,21 @@ func (t *trigger) SetFinalityAttestingRound(round uint64) {
 }
 
 // Revert sets the start of epoch back to true
-func (t *trigger) Revert(round uint64) {
+func (t *trigger) Revert(header data.HeaderHandler) {
+	if check.IfNil(header) || !header.IsStartOfEpochBlock() || header.GetEpoch() == 0 {
```
raduchis: Shouldn't it be possible to revert a StartOfEpochBlock?

process/block/metablock.go
```diff
@@ -588,7 +587,7 @@ func (mp *metaProcessor) RestoreBlockIntoPools(headerHandler data.HeaderHandler,
 	mp.blockTracker.RemoveLastNotarizedHeaders()
 
 	if metaBlock.IsStartOfEpochBlock() {
-		mp.epochStartTrigger.Revert(metaBlock.GetRound())
+		mp.epochStartTrigger.Revert(metaBlock)
```
raduchis: This line will do nothing, because the metaTrigger has a check for IsStartOfEpochBlock

process/peer/process.go
```diff
@@ -475,14 +475,17 @@ func (vs *validatorStatistics) updateShardDataPeerState(header data.HeaderHandle
...
-		prevShardData, shardInfoErr := process.GetShardHeader(
-			h.PrevHash,
-			vs.dataPool.Headers(),
-			vs.marshalizer,
-			vs.storageService,
-		)
+		prevShardData, shardInfoErr := vs.searchInMap(h.PrevHash, cacheMap)
 		if shardInfoErr != nil {
-			return shardInfoErr
+			prevShardData, shardInfoErr = process.GetShardHeader(
```
raduchis: If not nil, search again and then return?

epochStart/metachain/economics.go
sasurobert: there is no way for these 2 to be equal.

epochStart/metachain/trigger.go
sasurobert: reverting a trigger should happen only if header is of type start of epoch block.

process/block/metablock.go
sasurobert: it is only for double protection.

raduchis: ok

epochStart/metachain/trigger.go
raduchis: ok

process/peer/process.go
iulianpascalau: simplified, search only in provided map as you can not start processing if the data is not available in this temporary processing cache.

process/block/metablock.go
```diff
@@ -320,12 +321,6 @@ func (mp *metaProcessor) processEpochStartMetaBlock(
 		return err
 	}
 
-	defer func() {
```
iulianpascalau: Why did we remove this? In case we hit any of the following errors we do not revert the account state making impossible to (re)sync the next block

process/peer/process.go
sasurobert: resolved

process/block/metablock.go
sasurobert: it is called the defer from ProcessBlock, It would have been called twice.

iulianpascalau: oh

PR #1202, opened 2020-03-10
simplified heartbeat storage DTO

iulianpascalau:


node/heartbeat/proto/heartbeat.proto
```diff
@@ -18,21 +16,21 @@ message Heartbeat {
...
+	bool    IsValidator                  = 11;
+	int64	LastUptimeDowntime           = 12;
+	int64   GenesisTime                  = 13;
 }
 
 message DbTimeStamp {
-	google.protobuf.Timestamp TS                    = 1 ;
-}
+	int64 Timestamp = 1;
+}
```
AdoAdoAdo: newline characters

iulianpascalau: ok, regarding proxy, will launch a system test and will check. Expected to work accordingly as these structures are use only on local storage.

sasurobert: add new line to end of file.

iulianpascalau: done

iulianpascalau:
> check also that proxy is happy

Tested, works as expected.

PR #1203, opened 2020-03-10
En 5859 reward address nil

raduchis:
also waiting nodes are initialized in the peer processor


sharding/indexHashedNodesCoordinator.go
```diff
@@ -306,6 +306,14 @@ func (ihgs *indexHashedNodesCoordinator) GetValidatorWithPublicKey(
 		}
 	}
 
+	for shardId, shardWaiting := range nodesConfig.waitingMap {
```
iulianpascalau: Isn't this time consuming? This will be used in the upcoming network sharding feature branch as a way to know the shard ID of the public key provided. It will be called each time a connection is made (currently about 5 connections/sec are done)

raduchis: fixed by using a map for caching

sharding/indexHashedNodesCoordinator.go
```diff
@@ -481,14 +526,19 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
...
-		publicKeysPrevEpoch, err = ihgs.GetAllValidatorsPublicKeys(epoch - 1)
+		publicKeysPrevEpoch, err = ihgs.GetAllEligibleValidatorsPublicKeys(epoch - 1)
 		if err == nil {
 			prevEpochConfigExists = true
 		} else {
 			log.Warn("get consensus whitelisted nodes", "error", err.Error())
 		}
 	}
 
+	publicKeysNewEpoch, err := ihgs.GetAllEligibleValidatorsPublicKeys(epoch)
```
golangcibot: ineffectual assignment to `publicKeysNewEpoch` (from `ineffassign`)

PR #1219, opened 2020-03-13
Correlation of header tracker and cacher pools

SebastianMarian:
* Fixed an edge situation when header pools are full and many requests are done from blockProcessor but they are rejected by block tracker due an un-correlated out of range size for these pools.
* Fixed a situation when repetitive requests of the same header nonce/hash have been done from bootstrapper
* Fixed situation when after a forced roll back node would not be considered synced because method requestHeadersIfSyncIsStuck is called each round and this caused that node would received higher headers nonces and actually the mechanism of forced roll back would not worked as expected


process/sync/baseSync.go
```diff
@@ -845,6 +852,7 @@ func (boot *baseBootstrap) init() {
 }
 
 func (boot *baseBootstrap) requestHeaders(fromNonce uint64, toNonce uint64) {
+	boot.mutRequestHeaders.Lock()
 	numRequestedHeaders := 0
```
sasurobert: maybe add in defer the UnLock()

process/track/baseBlockTrack.go
```diff
@@ -106,7 +106,7 @@ func (bbt *baseBlockTrack) receivedMetaBlock(headerHandler data.HeaderHandler, m
 	bbt.blockProcessor.ProcessReceivedHeader(metaBlock)
 }
 
-func (bbt *baseBlockTrack) shouldAddHeader(headerHandler data.HeaderHandler) bool {
+func (bbt *baseBlockTrack) ShouldAddHeader(headerHandler data.HeaderHandler) bool {
```
sasurobert: add comment to public method.

process/sync/baseSync.go

process/track/baseBlockTrack.go
SebastianMarian: Done

PR #1225, opened 2020-03-14
Fix counting of pending miniblocks

SebastianMarian:
* Fixed counting of pending miniblocks created from meta to shards at the end of each epoch (rewards miniblocks)

### DIFF
--- a/epochStart/metachain/pendingMiniBlocks.go
+++ b/epochStart/metachain/pendingMiniBlocks.go
@@ -12,6 +12,7 @@ import (
 
 type pendingMiniBlocksHeaders struct {
 	mutPending            sync.RWMutex
+	//TODO: Also this map should be saved in BootStorer
 	mapMiniBlockHeaders   map[string]block.ShardMiniBlockHeader
 	mapShardNumMiniBlocks map[uint32]uint32
 }
@@ -28,8 +29,11 @@ func (p *pendingMiniBlocksHeaders) getAllCrossShardMiniBlocks(metaBlock *block.M
 	crossShardMiniBlocks := make(map[string]block.ShardMiniBlockHeader)
 
 	for _, mbHeader := range metaBlock.MiniBlockHeaders {
+		if mbHeader.SenderShardID != core.MetachainShardId && mbHeader.ReceiverShardID == core.MetachainShardId {
+			continue
+		}
 		//TODO: Activate the following code when EN-5464-peer-rating-process-in-shard will be merged in dev
-		//if mbHeader.ReceiverShardID == core.AllShardId {
+		//if mbHeader.SenderShardID == core.MetachainShardId && mbHeader.ReceiverShardID == core.AllShardId {
 		//	continue
 		//}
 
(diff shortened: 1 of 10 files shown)


epochStart/metachain/pendingMiniBlocks.go
```diff
@@ -1,257 +1,126 @@
 package metachain
```
sasurobert: i would move it to the trakcer package. no more need to stay in epochStart

epochStart/metachain/pendingMiniBlocks.go
```diff
@@ -1,257 +1,126 @@
...
 
-		delete(p.mapMiniBlockHeaders, key)
-		p.decrementNumMiniBlocks(mbHeader.ReceiverShardID)
-
-		var buff []byte
-		buff, err = p.marshalizer.Marshal(&mbHeader)
-		if err != nil {
-			return err
+		delete(p.mapMiniBlockHeaders, mbHash)
+		if p.mapShardNumMiniBlocks[mbHeader.ReceiverShardID] > 0 {
```
AdoAdoAdo: can we log an error in case p.mapShardNumMiniBlocks[] == 0 and we deleted?

so I would instead do: if p.mapShard.....[] == 0{
log error
continue
}
p.mapShardNum....[]--

although we should afterwards pay attention to this kind of errors in the logs...

epochStart/metachain/pendingMiniBlocks.go
```diff
@@ -1,257 +1,126 @@
...
-	crossShard := p.getAllCrossShardMiniBlocks(metaHdr)
+	crossShardMiniBlocks := p.getAllCrossShardMiniBlocks(metaBlock)
 
 	p.mutPending.Lock()
-	for mbHash, mbHeader := range crossShard {
+	for mbHash, mbHeader := range crossShardMiniBlocks {
 		if _, ok = p.mapMiniBlockHeaders[mbHash]; ok {
 			delete(p.mapMiniBlockHeaders, mbHash)
-			p.decrementNumMiniBlocks(mbHeader.ReceiverShardID)
+			if p.mapShardNumMiniBlocks[mbHeader.ReceiverShardID] > 0 {
```
AdoAdoAdo: can we log an error if we deleted but the num miniblocks was 0?

epochStart/metachain/pendingMiniBlocks.go
SebastianMarian: Done

epochStart/metachain/pendingMiniBlocks.go
SebastianMarian: Done

epochStart/metachain/pendingMiniBlocks.go
SebastianMarian: Actually is more used in process/block and also there is another interface for this implementation in process/interface.go.  So I moved it in one new folder(package) process/block/pendingMb/ and removed interface and implementation from epochStart

PR #1234, opened 2020-03-17
Tob crypto fixes

AdoAdoAdo:


crypto/signing/mcl/multisig/bls.go
```diff
@@ -281,8 +337,10 @@ func hashPublicKeyPoint(hasher hashing.Hasher, pubKeyPoint crypto.Point) ([]byte
 		return nil, err
 	}
 
-	// H1(pubkey_i)
-	h := hasher.Compute(string(pointBytes))
+	concatPkWithPKs := append(concatPubKeys, pointBytes...)
```
ccorcoveanu: Shouldn't it be the other way around? pki, {pk1, pk2....}

crypto/signing/mcl/multisig/bls.go
```diff
@@ -133,20 +117,37 @@ func (bms *BlsMultiSigner) VerifyAggregatedSig(
...
+	prepPubKeysPoints := make([]bls.PublicKey, len(pubKeys))
+
+	concatPKs, err := concatPubKeys(pubKeys)
+	if err != nil {
+		return nil, err
+	}
+
+	pubKeysPoints := make([]crypto.Point, len(pubKeys))
+	for i, pubKey := range pubKeys {
+		if check.IfNil(pubKey) {
```
ccorcoveanu: Why not make this check in the next for and get the point there?

crypto/signing/mcl/multisig/bls.go
AdoAdoAdo: Although it shouldn't matter, indeed the specification is pki, {pk1,...pkn}
done

crypto/signing/mcl/multisig/bls.go
AdoAdoAdo: right, done

PR #1244, opened 2020-03-18
Dev and hardfork to start in epoch mar18

sasurobert:


core/converters.go
```diff
@@ -126,6 +126,10 @@ func IsUnknownEpochIdentifier(identifier []byte) (bool, error) {
 // CommunicationIdentifierBetweenShards is used to generate the identifier between shardID1 and shardID2
 // identifier is generated such as the first shard from identifier is always smaller or equal than the last
 func CommunicationIdentifierBetweenShards(shardId1 uint32, shardId2 uint32) string {
+	if shardId1 == AllShardId || shardId2 == AllShardId {
```
iulianpascalau: I do not think this is ok

sasurobert: this is exactly a copy from development.

PR #1245, opened 2020-03-18
Fixed sender and receiver shard used to process txs

SebastianMarian:
* Fixed sender and receiver shard used to process txs in validator side
* Added unit test


process/block/preprocess/export_test.go
```diff
@@ -92,3 +93,40 @@ func (bsc *blockSizeComputation) NumMiniBlocks() uint32 {
...
+		},
+	}
+}
+
+func (txs *transactions) GetTxInfoForCurrentBlock(txHash []byte) (data.TransactionHandler, uint32, uint32) {
+	txs.txsForCurrBlock.mutTxsForBlock.RLock()
+	defer txs.txsForCurrBlock.mutTxsForBlock.RUnlock()
+
+	txInfo := txs.txsForCurrBlock.txHashAndInfo[string(txHash)]
+	return txInfo.tx, txInfo.senderShardID, txInfo.receiverShardID
```
iulianpascalau: this will panic if txInfo is nil (not existing in map).

PR #1250, opened 2020-03-19
[mcl] Resync with upstream changes

traian-schiau:


crypto/signing/mcl/suiteBLS12_381.go
```diff
@@ -22,37 +20,33 @@ type SuiteBLS12 struct {
...
 }
 
-// blsSwapG Is enabled if compiling mcl/bls/bls-go-binary with BLS_SWAP_G flag
+// Is enabled if compiling mcl/bls/bls-go-binary with BLS_SWAP_G flag, curently this is not the case
 // Compiling with the flag will give Public Keys on G1 (48 bytes) and Signatures on G2 (96 bytes)
 // Compiling without the flag will give Public Keys on G2 and Signatures on G1
 // For Elrond the public keys for the validators are known during an epoch and also are not set on blocks
 // BLS signatures are however set on every block header, so in order to optimise the header size flag will be false
 // to have smaller signatures, so on G1(48 bytes)
+const blsSwapG = false
```
AdoAdoAdo: maybe we should remove the constant as it is no longer used, but keep the comment (adapted) so that the configuration we are using is still clear.

crypto/signing/mcl/suiteBLS12_381.go
```diff
@@ -22,37 +20,33 @@ type SuiteBLS12 struct {
...
-const g1str = "1 3685416753713387016781088315183077757961620795782546409894578378688607592378376318836054947676345821548104185464507 1339506544944476473020471379941921221584933875938349620426543736416511423956333506472724655353366534992391756441569"
-
-var basePointG1Str atomic.Value
-var basePointG2Str atomic.Value
+var (
+	g2str  string = "1 352701069587466618187139116011060144890029952792775240219908644239793785735715026873347600343865175952761926303160 3059144344244213709971259814753781636986470325476647558659373206291635324768958432433509563104347017837885763365758 1985150602287291935568054521177171638300868978215655730859378665066344726373823718423869104263333984641494340347905 927553665492332455747201965776037880757740193453592970025027978793976877002675564980949289727957565575433344219582"
+	g1str  string = "1 3685416753713387016781088315183077757961620795782546409894578378688607592378376318836054947676345821548104185464507 1339506544944476473020471379941921221584933875938349620426543736416511423956333506472724655353366534992391756441569"
```
AdoAdoAdo: if we no longer support  both configs, I think we can safely remove g2str as this is read from the lib

Probably we can also remove g1str as in this case there would be no reason to create sigPoints starting from base point, but I would leave this for a future refactor as there will be some changes needed in the wrappers.

crypto/signing/mcl/suiteBLS12_381.go
```diff
@@ -146,14 +126,12 @@ func (s *SuiteBLS12) IsInterfaceNil() bool {
 
 // BaseG1 returns the generator point for G1
 func BaseG1() string {
-	v := basePointG1Str.Load()
-	vStr := v.(string)
-	return vStr
+	doInit.Do(blsInit)
```
AdoAdoAdo: I think you can remove from here the init

crypto/signing/mcl/suiteBLS12_381.go
```diff
@@ -146,14 +126,12 @@ func (s *SuiteBLS12) IsInterfaceNil() bool {
...
+	return g1str
 }
 
 // BaseG2 returns the generator point for G2
 func BaseG2() string {
-	v := basePointG2Str.Load()
-	vStr := v.(string)
-	return vStr
+	doInit.Do(blsInit)
```
AdoAdoAdo: I think you can remove from here the init

crypto/signing/mcl/suiteBLS12_381.go
```diff
@@ -94,21 +88,7 @@ func (s *SuiteBLS12) CreateScalar() crypto.Scalar {
...
-	}
-
-	if sc.IsZero() || !sc.IsValid() {
-		return nil, crypto.ErrInvalidPrivateKey
-	}
-
-	point := s.G2.CreatePointForScalar(scalar)
-
-	return point, nil
+	return s.G2.CreatePointForScalar(scalar), nil
```
AdoAdoAdo: I guess the removal of the checks was unintended

crypto/signing/mcl/suiteBLS12_381.go
traian-schiau: done

crypto/signing/mcl/suiteBLS12_381.go
traian-schiau: done

crypto/signing/mcl/suiteBLS12_381.go
traian-schiau: future

crypto/signing/mcl/suiteBLS12_381.go
traian-schiau: done

crypto/signing/mcl/suiteBLS12_381.go
traian-schiau: undone

PR #1257, opened 2020-03-20
Include logger name and other correlation elements in log line

andreibancioiu:
Logger name & epoch & round & sub-round.

```
DEBUG[2020-03-21 17:33:09.235] [p2p/libp2p]         [0/1/(START_ROUND)] network connection metrics               connections/s = 0 disconnections/s = 0 
TRACE[2020-03-21 17:33:09.435] [facade]             [0/1/(START_ROUND)] calling reset on WS source limiter       
TRACE[2020-03-21 17:33:10.435] [facade]             [0/1/(START_ROUND)] calling reset on WS source limiter       
TRACE[2020-03-21 17:33:11.435] [facade]             [0/1/(START_ROUND)] calling reset on WS source limiter       
DEBUG[2020-03-21 17:33:11.703] [consensus/spos]     [0/1/(START_ROUND)] extend function is called                subround = (START_ROUND) 

```


core/atomic/int64.go
```diff
@@ -0,0 +1,18 @@
+package atomic
```
iulianpascalau: why not using directly atomic? The rest of the project uses atomic directly.

core/atomic/string.go
```diff
@@ -0,0 +1,23 @@
+package atomic
```
iulianpascalau: same, why not using directly atomic Value?

logger/plainFormatter.go
```diff
@@ -14,13 +14,30 @@ func (pf *PlainFormatter) Output(line LogLineHandler) []byte {
 	}
 
 	level := LogLevel(line.GetLogLevel())
+	timestamp := displayTime(line.GetTimestamp())
+	loggerName := formatLoggerName(line.GetLoggerName())
+	message := formatMessage(line.GetMessage())
+	args := formatArgsNoAnsi(line.GetArgs()...)
+
+	const formatStringWithCorrelationElements = "%s[%s] %s %s %s %s\n"
```
iulianpascalau: why not taking out these consts on the package level?

logger/correlation.go
```diff
@@ -0,0 +1,68 @@
...
+
+type logCorrelation struct {
+	enabled  bool
+	epoch    atomic.Uint32
+	round    atomic.Int64
+	subRound atomic.String
+}
+
+func (correlation *logCorrelation) enable() {
+	// No need for atomic updates (it should only be called once, in main)
```
iulianpascalau: we should be able to call this as an option from logviewer, also.

logger/consoleFormatter.go
```diff
@@ -24,14 +24,30 @@ func (cf *ConsoleFormatter) Output(line LogLineHandler) []byte {
 
 	level := LogLevel(line.GetLogLevel())
 	levelColor := getLevelColor(level)
+	timestamp := displayTime(line.GetTimestamp())
+	loggerName := formatLoggerName(line.GetLoggerName())
```
iulianpascalau: I think we should be able to turn on/off this string, just like correlation.

logger/correlation.go
```diff
@@ -0,0 +1,68 @@
+package logger
+
+import (
+	atomic "github.com/ElrondNetwork/elrond-go/core/atomic"
+)
+
+var correlation logCorrelation
```
iulianpascalau: logger just got to know about epochs/rounds and subrounds.

logger/correlation.go
```diff
@@ -0,0 +1,68 @@
...
+
+type logCorrelation struct {
```
iulianpascalau: missing tests

logger/consoleFormatter.go
```diff
@@ -24,14 +24,30 @@ func (cf *ConsoleFormatter) Output(line LogLineHandler) []byte {
...
+	args := formatArgs(levelColor, line.GetArgs()...)
 
-	return []byte(fmt.Sprintf("\033[%s%s\033[0m[%s] %s %s\n",
-		levelColor,
-		level,
-		displayTime(line.GetTimestamp()),
-		formatMessage(line.GetMessage()),
-		formatArgs(levelColor, line.GetArgs()...),
-	),
+	const formatStringWithCorrelationElements = "\033[%s%s\033[0m[%s] %s %s %s %s\n"
```
iulianpascalau: why not taking out these consts on the package level?

logger/consoleFormatter.go
camilbancioiu: Maybe `loggerName` should be optional - it provides a lot of information, but takes up space too. If it's not too difficult now, add a configurable boolean here https://github.com/ElrondNetwork/elrond-go/blob/master/cmd/node/config/prefs.toml, called `LoggerNameInConsole` or something similar.

Once we will write the log integration tool, the `loggerName` will not be as necessary anymore.

logger/consoleFormatter.go
```diff
@@ -45,7 +61,7 @@ func formatArgs(levelColor string, args ...string) string {
 
 	argString := ""
 	for index := 1; index < len(args); index += 2 {
-		argString += fmt.Sprintf("\033[%s%s\033[0m=%s ", levelColor, args[index-1], args[index])
+		argString += fmt.Sprintf("\033[%s%s\033[0m = %s ", levelColor, args[index-1], args[index])
```
camilbancioiu: The original version made "empty" values more visible, because you would expect the value to start immediately after `=`. Would rather have the original.

core/atomic/int64.go
andreibancioiu: Actually, the abstractions in `core/atomic` make the code a bit more readable and less error-prone regarding atomic operations (for example, higher level semantics such as `flag` and `counter` are always better than raw int variables). We had used the abstractions when implementing `TxCache` counters and so on. The rest of the project should use these abstraction to improve code readability.

core/atomic/string.go
andreibancioiu: Same as above. It's important to encapsulate and hide the `casting to string` logic. Improved readability, better semantics, less bugs.

logger/plainFormatter.go
andreibancioiu: Fixed - moved to `constants.go`. Of course, now both `consoleFormatter` and `plainFormatter` have leaked their privates to the whole package :)

logger/consoleFormatter.go
andreibancioiu: Fixed (comment above).

logger/correlation.go
andreibancioiu: Fixed :)

logger/consoleFormatter.go
andreibancioiu: Done.

logger/correlation.go
andreibancioiu: This is true. A compromise. Less agnostic logger, a bit more tailored for our needs. In the future, we can re-design correlation elements as a generic map (without concepts such as epochs / rounds / sub-rounds), but let's see first if / how it improves our bug hunts.

logger/consoleFormatter.go
andreibancioiu: @bogdan-rosianu @iulianpascalau will decide this :)

core/atomic/int64.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "sync/atomic"
+
+// Int64 is a wrapper for atomic operations on int64
+type Int64 struct {
```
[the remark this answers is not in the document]
andreibancioiu: Usage will then be `atomic.AtomicInt64`. I've removed the `atomic` prefix in a past fix after review (for `TxCache`, where the other atomic components - Flag, Counter - were used).

core/atomic/string.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "sync/atomic"
+
+// String is a wrapper for atomic operations on String
+type String struct {
```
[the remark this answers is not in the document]
andreibancioiu: (see above)

logger/correlation.go
andreibancioiu: Added some tests.

logger/commonConverters.go
```diff
@@ -1,23 +1,57 @@
...
 	"strings"
 	"time"
 )
 
-const msgFixedLength = 40
+const bracketsLength = len("[]")
+const loggerNameFixedLength = 20
+const correlationElementsFixedLength = 10
+const messageFixedLength = 40
+const ellipsisString = "..."
```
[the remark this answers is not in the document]
andreibancioiu: Saved.

logger/consoleFormatter.go
andreibancioiu: Made it optional.

logger/consoleFormatter.go
bogdan-rosianu: I suggested this change (actually did this change) for a better experience when selecting a value (double-clicking the value now selects `key=value` but with this change it will select only the `value`).  But this is still debatable, we might wait for other opinions

logger/correlation.go
```diff
@@ -0,0 +1,68 @@
...
+	enabled  bool
+	epoch    atomic.Uint32
+	round    atomic.Int64
+	subRound atomic.String
+}
```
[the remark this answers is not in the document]
andreibancioiu: Prepared it as correlation element. But I've only set it in Node's `main` for now - that is, at start-up.

logger/constants.go
```diff
@@ -0,0 +1,4 @@
+package logger
+
+const formatColoredString = "\033[%s%s\033[0m[%s] %s %s %s %s\n"
+const formatPlainString = "%s[%s] %s %s %s %s\n"
```
camilbancioiu: Add newline at EOF.

logger/correlation.go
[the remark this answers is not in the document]
camilbancioiu: `shardID` will change when the epoch changes, but can't find a good place for updating it at the moment. @iulianpascalau?

logger/constants.go
andreibancioiu: Fixed.

core/atomic/int64.go
iulianpascalau: Ok, then we need those strucs tested.

logger/plainFormatter.go
iulianpascalau: agree, but we do not have to scan later the code file in search of some consts

logger/correlation.go
iulianpascalau: ok

api/logs/logSender.go
```diff
@@ -87,14 +87,19 @@ func (ls *logSender) StartSendingBlocking() {
...
-func (ls *logSender) waitForPatternMessage() error {
+func (ls *logSender) waitForProfile() error {
 	_, message, err := ls.conn.ReadMessage()
 	if err != nil {
 		return err
 	}
 
-	ls.log.Info("websocket log pattern received", "pattern", string(message))
-	err = logger.SetLogLevel(string(message))
+	ls.log.Info("websocket log profile received", "profile", string(message))
```
iulianpascalau: I think we might improve this log level message (string on some particular bytes is too criptic).
Maybe move it on the line 101 and have something like:
```GO
ls.log.Info("websocket log profile received",
"pattern", profile.LogLevelPatterns,
"with correlation", profile.WithCorrelation,
"with logger name", profile.WithLoggerName,
)

Comment (andreibancioiu) [reply api/logs/logSender.go]: Fixed.

File: logger/profile.go [hunk 16]
```diff
@@ -52,3 +53,11 @@ func (profile *Profile) Apply() error {
 	ToggleLoggerName(profile.WithLoggerName)
 	return nil
 }
+
+func (profile *Profile) String() string {
```
iulianpascalau: :+1:

logger/proto/logLineMessage.proto
```diff
@@ -13,4 +13,13 @@ message LogLineMessage{
     repeated string Args = 3;
     int64 Timestamp = 4;
     string LoggerName = 5;
+    LogCorrelationMessage Correlation = 6 [(gogoproto.nullable) = false];
 }
+
+message LogCorrelationMessage{
+    string Shard = 1;
+    uint32 Epoch = 2;
+    int64 Round = 3;
```
iulianpascalau: please ident this

PR #1261, opened 2020-03-20
bugfix/Revert accumulated fees when necessary

sasurobert:


process/block/postprocess/feeHandler.go
```diff
@@ -36,17 +39,34 @@ func (f *feeHandler) GetAccumulatedFees() *big.Int {
...
 // ProcessTransactionFee adds the tx cost to the accumulated amount
-func (f *feeHandler) ProcessTransactionFee(cost *big.Int) {
+func (f *feeHandler) ProcessTransactionFee(cost *big.Int, txHash []byte) {
 	if cost == nil {
 		log.Debug("nil cost in ProcessTransactionFee", "error", process.ErrNilValue.Error())
 		return
 	}
 
 	f.mut.Lock()
+	f.mapHashFee[string(txHash)] = big.NewInt(0).Set(cost)
```
andreibancioiu: Do we need to clone the value or `cost` would have sufficed?

process/block/postprocess/feeHandler.go
```diff
@@ -36,17 +39,34 @@ func (f *feeHandler) GetAccumulatedFees() *big.Int {
...
 
+// RevertFees reverts the accumulated fees for txHashes
+func (f *feeHandler) RevertFees(txHashes [][]byte) {
+	f.mut.Lock()
+	defer f.mut.Unlock()
+
+	for _, txHash := range txHashes {
+		cost, ok := f.mapHashFee[string(txHash)]
+		if !ok {
+			continue
```
andreibancioiu: ~~Not sure - perhaps add a `log.Trace()`?~~

process/block/postprocess/feeHandler.go
```diff
@@ -36,17 +39,34 @@ func (f *feeHandler) GetAccumulatedFees() *big.Int {
 }
 
 // ProcessTransactionFee adds the tx cost to the accumulated amount
-func (f *feeHandler) ProcessTransactionFee(cost *big.Int) {
+func (f *feeHandler) ProcessTransactionFee(cost *big.Int, txHash []byte) {
```
andreibancioiu: `txHash` will be nil for some calls. Do we need to handle this differently?

process/mock/feeAccumulatorStub.go
```diff
@@ -25,7 +33,7 @@ func (f *FeeAccumulatorStub) GetAccumulatedFees() *big.Int {
 }
 
 // ProcessTransactionFee -
-func (f *FeeAccumulatorStub) ProcessTransactionFee(cost *big.Int) {
+func (f *FeeAccumulatorStub) ProcessTransactionFee(cost *big.Int, txHash []byte) {
 	if f.ProcessTransactionFeeCalled != nil {
 		f.ProcessTransactionFeeCalled(cost)
```
SebastianMarian: Apply the same signature to the mock method ProcessTransactionFeeCalled

integrationTests/mock/unsignedTxHandlerMock.go
```diff
@@ -47,7 +55,7 @@ func (ut *UnsignedTxHandlerMock) CleanProcessedUTxs() {
 }
 
 // ProcessTransactionFee -
-func (ut *UnsignedTxHandlerMock) ProcessTransactionFee(cost *big.Int) {
+func (ut *UnsignedTxHandlerMock) ProcessTransactionFee(cost *big.Int, txHash []byte) {
 	if ut.ProcessTransactionFeeCalled == nil {
```
SebastianMarian: Apply the same signature to the mock method ProcessTransactionFeeCalled

process/block/postprocess/feeHandler.go
sasurobert: txHash will not be nil, should not be nil. I added everywhere.

process/smartContract/process.go
```diff
@@ -291,7 +291,7 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
 		return nil
 	}
 
-	sc.txFeeHandler.ProcessTransactionFee(feeForValidators)
+	sc.txFeeHandler.ProcessTransactionFee(feeForValidators, nil)
```
SebastianMarian: Every call of ProcessTransactionFee method is done with nil parameters in whole project Why?

process/mock/feeAccumulatorStub.go
sasurobert: done

integrationTests/mock/unsignedTxHandlerMock.go
sasurobert: done

process/smartContract/process.go
sasurobert: fixed.

process/block/preprocess/transactions.go
```diff
@@ -1076,6 +1074,7 @@ func (txs *transactions) ProcessMiniBlock(miniBlock *block.MiniBlock, haveTime f
 			return processedTxHashes, err
 		}
 
+		processedTxHashes = append(processedTxHashes, miniBlockTxHashes[index])
```
andreibancioiu: Rename `index` to `txHash` (optional).

process/smartContract/process.go
```diff
@@ -216,9 +216,16 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
 		return err
 	}
 
+	var txHash []byte
+	txHash, err = core.CalculateHash(sc.marshalizer, sc.hasher, tx)
```
andreibancioiu: Optional - add TODO to get the hash from `TransactionWrapper` (or `TransactionHandler.GetHash()` - in the future, when available).

Optional - why not using `computeTransactionHash()` already defined?

process/smartContract/process.go
```diff
@@ -441,14 +445,20 @@ func (sc *scProcessor) DeploySmartContract(
 		return process.ErrWrongTransaction
 	}
 
+	txHash, err := core.CalculateHash(sc.marshalizer, sc.hasher, tx)
```
andreibancioiu: Optional - why not using `computeTransactionHash()` already defined?

process/smartContract/process.go
```diff
@@ -1052,9 +1054,15 @@ func (sc *scProcessor) ProcessSmartContractResult(scr *smartContractResult.Smart
 	}
 
 	var err error
+	txHash, err := core.CalculateHash(sc.marshalizer, sc.hasher, scr)
```
andreibancioiu: Optional - as above (`computeTransactionHash`).

process/block/preprocess/transactions.go
sasurobert: it is actually the index. that is a list of hashes.

process/smartContract/process.go
sasurobert: discussed with Iulian - we can have a private field in the transaction / scr and other structures with Set and GetHash, not possible right now because of the way proto is integrated.

process/block/preprocess/transactions.go
SebastianMarian: You should move this line back, and use another variable for your scope, as if it would be an error in ProcessTransaction at any point, then on defer, the gas consumed would be reverted only for the txs which has been  processed and for the others the gas would remain not reverted.  Actually computeGasConsumed in this point has been executed for all of them.

process/transaction/shardProcess.go
```diff
@@ -285,12 +280,17 @@ func (txProc *txProcessor) processMoveBalance(
...
+		return err
+	}
+
+	err = txProc.createReceiptWithReturnedGas(txHash, tx, acntSrc)
 	if err != nil {
 		return err
 	}
 
-	txProc.txFeeHandler.ProcessTransactionFee(txFee)
+	txProc.txFeeHandler.ProcessTransactionFee(txFee, nil)
```
SebastianMarian: Here should be nil?

process/smartContract/process.go
sasurobert: deleted that function

process/smartContract/process.go
sasurobert: deleted that function

process/block/preprocess/transactions.go
sasurobert: done.

process/transaction/shardProcess.go
sasurobert: done.

PR #1262, opened 2020-03-20
EN-5829: read epoch from storage if storage pruning is disabled

bogdan-rosianu:


storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
...
+	bootstrapData, err := bootStorer.Get(highestRound)
+	if err != nil {
+		return 0, err
+	}
+
+	log.Debug("found epoch in BootstrapData", "epoch", bootstrapData.LastHeader.Epoch)
+	return bootstrapData.LastHeader.Epoch, nil
+}
+
+func getShardFromDirectory(path string, defaultShardString string) (string, error) {
```
AdoAdoAdo: what happens when the node has disabled pruning, but is a validator and gets shuffled between shards/
Would it create for each shard a new directory?
Then after a number of epochs he will have a directory for each shard. 
This implementation doesn't seem to take this into account, and will return the shardId for the first path which has a chance of only 1/(nbShards+meta) of being correct.

I think you need to search in all shard dirs for the highest round and then take the epoch from the db with highest round.

storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
...
+
+func getLastEpochFromStorage(
+	config config.Config,
+	marshalizer marshal.Marshalizer,
+	parentDir string,
+	defaultEpochString string,
+	defaultShardString string,
+) (uint32, error) {
+	persisterFactory := NewPersisterFactory(config.BootstrapStorage.DB)
+	pathWithoutShard := parentDir + string(os.PathSeparator) + defaultEpochString + "_0"
```
AdoAdoAdo: why not filepath.join

storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
...
+	shardIdStr, err := getShardFromDirectory(pathWithoutShard, defaultShardString)
+	if err != nil {
+		return 0, err
+	}
+
+	persisterPath := pathWithoutShard + string(os.PathSeparator) + defaultShardString
```
AdoAdoAdo: can you use filepath.join?

cmd/node/main.go
```diff
@@ -537,10 +537,13 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	var currentEpoch uint32
 	var errNotCritical error
 	currentEpoch, errNotCritical = storageFactory.FindLastEpochFromStorage(
+		*generalConfig,
+		&marshal.GogoProtoMarshalizer{}, // TODO: in another branch, this is not hardcoded because coreComponent is initialized before
```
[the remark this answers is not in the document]
iulianpascalau: Consider initiating core component before this call.

storage/factory/common.go
```diff
@@ -90,7 +97,17 @@ func FindLastEpochFromStorage(
 		epochDirs = append(epochDirs, file.Name())
 	}
 
-	return getLastEpochFromDirNames(epochDirs)
+	lastEpoch, err := getLastEpochFromDirNames(epochDirs)
+	if err != nil {
+		return 0, err
+	}
+
+	if lastEpoch == 0 {
```
iulianpascalau: what happens in a setting with storage pruning on and the node was shut down in epoch 0?
Why not injecting the setting that tells you if storage pruning is on or off instead of making assumptions?

storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
 
 	return epochsInDirName[0], nil
 }
+
+func getLastEpochFromStorage(
```
iulianpascalau: this function does 3 things: computes a file path, instantiate a storer (cacher + persister) and search some data inside.

storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
...
+	return bootstrapData.LastHeader.Epoch, nil
+}
+
+func getShardFromDirectory(path string, defaultShardString string) (string, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+
+	files, err := f.Readdir(-1)
```
iulianpascalau: magic number

storage/factory/common.go
```diff
@@ -117,3 +134,77 @@ func getLastEpochFromDirNames(epochDirs []string) (uint32, error) {
...
+		fileName := file.Name()
+		splitSlice := strings.Split(fileName, defaultShardString)
+		if len(splitSlice) < 2 {
+			continue
+		}
+
+		return splitSlice[1], nil
+	}
+
+	return "", errors.New("no shard info found in storage directories structure")
```
iulianpascalau: might extract this in errors.go

storage/factory/common.go
iulianpascalau: tests?

storage/factory/common.go
bogdan-rosianu: You can't tell this because the config may have changed since last run. So you don't know if the last time the node was online it had storage pruning off. If the node shuts down and goes back online in epoch 0, the code will be executed. Anyway, also in this PR I have to return also the last round, so each time you'll have to open the persister

storage/factory/common.go
bogdan-rosianu: done

storage/factory/common.go
bogdan-rosianu: pretty hard to do that, I have to agree. However, this is a temporary solution until we create a static component which holds all the needed data

bogdan-rosianu: extracted a function

cmd/node/main.go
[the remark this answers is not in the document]

storage/factory/common.go
bogdan-rosianu: done. Now the function searches inside all persister and returns the highest value. + the last round

storage/factory/common.go
bogdan-rosianu: done

storage/factory/common.go
bogdan-rosianu: done

storage/factory/common.go
iulianpascalau: maybe `allFiles`

storage/factory/common.go
iulianpascalau: I know, in order for the tests to be written in an easier way, the file/directory specific functions can be extracted in functions that can be mocked in tests. In this way we can reach as much as 100% code coverage, effectively testing each branch in each situation. Can be done in a future PR.

PR #1263, opened 2020-03-20
Adaptive block size throttle

SebastianMarian:
Main part of this PR contains the removed code from PR: https://github.com/ElrondNetwork/elrond-go/pull/1241 in which has been deleted throttle package and one mock file. The main logic of the restored package has not been changed but only adapted to the current scope by refactoring some methods/parameters names.


cmd/node/config/config.toml
```diff
@@ -362,3 +362,7 @@
     CheckpointRoundsModulus = 100
     AccountsStatePruningEnabled = true
     PeerStatePruningEnabled = true
+
+[BlockSizeThrottleConfig]
+    MinSizeInBytes = 104857 # 10% from 1MB
```
iulianpascalau: :+1:

process/block/preprocess/interfaces.go
```diff
@@ -25,3 +25,10 @@ type BlockSizeComputationHandler interface {
 	IsMaxBlockSizeReached(numNewMiniBlocks int, numNewTxs int) bool
 	IsInterfaceNil() bool
 }
+
+// BlockSizeThrottler defines the functionality of adapting the node to the network speed/latency when it should send a
+// block to its peers which should be received in a limited time frame
+type BlockSizeThrottler interface {
```
iulianpascalau: :+1:

process/throttle/block.go
```diff
@@ -0,0 +1,168 @@
...
+			bst.statistics[i].succeed = true
+			break
+		}
+	}
+	bst.mutThrottler.Unlock()
+}
+
+// ComputeCurrentMaxSize computes the current max size in bytes which could be used in one block, taking into consideration the previous results
+func (bst *blockSizeThrottle) ComputeCurrentMaxSize() {
+	//TODO: This is the first basic implementation, which will adapt the max size which could be used in one block,
```
iulianpascalau: Migh remove the TODO part, as this is the second iteration of the throttler :)

process/block/pendingMb/pendingMiniBlocks.go
```diff
@@ -29,10 +29,7 @@ func (p *pendingMiniBlocks) getAllCrossShardMiniBlocksHashes(metaBlock *block.Me
 	crossShardMiniBlocks := make(map[string]uint32)
 
 	for _, mbHeader := range metaBlock.MiniBlockHeaders {
-		if mbHeader.SenderShardID != core.MetachainShardId && mbHeader.ReceiverShardID == core.MetachainShardId {
-			continue
-		}
-		if mbHeader.SenderShardID == core.MetachainShardId && mbHeader.ReceiverShardID == core.AllShardId {
+		if !shouldConsiderCrossShardMiniBlock(mbHeader.SenderShardID, mbHeader.ReceiverShardID) {
```
iulianpascalau: these conditions are not equal to the old variant. Is this intended?

SebastianMarian: Actually these conditions should been applied in the same way in both iterations. The basic idea with method getAllCrossShardMiniBlocksHashes is that it extracts from one metablock all the cross shard miniblocks (dst != rcv no matter which is dst or rcv) which would appear 2 times in a while. First time when you see it you would add it as pending miniblock and the second time you would removed it from pending. So, there are some exception when a cross mini block would not appear 2 times. and these cases are: 1) system smart contract (miniblocks from shard to meta) and 2) PeerBlock miniblocks (from meta to AllShardID)

iulianpascalau: ok

PR #1271, opened 2020-03-23
epoch trigger - sort metaheaders according to nonce.

sasurobert:
make it easier to search in local pool.
epoch trigger - subscribe to notifier
shardBlock - do not delete in add
metaBlock - do not delete in add

### DIFF
--- a/epochStart/shardchain/trigger.go
+++ b/epochStart/shardchain/trigger.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"sort"
 	"sync"
 
 	"github.com/ElrondNetwork/elrond-go/core"
@@ -80,6 +81,11 @@ type trigger struct {
 	isEpochStart        bool
 
 	validatorInfoProcessor process.ValidatorInfoProcessorHandler
+}
+
+type metaInfo struct {
+	meta *block.MetaBlock
+	hash string
 }
 
 // NewEpochStartTrigger creates a trigger to signal start of epoch
@@ -171,6 +177,8 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
 		validatorInfoProcessor:      args.ValidatorInfoProcessor,
 	}
 
+	newTrigger.headersPool.RegisterHandler(newTrigger.receivedMetaBlock)
+
 	err := newTrigger.saveState(newTrigger.triggerStateKey)
 	if err != nil {
 		return nil, err
@@ -261,16 +269,16 @@ func (t *trigger) setLowerEpochFinalityAttestingRoundIfNeeded(metaHdr *block.Met
 }
 
 // ReceivedHeader saves the header into pool to verify if end-of-epoch conditions are fulfilled only called with validated metablocks
-func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
+func (t *trigger) receivedMetaBlock(headerHandler data.HeaderHandler, metaBlockHash []byte) {
 	t.mutTrigger.Lock()
 	defer t.mutTrigger.Unlock()
 
-	metaHdr, ok := header.(*block.MetaBlock)
+	metaHdr, ok := headerHandler.(*block.MetaBlock)
 	if !ok {
 		return
 	}
 
-	if t.isEpochStart && header.GetEpoch() == t.epoch {
+	if t.isEpochStart && headerHandler.GetEpoch() == t.epoch {
 		t.setLowerEpochFinalityAttestingRoundIfNeeded(metaHdr)
 		return
 	}
@@ -283,19 +291,14 @@ func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
 		return
 	}
 
-	hdrHash, err := core.CalculateHash(t.marshalizer, t.hasher, metaHdr)
-	if err != nil {
+	if _, ok = t.mapHashHdr[string(metaBlockHash)]; ok {
+		return
+	}
+	if _, ok = t.mapEpochStartHdrs[string(metaBlockHash)]; ok {
 		return
 	}
 
-	if _, ok = t.mapHashHdr[string(hdrHash)]; ok {
-		return
-	}
-	if _, ok = t.mapEpochStartHdrs[string(hdrHash)]; ok {
-		return
-	}
-
-	t.updateTriggerFromMeta(metaHdr, hdrHash)
+	t.updateTriggerFromMeta(metaHdr, metaBlockHash)
 }
 
 // call only if mutex is locked before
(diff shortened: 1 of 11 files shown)


epochStart/shardchain/trigger.go
```diff
@@ -283,42 +291,49 @@ func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
...
 
+	t.mapHashHdr[string(hdrHash)] = metaHdr
+	t.mapNonceHashes[metaHdr.Nonce] = append(t.mapNonceHashes[metaHdr.Nonce], string(hdrHash))
+
+	sortedMetas := make([]metaInfo, 0, len(t.mapEpochStartHdrs))
 	for hash, meta := range t.mapEpochStartHdrs {
-		canActivateEpochStart, finalityAttestingRound := t.checkIfTriggerCanBeActivated(hash, meta)
-		if canActivateEpochStart && t.epoch < meta.Epoch {
-			t.epoch = meta.Epoch
+		metaInfo := metaInfo{
```
SebastianMarian: metaInfo := &metaInfo{

epochStart/shardchain/trigger.go
```diff
@@ -283,42 +291,49 @@ func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
...
-	} else {
-		t.mapHashHdr[string(hdrHash)] = metaHdr
-		t.mapNonceHashes[metaHdr.Nonce] = append(t.mapNonceHashes[metaHdr.Nonce], string(hdrHash))
 	}
 
+	t.mapHashHdr[string(hdrHash)] = metaHdr
+	t.mapNonceHashes[metaHdr.Nonce] = append(t.mapNonceHashes[metaHdr.Nonce], string(hdrHash))
+
+	sortedMetas := make([]metaInfo, 0, len(t.mapEpochStartHdrs))
 	for hash, meta := range t.mapEpochStartHdrs {
```
SebastianMarian: hash, hdr

process/block/metablock.go
```diff
@@ -1506,6 +1496,10 @@ func (mp *metaProcessor) receivedShardHeader(headerHandler data.HeaderHandler, s
 		return
 	}
 
+	if !mp.blockTracker.ShouldAddHeader(shardHeader) {
```
SebastianMarian: Actually you can remove this condition at all as it was used only to delete the out of range received headers. The conditions which protect to not request miniblocks in advance is right above.

process/block/shardblock.go
```diff
@@ -1359,6 +1348,10 @@ func (sp *shardProcessor) receivedMetaBlock(headerHandler data.HeaderHandler, me
 		return
 	}
 
+	if !sp.blockTracker.ShouldAddHeader(metaBlock) {
```
SebastianMarian: You can remove this check at all. The same situation as in metablock.go above

epochStart/shardchain/trigger.go
```diff
@@ -261,16 +269,16 @@ func (t *trigger) setLowerEpochFinalityAttestingRoundIfNeeded(metaHdr *block.Met
 }
 
 // ReceivedHeader saves the header into pool to verify if end-of-epoch conditions are fulfilled only called with validated metablocks
```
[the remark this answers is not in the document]
sasurobert: done.

epochStart/shardchain/trigger.go
```diff
@@ -283,42 +291,49 @@ func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
...
-
-	} else {
-		t.mapHashHdr[string(hdrHash)] = metaHdr
-		t.mapNonceHashes[metaHdr.Nonce] = append(t.mapNonceHashes[metaHdr.Nonce], string(hdrHash))
 	}
 
+	t.mapHashHdr[string(hdrHash)] = metaHdr
+	t.mapNonceHashes[metaHdr.Nonce] = append(t.mapNonceHashes[metaHdr.Nonce], string(hdrHash))
+
+	sortedMetas := make([]metaInfo, 0, len(t.mapEpochStartHdrs))
```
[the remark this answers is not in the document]

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go

process/block/metablock.go
sasurobert: done

process/block/shardblock.go
sasurobert: done

epochStart/shardchain/trigger.go
```diff
@@ -283,42 +292,49 @@ func (t *trigger) ReceivedHeader(header data.HeaderHandler) {
...
 
-	for hash, meta := range t.mapEpochStartHdrs {
-		canActivateEpochStart, finalityAttestingRound := t.checkIfTriggerCanBeActivated(hash, meta)
-		if canActivateEpochStart && t.epoch < meta.Epoch {
-			t.epoch = meta.Epoch
+	sort.Slice(sortedMetaInfo, func(i, j int) bool {
+		return sortedMetaInfo[i].hdr.Epoch < sortedMetaInfo[j].hdr.Epoch
+	})
+
+	for _, metaInfo := range sortedMetaInfo {
```
ccorcoveanu: can you rename one of the `metaInfo` so the warning with shadow declaration is removed?

process/block/shardblock.go
```diff
@@ -1349,7 +1339,6 @@ func (sp *shardProcessor) receivedMetaBlock(headerHandler data.HeaderHandler, me
 		return
 	}
 
-	sp.epochStartTrigger.ReceivedHeader(metaBlock)
 	if sp.epochStartTrigger.IsEpochStart() {
```
SebastianMarian: Here we can have a deadlock, as this channel "chRcvEpochStart" is not buffered and it is consumed only in the method requestEpochStartInfo called from ProcessBlock. This means that if node will not reach to that point all the metablocks received meantime will be stuck in this point, as now you will not call the sp.epochStartTrigger.ReceivedHeader(metaBlock) method synchronous, from here, anymore.
I will suggest to remove this channel at all and also these 3 lines from here, and to change the method requestEpochStartInfo like this:

func (sp *shardProcessor) requestEpochStartInfo(header *block.Header, haveTime func() time.Duration) error {
	haveMissingMetaHeaders := header.IsStartOfEpochBlock() && !sp.epochStartTrigger.IsEpochStart()
	if !haveMissingMetaHeaders {
		return nil
	}

	for {
		time.Sleep(time.Millisecond)
		
		if haveTime() < 0 {
			break
		}
		
		if sp.epochStartTrigger.IsEpochStart() {
			return nil
		}
	}

	return process.ErrTimeIsOut
}

iulianpascalau: As an alternative, as to not have a blocking write, @sasurobert could have written something like:
```GO
select{
    case sp.chRcvEpochStart <- true:
    default:
}
```

PR #1273, opened 2020-03-23
EN-5897: shuffled out handler and notifier

bogdan-rosianu:


cmd/node/main.go
```diff
@@ -1192,6 +1202,28 @@ func createNodesCoordinator(
 		return nil, err
 	}
 
+	thresholdEpochDuration := 0.25
```
sasurobert: magic number - add to config.

cmd/node/main.go
```diff
@@ -1192,6 +1202,28 @@ func createNodesCoordinator(
...
+	maxDurationBeforeStopProcess := int64(nodesConfig.RoundDuration) * roundsPerEpoch
+	maxDurationBeforeStopProcess = int64(thresholdEpochDuration * float64(maxDurationBeforeStopProcess))
+	intRandomizer := &random.ConcurrentSafeIntRandomizer{}
+	randDurationBeforeStop, err := intRandomizer.Intn(int(maxDurationBeforeStopProcess))
+	if err != nil {
+		return nil, err
+	}
+	endOfProcessingHandler := func() error {
+		go func() {
```
sasurobert: no need for go func

cmd/node/main.go
sasurobert: also this threshold shoud not be considered if epoch is short. define a constant for short.

sharding/shuffledOutTrigger.go
```diff
@@ -0,0 +1,63 @@
...
+type shuffledOutTrigger struct {
+	ownPubKey         []byte
+	currentShardID    uint32
+	handlers          []func(newShardID uint32)
+	mutHandlers       sync.RWMutex
+	endProcessHandler func() error
+}
+
+// NewShuffledOutTrigger returns a new instance of shuffledOutTrigger
+func NewShuffledOutTrigger(ownPubKey []byte, currentShardID uint32, endProcessHandler func() error) (*shuffledOutTrigger, error) {
```
sasurobert: +1 nice and small.

cmd/node/main.go
bogdan-rosianu: added to config. Won't add that constant for short epochs, we can change the threshold in config

cmd/node/main.go
bogdan-rosianu: it is better this way. Otherwise, imagine when you start the node in a wrong shard. the channel will be written in it. but the checks for that channel are done only at the end of startNode() method in main. Also the print would be placed in a wrong place. 
Also, if I wouldn't use a goroutine, the function would be blocking. So what need for a random wait time before node stop ? if the node doesn't do anything meanwhile.

cmd/node/main.go
```diff
@@ -949,8 +955,11 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	log.Info("application is now running")
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
-	<-sigs
-	log.Info("terminating at user's signal...")
+	select {
+	case <-sigs:
+		log.Info("terminating at user's signal...")
+	case <-chanStopNodeProcess:
```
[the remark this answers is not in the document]
bogdan-rosianu: added a log which tells about terminating at internal stop signal, because I injected that channel inside the node.go and we can use it in more places.

PR #1276, opened 2020-03-23
Fix-misleading-prints-for-headers-pool-info

SebastianMarian:
* Fixed missleading prints for headers pool info
* Fixed value of maxNumHeadersToKeepPerShard from blockTracker
* Fixed marshal problems for empty slices instead nil in bootstrapStorer when protobuf is used
* Removed unnecessary code


process/block/metablock.go
```diff
@@ -1140,17 +1150,33 @@ func (mp *metaProcessor) CommitBlock(
...
+		log.Trace("pools info",
+			"shard", shardID,
+			"num headers", mp.dataPool.Headers().GetNumHeaders(shardID))
+	}
+
+	log.Trace("pools info",
+		"shard", core.MetachainShardId,
+		"num headers", mp.dataPool.Headers().GetNumHeaders(core.MetachainShardId))
+
+	numShardsToKeepHeaders := int(mp.shardCoordinator.NumberOfShards()) + 1
```
andreibancioiu: Please add a comment why +1 (metachain)?

process/block/metablock.go
```diff
@@ -1140,17 +1150,33 @@ func (mp *metaProcessor) CommitBlock(
...
+	log.Debug("pools info",
+		"total headers", mp.dataPool.Headers().Len(),
+		"headers pool capacity", mp.dataPool.Headers().MaxSize() * numShardsToKeepHeaders,
```
andreibancioiu: Optional: extract new variable, `capacity`.

process/block/shardblock.go
```diff
@@ -904,19 +904,33 @@ func (sp *shardProcessor) CommitBlock(
...
+func (sp *shardProcessor) displayPoolsInfo() {
+	log.Trace("pools info",
+		"shard", sp.shardCoordinator.SelfId(),
+		"num headers", sp.dataPool.Headers().GetNumHeaders(sp.shardCoordinator.SelfId()))
+
+	log.Trace("pools info",
+		"shard", core.MetachainShardId,
+		"num headers", sp.dataPool.Headers().GetNumHeaders(core.MetachainShardId))
+
+	numShardsToKeepHeaders := 2
```
andreibancioiu: Comment why 2? Self + metachain?

process/block/metablock.go
SebastianMarian: Yes. Added comment

process/block/shardblock.go
SebastianMarian: Yes. Added comment

process/block/metablock.go

process/block/metablock.go
```diff
@@ -1879,9 +1907,5 @@ func (mp *metaProcessor) getPendingMiniBlocks() []bootstrapStorage.PendingMiniBl
 		}
 	}
```
AdoAdoAdo: why is this removed? 
are we sure that we can never return empty pendingMiniBocksInfo?
if these are afterwards serialized, there might be some issues

PR #1277, opened 2020-03-23
Fix observer shardid on restart

AdoAdoAdo:


sharding/indexHashedNodesCoordinator.go
```diff
@@ -614,13 +616,13 @@ func (ihgs *indexHashedNodesCoordinator) GetConsensusWhitelistedNodes(
 
 func (ihgs *indexHashedNodesCoordinator) computeShardForSelfPublicKey(nodesConfig *epochNodesConfig) uint32 {
 	pubKey := ihgs.selfPubKey
-	selfShard := uint32(0)
+	selfShard := ihgs.shardIDAsObserver
```
iulianpascalau: :+1: how many times I've read this function because of this.

PR #1281, opened 2020-03-24
Go dump route

iulianpascalau:
added `node/godump` api route able to send through logger go routines stack trace.


api/node/routes.go
```diff
@@ -51,9 +55,21 @@ func Routes(router *gin.RouterGroup) {
 	router.GET("/statistics", Statistics)
 	router.GET("/status", StatusMetrics)
 	router.GET("/p2pstatus", P2pStatusMetrics)
+	router.GET("/godump", GoRoutinesDump)
```
andreibancioiu: Perhaps rename to `/goroutines-dump`?

api/node/routes.go
```diff
@@ -51,9 +55,21 @@ func Routes(router *gin.RouterGroup) {
...
 	// placeholder for custom routes
 }
 
+// GoRoutinesDump will dump go routines info
+func GoRoutinesDump(c *gin.Context) {
+	log.Warn("go routines dump")
+
+	buff := make([]byte, 1<<27)
```
andreibancioiu: :)

`buffSize := 128 * 1024 * 1024`, for improved readability? Make it **smaller**? 128 MB is a bit much to dump in the logger anyway.

api/node/routes.go
```diff
@@ -51,9 +55,21 @@ func Routes(router *gin.RouterGroup) {
...
+	num := runtime.Stack(buff, true)
+	log.Warn(string(buff[:num]))
+
+	c.JSON(http.StatusOK, gin.H{"go routines dump": true})
```
andreibancioiu: Why not respond with the dump?

iulianpascalau: will be overkill for the http server

api/node/routes.go
iulianpascalau: We had a previous situation where we had thousands of go routines. Having a large buffer will allow us to tell which go routines repeat frequently (that can be a sign of a go routine leakage). This is just the maximum buffer, when logging I do a slice. Will change to 128 * 1024 * 1024 for readability concerns.

api/node/routes.go
iulianpascalau: to much to type and memorize. I would like to keep it as simple as possible

iulianpascalau:
needs more refactoring

PR #1296, opened 2020-03-25
bugFix/ protect transaction in case of metachain.

sasurobert:
MetaChain can validate and process only smart contract transactions. However, there was no check that only these types of transactions will reach the metachain. Added check when processing these txs at the sender that it must be with txData. Added also a fallback mechanism to metachain, that if he cannot process a transaction, he will send the txValue back - consumes all gas.


process/transaction/shardProcess.go
```diff
@@ -262,6 +282,11 @@ func (txProc *txProcessor) processMoveBalance(
 		return err
 	}
 
+	err = txProc.checkIfValidTxToMetaChain(tx, acntSrc, adrDst)
```
SebastianMarian: I would rename this method to checkTxValidity as it is called with any kind of tx no matter if the destination is meta or not.

process/transaction/shardProcess.go
```diff
@@ -250,6 +248,28 @@ func (txProc *txProcessor) processTxFee(
...
+) error {
+
+	destShardId := txProc.shardCoordinator.ComputeId(adrDst)
+	if destShardId != core.MetachainShardId {
+		return nil
+	}
+
+	// it is not allowed to send transactions to metachain if those are not of type smart contract
+	if len(tx.GetData()) == 0 {
+		receiptErr := txProc.executingFailedTransaction(tx, acntSnd, process.ErrInvalidMetaTransaction)
```
SebastianMarian: Why not directly: return txProc.executingFailedTransaction(tx, acntSnd, process.ErrInvalidMetaTransaction) ?

process/transaction/metaProcess.go
```diff
@@ -88,12 +95,23 @@ func (txProc *metaTxProcessor) ProcessTransaction(tx *transaction.Transaction) e
 
 	switch txType {
 	case process.SCDeployment:
-		return txProc.processSCDeployment(tx, adrSrc)
+		err = txProc.processSCDeployment(tx, adrSrc)
```
SebastianMarian: Why not directly: return txProc.processSCDeployment(tx, adrSrc) ?

process/transaction/metaProcess.go
```diff
@@ -88,12 +95,23 @@ func (txProc *metaTxProcessor) ProcessTransaction(tx *transaction.Transaction) e
...
+		return err
 	case process.SCInvoking:
-		return txProc.processSCInvoking(tx, adrSrc, adrDst)
+		err = txProc.processSCInvoking(tx, adrSrc, adrDst)
```
SebastianMarian: Why not directly: return txProc.processSCInvoking(tx, adrSrc, adrDst) ?

process/transaction/shardProcess.go
sasurobert: there is already a checkTxValidity method - this is checking only in case of metachain calls.

process/transaction/shardProcess.go
sasurobert: done.

process/transaction/metaProcess.go
sasurobert: done

process/transaction/metaProcess.go
sasurobert: done

process/transaction/shardProcess.go
SebastianMarian: Ok. Anyway I think that a better name could be found like: checkTxValidityIfDestIsMetachain or something like this

PR #1301, opened 2020-03-25
Fix antiflood num messages counting on multidata interceptor

iulianpascalau:
- reworked prints in peerShardMapper.go
- added protection on multiDataInterceptor.go against large data entities inside a message
- small refactorization inside antiflood package
- removed one dependency between consensus -> main area
- changed antiflood config values


process/throttle/antiflood/floodPreventers/topicFloodPreventer.go
```diff
@@ -57,16 +57,14 @@ func (tfp *topicFloodPreventer) Accumulate(identifier string, topic string) bool
...
 
 	_, ok = tfp.counterMap[topic][identifier]
-	if !ok {
-		tfp.counterMap[topic][identifier] = 1
-		return true
-	}
-
-	// if this was already in the map, just increment it
-	tfp.counterMap[topic][identifier]++
+	tfp.counterMap[topic][identifier] += numMessages
```
ccorcoveanu: This may be a good thing that we increment before the check, but just want to make sure this is the intended behaviour :). Do we want to process a maximum number of messages for `tfp.counterMap[topic][identifier]` or be very! restrictive and punish if the limit is exceeded in a single message? If a peer (for some reason) first sends a message with bulk data that is bigger than the limit, that message will not be processed and no other (single or small sized) messages will be processed until this counter is reset

process/throttle/antiflood/p2pAntiflood.go
```diff
@@ -75,22 +75,22 @@ func (af *p2pAntiflood) CanProcessMessage(message p2p.MessageP2P, fromConnectedP
 	return nil
 }
 
-// CanProcessMessageOnTopic signals if a p2p message can be processed or not for a given topic
-func (af *p2pAntiflood) CanProcessMessageOnTopic(peer p2p.PeerID, topic string) error {
+// CanProcessMessagesOnTopic signals if a p2p message can be processed or not for a given topic
+func (af *p2pAntiflood) CanProcessMessagesOnTopic(peer p2p.PeerID, topic string, numMessages uint32) error {
```
ccorcoveanu: If this is not returning a bool and it has side effects (it actually increments some state) could we name it something else?

process/interceptors/multiDataInterceptor.go
```diff
@@ -78,11 +78,17 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
...
 	}
 	multiDataBuff := b.Data
-	if len(multiDataBuff) == 0 {
+	lenMultiData := len(multiDataBuff)
+	if lenMultiData == 0 {
 		mdi.throttler.EndProcessing()
 		return process.ErrNoDataInMessage
 	}
 
+	err = mdi.antifloodHandler.CanProcessMessagesOnTopic(fromConnectedPeer, mdi.topic, uint32(lenMultiData))
```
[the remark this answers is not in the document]
iulianpascalau: makes sense: the only way I see it is block body interceptor where we can get more than one data. The single data interceptor already calls preProcessMesage that calls CanProcessMessageOnTopic . Good catch!

process/throttle/antiflood/floodPreventers/topicFloodPreventer.go
iulianpascalau: It was intended as the peer should be "punished" by not disseminating it's large message (that can be exploited otherwise). A single transaction message, in the current implementation, is about 128kB large. That message will hold about 1000 messages (so under the current limit of 4000). Setting the limit to low will break any basic system test.

process/throttle/antiflood/p2pAntiflood.go
iulianpascalau: Will think about a better name

consensus/spos/sposFactory/sposFactory_test.go
```diff
@@ -3,9 +3,9 @@ package sposFactory_test
 import (
 	"testing"
 
+	"github.com/ElrondNetwork/elrond-go/consensus"
 	"github.com/ElrondNetwork/elrond-go/core"
```
sasurobert: goimports

process/interceptors/multiDataInterceptor.go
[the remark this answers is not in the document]
iulianpascalau: Block body interceptor/resolvers are in a refactoring process. Will be automatically solved as the new miniblock interceptor uses the multi data interceptor

process/interface.go
```diff
@@ -634,16 +634,16 @@ type BlockTracker interface {
 // FloodPreventer defines the behavior of a component that is able to signal that too many events occurred
 // on a provided identifier between Reset calls
 type FloodPreventer interface {
-	AccumulateGlobal(identifier string, size uint64) bool
-	Accumulate(identifier string, size uint64) bool
+	AccumulateGlobal(identifier string, size uint64) error
+	Accumulate(identifier string, size uint64) error
```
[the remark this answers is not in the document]
iulianpascalau: done

process/throttle/antiflood/p2pAntiflood.go
iulianpascalau: Thought of it, maybe ReceivedMessage might be a better match but it does not tell much as what it does. CanProcessMessage actually tells that the message should be processed or not. Agree it does change it's internal state but that is why it was constructed, no way to circumvent it. It has to decide if the message should be processed or not based on some state (that might change after this call). @sasurobert ?

consensus/spos/sposFactory/sposFactory_test.go
iulianpascalau: done

process/interceptors/multiDataInterceptor.go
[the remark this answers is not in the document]
ccorcoveanu: 👍

process/throttle/antiflood/p2pAntiflood.go
ccorcoveanu: This is like a part 2 of my earlier question regarding incrementing than doing the check :)
For example, I would have seen it like:
```go
func CanProcessMessage(uint size) bool {
  if currentState + size > limit {
    return false
  }
  return true
}

if CanProcessMessage(someSize) {
  increaseCurrentState(someSize)
  processMessage()
}
```

This way if the limit is 10 messages, I already sent 8, then I have 2 requests, one with 3 and one with 2 messages, the one with 2 will be processed. In the current way, neither will be processed, because `CanProcessMessage` will increase the state to 11, so until a state reset the peer is blocked.

But it's ok like this also, I just wanted to write a follow up to better explain what I meant 👍

iulianpascalau: Totally agree that the functionality can be split, somehow, because the function does 2 things, but how our project evolves, someone could just called CanProcessMessage without calling increaseCurrentState resulting in an ineffective mechanism that would have been hard to debug. So, at the expense of SRP, the functions does a checking and a state change.

cmd/node/config/config.toml
```diff
@@ -249,24 +249,26 @@
 [Antiflood]
     Enabled = true
     NumConcurrentResolverJobs = 50
-    PeerMaxMessagesPerSecond = 75
-    PeerMaxTotalSizePerSecond = 2097152
-    MaxMessagesPerSecond = 400
-    MaxTotalSizePerSecond = 4194304
+    [Antiflood.NetworkMaxInput]
```
ccorcoveanu, line 263: Nice split :)

iulianpascalau: Yeah, became a mess there, split in more sub-settings as to be more coherent.

node/node.go
```diff
@@ -647,10 +650,41 @@ func (n *Node) sendFromTxAccumulator() {
 			txs = append(txs, tx)
 		}
 
+		atomic.AddUint32(&n.txSentCounter, uint32(len(txs)))
+
 		n.sendBulkTransactions(txs)
 	}
 }
 
+func (n *Node) printTxSentCounter(numSeconds int, waitTime time.Duration) {
```
ccorcoveanu, line 660: Could you please document this function? I know it is not exported but it's really hard to figure out it's purpose

iulianpascalau: added description + TODO, the thing is not testable (yet)

PR #1306, opened 2020-03-26
fixed counter for termui

bogdan-rosianu:


process/block/metrics.go
```diff
@@ -125,8 +125,8 @@ func incrementCountAcceptedBlocks(
 
 	bitMap := header.GetPubKeysBitmap()
 
-	indexInBitmap := bitMap[myIndex/8]&(1<<uint8(myIndex%8)) == 1
```
iulianpascalau: could have worked also if the condition would have been `> 0` :)

bogdan-rosianu: done.. with `!= 0`

PR #1309, opened 2020-03-26
replaced github.com/cornelk/hashmap implementation with an internal, …

iulianpascalau:
replaced github.com/cornelk/hashmap implementation with an internal, mutex protected map implementation

### DIFF
--- a/core/container/mutexMap_test.go
+++ b/core/container/mutexMap_test.go
@@ -1,0 +1,265 @@
+package container
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewMutexMap(t *testing.T) {
+	t.Parallel()
+
+	mm := NewMutexMap()
+
+	assert.NotNil(t, mm)
+}
+
+//------- Get
+
+func TestMutexMap_GetAndSet(t *testing.T) {
+	t.Parallel()
+
+	mm := NewMutexMap()
+
+	val, ok := mm.Get("inexistent key")
+	assert.Nil(t, val)
+	assert.False(t, ok)
+
+	key := "existent key"
+	valToStore := 56
+	mm.Set(key, valToStore)
+
+	val, ok = mm.Get(key)
+	assert.Equal(t, valToStore, val)
+	assert.True(t, ok)
+}
+
+func TestMutexMap_GetConcurrent(t *testing.T) {
+	t.Parallel()
+
+	mm := NewMutexMap()
+	key := "existent key"
+	valToStore := 56
+	mm.Set(key, valToStore)
+
+	numIterations := 100
+	wg := &sync.WaitGroup{}
+	wg.Add(numIterations)
+	for i := 0; i < numIterations; i++ {
+		go func() {
+			val, ok := mm.Get(key)
+			assert.Equal(t, valToStore, val)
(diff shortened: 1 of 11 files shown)


core/container/mutexMap_test.go
```diff
@@ -0,0 +1,265 @@
...
+	keys := mm.Keys()
+	assert.Equal(t, 0, len(keys))
+
+	mm.Set(key1, struct{}{})
+	mm.Set(key2, struct{}{})
+	mm.Set(key3, struct{}{})
+
+	keys = mm.Keys()
+	require.Equal(t, 3, len(keys))
+	stringKeys := make([]string, len(keys))
```
andreibancioiu: From this line below, you only test for the existence of the keys inside the `mm.Keys()` slice. Sorting the keys and then performing the assertions in order does not improve the test (length assertion performed, and keys are unique). Is this correct? `require.Contains()` can be an alternative.

core/container/mutexMap_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	mm.Set(key1, struct{}{})
+	assert.Equal(t, 1, mm.Len())
+
+	mm.Set(key2, struct{}{})
+	assert.Equal(t, 2, mm.Len())
+
+	mm.Set(key3, struct{}{})
+	assert.Equal(t, 3, mm.Len())
+}
```
andreibancioiu: There are no tests with concurrent removals & getters & setters - paired or all of them running concurrently. I see `t.Parallel()` though, so `go test -race` should cover those.

process/factory/containers/preProcessorsContainer.go
```diff
@@ -91,17 +91,17 @@ func (ppc *preProcessorsContainer) Len() int {
 
 // Keys returns all the keys from the container
 func (ppc *preProcessorsContainer) Keys() []block.Type {
-	keys := make([]block.Type, 0)
-	for key := range ppc.objects.Iter() {
-		uint8key, ok := key.Key.(uint8)
+	keys := ppc.objects.Keys()
+	keysByte := make([]block.Type, 0, len(keys))
```
andreibancioiu: `keysAsBlockTypes` or `blockTypes`.

process/factory/containers/vmContainer.go
```diff
@@ -39,11 +40,11 @@ func (vmc *virtualMachinesContainer) Get(key []byte) (vmcommon.VMExecutionHandle
 // Add will add an object at a given key. Returns
 // an error if the element already exists
 func (vmc *virtualMachinesContainer) Add(key []byte, vm vmcommon.VMExecutionHandler) error {
-	if vm == nil {
+	if check.IfNilReflect(vm) {
```
andreibancioiu: So finally I've found an interface that doesn't define `IsInterfaceNil()` :))

process/factory/containers/intermediateTransactionHandlersContainer.go
```diff
@@ -91,17 +91,17 @@ func (ppc *intermediateTransactionHandlersContainer) Len() int {
 
 // Keys returns all the existing keys in the container
 func (ppc *intermediateTransactionHandlersContainer) Keys() []block.Type {
-	keys := make([]block.Type, 0)
-	for key := range ppc.objects.Iter() {
-		uint8key, ok := key.Key.(uint8)
+	keys := ppc.objects.Keys()
+	keysByte := make([]block.Type, 0, len(keys))
```
andreibancioiu: (As above)

PR #1318, opened 2020-03-30
updated travis.yml so it will build the termui and logviewer binaries

iulianpascalau:
updated travis.yml so it will build the termui and logviewer binaries


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 
-go: 1.13
+go: "1.13"
```
bogdan-rosianu: why ? it works both ways. 
https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use

iulianpascalau: Had a weak warning in goland.

PR #1321, opened 2020-03-30
changed epoch source in peer shard mapper

iulianpascalau:
changed epoch source in peer shard mapper


sharding/networksharding/peerShardMapper.go
```diff
@@ -72,7 +70,7 @@ func NewPeerShardMapper(
 		fallbackPkShard:  fallbackPkShard,
 		fallbackPidShard: fallbackPidShard,
 		nodesCoordinator: nodesCoordinator,
-		epochHandler:     epochHandler,
+		epoch:            uint32(0),
```
sasurobert: you should make epoch as an argument at constructor. - for start in epoch.

iulianpascalau: ok, will instantiate with the value stored in epochHandler.Epoch()

PR #1322, opened 2020-03-31
Fix trie snapshot and prune

BeniaminDrasovean:


data/trie/pruningBuffer.go
```diff
@@ -0,0 +1,53 @@
...
+
+	sb.buffer = append(sb.buffer, rootHash)
+	log.Trace("pruning buffer add", "rootHash", rootHash)
+}
+
+func (sb *pruningBuffer) remove(rootHash []byte) {
+	sb.mutOp.Lock()
+	defer sb.mutOp.Unlock()
+
+	for i := 0; i < len(sb.buffer); i++ {
```
sasurobert: instead of delete from existing buffer, create a new one without those being equal with rootHash

data/trie/pruningBuffer.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (sb *pruningBuffer) removeAll() [][]byte {
+	sb.mutOp.Lock()
+	defer sb.mutOp.Unlock()
+
+	log.Trace("pruning buffer", "len", len(sb.buffer))
+
+	buffer := sb.buffer
+	sb.buffer = make([][]byte, 0)
```
sasurobert: sb.buffer = sb.buffer[:0]

data/trie/pruningBuffer_test.go
```diff
@@ -0,0 +1,41 @@
...
+	assert.NotNil(t, newPruningBuffer())
+}
+
+func TestSnapshotsBuffer_Add(t *testing.T) {
+	sb := defaultPruningBuffer()
+
+	assert.Equal(t, 2, len(sb.buffer))
+}
+
+func TestSnapshotsBuffer_Remove(t *testing.T) {
```
sasurobert: add a test with remove at end of list. or having the same roothash multiple times.

data/trie/pruningBuffer.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type pruningBuffer struct {
+	mutOp  sync.RWMutex
+	buffer [][]byte
+}
+
+func newPruningBuffer() *pruningBuffer {
+	return &pruningBuffer{
+		buffer: make([][]byte, 0),
```
sasurobert: why not map[string]{} ?

data/trie/trieStorageManager.go
```diff
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	snapshotBufferLen = 10
-	pruningBufferLen  = 100
+	snapshotBufferLen = 100
```
sasurobert: if there are more than 100 smart contracts this might be problematic. I think these bufferLens should be not limited, or be super big in terms of millions ?

data/trie/trieStorageManager.go
```diff
@@ -184,33 +184,37 @@ func (tsm *trieStorageManager) ExitSnapshotMode() {
...
 	if tsm.snapshotInProgress > 0 {
+		tsm.pruningBuffer.add(rootHash)
 		return
 	}
 
-	if tsm.snapshotsBuffer.contains(rootHash[:len(rootHash)-1]) {
+	oldHashes := tsm.pruningBuffer.removeAll()
+	oldHashes = append(oldHashes, rootHash)
+
+	for i := range oldHashes {
```
sasurobert: I think this code should be called in ExistSnapshotMode as well - when snapshot buffer len is == 0.

data/trie/pruningBuffer.go
BeniaminDrasovean: Done.

data/trie/pruningBuffer_test.go
BeniaminDrasovean: Done.

data/trie/pruningBuffer.go
BeniaminDrasovean: Changed.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: This is a limitation for the buffered channel. If there are more than 100 smart contracts, all will be snapshoted, but the operation will be buffered, meaning that the first 100 will be added to the channel, then when one snapshot is done, another snapshot will be added.

data/trie/pruningBuffer.go
```diff
@@ -0,0 +1,48 @@
...
+	delete(sb.buffer, string(rootHash))
+}
+
+func (sb *pruningBuffer) removeAll() map[string]struct{} {
+	sb.mutOp.Lock()
+	defer sb.mutOp.Unlock()
+
+	log.Trace("pruning buffer", "len", len(sb.buffer))
+
+	buffer := sb.buffer
```
sasurobert: you need a deep copy here. with for range.

BeniaminDrasovean: Done.

PR #1333, opened 2020-04-01
Fix trigger for shards on missed peer miniblocks

SebastianMarian:


epochStart/shardchain/validatorInfoProcessor.go
```diff
@@ -164,58 +163,70 @@ func (vip *ValidatorInfoProcessor) processAllPeerMiniBlocks(metaBlock *block.Met
...
-		return nil
-	case <-time.After(time.Second):
-		return process.ErrTimeIsOut
+		return nil, nil
+	case <-time.After(waitTime):
+		return vip.getCopyOfAllMissingPeerMiniblocks(), process.ErrTimeIsOut
+	}
+}
+
+func (vip *ValidatorInfoProcessor) getCopyOfAllMissingPeerMiniblocks() map[string]*block.MiniBlock {
```
sasurobert, line 220: if these are missing then you have a nil pointer for every missing Peer miniblock ?

epochStart/shardchain/validatorInfoProcessor.go
```diff
@@ -164,58 +163,70 @@ func (vip *ValidatorInfoProcessor) processAllPeerMiniBlocks(metaBlock *block.Met
...
+	numMissingPeerMiniblocks := uint32(0)
+	vip.mutAllMissingPeerMiniblocks.Lock()
 
 	for _, mb := range metaBlock.MiniBlockHeaders {
 		if mb.Type != block.PeerBlock {
 			continue
 		}
 
-		vip.allPeerMiniblocks[string(mb.Hash)] = &miniBlockInfo{}
+		vip.mapAllMisingPeerMiniblocks[string(mb.Hash)] = nil
```
sasurobert, line 174: the naming is confusing - this is exactly allPeerMiniblock - those which we found - not those which are missing

epochStart/shardchain/validatorInfoProcessor.go
```diff
@@ -164,58 +163,70 @@ func (vip *ValidatorInfoProcessor) processAllPeerMiniBlocks(metaBlock *block.Met
 }
 
 func (vip *ValidatorInfoProcessor) computeMissingPeerBlocks(metaBlock *block.MetaBlock) {
-	missingNumber := uint32(0)
-	vip.mutMiniBlocksForBlock.Lock()
+	numMissingPeerMiniblocks := uint32(0)
+	vip.mutAllMissingPeerMiniblocks.Lock()
```
sasurobert, line 167: the mutex can be mutMiniBlocksForBlock

epochStart/shardchain/validatorInfoProcessor.go
sasurobert: here you are effectively returning all the miniblocks, those found and those missing as well. need a better naming.

epochStart/shardchain/trigger.go
```diff
@@ -454,6 +506,16 @@ func (t *trigger) checkIfTriggerCanBeActivated(hash string, metaHdr *block.MetaB
 	return isMetaHdrFinal, finalityAttestingRound
 }
 
+func (t *trigger) addMissingMiniblocks(missingMiniblocks map[string]*block.MiniBlock) {
+	t.mutMissingMiniblocks.Lock()
+	defer t.mutMissingMiniblocks.Unlock()
+
+	for hash, mb := range missingMiniblocks {
+		t.mapMisingMiniblocks[hash] = mb
+		log.Debug("trigger.addMissingMiniblocks", "hash", []byte(hash))
```
sasurobert, line 515: Thise are not only missing miniblocks - these are all of them.

epochStart/shardchain/trigger.go
```diff
@@ -441,8 +492,9 @@ func (t *trigger) checkIfTriggerCanBeActivated(hash string, metaHdr *block.MetaB
 	}
 
 	if metaHdr.IsStartOfEpochBlock() {
-		err := t.validatorInfoProcessor.ProcessMetaBlock(metaHdr, []byte(hash))
+		missingMiniblocks, err := t.validatorInfoProcessor.ProcessMetaBlock(metaHdr, []byte(hash))
```
sasurobert, line 495: not only missing are here - but all of them.

epochStart/shardchain/trigger.go
```diff
@@ -181,14 +190,56 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
...
+		t.updateMissingMiniblocks()
+	}
+}
+
+func (t *trigger) updateMissingMiniblocks() {
+	t.mutMissingMiniblocks.Lock()
+	defer t.mutMissingMiniblocks.Unlock()
+
+	for hash := range t.mapMisingMiniblocks {
+		if t.miniBlocksPool.Has([]byte(hash)) {
```
sasurobert, line 234: elements from the mapMissingMiniblocks should be deleted once the epoch has changed and was processed. No need to keep them. this can create out of memory problems.

epochStart/shardchain/trigger.go
```diff
@@ -181,14 +190,56 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
...
+		time.Sleep(sleepTime)
+
+		t.mutMissingMiniblocks.RLock()
+		if len(t.mapMisingMiniblocks) == 0 {
+			t.mutMissingMiniblocks.RUnlock()
+			continue
+		}
+
+		missingMiniblocks := make([][]byte, len(t.mapMisingMiniblocks))
+		for hash := range t.mapMisingMiniblocks {
```
sasurobert, line 217: missing are only those which are nil in the map.

epochStart/shardchain/trigger.go
```diff
@@ -181,14 +190,56 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
...
+
+	for hash := range t.mapMisingMiniblocks {
+		if t.miniBlocksPool.Has([]byte(hash)) {
+			delete(t.mapMisingMiniblocks, hash)
+		}
+	}
+
+	if len(t.mapMisingMiniblocks) == 0 {
+		log.Debug("trigger.updateMissingMiniblocks -> updateTriggerFromMeta")
+		t.updateTriggerFromMeta()
```
sasurobert, line 241: this should only be called if t.mutTrigger.Lock()

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Refactored

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Refactored

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Refactored

epochStart/shardchain/trigger.go
SebastianMarian: Now are only the missing ones

epochStart/shardchain/trigger.go
SebastianMarian: Now are only the missing ones

epochStart/shardchain/trigger.go
SebastianMarian: Actually the epoch could not be changed if there are more elements in this map. Anyway I would add a forced clean when epoch changed and processed

epochStart/shardchain/trigger.go
SebastianMarian: Refactored

epochStart/shardchain/trigger.go
SebastianMarian: Done

epochStart/shardchain/validatorInfoProcessor.go
```diff
@@ -38,12 +37,12 @@ type ValidatorInfoProcessor struct {
 	validatorStatisticsProcessor epochStart.ValidatorStatisticsProcessorHandler
 	requestHandler               epochStart.RequestHandler
 
-	allPeerMiniblocks     map[string]*miniBlockInfo
-	headerHash            []byte
-	metaHeader            data.HeaderHandler
-	chRcvAllMiniblocks    chan struct{}
-	mutMiniBlocksForBlock sync.Mutex
-	numMissing            uint32
+	mapAllMisingPeerMiniblocks  map[string]*block.MiniBlock
```
raduchis: add another s : miSsing

epochStart/shardchain/validatorInfoProcessor.go
```diff
@@ -164,60 +163,74 @@ func (vip *ValidatorInfoProcessor) processAllPeerMiniBlocks(metaBlock *block.Met
...
 }
 
+func (vip *ValidatorInfoProcessor) getAllMissingPeerMiniblocks() map[string]*block.MiniBlock {
+	vip.mutMiniBlocksForBlock.RLock()
+	defer vip.mutMiniBlocksForBlock.RUnlock()
+
+	mapMissingPeerMiniBlocks := make(map[string]*block.MiniBlock)
+	for hash, mb := range vip.mapAllPeerMiniblocks {
+		if mb == nil {
+			mapMissingPeerMiniBlocks[hash] = mb
```
raduchis, line 227: A map with <string, nil> will be sent back. Is that ok? Is the map used instead of a []string just to have an easy delete for a hash?

epochStart/shardchain/trigger.go
```diff
@@ -181,14 +190,69 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
...
+		}
+
+		missingMiniblocks := make([][]byte, len(t.mapMissingMiniblocks))
+		for hash := range t.mapMissingMiniblocks {
+			missingMiniblocks = append(missingMiniblocks, []byte(hash))
+			log.Debug("trigger.requestMissingMiniblocks", "hash", []byte(hash))
+		}
+		t.mutMissingMiniblocks.RUnlock()
+
+		go t.requestHandler.RequestMiniBlocks(core.MetachainShardId, missingMiniblocks)
```
raduchis, line 234: this is done also in the validatorInfoProcessor; is it really needed in both places?

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Done

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Yes, the intention was to send back only a map with hashes of the missing miniblocks(nil) and in trigger the map will help to easy delete with hash

epochStart/shardchain/trigger.go
SebastianMarian: Actually the problem was that if after 1 second of waiting in validatorInfoProcessor some of miniblocks are still missing there was no mechanism to request them again and the node would be stuck to this point. Now, in trigger if some missing hashes are returned they would be requested on this go routine and when they all are received the method updateTrigger... would be called again

epochStart/shardchain/trigger.go
```diff
@@ -181,14 +190,69 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
...
+		time.Sleep(waitTime)
+		t.updateMissingMiniblocks()
+	}
+}
+
+func (t *trigger) updateMissingMiniblocks() {
+	t.mutMissingMiniblocks.Lock()
+	for hash := range t.mapMissingMiniblocks {
```
sasurobert, line 242: this for is not needed if the delete from mapMissingMiniBLocks is done at receive time - this is time consuming.

sasurobert: if you make a register handler to the received miniblocks then this for is not needed - only if numMissingMiniBlocks == 0

epochStart/shardchain/validatorInfoProcessor.go
SebastianMarian: Anyway good point here, I refactored to return [][]byte as in trigger another map is created from this slice

epochStart/shardchain/trigger.go
```diff
@@ -26,6 +27,9 @@ import (
 
 var log = logger.GetOrCreate("epochStart/shardchain")
 
+// sleepTime defines the time in milliseconds between each iteration made in requestMissingMiniblocks method
+const sleepTime = 5 * time.Millisecond
```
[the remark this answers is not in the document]
SebastianMarian: Done

epochStart/shardchain/trigger.go
SebastianMarian: You can not do this there, as for each miniblock received you would call updateTrigger as numMissingMiniBlocks would be 0 fro most of time

epochStart/shardchain/trigger.go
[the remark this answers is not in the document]
SebastianMarian: Anyway there is only a len check and a continue for most of the time

epochStart/shardchain/validatorInfoProcessor.go
raduchis: ok

epochStart/shardchain/validatorInfoProcessor.go
raduchis: ok

PR #1338, opened 2020-04-02
Add cleaning component for cross txs

SebastianMarian:


process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -0,0 +1,285 @@
...
+	var crossTxInfo *crossTxInfo
+	for shardID := uint32(0); shardID < ctpc.shardCoordinator.NumberOfShards(); shardID++ {
+		isCrossTx, crossTxInfo = ctpc.isCrossTxFromShard(key, shardID, txType)
+		if isCrossTx {
+			break
+		}
+	}
+
+	if !isCrossTx {
+		isCrossTx, crossTxInfo = ctpc.isCrossTxFromShard(key, core.MetachainShardId, txType)
```
bogdan-rosianu, line 128: add a nil check for crossTxInfo as GoLand raises warnings

process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -0,0 +1,285 @@
...
+
+	_, ok := txStore.Peek(txHash)
+	if !ok {
+		return false, nil
+	}
+
+	return true, &crossTxInfo{txStore: txStore, senderShardID: shardID, receiverShardID: selfShardID}
+}
+
+func (ctpc *crossTxsPoolsCleaner) cleanCrossTxsPoolsIfNeeded() {
```
bogdan-rosianu, line 178: wouldn't it be more efficient if you call this on a goroutine once in a while, instead of trying to remove every transaction after it is added ? For example do the checks for percentAllowed and MaxRoundsToKeepUnprocessedTransactions somewhere else and call the clean mechanism only when needed (maybe this function can be the handler for a `PoolAlmostFullNotifier` or something like this)

process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -0,0 +1,285 @@
...
+		return
+	}
+
+	log.Trace("crossTxsPoolsCleaner.receivedUnsignedTx", "hash", key)
+	ctpc.processReceivedTx(key, unsignedTx)
+}
+
+func (ctpc *crossTxsPoolsCleaner) processReceivedTx(key []byte, txType int8) {
+	isCrossTx := false
+	var crossTxInfo *crossTxInfo
```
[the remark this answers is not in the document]
SebastianMarian: Done

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: Done

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: Good point

SebastianMarian: Refactored

bogdan-rosianu: It is better now, but can you also implement a weighted semaphore instead of just a mutex ? 
Imagine this situation: you receive a bulk of 1000 transactions. all of them are valid. for each transaction, you will call the cleanup method which will clean older transactions. but you will clean the pool from the first transaction, the rest of 999 would probably call the cleanup method with no change. 
By using a semaphore, you guarantee that the cleanup method is used only once per bulk.

SebastianMarian: Actually the cleanup... method is not called anymore from received... method (I have removed it from there), but only from this go routine and only once per minute. So no matter what it will clean all what is needed to be cleaned from one shot. One more thing is that this cleaner will remove txs based on their age and not on their count.

bogdan-rosianu: You're right, sorry. It's ok

process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+
+		txHandler, ok := value.(data.TransactionHandler)
+		if !ok {
+			log.Debug("cleanCrossTxsPoolsIfNeeded", "error", process.ErrWrongTypeAssertion,
+				"hash", []byte(hash),
+				"round", crossTxInfo.round,
+				"sender shard", crossTxInfo.senderShardID,
+				"receiver shard", crossTxInfo.receiverShardID,
+				"type", getTxTypeName(crossTxInfo.txType))
```
AdoAdoAdo, line 222: you don't need to delete this tx from pool/local crossTxInfo?

process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -0,0 +1,306 @@
...
+func (ctpc *crossTxsPoolsCleaner) cleanCrossTxsPoolsIfNeeded() int {
+	ctpc.mutMapCrossTxsRounds.Lock()
+	defer ctpc.mutMapCrossTxsRounds.Unlock()
+
+	selfShardID := ctpc.shardCoordinator.SelfId()
+	numPendingMiniBlocks := ctpc.blockTracker.GetNumPendingMiniBlocks(selfShardID)
+	numTxsCleaned := 0
+
+	for hash, crossTxInfo := range ctpc.mapCrossTxsRounds {
+		value, ok := crossTxInfo.txStore.Get([]byte(hash))
```
AdoAdoAdo, line 203: do we even need to get the tx from txstore and verify that it is valid etc?
don't we just need to remove if enough time passed?
I would just check the time in the crossTxInfo and if it is time to remove do the remove and don't care
no need to access multiple times the store if it is not needed.

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: Actually if this happens, wrong type assertion error, probably there is a kind of garbage in pool. But yes, I would add these both removal here.

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: It was done this way only to get the txs and to print some info about these which are going to be removed, but we can go without these info and to simplify this part.

PR #1339, opened 2020-04-02
Fix validators config between epochs

AdoAdoAdo:


sharding/common.go
```diff
@@ -4,6 +4,32 @@ import (
...
+	}
+	return resultMap, nil
+}
+
+func cloneValidatorsList(validatorsList []Validator) ([]Validator, error) {
+	var err error
+	var clone interface{}
+	resultList := make([]Validator, len(validatorsList))
+	for i, v := range validatorsList {
+		clone, err = v.Clone()
```
SebastianMarian, line 24: Why not: clone, err := v.Clone() and remove lines 20 and 21

sharding/interface.go
```diff
@@ -21,6 +21,7 @@ type Validator interface {
 	Address() []byte
 	Chances() uint32
 	SetChances(chances uint32)
+	Clone() (interface{}, error)
```
raduchis, line 24: Could the method return a Validator? It should always be a Validator

sharding/validator.go
```diff
@@ -51,6 +51,16 @@ func (v *validator) SetChances(chances uint32) {
 	v.mutChances.Unlock()
 }
 
+// Clone clones the validator
+func (v *validator) Clone() (interface{}, error) {
+	if v == nil {
+		return nil, ErrNilValidator
+	}
+	v2 := *v
```
raduchis, line 59: This will not copy the Pk and Address which are []byte but the pointer to them. Should it copy them? Could the PK and Address get changed like v.PK()[7] = 5? If not there is no need to copy it.

raduchis: ok

AdoAdoAdo: Correct it will not deep copy the Pk and Addr but this is fine as they will not change.

sharding/common.go
AdoAdoAdo: done

PR #1351, opened 2020-04-04
Fixed wrong type assertion for received transactions

SebastianMarian:
* Fixed wrong type assertion for received transactions
* Added guard checks for message type and message round in consensus topic interceptor
* Fixed memory leak in miniblocks/cross txs pools cleaner when pending miniblocks would not reached to 0 for a long time
* Added guard check for received header hash size from consensus message in consensus topic interceptor

### DIFF
--- a/process/block/poolsCleaner/crossTxsPoolsCleaner.go
+++ b/process/block/poolsCleaner/crossTxsPoolsCleaner.go
@@ -35,7 +35,6 @@ type txInfo struct {
 // crossTxsPoolsCleaner represents a pools cleaner that checks and cleans cross txs which should not be in pool anymore
 type crossTxsPoolsCleaner struct {
 	addressConverter         state.AddressConverter
-	blockTracker             BlockTracker
 	blockTransactionsPool    dataRetriever.ShardedDataCacherNotifier
 	rewardTransactionsPool   dataRetriever.ShardedDataCacherNotifier
 	unsignedTransactionsPool dataRetriever.ShardedDataCacherNotifier
@@ -50,7 +49,6 @@ type crossTxsPoolsCleaner struct {
 // NewCrossTxsPoolsCleaner will return a new cross txs pools cleaner
 func NewCrossTxsPoolsCleaner(
 	addressConverter state.AddressConverter,
-	blockTracker BlockTracker,
 	dataPool dataRetriever.PoolsHolder,
 	rounder process.Rounder,
 	shardCoordinator sharding.Coordinator,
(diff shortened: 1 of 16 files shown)


process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -115,7 +128,19 @@ func (ctpc *crossTxsPoolsCleaner) receivedRewardTx(key []byte, value interface{}
...
+
+	tx, ok := value.(data.TransactionHandler)
+	if !ok {
+		log.Warn("crossTxsPoolsCleaner.receivedRewardTx", "error", process.ErrWrongTypeAssertion)
+		return
+	}
+
+	senderShardID, receiverShardID, err := ctpc.computeSenderAndReceiverShards(tx)
+	if err != nil {
+		log.Debug("crossTxsPoolsCleaner.receivedRewardTx", "error", err.Error())
```
AdoAdoAdo, line 140: shouldn't you return here on error as well?

process/block/poolsCleaner/crossTxsPoolsCleaner.go
```diff
@@ -124,17 +149,28 @@ func (ctpc *crossTxsPoolsCleaner) receivedUnsignedTx(key []byte, value interface
...
 	if !ok {
-		log.Warn("crossTxsPoolsCleaner.processReceivedTx", "error", process.ErrWrongTypeAssertion)
+		log.Warn("crossTxsPoolsCleaner.receivedUnsignedTx", "error", process.ErrWrongTypeAssertion)
 		return
 	}
 
-	if wrappedTx.SenderShardID == ctpc.shardCoordinator.SelfId() {
+	senderShardID, receiverShardID, err := ctpc.computeSenderAndReceiverShards(tx)
+	if err != nil {
+		log.Debug("crossTxsPoolsCleaner.receivedUnsignedTx", "error", err.Error())
```
AdoAdoAdo, line 161: return on error?

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: Yep, forgot to do it

process/block/poolsCleaner/crossTxsPoolsCleaner.go
SebastianMarian: Yep, forgot to do it

PR #1357, opened 2020-04-06
Pass protocol built-in functions to Arwen

andreibancioiu:
- Passed list of built-in functions such as "ClaimDeveloperRewards", "ChangeOwnerAddress" to Arwen.
 - Improved some tests


process/factory/shard/vmContainerFactory.go
```diff
@@ -26,6 +27,7 @@ type vmContainerFactory struct {
 	cryptoHook         vmcommon.CryptoHook
 	blockGasLimit      uint64
 	gasSchedule        map[string]map[string]uint64
+	builtinFunctions   []string
```
sasurobert, line 30: in the PR https://github.com/ElrondNetwork/elrond-go/pull/1365 the keys for the builtInFunctions can be read from the blockchainhookHandler interface as it returns the container of the builtInFunctions. Can you change the PR to target that one ? Plus to make a config.toml for these functions, read from it, and push it everywhere ?

PR #1360, opened 2020-04-06
fix out of range case

bogdan-rosianu:


process/block/metrics.go
```diff
@@ -126,6 +126,14 @@ func incrementCountAcceptedBlocks(
 	}
 
 	bitMap := header.GetPubKeysBitmap()
+	indexOutOfBounds := myIndex/8 >= len(bitMap)
+	if indexOutOfBounds {
+		log.Debug("process blocks metrics: index out of bounds",
```
sasurobert, line 131: Trace ?

bogdan-rosianu: This shouldn't happen too often, but ok, I'll move it to trace

process/block/metrics.go
```diff
@@ -126,6 +126,14 @@ func incrementCountAcceptedBlocks(
 	}
 
 	bitMap := header.GetPubKeysBitmap()
+	indexOutOfBounds := myIndex/8 >= len(bitMap)
```
iulianpascalau, line 129: the big question here is: why this could happen in a normal case?

bogdan-rosianu: normally it shouldn't happen. but there was a spam which had hardcoded fields of the block which somehow passed the interceptor

statusHandler/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -190,21 +190,22 @@ func (wr *WidgetsRender) prepareChainInfo() {
 	numRows := 10
 	rows := make([][]string, numRows)
 
-	syncStatus := wr.presenter.GetIsSyncing()
-	syncingStr := fmt.Sprintf("undefined %d", syncStatus)
+	synchronizedRound := wr.presenter.GetSynchronizedRound()
+	currentRound := wr.presenter.GetCurrentRound()
+	syncingStr := "undefined"
```
golangcibot, line 195: ineffectual assignment to `syncingStr` (from `ineffassign`)

PR #1362, opened 2020-04-06
Fix heartbeat exploit between peer ID and public key 

iulianpascalau:
- removed heartbeat exploit between pid and pk
- removed hardfork trigger exploit by adding a timestamp for the hardfork message


node/heartbeat/monitor.go
```diff
@@ -247,12 +252,27 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
+	isHardforkTrigger, err := m.hardforkTrigger.TriggerReceived(message.Data(), hbRecv.Payload, hbRecv.Pubkey)
 	if isHardforkTrigger {
-		errHardforkTrigger := m.hardforkTrigger.TriggerReceived(hbRecv.Pubkey, hbRecv.Payload)
-		if errHardforkTrigger != nil {
-			return errHardforkTrigger
-		}
+		return err
+	}
+
+	if !bytes.Equal(hbRecv.Pid, message.Peer().Bytes()) {
```
raduchis, line 260: the hbRecv is created from the message, and the pid is set to the Peer.Bytes() if the pid does not exist in a map. When can they be different?

iulianpascalau: The whole idea of this fix was to include the peer id (pid) inside the message that will get signed. If a node decides to "replay" or re-broadcast a message containing a pid other then the self, it will immediately be blacklisted by other peer. This is done as to prevent some malicious nodes to replay other node's heartbeat message in order to confuse and disrupt the network sharding's data collector.

node/heartbeat/proto/heartbeat.proto
```diff
@@ -12,6 +12,7 @@ message Heartbeat {
 	uint32 ShardID         = 4;
 	string VersionNumber   = 5;
 	string NodeDisplayName = 6;
+	bytes  Pid 			   = 7;
```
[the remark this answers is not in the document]
iulianpascalau: Indeed. Fixed

node/heartbeat/errors.go
```diff
@@ -73,3 +73,9 @@ var ErrNilAntifloodHandler = errors.New("nil antiflood handler")
 
 // ErrNilHardforkTrigger signals that a nil hardfork trigger has been provided
 var ErrNilHardforkTrigger = errors.New("nil hardfork trigger")
+
+// ErrHearbeatPidMismatch signals that a received hearbeat did not come from the correct originator
+var ErrHearbeatPidMismatch = errors.New("heartbeat peer id mismatch")
```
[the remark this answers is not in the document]
iulianpascalau: fixed

node/heartbeat/monitor.go
```diff
@@ -252,6 +252,8 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
 		return err
 	}
 
+	//TODO check if hardfork trigger can be set otherwise (not requiring a pk that will actually send the trigger
```
sasurobert, line 255: it is not whenever it wants - it is the PK managed by an off-chain governance, people can opt out of this. They can set their own PK to activate this trigger. Later on this trigger, can be done in on-chain governance.

iulianpascalau: Agree, TODO will remain as to remind us to re-think of a better solution (if we found one)

node/heartbeat/errors.go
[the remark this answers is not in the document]
mihaib79: Hearbeat :)

PR #1374, opened 2020-04-07
Add consensus messages validity checks

SebastianMarian:


consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,268 @@
...
+		return fmt.Errorf("%w : received public key from consensus topic has an invalid size: %d",
+			ErrInvalidPublicKeySize,
+			len(cnsMsg.PubKey))
+	}
+
+	if len(cnsMsg.Signature) != core.SignatureSizeInBytes {
+		return fmt.Errorf("%w : received signature from consensus topic has an invalid size: %d",
+			ErrInvalidSignatureSize,
+			len(cnsMsg.Signature))
+	}
```
mihaib79: Aren't these already checked in checkConsensusMessageValidityForMessageType ?

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+			err.Error())
+	}
+
+	return nil
+}
+
+func (wrk *Worker) checkConsensusMessageValidityForMessageType(cnsMsg *consensus.Message) error {
+	msgType := consensus.MessageType(cnsMsg.MsgType)
+
+	if wrk.consensusService.IsMessageWithBlockBodyAndHeader(msgType) {
```
iulianpascalau, line 90: this could have looked better with a switch on the cnsMsg.MsgType

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
+			"SignatureShare = %s PubKeysBitmap = %s AggregateSignature = %s LeaderSignature = %s",
+			ErrInvalidMessage,
+			logger.DisplayByteSlice(cnsMsg.SignatureShare),
+			logger.DisplayByteSlice(cnsMsg.PubKeysBitmap),
+			logger.DisplayByteSlice(cnsMsg.AggregateSignature),
+			logger.DisplayByteSlice(cnsMsg.LeaderSignature))
+	}
+
+	if len(cnsMsg.Body) > core.MegabyteSize {
```
iulianpascalau, line 131: This could not happen

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("%w : received body from consensus topic has an invalid size: %d",
+			ErrInvalidBodySize,
+			len(cnsMsg.Body))
+	}
+
+	if cnsMsg.Header == nil || len(cnsMsg.Header) > core.MegabyteSize {
```
iulianpascalau, line 137: Better as to consider also the empty slice:
```GO
headerLen := len(cnsMsg.Header)
if headerLen == 0 || headerLen > core.MegabyteSize {
```

len(cnsMsg.Header) > core.MegabyteSize cannot happen

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+			"len(Body) = %d SignatureShare = %s PubKeysBitmap = %s AggregateSignature = %s LeaderSignature = %s",
+			ErrInvalidMessage,
+			len(cnsMsg.Body),
+			logger.DisplayByteSlice(cnsMsg.SignatureShare),
+			logger.DisplayByteSlice(cnsMsg.PubKeysBitmap),
+			logger.DisplayByteSlice(cnsMsg.AggregateSignature),
+			logger.DisplayByteSlice(cnsMsg.LeaderSignature))
+	}
+
+	if cnsMsg.Header == nil || len(cnsMsg.Header) > core.MegabyteSize {
```
iulianpascalau, line 191: check for len(cnsMsg.Header)

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+			"len(Body) = %d len(Header) = %d PubKeysBitmap = %s AggregateSignature = %s LeaderSignature = %s",
+			ErrInvalidMessage,
+			len(cnsMsg.Body),
+			len(cnsMsg.Header),
+			logger.DisplayByteSlice(cnsMsg.PubKeysBitmap),
+			logger.DisplayByteSlice(cnsMsg.AggregateSignature),
+			logger.DisplayByteSlice(cnsMsg.LeaderSignature))
+	}
+
+	if len(cnsMsg.SignatureShare) != core.SignatureSizeInBytes {
```
iulianpascalau, line 218: we should not have the const core.SignatureSizeInBytes as it defeats the purpose of changing the signature type from config.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	err := wrk.checkConsensusMessageValidityForMessageType(cnsMsg)
+	if err != nil {
+		return err
+	}
+
+	msgType := consensus.MessageType(cnsMsg.MsgType)
+
+	isBlockHeaderHashSizeInvalid := wrk.consensusService.IsMessageWithBlockBody(msgType) && cnsMsg.BlockHeaderHash != nil ||
```
iulianpascalau, line 26: this should be broken in more vars

core/constants.go
```diff
@@ -327,3 +327,9 @@ const NodesCoordinatorRegistryKeyPrefix = "indexHashed_"
 
 // HashSizeInBytes represents the size in bytes of a hash
 const HashSizeInBytes = 32
+
+// SignatureSizeInBytes represents the size in bytes of a BLS signature
+const SignatureSizeInBytes = 48
```
iulianpascalau, line 332: These consts should not be here. It will be impossible to change the public key type and the signature type from config

consensus/spos/consensusMessageValidator.go
sasurobert, line 26: there is a need for paranthesis here - it does not work as expected. on set till || and the other set after ||

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (wrk *Worker) checkMessageWithBlockBodyAndHeaderValidity(cnsMsg *consensus.Message) error {
+	isMessageInvalid := cnsMsg.SignatureShare != nil ||
+		cnsMsg.PubKeysBitmap != nil ||
+		cnsMsg.AggregateSignature != nil ||
+		cnsMsg.LeaderSignature != nil
+
+	if isMessageInvalid {
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
```
sasurobert, line 122: is this error message not too long ? maybe do a log trace before then send and simpler wrapped error.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (wrk *Worker) checkMessageWithBlockBodyValidity(cnsMsg *consensus.Message) error {
+	isMessageInvalid := cnsMsg.Header != nil ||
+		cnsMsg.SignatureShare != nil ||
+		cnsMsg.PubKeysBitmap != nil ||
+		cnsMsg.AggregateSignature != nil ||
+		cnsMsg.LeaderSignature != nil
+
+	if isMessageInvalid {
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
```
sasurobert, line 154: is this error message not too long ? maybe do a log trace before then send and simpler wrapped error.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (wrk *Worker) checkMessageWithBlockHeaderValidity(cnsMsg *consensus.Message) error {
+	isMessageInvalid := cnsMsg.Body != nil ||
+		cnsMsg.SignatureShare != nil ||
+		cnsMsg.PubKeysBitmap != nil ||
+		cnsMsg.AggregateSignature != nil ||
+		cnsMsg.LeaderSignature != nil
+
+	if isMessageInvalid {
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
```
sasurobert, line 181: is this error message not too long ? maybe do a log trace before then send and simpler wrapped error.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+func (wrk *Worker) checkMessageWithSignatureValidity(cnsMsg *consensus.Message) error {
+	isMessageInvalid := cnsMsg.Body != nil ||
+		cnsMsg.Header != nil ||
+		cnsMsg.PubKeysBitmap != nil ||
+		cnsMsg.AggregateSignature != nil ||
+		cnsMsg.LeaderSignature != nil
+
+	if isMessageInvalid {
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
+			"len(Body) = %d len(Header) = %d PubKeysBitmap = %s AggregateSignature = %s LeaderSignature = %s",
```
sasurobert, line 209: is this error message not too long ? maybe do a log trace before then send and simpler wrapped error.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (wrk *Worker) checkMessageWithFinalInfoValidity(cnsMsg *consensus.Message) error {
+	isMessageInvalid := cnsMsg.Body != nil ||
+		cnsMsg.Header != nil ||
+		cnsMsg.SignatureShare != nil
+
+	if isMessageInvalid {
+		return fmt.Errorf("%w : received message from consensus topic is invalid: "+
+			"len(Body) = %d len(Header) = %d SignatureShare = %s",
```
sasurobert, line 234: is this error message not too long ? maybe do a log trace before then send and simpler wrapped error.

consensus/spos/consensusMessageValidator.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("%w : received leader signature from consensus topic has an invalid size: %d",
+			ErrInvalidSignatureSize,
+			len(cnsMsg.LeaderSignature))
+	}
+
+	return nil
+}
+
+func (wrk *Worker) getPubKeyBitmapSize() int {
+	sizeConsensus := wrk.consensusState.consensusGroupSize
```
sasurobert, line 263: this should be computed only once at constructor.

consensus/spos/consensusMessageValidator.go
SebastianMarian: Actually we don't have access to the message type as it is defined in bls package and not in spos. This is why we used consensusService interface for this.

consensus/spos/consensusMessageValidator.go
SebastianMarian: It couldn't happen in the current setting of the p2p, but if in the future this limit would be changed, this condition would prevent spam on this field greater than 1MB

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

SebastianMarian: It couldn't happen in the current setting of the p2p, but if in the future this limit would be changed, this condition would prevent spam on this field greater than 1MB

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: I would pull it in config.toml in [Consensus] section

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

core/constants.go
SebastianMarian: I would pull them in config.toml in [Consensus] section

consensus/spos/consensusMessageValidator.go
SebastianMarian: Refactored

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

consensus/spos/consensusMessageValidator.go
mihaib79: funny fact in GO - || has precedence over &&

consensus/spos/consensusMessageValidator.go
mihaib79: funny fact in GO - || has precedence over &&

consensus/spos/consensusMessageValidator.go
iulianpascalau: If we allow p2p messages of 2 or more MB, we certainly want a larger header to be processed.

consensus/spos/consensusMessageValidator.go
iulianpascalau: I should have made it more clear:
HashSizeInBytes is not required, as Hasher interface has the Size() method
PublicKeySizeInBytes can be recovered from the config, from the Address and BLSPublicKey area. 
SignatureSizeInBytes might be included in the aforementioned Address and BLSPublicKey as to be changed on the same time.

consensus/spos/consensusMessageValidator.go
SebastianMarian: Depends on the time which will take the broadcast header to reach to every node in a round time frame

consensus/spos/consensusMessageValidator.go
SebastianMarian: Done

PR #1375, opened 2020-04-07
Enable-start-in-epoch

sasurobert:


data/blockchain/baseBlockchain.go
```diff
@@ -84,37 +83,3 @@ func (bbc *baseBlockChain) SetCurrentBlockHeaderHash(hash []byte) {
 	bbc.currentBlockHeaderHash = hash
 	bbc.mut.Unlock()
 }
-
-// GetCurrentBlockBody returns the tx block body pointer
-func (bbc *baseBlockChain) GetCurrentBlockBody() data.BodyHandler {
```
SebastianMarian, line 89: You can also remove currentBlockBody variable from baseBlockChain struct

PR #1376, opened 2020-04-07
ratings config updates april

raduchis:
updated ratings config for 400 nodes per shard as for the mainnet
the consecutiveMissedBlocks are not reset every epoch any more.


cmd/node/main.go
```diff
@@ -819,6 +828,11 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	err = ioutil.WriteFile(statsFile, []byte(sessionInfoFileOutput), os.ModePerm)
 	log.LogIfError(err)
 
+	computedRatingsData := filepath.Join(statsFolder, "ratings.info")
+	computedRatingsDataStr := createStringFromRatingsData(ratingsData)
+	err = ioutil.WriteFile(computedRatingsData, []byte(computedRatingsDataStr), os.ModePerm)
```
sasurobert, line 833: is this really needed ?

cmd/node/main.go
```diff
@@ -1052,6 +1066,32 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	return nil
 }
 
+func createStringFromRatingsData(ratingsData *rating.RatingsData) string {
```
sasurobert, line 1069: this should be only for a log.Debug and could be done at the constructor of NewRatingsData

process/rating/ratingsData.go
```diff
@@ -18,70 +21,177 @@ type RatingsData struct {
...
+}
 
-	chances := make([]process.SelectionChance, 0)
-	for _, chance := range settings.General.SelectionChances {
-		chances = append(chances, &SelectionChance{
-			MaxThreshold:  chance.MaxThreshold,
-			ChancePercent: chance.ChancePercent,
-		})
+func computeRatingStep(
+	shardSize uint32,
```
sasurobert, line 143: if there are so many input it would be better to give an argument DTO

process/rating/ratingsData.go
```diff
@@ -18,70 +21,177 @@ type RatingsData struct {
...
+
+	ratingFromProposer := float32(ratingDifference) * (proposerValidatorImportance / totalImportance)
+	ratingFromValidator := float32(ratingDifference) * (1 / totalImportance)
+
+	proposerIncrease := ratingFromProposer / proposerProbability
+	validatorIncrease := ratingFromValidator / validatorProbability
+	proposerDecrease := proposerIncrease * proposerDecreaseFactor
+	validatorDecrease := validatorIncrease * validatorDecreaseFactor
+
+	if proposerIncrease > math.MaxInt32 {
```
sasurobert, line 170: this could never happen. Over and UnderFlow protection has to be done before the operation. You could do all the computation with big.Int as it is only called in constructor.

cmd/node/main.go
AdoAdoAdo, line 834: I actually like it that we persist the computed rating parameters to a file
The rating is important and this way it is easy to check afterwards.

process/rating/ratingsData.go
raduchis: proposerIncrease is a float32 and it can happen. I wrote some unit Tests

cmd/node/main.go
raduchis: Currently it is for debugging purposes to see what was computed based on the shardSize, consensusSize, timeToReachMaxRating and proposerValidatorImportance. It is like the session.info.

cmd/node/main.go
raduchis: I would like it to be for all logging levels and be persistent in the folder. I can add a TODO to remove it in the future when it is not needed any more.

process/rating/ratingsData.go
raduchis: fixed

cmd/node/main.go
raduchis: added todo to remove and add just log.debug

process/rating/ratingsData.go
AdoAdoAdo: I think float32 goes somewhere up to 2^127, although not sure about the precision, so the check should be ok.

PR #1378, opened 2020-04-08
Fix block size throttle edge case for cross mbs dest me

SebastianMarian:
* Fixed block size throttling edge case for cross shard miniblocks with destination in self shard
* Fixed computing of sender and receiver shards for rewards transactions in crossTxsPoolsCleaner


cmd/node/config/config.toml
```diff
@@ -374,8 +374,7 @@
     PeerStatePruningEnabled = true
 
 [BlockSizeThrottleConfig]
-    #TODO fix MinSizeInBytes when executing large cross-shard miniblocks
-    MinSizeInBytes = 943718 # 943718 is 90% from 1MB or 104857 is 10% from 1MB
+    MinSizeInBytes = 104857 # 943718 is 90% from 1MB or 104857 is 10% from 1MB
```
sasurobert, line 377: change comment

PR #1379, opened 2020-04-08
Merge dev in feat bech32 2020.04.08

iulianpascalau:


process/block/baseProcess.go
```diff
@@ -66,6 +66,8 @@ type baseProcessor struct {
 
 	appStatusHandler       core.AppStatusHandler
 	stateCheckpointModulus uint
+	blockProcessor         blockProcessor
+	txCounter              *transactionCounter
```
golangcibot, line 70: `txCounter` is unused (from `structcheck`)

iulianpascalau: This is actually used in both shardProcessor and metaProcessor. Moved the txCounter as to silent the warning.

iulianpascalau: Fixed in a next PR.

PR #1380, opened 2020-04-08
Keep meta and shard in sync

AdoAdoAdo:


process/block/hdrForBlock.go
```diff
@@ -35,6 +35,13 @@ func (hfb *hdrForBlock) resetMissingHdrs() {
 	hfb.mutHdrsForBlock.Unlock()
 }
 
+func (hfb *hdrForBlock) getHeaderInfo(hash string) (*hdrInfo, bool) {
```
golangcibot, line 38: func `(*hdrForBlock).getHeaderInfo` is unused (from `unused`)

integrationTests/vm/systemVM/systemVm_test.go
```diff
@@ -111,6 +112,7 @@ func TestStakingUnstakingAndUnboundingOnMultiShardEnvironmentWithValidatorStatis
 		t.Skip("this is not a short test")
 	}
 
+	_ = logger.SetLogLevel("process:TRACE")
```
sasurobert, line 115: delete this

process/block/metablock.go
```diff
@@ -1205,9 +1205,68 @@ func (mp *metaProcessor) updateState(lastMetaBlock data.HeaderHandler) {
...
+		shardHeader, ok := headerInfo.hdr.(*block.Header)
+		if !ok {
+			log.Debug("getLastSelfNotarizedHeaderByShard",
+				"error", process.ErrWrongTypeAssertion,
+				"hash", shardData.HeaderHash)
+			continue
+		}
+
+		for _, metaHash := range shardHeader.MetaBlockHashes {
+			metaHeader, err := process.GetMetaHeader(metaHash, mp.dataPool.Headers(), mp.marshalizer, mp.store)
```
sasurobert, line 1245: I think having LastIncludedMetaNonce already in the previous metablock (the one from blockchain) + the information in the current metablock, the new LastIncludedMetaNonce for every shard can be calculated without reading it from storage. Effectively it would be LastIncludedMetaNonce + len(shardHeader.MetaBlockHashes) - 1 if the last included metahash from the previous shardHeader is equal with the first of the current one.

process/track/baseBlockTrack.go
```diff
@@ -518,7 +518,20 @@ func (bbt *baseBlockTrack) GetTrackedHeadersWithNonce(shardID uint32, nonce uint
 // IsShardStuck returns true if the given shard is stuck
 func (bbt *baseBlockTrack) IsShardStuck(shardID uint32) bool {
 	numPendingMiniBlocks := bbt.blockBalancer.GetNumPendingMiniBlocks(shardID)
-	isShardStuck := numPendingMiniBlocks >= process.MaxNumPendingMiniBlocks
+	shardProcessedMetaNonce := bbt.blockBalancer.GetLastShardProcessedMetaNonce(shardID)
+
+	isMetaDifferenceTooLarge := false
+	if shardID != core.MetachainShardId {
+		lastMetaHeader, _, err := bbt.GetLastCrossNotarizedHeader(core.MetachainShardId)
```
SebastianMarian, line 525: I think that GetLastCrossNotarizedHeader should be replaced with GetTrackedHeaders, for two reasons: 1) First method will return last cross notarized meta header from the node point of view and is related to where the node reached with the processed meta headers (so if this shard also has not kept the sync with meta it would consider that another shard which is closer with processing to the same meta header to be also synced which is false) 2) If IsShardStuck method would be called by metachain when it creates miniblocks for shards, like rewards or maybe other future miniblocks type, and it wants to check if the given shard is stuck or not, the call of method GetLastCrossNotarizedHeader(core.MetachainShardId) it would return only the genesis meta header, as this notarization does not exists (metachain will store cross notarization only for shards but not for himself. For own shard notarization it is used GetLastSelfNotarizedHeader method which is backed by another map)

SebastianMarian: So I propose that lines between 525-531 to be replaced with this code:

		metaHeaders, _ := bbt.GetTrackedHeaders(core.MetachainShardId)
		numMetaHeaders := len(metaHeaders)
		if numMetaHeaders > 0 {
			lastMetaHeader := metaHeaders[numMetaHeaders - 1]
			metaDiff := lastMetaHeader.GetNonce() - shardProcessedMetaNonce
			isMetaDifferenceTooLarge = metaDiff > process.MaxMetaNoncesBehind
		}

process/block/metablock.go
```diff
@@ -1205,9 +1205,68 @@ func (mp *metaProcessor) updateState(lastMetaBlock data.HeaderHandler) {
...
 }
 
-func (mp *metaProcessor) getLastSelfNotarizedHeaderByShard(_ uint32) (data.HeaderHandler, []byte) {
-	//TODO: Implement mechanism to extract last meta header notarized by the given shard if this info will be needed later
-	return nil, nil
+func (mp *metaProcessor) getLastSelfNotarizedHeaderByShard(
+	metaBlock *block.MetaBlock,
+	shardID uint32,
+) (data.HeaderHandler, []byte) {
+	maxNotarizedNonce := uint64(0)
```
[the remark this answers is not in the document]
AdoAdoAdo: done

process/track/baseBlockTrack.go
AdoAdoAdo: done

integrationTests/vm/systemVM/systemVm_test.go
AdoAdoAdo: done

PR #1388, opened 2020-04-09
Snapshot empty trie

BeniaminDrasovean:
Fix bug when trying to snapshot empty trie.
Fix TestRollbackBlockAndCheckThatPruningIsCancelledOnAccountsTrie
Move trie storage manager constants in config.


data/trie/trieStorageManager.go
```diff
@@ -359,7 +361,7 @@ func (tsm *trieStorageManager) getSnapshotDb(newDb bool) data.DBWriteCacher {
 		return nil
 	}
 
-	if len(tsm.snapshots) > maxSnapshots {
+	if uint8(len(tsm.snapshots)) > tsm.maxSnapshots {
```
sasurobert, line 364: what happens if in the same shard there are more smart contracts than maxSnapshots ? when snapshoting the number of tsm.snapshots will be higher.

BeniaminDrasovean: The data tries will be saved in the same db as the main trie, so each snapshot creates only one db, and each trie is saved in the same db. For the data tries, SetChekpoint() method is called, which will not trigger a new db creation.

PR #1393, opened 2020-04-09
Skip some tests.

andreibancioiu:


integrationTests/consensus/consensus_test.go
```diff
@@ -186,6 +187,10 @@ func runFullConsensusTest(t *testing.T, consensusType string) {
 }
 
 func TestConsensusBLSFullTest(t *testing.T) {
+	if os.Getenv("SKIPCI") != "" {
```
sasurobert, line 190: maybe add TODO to take this out when TC is great again

PR #1396, opened 2020-04-10
Compute when to sync epoch start

sasurobert:


epochStart/bootstrap/process.go
```diff
@@ -181,27 +186,21 @@ func NewEpochStartBootstrap(args ArgsEpochStartBootstrap) (*epochStartBootstrap,
...
 	if isCurrentTimeBeforeGenesis {
 		return true
 	}
 
-	configuredDurationOfEpoch := startTime.Add(e.computedDurationOfEpoch())
-	isEpochZero := time.Since(configuredDurationOfEpoch) < 0
-
-	return isEpochZero
+	currentRound := e.rounder.Index()
+	return float64(currentRound) < float64(e.generalConfig.EpochStartConfig.RoundsPerEpoch)*gracePeriodInPercentage
```
iulianpascalau, line 203: this won't work with the early start of epoch. Can remain like this for time being.

epochStart/bootstrap/process.go
```diff
@@ -39,9 +39,12 @@ import (
 
 var log = logger.GetOrCreate("epochStart/bootstrap")
 
-const timeToWait = 10 * time.Second
+const timeToWait = time.Minute
+const trieSyncWaitTime = 5 * time.Minute
 const timeBetweenRequests = 100 * time.Millisecond
 const maxToRequest = 100
+const gracePeriodInPercentage = float64(0.25)
```
SebastianMarian, line 46: This is equivalent to ShuffledOutRestartThreshold from config.toml [EpochStartConfig] section ?

epochStart/bootstrap/process.go
```diff
@@ -285,6 +268,21 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
 	return e.requestAndProcessing()
 }
 
+func (e *epochStartBootstrap) computeIfCurrentEpochIsSaved() bool {
+	e.initializeFromLocalStorage()
```
iulianpascalau, line 272: We have a bug in the method initializeFromLocalStorage that can appear if we have an epoch DB saved but on the wrong shard. So, the method `getParentDirAndLastEpoch` should take into account also the shard ID

epochStart/bootstrap/process.go
SebastianMarian, line 203: We don't need to use here MinRoundsBetweenEpochs from config.toml [EpochStartConfig] section instead 
float64(e.generalConfig.EpochStartConfig.RoundsPerEpoch)*gracePeriodInPercentage ?

epochStart/bootstrap/process.go
```diff
@@ -285,6 +268,21 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
...
+		return false
+	}
+
+	computedRound := e.rounder.Index()
+	if computedRound-e.baseData.lastRound < roundGracePeriod {
+		return true
+	}
+
+	roundsSinceEpochStart := computedRound - int64(e.baseData.epochStartRound)
+	return float64(roundsSinceEpochStart) < float64(e.generalConfig.EpochStartConfig.RoundsPerEpoch)*gracePeriodInPercentage
```
SebastianMarian, line 283: We don't need to use here MinRoundsBetweenEpochs from config.toml [EpochStartConfig] section instead
float64(e.generalConfig.EpochStartConfig.RoundsPerEpoch)*gracePeriodInPercentage ?

epochStart/bootstrap/process.go
```diff
@@ -285,6 +268,21 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
...
 }
 
+func (e *epochStartBootstrap) computeIfCurrentEpochIsSaved() bool {
+	e.initializeFromLocalStorage()
+	if !e.baseData.storageExists {
+		return false
+	}
+
+	computedRound := e.rounder.Index()
+	if computedRound-e.baseData.lastRound < roundGracePeriod {
```
SebastianMarian, line 278: Is this condition safe, we really need it?

storage/factory/openStorage.go
```diff
@@ -271,37 +284,85 @@ func getLastEpochAndRoundFromStorage(
...
-		bootstrapData, errGet := getBootstrapDataForPersisterPath(persisterFactory, persisterPath, marshalizer)
+		bootstrapData, storer, errGet := getBootstrapDataForPersisterPath(persisterFactory, persisterPath, marshalizer)
 		if errGet != nil {
 			continue
 		}
 
 		if bootstrapData.LastRound > highestRoundInStoredShards {
+			shardID := uint32(0)
+			var err error
+			shardID, err = convertShardIDToUint32(shardIdStr)
```
SebastianMarian, line 295: Why not shardID, errConvert := and you can remove both lines above

storage/factory/openStorage.go
```diff
@@ -271,37 +284,85 @@ func getLastEpochAndRoundFromStorage(
...
+			if err != nil {
+				continue
```
SebastianMarian, line 297: Maybe we can add a log here

storage/factory/openStorage.go
```diff
@@ -271,37 +284,85 @@ func getLastEpochAndRoundFromStorage(
...
+			}
+			epochStartRound, err = loadEpochStartRound(shardID, bootstrapData.EpochStartTriggerConfigKey, storer)
+			if err != nil {
+				continue
```
SebastianMarian, line 301: Maybe we can add a log here

epochStart/bootstrap/process.go
sasurobert: @bogdan-rosianu is working on refactoring openStorage.go file .

cmd/node/main.go
```diff
@@ -664,6 +665,15 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 		return err
 	}
 
+	rounder, err := round.NewRound(
```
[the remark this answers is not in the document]
sasurobert: in epoch start in does not do much. It is a simple component, I do not see it problematic of being duplicated.

epochStart/bootstrap/process.go
sasurobert: yep. but that is hard to know in this case.

cmd/node/main.go
SebastianMarian, line 668: Please pass this rounder to NewProcessComponentsFactoryArgs right below in line 906, and remove its second creation from func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, error), to have only one instance of rounder in the whole project

epochStart/bootstrap/process.go
sasurobert: MinRoundsBetweenEpoch is used only for forceEndOfEpoch.

epochStart/bootstrap/process.go
sasurobert: MinRoundsBetweenEpoch is used only for forceEndOfEpoch.

epochStart/bootstrap/process.go
sasurobert: the condition is safe. I just added if for any case the round is relatively close to one in the DB there is no reason to start syncing.

epochStart/bootstrap/process.go
sasurobert: right now yes. but I do not know what gracePeriod should we choose.

storage/factory/openStorage.go
sasurobert: done

storage/factory/openStorage.go
sasurobert: done

storage/factory/openStorage.go
sasurobert: done

cmd/node/main.go
sasurobert: done

cmd/node/factory/structs.go
```diff
@@ -548,6 +547,7 @@ type processComponentsFactoryArgs struct {
 	nodesConfig               *sharding.NodesSetup
 	gasSchedule               map[string]map[string]uint64
 	syncer                    ntp.SyncTimer
```
SebastianMarian, line 549: Delete syncer var as it is unused now

epochStart/bootstrap/process.go
iulianpascalau: agree, probably we will have a dedicated component for this

epochStart/bootstrap/process.go
iulianpascalau: great!

epochStart/bootstrap/process.go
SebastianMarian: I think we should add a TODO here to not forget later to replace these constants, gracePeriodInPercentage and roundGracePeriod, with values read from config.toml

epochStart/bootstrap/process.go
SebastianMarian: Yes I know, and I just thinking to this case, just to work even in this situation. But anyway, you can leave it like this right now, if this is not a good suggestion

epochStart/bootstrap/process.go
SebastianMarian: Yes I know, and I just thinking to this case, just to work even in this situation. But anyway, you can leave it like this right now, if this is not a good suggestion

epochStart/bootstrap/process.go
SebastianMarian: Agree, but probably would be catch anyway by the next computation

epochStart/bootstrap/process.go
sasurobert: I think this should not be configurable. we have to find the best values.

epochStart/bootstrap/process.go
sasurobert: it is impossible to know at start in epoch if there was a force start in epoch or not.

epochStart/bootstrap/process.go
sasurobert: t is impossible to know at start in epoch if there was a force start in epoch or not.

cmd/node/factory/structs.go
sasurobert: done.

storage/factory/openStorage.go
```diff
@@ -365,17 +382,22 @@ func getBootstrapDataForPersisterPath(
 		return nil, nil, err
 	}
 
-	defer func() {
-		errClose := persister.Close()
```
iulianpascalau, line 308: This was really wrong

PR #1399, opened 2020-04-11
fix the order of processing epoch start metablock in case of conflicts.

sasurobert:
- We observed an edge case in processing and validating epoch start block and the consequent blocks for the new epoch coming from epoch. There is an edge case where the epoch start block is created by a malicious leader, he gathers the signatures, but will release that only later. 
- The specific scenario was having a set of metablocks with (R:543,N:543), (R:544, N:544), (R:545, N:545, NewEpoch) and the other chain of the same (R:543, N:543) and (R:545:N:544, NewEpoch). In this case the previous hash of the epoch start metablock is different for the 2 different cases, thus resulting a different nodesConfig after shuffling. 
- The fix is keeping a defined order of processing, shuffling nodes according to the received metablocks, in this case as the longer chain is the first one, the processing order should be according to nonce - thus resulting the processing of the start of epoch block with the higher nonce last.
- Put check epoch correctness cross chain earlier in order to take a better decision to reconstruct the chain - in case of shards when they missed the new epoch. 
- Fixed trie node resolver to request only intra shard.
- Process epoch start metablock only until it is not finalized for each epoch

### DIFF
--- a/dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
+++ b/dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
@@ -267,13 +267,18 @@ func (brcf *baseResolversContainerFactory) createOneResolverSenderWithSpecifiedN
 	return resolverSender, nil
 }
 
-func (brcf *baseResolversContainerFactory) createTrieNodesResolver(topic string, trieId string) (dataRetriever.Resolver, error) {
+func (brcf *baseResolversContainerFactory) createTrieNodesResolver(
+	topic string,
+	trieId string,
+	numCrossShard int,
+	numIntraShard int,
+) (dataRetriever.Resolver, error) {
 	resolverSender, err := brcf.createOneResolverSenderWithSpecifiedNumRequests(
 		topic,
 		EmptyExcludePeersOnTopic,
 		defaultTargetShardID,
-		0,
-		numIntraShardPeers+numCrossShardPeers,
+		numCrossShard,
+		numIntraShard,
 	)
 	if err != nil {
 		return nil, err
--- a/dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
+++ b/dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
@@ -230,6 +230,16 @@ func (brcf *baseResolversContainerFactory) createOneResolverSender(
 	excludedTopic string,
 	targetShardId uint32,
 ) (dataRetriever.TopicResolverSender, error) {
+	return brcf.createOneResolverSenderWithSpecifiedNumRequests(topic, excludedTopic, targetShardId, numCrossShardPeers, numIntraShardPeers)
+}
+
+func (brcf *baseResolversContainerFactory) createOneResolverSenderWithSpecifiedNumRequests(
+	topic string,
+	excludedTopic string,
+	targetShardId uint32,
+	numCrossShard int,
+	numIntraShard int,
+) (dataRetriever.TopicResolverSender, error) {
 
 	peerListCreator, err := topicResolverSender.NewDiffPeerListCreator(brcf.messenger, topic, brcf.intraShardTopic, excludedTopic)
 	if err != nil {
@@ -244,8 +254,8 @@ func (brcf *baseResolversContainerFactory) createOneResolverSender(
 		Randomizer:         brcf.intRandomizer,
 		TargetShardId:      targetShardId,
 		OutputAntiflooder:  brcf.outputAntifloodHandler,
-		NumCrossShardPeers: numCrossShardPeers,
-		NumIntraShardPeers: numIntraShardPeers,
+		NumCrossShardPeers: numCrossShard,
+		NumIntraShardPeers: numIntraShard,
 	}
 	//TODO instantiate topic sender resolver with the shard IDs for which this resolver is supposed to serve the data
 	// this will improve the serving of transactions as the searching will be done only on 2 sharded data units
@@ -258,20 +268,26 @@ func (brcf *baseResolversContainerFactory) createOneResolverSender(
 }
 
 func (brcf *baseResolversContainerFactory) createTrieNodesResolver(topic string, trieId string) (dataRetriever.Resolver, error) {
-	resolverSender, err := brcf.createOneResolverSender(topic, EmptyExcludePeersOnTopic, defaultTargetShardID)
+	resolverSender, err := brcf.createOneResolverSenderWithSpecifiedNumRequests(
+		topic,
+		EmptyExcludePeersOnTopic,
+		defaultTargetShardID,
+		0,
+		numIntraShardPeers+numCrossShardPeers,
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	trie := brcf.triesContainer.Get([]byte(trieId))
-	arg := resolvers.ArgTrieNodeResolver{
+	argTrie := resolvers.ArgTrieNodeResolver{
 		SenderResolver:   resolverSender,
 		TrieDataGetter:   trie,
 		Marshalizer:      brcf.marshalizer,
 		AntifloodHandler: brcf.inputAntifloodHandler,
 		Throttler:        brcf.throttler,
 	}
-	resolver, err := resolvers.NewTrieNodeResolver(arg)
+	resolver, err := resolvers.NewTrieNodeResolver(argTrie)
 	if err != nil {
 		return nil, err
 	}
--- a/epochStart/bootstrap/epochStartMetaBlockProcessor.go
+++ b/epochStart/bootstrap/epochStartMetaBlockProcessor.go
@@ -196,8 +196,7 @@ func (e *epochStartMetaBlockProcessor) getMostReceivedMetaBlock() (*block.MetaBl
 	e.mutReceivedMetaBlocks.RLock()
 	defer e.mutReceivedMetaBlocks.RUnlock()
 
-	const hashNotAvailable = "N/A"
-	mostReceivedHash := hashNotAvailable
+	var mostReceivedHash string
 	maxLength := minNumOfPeersToConsiderBlockValid - 1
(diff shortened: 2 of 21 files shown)


epochStart/shardchain/trigger.go
```diff
@@ -99,6 +99,28 @@ type metaInfo struct {
 	hash string
 }
 
+type metaInfloSlice []*metaInfo
+
+// Len will return the length of the metaInfoList
```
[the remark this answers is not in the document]

epochStart/shardchain/trigger.go
```diff
@@ -99,6 +99,28 @@ type metaInfo struct {
 	hash string
 }
 
+type metaInfloSlice []*metaInfo
```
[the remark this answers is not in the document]

dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
```diff
@@ -258,20 +268,26 @@ func (brcf *baseResolversContainerFactory) createOneResolverSender(
 }
 
 func (brcf *baseResolversContainerFactory) createTrieNodesResolver(topic string, trieId string) (dataRetriever.Resolver, error) {
-	resolverSender, err := brcf.createOneResolverSender(topic, EmptyExcludePeersOnTopic, defaultTargetShardID)
+	resolverSender, err := brcf.createOneResolverSenderWithSpecifiedNumRequests(
+		topic,
+		EmptyExcludePeersOnTopic,
+		defaultTargetShardID,
+		0,
+		numIntraShardPeers+numCrossShardPeers,
```
iulianpascalau, line 276: :+1:

epochStart/bootstrap/process.go
```diff
@@ -466,6 +466,11 @@ func (e *epochStartBootstrap) requestAndProcessing() (Parameters, error) {
 		}
 	}
 
+	err = e.messenger.CreateTopic(core.ConsensusTopic+e.shardCoordinator.CommunicationIdentifier(e.shardCoordinator.SelfId()), true)
```
iulianpascalau, line 469: Although this solves the issue, we might think of a solution in which, after the creation of the shard coordinator instance, the node automatically registers to the consensus topic. In this way we won't "forget" to create the consensus topic on other possible cases.

epochStart/shardchain/trigger.go
```diff
@@ -99,6 +99,28 @@ type metaInfo struct {
...
+func (m metaInfoSlice) Len() int { return len(m) }
+
+// Swap will interchange the objects on input indexes
+func (m metaInfoSlice) Swap(i, j int) { m[i], m[j] = m[j], m[i] }
+
+// Less will return true if object on index i should appear before object in index j
+// Sorting of headers should be by epoch, by nonce and by hash in ascending order
+// this will ensure that in case of equality for epoch, the metaHdr with higher nonce will
+// be processed last - that is  the correct one - as it finalizes the previous nonce
+func (m metaInfoSlice) Less(i, j int) bool {
```
iulianpascalau, line 114: :+1:

epochStart/shardchain/trigger.go
```diff
@@ -509,17 +529,15 @@ func (t *trigger) checkIfTriggerCanBeActivated(hash string, metaHdr *block.MetaB
...
-	if metaHdr.IsStartOfEpochBlock() {
-		missingMiniblocksHashes, blockBody, err := t.peerMiniBlocksSyncer.SyncMiniBlocks(metaHdr)
-		if err != nil {
-			t.addMissingMiniblocks(metaHdr.Epoch, missingMiniblocksHashes)
-			log.Warn("processMetablock failed", "error", err)
-			return false, 0
-		}
-
-		t.epochStartNotifier.NotifyAllPrepare(metaHdr, blockBody)
+	missingMiniblocksHashes, blockBody, err := t.peerMiniBlocksSyncer.SyncMiniBlocks(metaHdr)
```
iulianpascalau, line 532: so, we do not need this condition anymore?

integrationTests/testProcessorNode.go
```diff
@@ -724,6 +724,8 @@ func (tpn *TestProcessorNode) initInterceptors() {
 func (tpn *TestProcessorNode) initResolvers() {
 	dataPacker, _ := partitioning.NewSimpleDataPacker(TestMarshalizer)
 
+	_ = tpn.Messenger.CreateTopic(core.ConsensusTopic+tpn.ShardCoordinator.CommunicationIdentifier(tpn.ShardCoordinator.SelfId()), true)
```
iulianpascalau, line 727: this should do it

node/node.go
```diff
@@ -617,7 +617,7 @@ func (n *Node) createConsensusTopic(messageProcessor p2p.MessageProcessor) error
 
 	n.consensusTopic = core.ConsensusTopic + n.shardCoordinator.CommunicationIdentifier(n.shardCoordinator.SelfId())
 	if n.messenger.HasTopicValidator(n.consensusTopic) {
-		return ErrValidatorAlreadySet
+		return n.messenger.RegisterMessageProcessor(n.consensusTopic, messageProcessor)
```
iulianpascalau, line 620: this should be reverted. The check is done on the HasTopicValidator, not HasTopic

node/node.go
```diff
@@ -617,7 +617,7 @@ func (n *Node) createConsensusTopic(messageProcessor p2p.MessageProcessor) error
...
 	}
 
 	if !n.messenger.HasTopic(n.consensusTopic) {
```
iulianpascalau, line 623: this if should be placed first, on line 619.

epochStart/shardchain/trigger.go
sasurobert: it is ensured that it is called only be start of epoch metablocks.

iulianpascalau: ok

epochStart/shardchain/trigger.go
```diff
@@ -171,6 +194,7 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
 		mapHashHdr:                  make(map[string]*block.MetaBlock),
 		mapNonceHashes:              make(map[uint64][]string),
 		mapEpochStartHdrs:           make(map[string]*block.MetaBlock),
+		mapFinalizedEpochs:          make(map[uint32]struct{}, 0),
```
golangcibot, line 197: S1019: should use make(map[uint32]struct{}) instead (from `gosimple`)

epochStart/shardchain/trigger.go
```diff
@@ -724,6 +754,7 @@ func (t *trigger) SetProcessed(header data.HeaderHandler, _ data.BodyHandler) {
 	t.mapHashHdr = make(map[string]*block.MetaBlock)
 	t.mapNonceHashes = make(map[uint64][]string)
 	t.mapEpochStartHdrs = make(map[string]*block.MetaBlock)
+	t.mapFinalizedEpochs = make(map[uint32]struct{}, 0)
```
golangcibot, line 757: S1019: should use make(map[uint32]struct{}) instead (from `gosimple`)

epochStart/shardchain/trigger.go
```diff
@@ -742,6 +754,7 @@ func (t *trigger) SetProcessed(header data.HeaderHandler, _ data.BodyHandler) {
...
```
SebastianMarian: What about RevertStateToBlock, do we need to set this map back?

epochStart/shardchain/trigger.go
```diff
@@ -424,6 +430,10 @@ func (t *trigger) updateTriggerFromMeta() {
 	sort.Sort(sortedMetaInfo)
 
 	for _, currMetaInfo := range sortedMetaInfo {
+		if _, ok := t.mapFinalizedEpochs[currMetaInfo.hdr.Epoch]; ok {
```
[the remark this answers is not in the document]
sasurobert: the check is needed. In case no of the earlier metablocks were finalized. You stop at the first one for the set epoch.

epochStart/shardchain/trigger.go
sasurobert: none of these maps are set in the revert - they are repopulated.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -191,6 +192,26 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
 	}
 }
 
+func (e *epochStartMetaBlockProcessor) getMostReceivedMetaBlock() (*block.MetaBlock, error) {
+	e.mutReceivedMetaBlocks.RLock()
+	defer e.mutReceivedMetaBlocks.RUnlock()
+
+	const hashNotAvailable = "N/A"
+	mostReceivedHash := hashNotAvailable
```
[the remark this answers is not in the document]
sasurobert: done

epochStart/shardchain/trigger.go
SebastianMarian: ok

epochStart/shardchain/trigger.go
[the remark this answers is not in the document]
SebastianMarian: ok

epochStart/shardchain/trigger.go
```diff
@@ -454,26 +454,32 @@ func (t *trigger) updateTriggerFromMeta() {
...
+	}
+}
+
+func (t *trigger) saveEpochStartMeta(metaHdr *block.MetaBlock) {
+	if check.IfNil(metaHdr) {
+		return
+	}
 
-			log.Debug("saving epoch start meta with", "key", epochStartIdentifier)
+	epochStartIdentifier := core.EpochStartIdentifier(metaHdr.Epoch)
```
SebastianMarian, line 467: You still can add this log: log.Debug("saving epoch start meta with", "key", epochStartIdentifier)

sasurobert: It was only for my simple purpose. No need

PR #1400, opened 2020-04-11
Fix edge case when forced fork and reset probable highest nonce are prevented

SebastianMarian:
* Fixed situation when sync would be stuck, because forced fork and reset probable highest nonce are prevented by two consecutive blocks synced, first one without errors and the next one with errors. Now, synced blocks with errors are counted per nonce.

### DIFF
--- a/process/sync/baseSync.go
+++ b/process/sync/baseSync.go
@@ -601,18 +601,10 @@ func (boot *baseBootstrap) syncBlock() error {
 		"nonce", hdr.GetNonce(),
 	)
 
-	boot.resetNonceSyncedWithErrors(hdr.GetNonce())
 	boot.cleanNoncesSyncedWithErrorsBehindFinal()
 	boot.requestsWithTimeout = 0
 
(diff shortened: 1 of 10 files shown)


process/sync/baseSync.go
```diff
@@ -486,7 +485,21 @@ func (boot *baseBootstrap) doJobOnSyncBlockFail(headerHandler data.HeaderHandler
 		}
 	}
 
-	allowedSyncWithErrorsLimitReached := boot.syncWithErrors >= process.MaxSyncWithErrorsAllowed
+	boot.resetProbableHighestNonceIfNeeded(headerHandler)
+}
+
+func (boot *baseBootstrap) resetProbableHighestNonceIfNeeded(headerHandler data.HeaderHandler) {
```
sasurobert, line 491: Maybe a set of unit tests? :)

SebastianMarian: Done

PR #1403, opened 2020-04-13
return metablock if at least num of peers send it

bogdan-rosianu:
when the context expires for getting the epoch start metablock, first check if at least a minimum number of peers sent a metablock and return it instead of returning a time out error.
+ unit test

### DIFF
--- a/epochStart/bootstrap/epochStartMetaBlockProcessor.go
+++ b/epochStart/bootstrap/epochStartMetaBlockProcessor.go
@@ -201,6 +201,7 @@ func (e *epochStartMetaBlockProcessor) getMostReceivedMetaBlock() (*block.MetaBl
 	maxLength := minNumOfPeersToConsiderBlockValid - 1
 	for hash, entry := range e.mapMetaBlocksFromPeers {
 		if len(entry) > maxLength {
+			maxLength = len(entry)
 			mostReceivedHash = hash
 		}
 	}
--- a/epochStart/bootstrap/epochStartMetaBlockProcessor.go
+++ b/epochStart/bootstrap/epochStartMetaBlockProcessor.go
@@ -21,6 +21,7 @@ const durationBetweenChecks = 200 * time.Millisecond
 const durationBetweenReRequests = 1 * time.Second
 const durationBetweenCheckingNumConnectedPeers = 500 * time.Millisecond
 const minNumConnectedPeers = 6
+const minNumOfPeersToConsiderBlockValid = 3
 
 var _ process.InterceptorProcessor = (*epochStartMetaBlockProcessor)(nil)
 
@@ -177,7 +178,7 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
 		case <-e.chanConsensusReached:
 			return e.metaBlock, nil
 		case <-ctx.Done():
-			return nil, epochStart.ErrTimeoutWaitingForMetaBlock
+			return e.getMostReceivedMetaBlock()
 		case <-chanRequests:
 			err = e.requestMetaBlock()
 			if err != nil {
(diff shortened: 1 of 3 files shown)


epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -191,6 +192,26 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
...
 
+func (e *epochStartMetaBlockProcessor) getMostReceivedMetaBlock() (*block.MetaBlock, error) {
+	e.mutReceivedMetaBlocks.RLock()
+	defer e.mutReceivedMetaBlocks.RUnlock()
+
+	const hashNotAvailable = "N/A"
+	mostReceivedHash := hashNotAvailable
+	maxLength := minNumOfPeersToConsiderBlockValid - 1
+	for hash, entry := range e.mapMetaBlocksFromPeers {
+		if len(entry) > maxLength {
```
sasurobert, line 203: this is minLength

bogdan-rosianu: no, because it refers to the length of the peers slice for a metablock

bogdan-rosianu: and we are interested to get the hash for the longest slice -> the most received

sasurobert: maxLength should be updated.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -191,6 +192,26 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
...
+			mostReceivedHash = hash
```
sasurobert, line 204: add line maxLength = len(entry)

bogdan-rosianu: done

epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -191,6 +192,27 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
...
 }
 
+func (e *epochStartMetaBlockProcessor) getMostReceivedMetaBlock() (*block.MetaBlock, error) {
+	e.mutReceivedMetaBlocks.RLock()
+	defer e.mutReceivedMetaBlocks.RUnlock()
+
+	const hashNotAvailable = "N/A"
+	mostReceivedHash := hashNotAvailable
+	maxLength := minNumOfPeersToConsiderBlockValid - 1
+	for hash, entry := range e.mapMetaBlocksFromPeers {
```
raduchis, line 202: Could it happen that the mapMetaBlocksFromPeers contains multiple entries from the same peer? Does it somehow influnece the selected metablock if we just have the minimum number.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -191,6 +192,27 @@ func (e *epochStartMetaBlockProcessor) GetEpochStartMetaBlock(ctx context.Contex
...
+			maxLength = len(entry)
+			mostReceivedHash = hash
+		}
+	}
+
+	if mostReceivedHash == hashNotAvailable {
+		return nil, epochStart.ErrTimeoutWaitingForMetaBlock
+	}
+
+	return e.mapReceivedMetaBlocks[mostReceivedHash], nil
```
raduchis, line 213: this returns the first miniblock of a specific length. Is it ok if multiple blocks have the same number of appearances to return the first? e.g.  we have 20 peers but 2 blocks have 9 appearances.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
bogdan-rosianu: no, there is a check when entries are added to that map so that the slice of an entry doesn't contain duplicate peers

epochStart/bootstrap/epochStartMetaBlockProcessor.go
bogdan-rosianu: You're right, this can happen. I don't know what to say about this.. maybe @sasurobert can provide an input

sasurobert: this is only to resolve an edge case where you do not get the same block from 2/3+1 of the connected peers. In that case the best thing you can do is to choose the best of the rest. If it is an equality - you can't decide which is better. It should not happen. Anything you decide in that moment is ok. at least you decided.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
raduchis: ok, clarified that the slice for the same hash does not contain dupplicates, but for different hashes we could have the same peer.

epochStart/bootstrap/epochStartMetaBlockProcessor.go
raduchis: ok

PR #1409, opened 2020-04-13
bugFix-request-epoch-start-info

sasurobert:
When the node is in sync, there might be scenarios when the epoch start trigger for a certain epoch was already saved and the requestEpochStartInfo function in process/shardBlock would return timeout error. There is no need to request epoch start info if the nodes current trigger is equal or higher then the epoch of the current header.
Added Epoch and MetaEpoch in the trigger. In order to treat differently those 2, making it clear what is in sync and what is received async.


epochStart/shardchain/trigger.go
```diff
@@ -735,6 +744,15 @@ func (t *trigger) getHeaderWithNonceAndPrevHash(nonce uint64, prevHash []byte) (
 	return t.getHeaderWithNonceAndHash(nonce, b.Data[0])
 }
 
+func (t *trigger) getAllFinishedStartOfEpochMetaHdrs() []*block.MetaBlock {
+	finishedMetaHdrs := make([]*block.MetaBlock, 0, len(t.mapFinalizedEpochs))
```
SebastianMarian, line 748: You can define this var like this: finishedMetaHdrs := make([]*block.MetaBlock, len(t.mapFinalizedEpochs)) and use 
finishedMetaHdrs[index] instead append in the next for

epochStart/shardchain/trigger.go
```diff
@@ -752,18 +770,22 @@ func (t *trigger) SetProcessed(header data.HeaderHandler, _ data.BodyHandler) {
...
+	if t.metaEpoch < t.epoch {
+		t.metaEpoch = t.epoch
+	}
+
 	t.isEpochStart = false
 	t.newEpochHdrReceived = false
 	t.epochMetaBlockHash = shardHdr.EpochStartMetaHash
 	t.epochStartShardHeader = shardHdr
-	currEpochStartMeta := t.mapEpochStartHdrs[string(t.epochMetaBlockHash)]
+	finishedEpochStartMetas := t.getAllFinishedStartOfEpochMetaHdrs()
```
SebastianMarian, line 781: finishedStartOfEpochMetaHdrs instead finishedEpochStartMetas ?

epochStart/shardchain/trigger.go
sasurobert: no, as the index is efectively equal with epoch. so you should have a separate index counter.

epochStart/shardchain/trigger.go

epochStart/shardchain/trigger.go
SebastianMarian: ok

PR #1415, opened 2020-04-14
Fixed creation of interceptors when starting the bootstrap from storage

iulianpascalau:
created a data pool in start in epoch as to not exit with error when trying to create the interceptors container. This removes the debug line error:
`DEBUG[2020-04-14 16:42:24.507] [..hStart/bootstrap] [/0/0/]        could not start from storage - will try sync for start in epoch error = nil data pool holder`
Fixing restart with DB for observers.
Fixing shuffling out and shard coordinator creation from new shard ID.


epochStart/bootstrap/process.go
```diff
@@ -247,6 +247,17 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
 		return e.prepareEpochZero()
 	}
 
+	e.dataPool, err = factoryDataPool.NewDataPoolFromConfig(
```
sasurobert, line 250: maybe move only where it is used. It can stay like this, it is not cricital

iulianpascalau: keeping this here will ensure that the next calls to interceptor container will have the data pool, it will be less prone to errors :)

epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -155,23 +157,54 @@ func (e *epochStartBootstrap) prepareEpochFromStorage() (Parameters, error) {
...
-	for _, validator := range epochConfig.WaitingValidators[shardIdForConfig] {
-		if bytes.Equal(pubKey, validator.PubKey) {
-			return false
+	newShardId, isWaitingForShard := checkIfIsValidatorForEpoch(pubKey, epochConfig.WaitingValidators)
+	if isWaitingForShard {
+		isShuffledOut := newShardId != e.baseData.shardId
+		return newShardId, isShuffledOut
+	}
+
+	newShardId, isEligibleForShard := checkIfIsValidatorForEpoch(pubKey, epochConfig.WaitingValidators)
```
raduchis, line 170: epochconfig.EligibleValidators and somehow no tests have failed.

cmd/node/main.go
```diff
@@ -723,27 +723,33 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
...
 		return err
 	}
+
+	shardCoordinator := genesisShardCoordinator
+	currentEpoch := uint32(0)
+	storerEpoch := currentEpoch
+
 	bootstrapParameters, err := bootstrapper.Bootstrap()
 	if err != nil {
 		log.Error("boostrap return error", "error", err)
```
iulianpascalau, line 733: If the node fails to bootstrap (might not gather the p2p connections in the bounded time) it will start from genesis and then stop (as it won't be able to sync from the network - wrong epoch) and the validator's owner won't be able to tell if the validator performed well or not. The return error is a must, at least for now, until we have a retry mechanism.

epochStart/bootstrap/fromLocalStorage.go
iulianpascalau, line 170: duplicate call `checkIfIsValidatorForEpoch(pubKey, epochConfig.WaitingValidators)` something is wrong in the second call parameter

epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -155,23 +157,54 @@ func (e *epochStartBootstrap) prepareEpochFromStorage() (Parameters, error) {
...
+	if isEligibleForShard {
+		isShuffledOut := newShardId != e.baseData.shardId
+		return newShardId, isShuffledOut
+	}
+
+	return e.baseData.shardId, false
+}
+
+func checkIfIsValidatorForEpoch(
```
iulianpascalau, line 179: bad naming: `checkIfPubkeyIsInMap`

epochStart/bootstrap/fromLocalStorage.go
sasurobert: done, thanks.

epochStart/bootstrap/fromLocalStorage.go
sasurobert: resolved

epochStart/bootstrap/fromLocalStorage.go
sasurobert: thanks. resolved.

cmd/node/main.go
sasurobert: reverted

epochStart/bootstrap/fromLocalStorage.go
raduchis: ok

PR #1416, opened 2020-04-14
Fix processing txs from me using initial balances

SebastianMarian:
Implemented a new mechanism for processing "txs from me" which take into consideration only the initial balances of the sender addresses accounts, without the possible increased amount coming after processing "txs dest me" in the same block.


process/block/preprocess/smartContractResults.go
```diff
@@ -496,6 +507,8 @@ func (scr *smartContractResults) ProcessMiniBlock(miniBlock *block.MiniBlock, ha
 			return processedTxHashes, err
 		}
 
+		scr.setInitialBalanceForAddress(miniBlockScrs[index].GetRcvAddr())
```
sasurobert, line 510: name is confusing. it does not set balance - it initialises the account from the current trie.

process/block/preprocess/balanceComputation.go
```diff
@@ -0,0 +1,106 @@
...
+
+// SetBalanceToAddress method sets balance to an address
+func (bc *balanceComputation) SetBalanceToAddress(address []byte, value *big.Int) {
+	bc.mutAddressBalance.Lock()
+	bc.mapAddressBalance[string(address)] = big.NewInt(0).Set(value)
+	bc.mutAddressBalance.Unlock()
+}
+
+// GetBalanceOfAddress method gets balance of an address
+func (bc *balanceComputation) GetBalanceOfAddress(address []byte) (*big.Int, error) {
```
sasurobert, line 37: you could eliminate some of the functions.

process/block/preprocess/smartContractResults.go
SebastianMarian: refactored

process/block/preprocess/balanceComputation.go
SebastianMarian: This method is used now for debug info

process/block/preprocess/basePreProcess.go
```diff
@@ -307,3 +312,47 @@ func (bpp *basePreProcess) computeGasConsumedByTx(
...
+
+	account, ok := accountHandler.(state.UserAccountHandler)
+	if !ok {
+		return nil, process.ErrWrongTypeAssertion
+	}
+
+	return account.GetBalance(), nil
+}
+
+func (bpp *basePreProcess) getTxTotalCost(txHandler data.TransactionHandler) *big.Int {
```
iulianpascalau, line 349: This is a bit misleading, I thought it will invoke the txFeeHandler. Might rename this to getMaxTxTotalCost

process/block/preprocess/interfaces.go
```diff
@@ -33,3 +35,14 @@ type BlockSizeThrottler interface {
 	GetCurrentMaxSize() uint32
 	IsInterfaceNil() bool
 }
+
+// BalanceComputationHandler defines the functionality for addresses balances computation
```
iulianpascalau, line 39: might add: ...used in preventing executing too many debit transactions after the proposer executed a credit transaction on the same account in the same block

process/block/preprocess/basePreProcess.go
```diff
@@ -307,3 +312,47 @@ func (bpp *basePreProcess) computeGasConsumedByTx(
...
 }
+
+func (bpp *basePreProcess) saveAccountBalanceForAddress(address []byte) {
+	if bpp.balanceComputation.HasAddressBalanceSet(address) {
+		return
+	}
+
+	balance, err := bpp.getBalanceForAddress(address)
+	if err != nil {
+		//TODO: Analyze if it is better to call return here
```
iulianpascalau, line 323: I think this is ok. If there is an error here (because the account does not exists, (adress conversion will not raise error & type assertion should not occur) ) the account is snapshot with balance 0 meaning there is no possibility of a debit tx occurring withing this block.

process/block/preprocess/balanceComputation.go
```diff
@@ -0,0 +1,87 @@
...
+			currValue.Sub(currValue, value)
+			return true
+		}
+	}
+
+	return false
+}
+
+// HasAddressBalanceSet method returns if address has balance set or not
+func (bc *balanceComputation) HasAddressBalanceSet(address []byte) bool {
```
iulianpascalau, line 63: IsAddressSet?

process/block/preprocess/balanceComputation.go
```diff
@@ -0,0 +1,87 @@
...
+	bc.mutAddressBalance.RLock()
+	defer bc.mutAddressBalance.RUnlock()
+
+	_, ok := bc.mapAddressBalance[string(address)]
+	return ok
+}
+
+// IsBalanceInAddress method returns if in given address is at least given balance
+func (bc *balanceComputation) IsBalanceInAddress(address []byte, value *big.Int) bool {
```
iulianpascalau, line 72: AddressHasEnoughBalance?

process/block/preprocess/transactions.go
```diff
@@ -954,6 +968,16 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 			continue
 		}
 
+		if txs.balanceComputation.HasAddressBalanceSet(tx.GetSndAddr()) {
+			ok = txs.balanceComputation.SubBalanceFromAddress(tx.GetSndAddr(), txTotalCost)
+			if !ok {
+				log.Error("createAndProcessMiniBlocksFromMe.SubBalanceFromAddress",
```
iulianpascalau, line 974: This should really not happen :)

process/block/preprocess/transactions.go
```diff
@@ -1103,6 +1128,8 @@ func (txs *transactions) ProcessMiniBlock(miniBlock *block.MiniBlock, haveTime f
 			return processedTxHashes, err
 		}
 
+		txs.saveAccountBalanceForAddress(miniBlockTxs[index].GetRcvAddr())
```
iulianpascalau, line 1131: why saving this here?

process/block/preprocess/balanceComputation.go
```diff
@@ -0,0 +1,87 @@
+package preprocess
+
+import (
+	"math/big"
+	"sync"
+)
+
+type balanceComputation struct {
+	mapAddressBalance map[string]*big.Int
```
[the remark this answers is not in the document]
SebastianMarian: Yep, but from whole mutexMap functionality I would only can used Set method, and I should have implemented Has, Sub, Add..

process/block/preprocess/basePreProcess.go
SebastianMarian: done

process/block/preprocess/interfaces.go
SebastianMarian: done

process/block/preprocess/basePreProcess.go
SebastianMarian: Yep, this is how I thought about. Removed the TODO

process/block/preprocess/balanceComputation.go
SebastianMarian: done

process/block/preprocess/balanceComputation.go
SebastianMarian: done

process/block/preprocess/transactions.go
SebastianMarian: Yep, this is why I put this message with log level error

process/block/preprocess/transactions.go
SebastianMarian: Actually here is the point where cross txs dest me would be executed by the proposer when it creates a block. I need to take the initial balance of the receiver address before the tx is executed

process/block/preprocess/balanceComputation.go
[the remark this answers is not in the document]
iulianpascalau: I meant the container could have been used as an internal, unexported field (not that this struct should implement the container interface)
```GO
type balanceComputation struct {
	mapAddressBalance *container.mutexMap
}
//.......
```

process/block/preprocess/transactions.go
iulianpascalau: ok

process/block/preprocess/transactions.go
```diff
@@ -874,9 +877,8 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 			}
 		}
 
-		txMaxTotalCost := txs.getTxMaxTotalCost(tx)
-
 		if txs.balanceComputation.IsAddressSet(tx.GetSndAddr()) {
+			txMaxTotalCost = txs.getTxMaxTotalCost(tx)
```
iulianpascalau, line 881: I know that it works like this but the code might easily break in panic if someone decides to change the condition on L880 & L973. Strongly suggest to extract `txs.balanceComputation.IsAddressSet(tx.GetSndAddr())` in a bool var that will be used on the both `if` lines

SebastianMarian: Refactored

iulianpascalau: great

PR #1422, opened 2020-04-15
Changed key filename

iulianpascalau:
- changed key filename
- updated scripts & README.md
- added default values in nodesSetup.json and genesis.json so the node can start with provided default config

Related to https://github.com/ElrondNetwork/elrond-deploy-go/pull/22


cmd/keygenerator/main.go
```diff
@@ -56,15 +56,15 @@ VERSION:
...
+		Usage:       "What king of keys should generate. Available options: validator, wallet, both",
+		Value:       "validator",
 		Destination: &argsConfig.keyType,
 	}
 
 	argsConfig = &cfg{}
 
-	initialBalancesSkFileName = "initialBalancesSk.pem"
-	initialNodesSkFileName    = "initialNodesSk.pem"
+	initialBalancesSkFileName = "walletKey.pem"
```
SebastianMarian, line 66: walletKeyPemFileName ?

cmd/keygenerator/main.go
```diff
@@ -56,15 +56,15 @@ VERSION:
...
+	initialNodesSkFileName    = "validatorKey.pem"
```
SebastianMarian, line 67: validatorKeyPemFileName ?

cmd/keygenerator/main.go
iulianpascalau: yes, missed that :+1:

iulianpascalau: done

cmd/keygenerator/main.go
iulianpascalau: done

PR #1423, opened 2020-04-15
display error message fix

miiu96:


cmd/node/main.go
```diff
@@ -1319,7 +1319,7 @@ func loadExternalConfig(filepath string) (*config.ExternalConfig, error) {
 	cfg := &config.ExternalConfig{}
 	err := core.LoadTomlFile(cfg, filepath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot load external config %w", err)
```
bogdan-rosianu, line 1322: separate the main error message and the resulted err.
`return nil, fmt.Errorf("cannot load external config: %w", err)` for example

core/indexer/elasticsearch.go
```diff
@@ -64,7 +66,7 @@ func NewElasticIndexer(arguments ElasticIndexerArgs) (Indexer, error) {
 	}
 	client, err := newElasticSearchDatabase(databaseArguments)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot create indexer %w", err)
```
bogdan-rosianu, line 69: also here

PR #1425, opened 2020-04-15
Automatic recover of corrupted DB file

iulianpascalau:
added a call to leveldb subsystem to recover the db file in case there is an error opening the existing file

### DIFF
--- a/storage/leveldb/common.go
+++ b/storage/leveldb/common.go
@@ -2,7 +2,6 @@ package leveldb
 
 import (
 	"fmt"
-	"reflect"
 
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/errors"
@@ -15,7 +14,7 @@ func openLevelDB(path string, options *opt.Options) (*leveldb.DB, error) {
 		return db, nil
 	}
 
-	if reflect.TypeOf(errOpen) == reflect.TypeOf(&errors.ErrCorrupted{}) {
+	if errors.IsCorrupted(errOpen) {
 		var errRecover error
 		log.Warn("corrupted DB file",
 			"path", path,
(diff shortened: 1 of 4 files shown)


storage/leveldb/common.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+)
+
+func openLevelDB(path string, options *opt.Options) (*leveldb.DB, error) {
+	db, errOpen := leveldb.OpenFile(path, options)
+	if errOpen != nil {
```
AdoAdoAdo, line 12: this may be a different error, such as file already opened
in this case recover should not be done

raduchis: https://godoc.org/github.com/syndtr/goleveldb/leveldb#Open

Open will return an error with type of ErrCorrupted if corruption detected in the DB. Use errors.IsCorrupted to test whether an error is due to corruption. Corrupted DB can be recovered with Recover function

iulianpascalau: done, added check

storage/leveldb/leveldb_test.go
```diff
@@ -26,6 +30,49 @@ func TestDB_InitNoError(t *testing.T) {
...
+	err = os.Remove(path.Join(dir, "MANIFEST-000000"))
+	require.Nil(t, err)
+
+	dbRecovered, err := leveldb.NewDB(dir, 10, 1, 10)
+	if err != nil {
+		assert.Fail(t, fmt.Sprintf("should have not errored %s", err.Error()))
+		return
+	}
+
+	valRecovered, err := dbRecovered.Get(key)
```
golangcibot, line 56: SA4006: this value of `err` is never used (from `staticcheck`)

iulianpascalau: fixed

PR #1429, opened 2020-04-16
bugfix/computing-finality-attesting-round

sasurobert:
finality attesting round in case of epoch start should be set to the lower round, until there is no block which creates on top of the finality attesting block.


epochStart/shardchain/trigger.go
```diff
@@ -367,7 +367,27 @@ func (t *trigger) SetAppStatusHandler(handler core.AppStatusHandler) error {
...
+func (t *trigger) changeEpochFinalityAttestingRoundIfNeeded(
+	metaHdr *block.MetaBlock,
+) {
+	hash := t.mapFinalizedEpochs[metaHdr.Epoch]
+	epochStartMetaHdr := t.mapEpochStartHdrs[hash]
+	if check.IfNil(epochStartMetaHdr) {
+		return
+	}
+
+	isHeaderOnTopOfFinalityAttestingRound := metaHdr.Nonce == epochStartMetaHdr.Nonce+2
```
iulianpascalau, line 379: magic number 2. Might use [something].Finality constant defined somewhere in process. Might pull that constant in core as it is specific to our protocol.

epochStart/shardchain/trigger.go
```diff
@@ -367,7 +367,27 @@ func (t *trigger) SetAppStatusHandler(handler core.AppStatusHandler) error {
 	return nil
 }
 
-func (t *trigger) setLowerEpochFinalityAttestingRoundIfNeeded(metaHdr *block.MetaBlock) {
+func (t *trigger) changeEpochFinalityAttestingRoundIfNeeded(
```
iulianpascalau, line 370: some unit tests?

epochStart/shardchain/trigger.go
```diff
@@ -741,7 +773,20 @@ func (t *trigger) getHeaderWithNonceAndPrevHash(nonce uint64, prevHash []byte) (
 		return nil, marshal.ErrUnmarshallingBadSize
 	}
 
-	return t.getHeaderWithNonceAndHash(nonce, b.Data[0])
+	metaHdrs := make([]*block.MetaBlock, 0)
```
SebastianMarian, line 776: I think that this new part of code is useless as always the return of the dataHdr, err := t.metaNonceHdrStorage.Get(nonceToByteSlice)
will have maximum one record. As this storage will overwrite the latest nonce committed. Also the batch used above is useless

epochStart/shardchain/trigger.go
sasurobert: done.

epochStart/shardchain/trigger.go
sasurobert: it is useless to search in storage at that moment. the finality attesting metaheader should be in the pool or in the local cache.

sasurobert: reverted as lower round search was done in previous iterations.

epochStart/shardchain/trigger.go
sasurobert: added.

epochStart/metachain/epochStartData.go
```diff
@@ -205,6 +207,10 @@ func (e *epochStartData) lastFinalizedFirstPendingListHeadersForShard(shardHdr *
 		prevShardHdr, err := process.GetShardHeader(currentHdr.GetPrevHash(), e.dataPool.Headers(), e.marshalizer, e.store)
 		if err != nil {
 			go e.requestHandler.RequestShardHeader(currentHdr.ShardID, currentHdr.GetPrevHash())
+			if e.epochStartTrigger.Epoch()-currentHdr.GetEpoch() > maxEpochDifference {
+				log.Warn("shard remained in too old epoch", "shardID", currentHdr.ShardID, "shard Epoch", currentHdr.Epoch, "meta Epoch", e.epochStartTrigger.Epoch())
```
iulianpascalau, line 211: "shard remained in an epoch that is too old"

sasurobert: done.

PR #1432, opened 2020-04-16
Rating Modifier for Voting Power computation

raduchis:
In order to have the some kind of voting power in the explorer the "ratingModifier" has been added. This computes the proportion between the chances for the currentRating and the chances for the startRating as a relative value.  

- changed startRating to 5000001 in order to have 20 chances (100%) because the threshold for 19 chances is 5000000
- changed the time to reach max rating to 72 hours as described in the economics document
- added RatingModifier property to peerAccount, ValidatorInfo and ValidatorApiResponse
- added unit tests


process/peer/process.go
```diff
@@ -401,13 +401,18 @@ func (vs *validatorStatistics) getValidatorDataFromLeaves(
 }
 
 func (vs *validatorStatistics) peerAccountToValidatorInfo(peerAccount state.PeerAccountHandler) *state.ValidatorInfo {
+	chance := vs.rater.GetChance(peerAccount.GetRating())
+	startRatingChance := vs.rater.GetChance(vs.rater.GetStartRating())
```
sasurobert, line 405: Do you really need to save it to the trie If it can be computed easily?

PR #1438, opened 2020-04-17
updated peerType and shardId for offline nodes

raduchis:
For the inactive validators the shardId and peertype were not updated during the offline period. This also sets the list=observer for a node that is below the ratingThreshold.

- added method to recompute the shardId and peerType for the offline nodes
- added display of the first epoch configuration in the nodesCoordinator
- added HideInactiveValidatorIntervalInHours in order to not display the inactive observers if they are offline for more than HideInactiveValidatorIntervalInHours


node/heartbeat/export_test.go
```diff
@@ -70,6 +70,18 @@ func (hbmi *heartbeatMessageInfo) GetTotalDownTime() time.Duration {
 	return hbmi.totalDownTime
 }
 
+func (hbmi *heartbeatMessageInfo) GetComputedShardId() uint32 {
```
iulianpascalau, line 73: Missing comments

node/heartbeat/heartbeatMessageInfo.go
```diff
@@ -163,6 +163,17 @@ func (hbmi *heartbeatMessageInfo) HeartbeatReceived(
 	hbmi.isActive = true
 }
 
+func (hbmi *heartbeatMessageInfo) UpdateShardAndPeerType(
```
iulianpascalau, line 166: missing comments

node/heartbeat/monitor.go
```diff
@@ -381,9 +381,30 @@ func (m *Monitor) computeAllHeartbeatMessages() {
...
+
+		peerType, shardId := m.computePeerTypeAndShardID([]byte(key))
+		if v.peerType != peerType || v.computedShardID != shardId {
+			v.UpdateShardAndPeerType(shardId, peerType)
+			inactiveHbChangedMap[key] = v
+		}
+	}
+
+	m.mutHeartbeatMessages.Unlock()
+	go m.SaveMultipleHeartbeatMessageInfos(inactiveHbChangedMap)
```
iulianpascalau, line 401: this is called from a function that is called GetHeartbeats. It does saving. Same as what happens inside computeAllHeartbeatMessages function

node/heartbeat/export_test.go
raduchis: it,s in export_test. we didn't comment these methods until now. but fixed

node/heartbeat/heartbeatMessageInfo.go
raduchis: fixed

node/heartbeat/monitor.go
raduchis: fixed

node/heartbeat/export_test.go
iulianpascalau: yeah, saw that now. Will try not ask this in the future :D

node/heartbeat/monitor.go
```diff
@@ -229,7 +237,9 @@ func (m *Monitor) SetAppStatusHandler(ash core.AppStatusHandler) error {
 		return ErrNilAppStatusHandler
 	}
 
+	m.mutAppStatusHandler.Lock()
```
iulianpascalau, line 240: :+1:

node/heartbeat/monitor.go
```diff
@@ -403,13 +415,11 @@ func (m *Monitor) computeInactiveHeartbeatMessages() {
...
 // GetHeartbeats returns the heartbeat status
 func (m *Monitor) GetHeartbeats() []PubKeyHeartbeat {
-	m.computeAllHeartbeatMessages()
-	m.computeInactiveHeartbeatMessages()
-
 	m.mutHeartbeatMessages.Lock()
 	status := make([]PubKeyHeartbeat, len(m.heartbeatMessages))
 	idx := 0
 	for k, v := range m.heartbeatMessages {
+		v.updateMutex.Lock()
```
iulianpascalau, line 422: We use double locking (prone to deadlocks). m.mutHeartbeatMessages.Lock() and v.updateMutex.Lock()
Why do we need the newly added mutex?

config/config.go
```diff
@@ -161,6 +161,7 @@ type HeartbeatConfig struct {
 	MinTimeToWaitBetweenBroadcastsInSec int
 	MaxTimeToWaitBetweenBroadcastsInSec int
 	DurationInSecToConsiderUnresponsive int
+	HbmiRefreshInterval                 uint32
```
iulianpascalau, line 164: HbmiRefreshIntervalInSec* as to be more specific

node/heartbeat/monitor.go
```diff
@@ -377,8 +387,10 @@ func (m *Monitor) computeAllHeartbeatMessages() {
 	m.mutHeartbeatMessages.Unlock()
 	go m.SaveMultipleHeartbeatMessageInfos(hbChangedStateToInactiveMap)
 
+	m.mutAppStatusHandler.Lock()
```
iulianpascalau, line 390: RLock

config/config.go
raduchis: fixed

raduchis: fixed

node/heartbeat/monitor.go
raduchis: removed. works also without because the compute and get are locked by the initial mutex

node/heartbeat/monitor.go
raduchis: fixed

node/heartbeat/monitor.go
```diff
@@ -237,9 +244,9 @@ func (m *Monitor) SetAppStatusHandler(ash core.AppStatusHandler) error {
 		return ErrNilAppStatusHandler
 	}
 
-	m.mutAppStatusHandler.Lock()
+	m.mutAppStatusHandler.RLock()
```
iulianpascalau, line 247: Should revert to Lock-Unlock as you set the pointer

node/heartbeat/monitor.go
```diff
@@ -416,10 +423,12 @@ func (m *Monitor) computeInactiveHeartbeatMessages() {
 // GetHeartbeats returns the heartbeat status
 func (m *Monitor) GetHeartbeats() []PubKeyHeartbeat {
 	m.mutHeartbeatMessages.Lock()
-	status := make([]PubKeyHeartbeat, len(m.heartbeatMessages))
-	idx := 0
+	status := make([]PubKeyHeartbeat, 0, len(m.heartbeatMessages))
 	for k, v := range m.heartbeatMessages {
-		v.updateMutex.Lock()
+		if m.shouldSkipValidator(v) {
+			continue
```
iulianpascalau, line 429: delete(m.heartbeat, k) ?
A node should simply not care about an unresponsive observers.

process/peer/validatorsProvider.go
```diff
@@ -85,6 +85,21 @@ func (vp *validatorsProvider) GetLatestValidators() map[string]*state.ValidatorA
 	return mapToReturn
 }
 
+// GetLatestValidatorInfos gets the latest configuration of validators per shard from the peerAccountsTrie
+func (vp *validatorsProvider) GetLatestValidatorInfos() (map[uint32][]*state.ValidatorInfo, error) {
```
sasurobert, line 89: Is this route protected? As this is costly as processing

process/peer/validatorsProvider.go
```diff
@@ -85,6 +85,21 @@ func (vp *validatorsProvider) GetLatestValidators() map[string]*state.ValidatorA
 	return mapToReturn
 }
 
+// GetLatestValidatorInfos gets the latest configuration of validators per shard from the peerAccountsTrie
```
iulianpascalau, line 88: Network sharding relies on the nodesCoordinator to get the validators on shard. It is better to use this instead?

node/heartbeat/monitor.go
raduchis: fixed

node/heartbeat/monitor.go
raduchis: yes, good point. It should be removed from the heartBeatMessages. If it is online it will be added again.

process/peer/validatorsProvider.go
raduchis: it will be used for updates in the cache in order to have the "latest" version of the validators: mainly the leaving that were processed by the metachain and will be removed in the next epoch.

process/peer/validatorsProvider.go
raduchis: I removed the method, because it will be used in the next version. It will be used to populate a cache and have the latest version for a validator. If it sent a transaction for unstake-ing in the middle of the epoch he will appear as leaving only in trie. It will be a combination between trie and nodesCoordinator.

iulianpascalau: ok

PR #1441, opened 2020-04-18
refactored nodes shuffler

raduchis:
The order in which the validators were removed from the eligible list could be different because of a range on a map with the shards. This combined with the optimized remove from list by index could remove different validators from the eligible and waiting lists.

- range on maps now changed to for with sorted keys
- distributeValidators tests fixed


sharding/xorValidatorShuffler_test.go
```diff
@@ -303,6 +303,7 @@ func Test_promoteWaitingToEligible(t *testing.T) {
 }
 
 func Test_removeValidatorFromListFirst(t *testing.T) {
+	t.Skip()
```
iulianpascalau, line 306: We should fix these 2 tests

sharding/xorValidatorShuffler_test.go
```diff
@@ -332,6 +333,7 @@ func Test_removeValidatorFromListLast(t *testing.T) {
 }
 
 func Test_removeValidatorFromListMiddle(t *testing.T) {
+	t.Skip()
```
sasurobert, line 336: fix test

sharding/xorValidatorShuffler_test.go
raduchis: fixed

sharding/xorValidatorShuffler_test.go
raduchis: fixed

sharding/xorValidatorShuffler.go
```diff
@@ -102,20 +102,28 @@ func (rxs *randXORShuffler) UpdateNodeLists(args ArgsUpdateNodes) (map[uint32][]
 }
 
 func removeFromWaiting(waitingNodes map[uint32][]Validator, leavingNodes []Validator) (map[uint32][]Validator, []Validator) {
```
iulianpascalau, line 104: is this only computing or computing and removing? Doesn't seem right that a function called removeFromWaiting does 2 computing and no actual removing.

sharding/xorValidatorShuffler.go
```diff
@@ -102,20 +102,28 @@ func (rxs *randXORShuffler) UpdateNodeLists(args ArgsUpdateNodes) (map[uint32][]
...
-	for shard, vList := range waitingNodes {
-		nbToRemove := len(vList)
-		if len(leavingNodes) < nbToRemove {
-			nbToRemove = len(leavingNodes)
-		}
-
-		vList, removedNodes := removeValidatorsFromList(vList, leavingNodes, nbToRemove)
-		leavingNodes, _ = removeValidatorsFromList(leavingNodes, removedNodes, len(removedNodes))
-		waitingNodes[shard] = vList
+	for shard := uint32(0); shard < uint32(len(waitingNodes)-1); shard++ {
```
iulianpascalau, line 105: this won't work as expected, and might panic, if we decide that we will call this function on a subset of all shard's validators. Why not take the available keys, sort them and then iterate on them?

sharding/xorValidatorShuffler.go
```diff
@@ -102,20 +102,28 @@ func (rxs *randXORShuffler) UpdateNodeLists(args ArgsUpdateNodes) (map[uint32][]
...
+		leavingNodes = computeLeaving(waitingNodes, leavingNodes, shard)
 	}
 
+	leavingNodes = computeLeaving(waitingNodes, leavingNodes, core.MetachainShardId)
+
 	return waitingNodes, leavingNodes
 }
 
+func computeLeaving(waitingNodes map[uint32][]Validator, leavingNodes []Validator, shard uint32) []Validator {
```
iulianpascalau, line 114: this doesn't just compute, it also updates the waiting nodes

sharding/xorValidatorShuffler.go
```diff
@@ -170,7 +178,8 @@ func shuffleNodesIntraShards(
 
 	waiting, stillRemainingInLeaving = removeFromWaiting(waiting, stillRemainingInLeaving)
 
-	for shard, validators := range eligible {
+	for shard := uint32(0); shard < uint32(len(eligible)-1); shard++ {
```
iulianpascalau, line 181: same thing as my previous comment regarding iterating on a map

sharding/xorValidatorShuffler_test.go
```diff
@@ -918,6 +922,61 @@ func TestRandXORShuffler_UpdateNodeListsWithNonExistentLeavingDoesNotRemove(t *t
...
+
+		args.Leaving = leavingValidators
+
+		initialEligible, initialWaiting, initialLeaving := shuffler.UpdateNodeLists(args)
+
+		for i := 0; i < 100; i++ {
+			newArgs := copyArgs(args)
+			eligible, waiting, leaving := shuffler.UpdateNodeLists(newArgs)
+
+			assert.True(t, areMapsEqual(initialEligible, eligible))
```
iulianpascalau, line 973: why do we need areMapsEqual?
```GO
assert.Equal(t, initialEligible, eligible)
```

sharding/xorValidatorShuffler.go
```diff
@@ -240,13 +258,21 @@ func shuffleOutNodes(
 	newEligible := make(map[uint32][]Validator)
 	var shardShuffledOut []Validator
 
-	for shard, validators := range eligible {
+	for shard := uint32(0); shard < uint32(len(eligible)-1); shard++ {
```
sasurobert, line 261: it think it would have been easier to received [][]Validator instead of map[uint32]validators, this way a lot of code is duplicated, plus shuffler needs to know about metachainshardID.

sharding/xorValidatorShuffler_test.go
```diff
@@ -918,6 +922,61 @@ func TestRandXORShuffler_UpdateNodeListsWithNonExistentLeavingDoesNotRemove(t *t
...
+			leavingValidators[i] = allValidators[randIndex]
+			allValidators = removeValidatorFromList(allValidators, randIndex)
+		}
+
+		args.Leaving = leavingValidators
+
+		initialEligible, initialWaiting, initialLeaving := shuffler.UpdateNodeLists(args)
+
+		for i := 0; i < 100; i++ {
+			newArgs := copyArgs(args)
```
iulianpascalau, line 970: Is this really necessary? Are we altering the provided slices from args?

sharding/xorValidatorShuffler_test.go
```diff
@@ -980,3 +1039,52 @@ func createShufflerArgs(eligiblePerShard int, waitingPerShard int, nbShards uint
...
+	}
+
+	return true
+}
+
+func copyArgs(args ArgsUpdateNodes) ArgsUpdateNodes {
+	newArg := ArgsUpdateNodes{
+		Eligible: copyMap(args.Eligible),
+		Waiting:  copyMap(args.Waiting),
+		NewNodes: append(make([]Validator, 0), args.NewNodes...),
```
iulianpascalau, line 1075: If this function is really necessary, why not using copy function?

sharding/xorValidatorShuffler.go
raduchis: good catch. there are multiple places in the application where we do a range over the maps and expect that the number of shards is equal to the len of the maps.

sharding/xorValidatorShuffler.go
raduchis: good catch. there are multiple places in the application where we do a range over the maps and expect that the number of shards is equal to the len of the maps.

sharding/xorValidatorShuffler_test.go
raduchis: seems easier to copy in one line instead of multiple lines

sharding/xorValidatorShuffler_test.go
raduchis: removed. the range problem seemed to happen every time if the maps were copied. But it also happens in 100 steps with the same map.

sharding/xorValidatorShuffler_test.go
raduchis: nope. seems it works with maps of slices of validators also.

sharding/xorValidatorShuffler.go
raduchis: renamed internal method removeNodes

sharding/xorValidatorShuffler.go
raduchis: renamed to removeNodes

sharding/xorValidatorShuffler.go
raduchis: fixed

sharding/xorValidatorShuffler_test.go
iulianpascalau: `copy(newArg.NewNodes, args.NewNodes)` and removed that line from the ArgsUpdateNodes initialization

sharding/indexHashedNodesCoordinator.go
```diff
@@ -882,3 +885,36 @@ func createValidatorInfoFromBody(
 
 	return allValidatorInfo, nil
 }
+
+func toOrderedSlice(validatorMap map[uint32][]Validator, numShardsPlusMeta uint32) [][]Validator {
+	slicedLists := make([][]Validator, numShardsPlusMeta)
+	for shardId, list := range validatorMap {
+		if shardId < numShardsPlusMeta-1 {
+			slicedLists[shardId] = list
```
iulianpascalau, line 893: This will break in panic if len(map[uint32][]Validator) > numShardsPlusMeta . Not saying that it is possible as the code is written (the method is unexported and the parameters might be carefully crafted) but the code just become unnecessarily fragile. 
Strongly suggest to do this in 3 easy steps:
1. collect keys (shard ids) into a slice of uint32
2. sort that slice
3. iterate the keys

Will work with every map, in any configuration.

A better, cleaner solution is to have a DTO that will have 2 fields: the validator list and the shardID, then you could do it in 2 easy steps:
1. iterate the map, contruct the DTO for each shard, append it to the slice
2. sort the resulting DTO slice by the shard ID field

The advantage of the latest is that the code will become easier to read, not to mention the fact that now we have to keep in mind that the last item is the metachain's ID. Also, there you will not need the numShardsPlusMeta variable injected just about everywhere.

sharding/indexHashedNodesCoordinator.go
```diff
@@ -496,15 +496,18 @@ func (ihgs *indexHashedNodesCoordinator) EpochStartPrepare(metaHdr data.HeaderHa
...
+		Eligible: toOrderedSlice(newNodesConfig.eligibleMap, newNodesConfig.nbShards),
+		Waiting:  toOrderedSlice(newNodesConfig.waitingMap, newNodesConfig.nbShards),
 		NewNodes: newNodesConfig.newList,
 		Leaving:  newNodesConfig.leavingList,
 		Rand:     randomness,
 		NbShards: newNodesConfig.nbShards,
 	}
 
-	eligibleMap, waitingMap, leaving := ihgs.shuffler.UpdateNodeLists(shufflerArgs)
+	eligibleLists, waitingLists, leaving := ihgs.shuffler.UpdateNodeLists(shufflerArgs)
```
iulianpascalau, line 507: Why does the UpdateNodeLists update a slice of slices? Why not a map directly?

sharding/interface.go
```diff
@@ -74,7 +74,7 @@ type ArgsUpdateNodes struct {
 // NodesShuffler provides shuffling functionality for nodes
 type NodesShuffler interface {
 	UpdateParams(numNodesShard uint32, numNodesMeta uint32, hysteresis float32, adaptivity bool)
-	UpdateNodeLists(args ArgsUpdateNodes) (map[uint32][]Validator, map[uint32][]Validator, []Validator)
+	UpdateNodeLists(args ArgsUpdateNodes) ([][]Validator, [][]Validator, []Validator)
```
iulianpascalau, line 77: As seen in other parts returning the same parameter type more than once and in a consecutive manner might lead to mistakenly return wrong variables (of the same type) leading into hard-to-catch bugs. Suggestion (for the next PRs) to return a DTO

sharding/xorValidatorShuffler.go
```diff
@@ -73,9 +71,9 @@ func (rxs *randXORShuffler) UpdateParams(
 //          b)  In case (shuffled out nodes + new nodes) < (nbShards * perShardHysteresis) then we can immediately
 //              execute the shard merge
 //          c)  No change in the number of shards then nothing extra needs to be done
-func (rxs *randXORShuffler) UpdateNodeLists(args ArgsUpdateNodes) (map[uint32][]Validator, map[uint32][]Validator, []Validator) {
-	eligibleAfterReshard := copyValidatorMap(args.Eligible)
-	waitingAfterReshard := copyValidatorMap(args.Waiting)
+func (rxs *randXORShuffler) UpdateNodeLists(args ArgsUpdateNodes) ([][]Validator, [][]Validator, []Validator) {
```
sasurobert, line 74: I think receiving as map and returning as map should remain, as it is cleaner for the component which calls on the top. Doing the translation from map to slice could happen here in this public method, and all the other method could remain with lists of list.

sharding/indexHashedNodesCoordinator.go
```diff
@@ -882,3 +885,36 @@ func createValidatorInfoFromBody(
...
+
+	return slicedLists
+}
+
+func toMap(slicedLists [][]Validator, numShardsPlusMeta uint32) map[uint32][]Validator {
+	validatorsMap := make(map[uint32][]Validator)
+
+	for shardId := uint32(0); shardId < numShardsPlusMeta; shardId++ {
+		validatorsList := slicedLists[shardId]
+		if validatorsList == nil {
```
sasurobert, line 908: this should not happen. all integration tests now have at least 1 node per shard.

sharding/xorValidatorShuffler.go
```diff
@@ -249,30 +240,27 @@ func (rxs *randXORShuffler) computeNewShards(
...
+	numShardsPlusMeta uint32,
+) ([]Validator, [][]Validator, []Validator) {
 	shuffledOut := make([]Validator, 0)
-	newEligible := make(map[uint32][]Validator)
+	newEligible := make([][]Validator, numShardsPlusMeta)
 	var shardShuffledOut []Validator
 
-	for shard := uint32(0); shard < uint32(len(eligible)-1); shard++ {
+	for shard := uint32(0); shard < numShardsPlusMeta; shard++ {
 		validators := eligible[shard]
```
iulianpascalau, line 254: same problem as stated above if the numShardsPlusMeta is larger than len(eligible) this will break.

sharding/indexHashedNodesCoordinator.go
sasurobert: this is a clear error should be treated

sharding/xorValidatorShuffler.go
```diff
@@ -386,31 +374,31 @@ func xorBytes(a []byte, b []byte) []byte {
...
 
-// copyValidatorMap creates a copy for the Validators map, creating copies for each of the lists for each shard
-func copyValidatorMap(validators map[uint32][]Validator) map[uint32][]Validator {
-	result := make(map[uint32][]Validator)
+// copyValidatorList creates a copy for the Validators list, creating copies for each of the lists for each shard
+func copyValidatorList(validators [][]Validator) [][]Validator {
+	result := make([][]Validator, len(validators))
 
 	for k, v := range validators {
 		elems := make([]Validator, 0)
```
iulianpascalau, line 404: `copy(result[k], v)` on the next line

sharding/xorValidatorShuffler.go
raduchis: not the case any more. using: sort keys on map

sharding/xorValidatorShuffler.go
raduchis: not the case any more

sharding/xorValidatorShuffler_test.go
raduchis: also a initialization with the len has to be done first. fixed

sharding/xorValidatorShuffler.go
raduchis: undone and used sorted keys on map

sharding/indexHashedNodesCoordinator.go
raduchis: undone. not the case any more

sharding/indexHashedNodesCoordinator.go
raduchis: undone. used sorted keys on map

sharding/indexHashedNodesCoordinator.go
raduchis: undone. fixed

sharding/xorValidatorShuffler.go
raduchis: undone. not the case any more

sharding/xorValidatorShuffler.go
raduchis: fixed

sharding/interface.go
raduchis: fixed. Added ResUpdateNodes

PR #1447, opened 2020-04-21
EN-6253: api routes config file

bogdan-rosianu:
Added an `api.toml` file which contains all the endpoints available in the node and options to enable or disable them. This way, if a node doesn't want to expose some (or all) of the API routes, this can be done now.


config/tomlConfig_test.go
```diff
@@ -207,3 +207,72 @@ func TestTomlExternalParser(t *testing.T) {
...
+         # /transaction/cost will receive a single transaction in JSON format and will return the estimated cost of it
+         { Name = "/transaction/cost", Open = true }
+	]
+ `
+
+	cfg := ApiRoutesConfig{}
+
+	err := toml.Unmarshal([]byte(testString), &cfg)
+
+	assert.Nil(t, err)
```
sasurobert, line 277: shoud add some asserts that the unmarshalled data is ok.

bogdan-rosianu: sure. that test was just for checking toml structs. forgot to update the test
done

api/address/routes.go
```diff
@@ -28,9 +30,17 @@ type accountResponse struct {
...
 // Routes defines address related routes
-func Routes(router *gin.RouterGroup) {
-	router.GET("/:address", GetAccount)
-	router.GET("/:address/balance", GetBalance)
+func Routes(router *gin.RouterGroup, routesConfig config.ApiRoutesConfig) {
+	addresRoutes, ok := routesConfig.APIPackages["address"]
+	if !ok {
+		return
+	}
+	if configparser.CheckEndpoint(":address", addresRoutes) {
```
ccorcoveanu, line 38: `CheckEndpoint` doesn't say much about what it's going to check. `IsEndpointActive` maybe?

api/configparser/configParser.go
```diff
@@ -0,0 +1,14 @@
+package configparser
+
+import "github.com/ElrondNetwork/elrond-go/config"
+
+// CheckEndpoint will return true if a configuration for the given endpoint exists and it is configured to be opened
+func CheckEndpoint(endpointToCheck string, routesConfig config.APIPackageConfig) bool {
```
ccorcoveanu, line 6: To simplify the config, we should not bloat it with every route. We could just simply define routes that we don't want active, this function (or `IsEndpointActive`) can return true if route is not defined

api/hardfork/routes.go
```diff
@@ -17,8 +19,14 @@ type TriggerHardforkHandler interface {
...
 
 // Routes defines node related routes
-func Routes(router *gin.RouterGroup) {
-	router.POST("/trigger", Trigger)
+func Routes(router *gin.RouterGroup, routesConfig config.ApiRoutesConfig) {
+	hardforkRoutes, ok := routesConfig.APIPackages["hardfork"]
+	if !ok {
+		return
+	}
+	if configparser.CheckEndpoint("trigger", hardforkRoutes) {
```
ccorcoveanu, line 27: We could refactor this a bit to look cleaner. Instead of having `if active -> then` on every route: We can have a general wrappers for `Get` and `Post` that make this check and have the condition only in one place

api/configparser/configParser.go
bogdan-rosianu: Yes, it would be easier. But that toml file can now be seen as a minimal API routes documentation because all routes are listed there and you'll have a better overview over what to open or not.

api/address/routes.go
bogdan-rosianu: done

api/hardfork/routes.go
bogdan-rosianu: done. made a wrapper over RouterGroup

PR #1453, opened 2020-04-22
Added num of shards in node/status route

iulianpascalau:
- added new info line in node/status route that will specify the number of shards without metachain
- added number of shards and shard ID to the automatic refreshing system in order to correctly specify the shard ID if the node does not restart


cmd/node/metrics/metrics.go
```diff
@@ -105,6 +107,7 @@ func StartStatusPolling(
 	pollingInterval time.Duration,
 	networkComponents *factory.Network,
 	processComponents *factory.Process,
+	shardCoordinator sharding.Coordinator,
```
sasurobert, line 110: there are no nil checks

iulianpascalau: yup, old code :(

iulianpascalau: done

PR #1463, opened 2020-04-23
storage: get from batch, remove with batch

AdoAdoAdo:
added a cache for batched data
Previously get was not possible to return data from batch, now it returns the data from cache.
Also remove had to go directly to disk and it could have been very disk intensive.
It is no longer required to correlate the sizes of cache and batch for storers.

### DIFF
--- a/storage/leveldb/batch.go
+++ b/storage/leveldb/batch.go
@@ -1,35 +1,60 @@
 package leveldb
 
 import (
+	"sync"
+
 	"github.com/syndtr/goleveldb/leveldb"
 )
 
+const removed = "removed"
+
 type batch struct {
-	batch *leveldb.Batch
+	batch         *leveldb.Batch
+	cachedData    map[string][]byte
+	mutCachedData sync.RWMutex
 }
 
 // NewBatch creates a batch
 func NewBatch() *batch {
 	return &batch{
-		batch: &leveldb.Batch{},
+		batch:         &leveldb.Batch{},
+		cachedData:    make(map[string][]byte),
+		mutCachedData: sync.RWMutex{},
 	}
 }
 
 // Put inserts one entry - key, value pair - into the batch
 func (b *batch) Put(key []byte, val []byte) error {
+	b.mutCachedData.Lock()
 	b.batch.Put(key, val)
+	b.cachedData[string(key)] = val
+	b.mutCachedData.Unlock()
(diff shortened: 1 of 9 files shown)


storage/leveldb/batch.go
```diff
@@ -1,35 +1,60 @@
...
 	"github.com/syndtr/goleveldb/leveldb"
 )
 
+const removed = "removed"
+
 type batch struct {
-	batch *leveldb.Batch
+	batch         *leveldb.Batch
+	cachedData    map[string][]byte
+	mutCachedData sync.RWMutex
```
iulianpascalau, line 14: since this mutex also protects against the simultaneous batch writes (which is very good), might rename it to mutBatch

storage/leveldb/leveldb.go
```diff
@@ -94,13 +95,7 @@ func (s *DB) batchTimeoutHandle() {
...
 
-// Put adds the value to the (key, val) storage medium
-func (s *DB) Put(key, val []byte) error {
-	err := s.batch.Put(key, val)
-	if err != nil {
-		return err
-	}
-
+func (s *DB) updateBatchWithIncrement() error {
 	s.mutBatch.Lock()
```
iulianpascalau, line 99: I think this mutex is no longer required, the batch struct just became concurrent safe and we are reusing the same batch

storage/leveldb/leveldbSerial.go
```diff
@@ -90,29 +91,34 @@ func (s *SerialDB) batchTimeoutHandle(ctx context.Context) {
...
+}
+
 // Put adds the value to the (key, val) storage medium
 func (s *SerialDB) Put(key, val []byte) error {
 	if s.isClosed() {
 		return storage.ErrSerialDBIsClosed
 	}
 
-	s.mutBatch.Lock()
+	s.mutBatch.RLock()
```
iulianpascalau, line 114: :+1: we need this mutex in this impl. Rlocks should improve performance

storage/leveldb/batch.go
AdoAdoAdo: done

storage/leveldb/leveldb.go
AdoAdoAdo: it is required as it also protects the current size of the batch

PR #1466, opened 2020-04-24
Fix trie snapshot concurrency bug

BeniaminDrasovean:


data/state/peerAccountsDB.go
```diff
@@ -55,6 +55,14 @@ func (adb *PeerAccountsDB) SnapshotState(rootHash []byte) {
 	adb.mainTrie.ExitSnapshotMode()
 }
 
+// SetStateCheckpoint triggers the checkpointing process of the state trie
+func (adb *PeerAccountsDB) SetStateCheckpoint(rootHash []byte) {
+	log.Trace("peerAccountsDB.SetStateCheckpoint", "root hash", rootHash)
+	adb.mainTrie.EnterSnapshotMode()
+	adb.mainTrie.SetCheckpoint(rootHash)
```
iulianpascalau, line 62: This already calls adb.mainTrie.EnterSnapshotMode()
Why do we still have exported functions for EnterSnapshotMode() and ExitSnapshotMode() ?

PR #1467, opened 2020-04-24
En 6192 trie sync on list of hashes

sasurobert:
Request and resolve trie by hashes


data/trie/sync.go
```diff
@@ -241,11 +241,13 @@ func trieNode(data interface{}) (node, error) {
...
 	ts.mutOperation.RLock()
 	numUnResolvedNodes := uint32(len(ts.nodesForTrie))
+	hashes := make([][]byte, 0)
 	for hash, nodeInfo := range ts.nodesForTrie {
 		if !nodeInfo.received {
-			ts.requestHandler.RequestTrieNodes(ts.shardId, []byte(hash), ts.topic)
+			hashes = append(hashes, []byte(hash))
 		}
 	}
+	ts.requestHandler.RequestTrieNodes(ts.shardId, hashes, ts.topic)
```
iulianpascalau, line 250: we might trim this slice to first n (n = const) hashes

dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -108,34 +108,35 @@ func (rrh *resolverRequestHandler) requestByHashes(destShardID uint32, hashes []
 		log.Trace("requestByHashes", "hash", txHash)
 	}
 
-	go func() {
-		dataSplit := &partitioning.DataSplit{}
-		var sliceBatches [][][]byte
-		sliceBatches, err = dataSplit.SplitDataInChunks(unrequestedHashes, rrh.maxTxsToRequest)
+	go rrh.requestHashesWithDataSplit(unrequestedHashes, txResolver)
+
+	rrh.addRequestedItem(unrequestedHashes)
```
iulianpascalau, line 113: addRequestedItem**s**

dataRetriever/resolvers/trieNodeResolver.go
```diff
@@ -75,28 +75,76 @@ func (tnRes *TrieNodeResolver) ProcessReceivedMessage(message p2p.MessageP2P, fr
...
+		if remainingSpace >= lenNextNodes {
+			remainingSpace -= uint64(len(nextNodes))
 		}
 
-		var buff []byte
-		buff, err = tnRes.marshalizer.Marshal(&batch.Batch{Data: serializedNodes})
-		if err != nil {
-			return err
+		if lenNextNodes == 0 || remainingSpace == 0 {
+			break
```
iulianpascalau, line 110: this is a good thing to stop until you have reached the maximum capacity to send
The condition needs to be changed. As we might have remainingSpace = 1 which is not enough to send something else. Maybe add an else branch on line 107 that will set a bool that will be used on line 109

data/trie/sync.go
sasurobert: would not trim. it is hard to adjust to a number. Requests and resolves are done in batches.

dataRetriever/requestHandlers/requestHandler.go
sasurobert: done

dataRetriever/resolvers/trieNodeResolver.go
sasurobert: good catch. done

PR #1470, opened 2020-04-24
Collect more sigs if possible

SebastianMarian:
* Implemented a mechanism through which the proposer will wait for more than 2/3+1 signatures in subround signature until max subround allocated time would be reached
* Fixed an edge case when final check point was not re-computed on rollback after forkDetector.RemoveHeader method was called
* Fixed an edge case on sync, when received requested missing miniblocks were not added in the same order in block.Body structure which should be passed to blockProcessor.ProcessBlock method


consensus/spos/bls/subroundSignature.go
```diff
@@ -239,3 +239,31 @@ func (sr *subroundSignature) signaturesCollected(threshold int) (bool, int) {
 
 	return n >= threshold, n
 }
+
+func (sr *subroundSignature) waitAllSignatures() {
```
sasurobert, line 243: it would be easier for this func to get a context with cancel - and that context can be with timeout. Calculated only once with the remainingTime func.

consensus/spos/bls/subroundSignature_test.go
```diff
@@ -311,19 +311,6 @@ func TestSubroundSignature_ReceivedSignature(t *testing.T) {
 	assert.False(t, r)
 
 	sr.SetSelfPubKey(sr.ConsensusGroup()[0])
```
sasurobert, line 313: add unit test when receiving enough, exactly 2/3+1, more signatures.

consensus/spos/bls/subroundSignature.go
```diff
@@ -239,3 +239,31 @@ func (sr *subroundSignature) signaturesCollected(threshold int) (bool, int) {
...
+	for {
```
iulianpascalau, line 244: I think this for is useless
Following Robert's suggestion the code might look like this:
```GO
func (sr *subroundSignature) waitAllSignatures(ctx context.Context) {
   remainingTime := sr.remainingTime()
   select{
       case <- ctx.Done:
            return
       case <-time.After(remainginTime): 
   }
   
   if sr.IsSubroundFinished(sr.Current()) {
        return 
   }

   select {
      case sr.ConsensusChannel() <- true:
      default:
   }

}
```

consensus/spos/bls/subroundSignature.go
SebastianMarian: Done

consensus/spos/bls/subroundSignature.go
SebastianMarian: Done

consensus/spos/bls/subroundSignature_test.go
SebastianMarian: Done

consensus/spos/bls/subroundSignature.go
```diff
@@ -239,3 +251,26 @@ func (sr *subroundSignature) signaturesCollected(threshold int) (bool, int) {
...
+	remainingTime := sr.remainingTime()
+
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(remainingTime):
+	}
+
+	select {
+	case sr.ConsensusChannel() <- true:
```
sasurobert, line 265: it is not so good combining a channel and a context. Why is this needed ? the func could exit before ctx.Done or time.After because of consensusChannel() <- true

SebastianMarian: Actually this is the channel which I need to trigger after remaining time expires and if until then leader would not received all signatures and finished subround signature meantime. This trigger will end the sub-round signature with the collected signatures no matter if they are all of them or not. So I think that I didn't need here a context approach but a little refactor to the initial implementation. I propose the below one:

SebastianMarian: func (sr *subroundSignature) waitAllSignatures() {
	remainingTime := sr.remainingTime()
	time.Sleep(remainingTime)

	if sr.IsSubroundFinished(sr.Current()) {
		return
	}

	select {
	case sr.ConsensusChannel() <- true:
	default:
	}
}

process/sync/baseSync.go
```diff
@@ -870,24 +870,51 @@ func (boot *baseBootstrap) requestMiniBlocksByHashes(hashes [][]byte) {
...
 func (boot *baseBootstrap) getMiniBlocksRequestingIfMissing(hashes [][]byte) (block.MiniBlockSlice, error) {
-	miniBlocks, missingMiniBlocksHashes := boot.miniBlocksProvider.GetMiniBlocksFromPool(hashes)
-	if len(missingMiniBlocksHashes) > 0 {
-		_ = core.EmptyChannel(boot.chRcvMiniBlocks)
-		boot.requestMiniBlocksByHashes(missingMiniBlocksHashes)
-		err := boot.waitForMiniBlocks()
-		if err != nil {
-			return nil, err
-		}
+	miniBlocks, miniBlocksHashes, missingMiniBlocksHashes := boot.miniBlocksProvider.GetMiniBlocksFromPool(hashes)
```
iulianpascalau, line 873: Could have made another PR for this changes.

process/sync/export_test.go
```diff
@@ -11,7 +11,7 @@ func (boot *ShardBootstrap) RequestHeaderWithNonce(nonce uint64) {
 	boot.requestHeaderWithNonce(nonce)
 }
 
-func (boot *ShardBootstrap) GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
+func (boot *ShardBootstrap) GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte, [][]byte) {
```
iulianpascalau, line 14: Same here, I would avoid returning the same parameter type in consecutive variables

consensus/spos/bls/subroundSignature.go
iulianpascalau: Just for the record: context and channels are go idiomatic. Since this function is launched on a go routine, it will end up having a context.

process/interface.go
```diff
@@ -725,8 +725,8 @@ type ValidityAttester interface {
 
 // MiniBlockProvider defines what a miniblock data provider should do
 type MiniBlockProvider interface {
-	GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte)
-	GetMiniBlocksFromPool(hashes [][]byte) (block.MiniBlockSlice, [][]byte)
+	GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte, [][]byte)
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

dataRetriever/provider/miniBlocks.go
```diff
@@ -45,32 +45,36 @@ func NewMiniBlockProvider(arg ArgMiniBlockProvider) (*miniBlockProvider, error)
 }
 
 // GetMiniBlocks method returns a list of deserialized mini blocks from a given hash list either from data pool or from storage
-func (mbp *miniBlockProvider) GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte) {
-	missingMiniBlocksHashes := make([][]byte, 0)
+func (mbp *miniBlockProvider) GetMiniBlocks(hashes [][]byte) (block.MiniBlockSlice, [][]byte, [][]byte) {
```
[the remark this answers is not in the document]
SebastianMarian: Refactored

process/sync/baseSync.go
SebastianMarian: Yep, but I wanted to have these hot fixes as soon as possible in dev

process/sync/export_test.go
SebastianMarian: Refactored

consensus/spos/bls/subroundSignature.go
```diff
@@ -76,7 +78,9 @@ func (sr *subroundSignature) doSignatureJob() bool {
 		return false
 	}
 
-	if !sr.IsSelfLeaderInCurrentRound() {
+	isSelfLeader := sr.IsSelfLeaderInCurrentRound()
+
```
sasurobert, line 82: delete empty line

process/sync/baseSync.go
iulianpascalau: ok

PR #1472, opened 2020-04-24
bugFix/epoch-start-miniblocks-pending-processed

sasurobert:
Recalibrated how processed and pending miniblocks are calculated in case of start in epoch. Added requests on processing of start in epoch to request the missing mteablocks.


process/block/shardblock.go
```diff
@@ -276,15 +274,28 @@ func (sp *shardProcessor) requestEpochStartInfo(header *block.Header, haveTime f
 
 	go sp.requestHandler.RequestMetaHeader(header.EpochStartMetaHash)
 
+	headersPool := sp.dataPool.Headers()
 	for {
-		time.Sleep(time.Millisecond)
+		time.Sleep(100 * time.Millisecond)
```
iulianpascalau, line 279: magic values

sasurobert: done

process/block/shardblock.go
```diff
@@ -276,15 +274,28 @@ func (sp *shardProcessor) requestEpochStartInfo(header *block.Header, haveTime f
...
 		if haveTime() < 0 {
 			break
 		}
 
 		if sp.epochStartTrigger.IsEpochStart() {
 			return nil
 		}
+
+		gotHdr, err := headersPool.GetHeaderByHash(header.EpochStartMetaHash)
```
SebastianMarian, line 290: epochStartMetaBlock

process/block/shardblock.go
```diff
@@ -276,15 +274,28 @@ func (sp *shardProcessor) requestEpochStartInfo(header *block.Header, haveTime f
...
+		if err != nil || check.IfNil(gotHdr) {
```
SebastianMarian, line 291: err != nil would is enough

process/block/shardblock.go
```diff
@@ -399,6 +411,21 @@ func (sp *shardProcessor) checkEpochCorrectness(
 	return nil
 }
 
+func (sp *shardProcessor) getOldEpochStartMetaHdr(header *block.Header) (data.HeaderHandler, error) {
```
SebastianMarian, line 417: Unused

process/block/shardblock.go
sasurobert: done

process/block/shardblock.go
sasurobert: done

process/block/shardblock.go
sasurobert: done

process/block/shardblock.go
```diff
@@ -399,6 +411,21 @@ func (sp *shardProcessor) checkEpochCorrectness(
...
+	metaBlock, err := sp.dataPool.Headers().GetHeaderByHash(header.EpochStartMetaHash)
```
SebastianMarian, line 417: I think this line could be replaced with:
	metaBlock, err := process.GetMetaHeader(header.EpochStartMetaHash, sp.dataPool.Headers(), sp.marshalizer, sp.store)

In this way also header.EpochStartMetaHash would be searched first in pool and then if it is not found there in cacher->storage

sasurobert: no need for that, as in case of storage it should be saved with the epoch trigger key.

SebastianMarian: ok

epochStart/metachain/epochStartData_test.go
```diff
@@ -378,3 +431,112 @@ func TestMetaProcessor_CreateEpochStartFromMetaBlockShouldWork(t *testing.T) {
 	err = epoch.VerifyEpochStartDataForMetablock(&block.MetaBlock{EpochStart: *epStart})
 	assert.Nil(t, err)
 }
+
+func TestMetaProcessor_CreateEpochStartFromMetaBlockEdgeCaseChecking(t *testing.T) {
```
iulianpascalau, line 435: :+1:

PR #1480, opened 2020-04-27
Integrate hardfork in genesis process

sasurobert:


genesis/process/genesisBlockCreator.go
```diff
@@ -29,11 +34,58 @@ func NewGenesisBlockCreator(arg ArgsGenesisBlockCreator) (*genesisBlockCreator,
...
-	return &genesisBlockCreator{
+	gbc := &genesisBlockCreator{
 		arg:                 arg,
 		shardCreatorHandler: CreateShardGenesisBlock,
 		metaCreatorHandler:  CreateMetaGenesisBlock,
-	}, nil
+	}
+
+	if arg.HardForkConfig.MustImport {
+		err := gbc.createHardForkImportHandler()
```
iulianpascalau, line 44: we should think of a way to invalidate the must import flag after the first successful sync of the network. Might use a max nonce until this flag is validated?

genesis/process/metaGenesisBlockCreator.go
```diff
@@ -65,6 +66,15 @@ func CreateMetaGenesisBlock(arg ArgsGenesisBlockCreator) (data.HeaderHandler, er
 		return nil, err
 	}
 
+	if arg.HardForkConfig.MustImport {
+		//TODO: think how to integrate when genesis is modified as well - how to integrate with imported data
+		// one example being - delete or not the validator statistics trie - and restart from 0, or should import from old
```
iulianpascalau, line 71: Answer & discussed: recreate the validator statistics trie from the (newly) nodesSetup.json. Or a new predictable nodesSetup_1.json or something like this.

genesis/process/shardGenesisBlockCreator.go
```diff
@@ -36,6 +37,11 @@ func CreateShardGenesisBlock(arg ArgsGenesisBlockCreator) (data.HeaderHandler, e
 	}
 
 	//TODO add here delegation process
+	if arg.HardForkConfig.MustImport {
+		//TODO think about how to integrate when genesis is modified as well for hardfork - when should set balances be called
```
iulianpascalau, line 41: Is a must to keep the old, original genesis block? No one will care about it except the entities that want to sync from the start.

process/block/baseProcess.go
```diff
@@ -118,7 +119,7 @@ func (bp *baseProcessor) checkBlockValidity(
 	currentBlockHeader := bp.blockChain.GetCurrentBlockHeader()
 
 	if check.IfNil(currentBlockHeader) {
-		if headerHandler.GetNonce() == 1 { // first block after genesis
+		if headerHandler.GetNonce() == bp.genesisNonce+1 { // first block after genesis
```
iulianpascalau, line 122: :+1:

genesis/process/genesisBlockCreator.go
sasurobert: after a hardfork everybody has to start and create the genesis block through this method - so it is ok. MustImport as a flag will be set only when there is a hardfork - for that you need to set export and import storage / folder.

genesis/process/shardGenesisBlockCreator.go
sasurobert: agreed. at this moment it we do not care about initial genesis.

genesis/process/shardGenesisBlockCreator.go
```diff
@@ -53,6 +59,59 @@ func CreateShardGenesisBlock(arg ArgsGenesisBlockCreator) (data.HeaderHandler, e
...
+		arg.ChainID,
+		arg.HardForkConfig.StartRound,
+		arg.HardForkConfig.StartNonce,
+		arg.HardForkConfig.StartEpoch,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	errNotCritical := processors.txCoordinator.SaveBlockDataToStorage(bodyHandler.(*block.Body))
```
SebastianMarian, line 104: I would extract bodyHandler.(*block.Body) into variable and check it against type assertion before saving it into storage

process/block/metablock.go
```diff
@@ -1626,8 +1641,11 @@ func (mp *metaProcessor) computeExistingAndRequestMissingShardHeaders(metaBlock
 	mp.hdrsForCurrBlock.mutHdrsForBlock.Lock()
 	defer mp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 
-	for i := 0; i < len(metaBlock.ShardInfo); i++ {
-		shardData := metaBlock.ShardInfo[i]
+	for _, shardData := range metaBlock.ShardInfo {
+		if shardData.Nonce == 0 {
+			continue
```
SebastianMarian, line 1646: Do we not need to execute lines 1163-1670 before continue? Maybe later we need to read from mp.hdrsForCurrBlock.hdrHashAndInfo the genesis shard header(to save it or other stuff) , or we should add this skip everywhere.

process/peer/process.go
```diff
@@ -690,6 +693,10 @@ func (vs *validatorStatistics) updateShardDataPeerState(header data.HeaderHandle
 	}
 
 	for _, h := range metaHeader.ShardInfo {
+		if h.Nonce == 0 {
```
SebastianMarian, line 696: == vs.genesisNonce

process/block/metablock.go
```diff
@@ -1626,8 +1641,11 @@ func (mp *metaProcessor) computeExistingAndRequestMissingShardHeaders(metaBlock
 	mp.hdrsForCurrBlock.mutHdrsForBlock.Lock()
 	defer mp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 
-	for i := 0; i < len(metaBlock.ShardInfo); i++ {
-		shardData := metaBlock.ShardInfo[i]
+	for _, shardData := range metaBlock.ShardInfo {
+		if shardData.Nonce == 0 {
```
SebastianMarian, line 1645: == mp.genesisNonce

genesis/process/metaGenesisBlockCreator.go
```diff
@@ -95,6 +105,47 @@ func CreateMetaGenesisBlock(arg ArgsGenesisBlockCreator) (data.HeaderHandler, er
...
+	metaHdr.SetTimeStamp(arg.GenesisTime)
+	metaHdr.EpochStart.Economics = block.Economics{
+		TotalSupply:            big.NewInt(0).Set(arg.Economics.GenesisTotalSupply()),
+		TotalToDistribute:      big.NewInt(0),
+		TotalNewlyMinted:       big.NewInt(0),
+		RewardsPerBlockPerNode: big.NewInt(0),
+		NodePrice:              big.NewInt(0).Set(arg.Economics.GenesisNodePrice()),
+	}
+
+	errNotCritical := processors.txCoordinator.SaveBlockDataToStorage(bodyHandler.(*block.Body))
```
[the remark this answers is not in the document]
sasurobert: done

genesis/process/shardGenesisBlockCreator.go
sasurobert: done

process/peer/process.go
sasurobert: good catch.

process/block/metablock.go
sasurobert: good catch.

process/block/metablock.go
sasurobert: putting to hdrsForCurrBlock is done at checkShardHeadersValidity

genesis/process/metaGenesisBlockCreator.go
sasurobert: would leave like this now, we can decide about this later.

PR #1488, opened 2020-04-29
displayRatings and signedBlocksThreshold

raduchis:
displayRatings failed on the first block of every epoch because the nodesConfiguration for the new epoch was not found. Moved displayRatings on metablocks after the nodesConfiguration is created by the nodesCoordinator. 
- fixed DisplayRatings on epochStartBlock

The signedBlocksThreshold was verified not just for the eligible list members but for all validators and for the waiting validators the signedBlocks was always 0 and a debug message appeared: "below signedBlocksThreshold"
- computeSignedThreshold is done only on eligible Validators


process/interface.go
```diff
@@ -228,6 +228,7 @@ type ValidatorStatisticsProcessor interface {
 	GetValidatorInfoForRootHash(rootHash []byte) (map[uint32][]*state.ValidatorInfo, error)
 	ProcessRatingsEndOfEpoch(validatorInfos map[uint32][]*state.ValidatorInfo) error
 	Commit() ([]byte, error)
+	DisplayRatings(_ uint32)
```
AdoAdoAdo, line 231: why no param name?

process/block/metablock.go
```diff
@@ -1003,6 +1003,8 @@ func (mp *metaProcessor) CommitBlock(
 
 	mp.commitEpochStart(header, body)
 
+	mp.validatorStatisticsProcessor.DisplayRatings(header.GetEpoch())
```
AdoAdoAdo, line 1006: This would display the ratings after each commit.
Shouldn't we only display after epoch start?

process/interface.go
raduchis: fixed

process/block/metablock.go
raduchis: yes, moved from UpdatePeerState that was done evey block to commitblock, that is done every block.

PR #1489, opened 2020-04-29
En 6125/ESDT transfer shard process

sasurobert:
Added the built-in function of ESDT transfer. Which will unmarshall the data under a certain key of the user and will modify with the value from the transaction if it is necessary or possible.


process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -13,17 +13,45 @@ import (
...
+	s.mapInvalidKeys = make(map[string]struct{}, len(mapInvalidKeys))
+	for key := range mapInvalidKeys {
+		trimmedKey := key[:numInitCharactersInProtectedKeys]
+		s.mapInvalidKeys[trimmedKey] = struct{}{}
+	}
+
+	return s, nil
+}
+
+func (k *saveKeyValueStorage) isAllowedToSaveUnderKey(key []byte) bool {
```
camilbancioiu, line 47: :+1:

camilbancioiu: This check should be used in two more places: as a blockchain hook, for Arwen, to stop SmartContracts for overwriting protected keys, and in `processSCOutputAccounts()` as a final check, before individual storage updates are saved to the DataTrie.

camilbancioiu: Or directly in the `DataTrieTracker()`, to prevent us from accidentally forgetting to check `isAllowedToSaveUnderKey()`.

sasurobert: good catch. I will add at processSCOutput I think.

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -13,17 +13,45 @@ import (
...
+		return nil, process.ErrNilInvalidKeysMap
+	}
+
+	s := &saveKeyValueStorage{
+		gasConfig:   gasConfig,
+		funcGasCost: funcGasCost,
+	}
+	s.mapInvalidKeys = make(map[string]struct{}, len(mapInvalidKeys))
+	for key := range mapInvalidKeys {
+		trimmedKey := key[:numInitCharactersInProtectedKeys]
```
SebastianMarian, line 40: If len of key is lower than numInitCharactersInProtectedKeys the code would not panicked here?

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return err
+	}
+
+	userAcnt.DataTrieTracker().SaveKeyValue(key, marshalledData)
+
+	return nil
+}
+
+func (e *esdtTransfer) getESDTDataFromKey(userAcnt state.UserAccountHandler, key []byte) *ESDigitalToken {
```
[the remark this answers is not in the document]
sasurobert: it is safer to not think about those returns, because it could self repair itself. it is safer this way, as you would rewrite the underlying data with the new one if it is allowed.

process/smartContract/builtInFunctions/proto/esdt.proto
```diff
@@ -0,0 +1,14 @@
...
+syntax = "proto3";
+
+package protoBuiltInFunctions;
+
+option go_package = "builtInFunctions";
+option (gogoproto.stable_marshaler_all) = true;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+// RewardTx holds the data for a reward transaction
```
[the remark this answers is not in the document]
sasurobert: copy paste problem :D

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: no, it would not panic - it goes to the index only if exist.

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -13,17 +13,45 @@ import (
...
+		s.mapInvalidKeys[trimmedKey] = struct{}{}
+	}
+
+	return s, nil
+}
+
+func (k *saveKeyValueStorage) isAllowedToSaveUnderKey(key []byte) bool {
+	if len(key) < numInitCharactersInProtectedKeys {
+		return true
```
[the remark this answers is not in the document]
sasurobert: yes. as those are no keys which we use to save special values.

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -13,17 +13,45 @@ import (
...
+	if mapInvalidKeys == nil {
+		return nil, process.ErrNilInvalidKeysMap
+	}
+
+	s := &saveKeyValueStorage{
+		gasConfig:   gasConfig,
+		funcGasCost: funcGasCost,
+	}
+	s.mapInvalidKeys = make(map[string]struct{}, len(mapInvalidKeys))
+	for key := range mapInvalidKeys {
```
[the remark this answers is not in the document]
sasurobert: no problem. We effectively stop the user right now to write some arbitrary values under a specific key - those keys can be modified only throguh ESDT processing right now. Later on we might have other set of keys as well.

process/smartContract/builtInFunctions/keyValueStorage.go
SebastianMarian: ok

process/smartContract/builtInFunctions/keyValueStorage.go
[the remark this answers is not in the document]
SebastianMarian: ok

process/smartContract/builtInFunctions/keyValueStorage.go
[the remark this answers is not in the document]
SebastianMarian: ok

process/smartContract/builtInFunctions/esdtTransfer.go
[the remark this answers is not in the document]
SebastianMarian: Yes, but what if I have a value of 10 behind the given key, and for some reasons the unmarshal or retrive fails, you return an esdtData with a value of 0, and then in addToESDTBalance you will add to this 0 the value received there, let's say 5, and then you will save the computation result (5 instead 15) in userAcnt. This will not cause a problem?

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -0,0 +1,119 @@
...
+	err = e.marshalizer.Unmarshal(esdtData, marshalledData)
+	if err != nil {
+		log.Debug("getESDTDataFromKey unmarshal error", "error", err)
+		return esdtData
+	}
+
+	return esdtData
+}
+
+// IsInterfaceNil return true if underlying object in nil
```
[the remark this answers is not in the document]
sasurobert: done.

process/smartContract/builtInFunctions/esdtTransfer.go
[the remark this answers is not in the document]
sasurobert: done.

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -13,17 +13,45 @@ import (
...
+
 type saveKeyValueStorage struct {
-	gasConfig   BaseOperationCost
-	funcGasCost uint64
+	gasConfig      BaseOperationCost
+	funcGasCost    uint64
+	mapInvalidKeys map[string]struct{}
 }
 
 // NewSaveKeyValueStorageFunc returns the key-value
```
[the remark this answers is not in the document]
sasurobert: done

process/smartContract/builtInFunctions/keyValueStorage.go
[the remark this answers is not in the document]
sasurobert: I think one better approach is to not let them write under KEYs which start with Elrond. As those are reserved for special functions.

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: yes it panics, added check.

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: I will skip those keys which are protected for these writes, maybe with a warning ? shouldn't we protect it at arwen level - write key value ?

sasurobert: done

process/smartContract/process.go
```diff
@@ -811,8 +811,11 @@ func (sc *scProcessor) processSCOutputAccounts(
 
 		for j := 0; j < len(storageUpdates); j++ {
 			storeUpdate := storageUpdates[j]
-			acc.DataTrieTracker().SaveKeyValue(storeUpdate.Offset, storeUpdate.Data)
+			if !process.IsAllowedToSaveUnderKey(storeUpdate.Offset) {
+				log.Trace("storeUpdate is not allowed", "acc", outAcc.Address, "key", storeUpdate.Offset, "data", storeUpdate.Data)
+			}
```
camilbancioiu, line 816: Missing `continue`?

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -96,24 +102,24 @@ func (e *esdtTransfer) addToESDTBalance(userAcnt state.UserAccountHandler, key [
...
 	marshalledData, err := userAcnt.DataTrieTracker().RetrieveValue(key)
 	if err != nil {
 		log.Debug("getESDTDataFromKey retrieve value error", "error", err)
-		return esdtData
+		return esdtData, nil
 	}
 
 	err = e.marshalizer.Unmarshal(esdtData, marshalledData)
 	if err != nil {
 		log.Debug("getESDTDataFromKey unmarshal error", "error", err)
```
SebastianMarian, line 115: This log could be removed now, as would be printed in the caller side

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -96,24 +102,24 @@ func (e *esdtTransfer) addToESDTBalance(userAcnt state.UserAccountHandler, key [
 	return nil
 }
 
-func (e *esdtTransfer) getESDTDataFromKey(userAcnt state.UserAccountHandler, key []byte) *ESDigitalToken {
+func (e *esdtTransfer) getESDTDataFromKey(userAcnt state.UserAccountHandler, key []byte) (*ESDigitalToken, error) {
 	esdtData := &ESDigitalToken{Value: big.NewInt(0)}
 	marshalledData, err := userAcnt.DataTrieTracker().RetrieveValue(key)
 	if err != nil {
 		log.Debug("getESDTDataFromKey retrieve value error", "error", err)
```
SebastianMarian, line 109: This log should be removed as in this case is not a error

process/smartContract/builtInFunctions/esdtTransfer.go
sasurobert: done

process/smartContract/process.go
sasurobert: done

process/block/preprocess/gasComputation.go
```diff
@@ -169,25 +170,28 @@ func (gc *gasComputation) ComputeGasConsumedByTx(
 		return 0, 0, process.ErrNilTransaction
 	}
 
-	txGasLimitConsumption := gc.economicsFee.ComputeGasLimit(txHandler)
+	moveBalanceConsumption := gc.economicsFee.ComputeGasLimit(txHandler)
 
 	txType := gc.txTypeHandler.ComputeTransactionType(txHandler)
-	isSCCall := txType == process.SCDeployment || txType == process.SCInvoking || txType == process.BuiltInFunctionCall
+	isSCCall := txType == process.SCDeployment ||
```
SebastianMarian, line 176: I think that condition: txType == process.SCDeployment || txType == process.SCInvoking is not really necessary now as you have added now core.IsSmartContractAddress(txHandler.GetRcvAddr()) which covers these both too

process/block/preprocess/gasComputation.go
```diff
@@ -165,20 +170,28 @@ func (gc *gasComputation) ComputeGasConsumedByTx(
...
-		if txSenderShardId != txReceiverShardId && txGasLimitConsumption < txHandler.GetGasLimit() {
-			gasConsumedByTxInSenderShard := txGasLimitConsumption
-			gasConsumedByTxInReceiverShard := txHandler.GetGasLimit() - txGasLimitConsumption
+	moveBalanceConsumption := gc.economicsFee.ComputeGasLimit(txHandler)
+
+	txType := gc.txTypeHandler.ComputeTransactionType(txHandler)
+	isSCCall := txType == process.SCDeployment ||
+		txType == process.SCInvoking ||
+		txType == process.BuiltInFunctionCall ||
+		(core.IsSmartContractAddress(txHandler.GetRcvAddr()) && len(txHandler.GetData()) > 0)
```
iulianpascalau, line 179: Isn't this process.SCInvoking ?

process/coordinator/transactionType.go
```diff
@@ -56,39 +56,43 @@ func NewTxTypeHandler(
 }
 
 // ComputeTransactionType calculates the transaction type
-func (tth *txTypeHandler) ComputeTransactionType(tx data.TransactionHandler) (process.TransactionType, error) {
+func (tth *txTypeHandler) ComputeTransactionType(tx data.TransactionHandler) process.TransactionType {
```
iulianpascalau, line 59: :+1:

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -55,11 +55,14 @@ func (e *esdtTransfer) ProcessBuiltinFunction(
...
-	esdtTokenKey := append(e.keyPrefix, vmInput.Arguments[0]...)
 	value := big.NewInt(0).SetBytes(vmInput.Arguments[1])
+	if value.Cmp(zero) <= 0 {
+		return nil, process.ErrNegativeValue
+	}
 
-	log.Trace("esdtTranfer", "snd", vmInput.CallerAddr, "rcv", vmInput.RecipientAddr, "value", value)
+	gasRemaining := uint64(0)
+	esdtTokenKey := append(e.keyPrefix, vmInput.Arguments[0]...)
+	log.Trace("esdtTransfer", "snd", vmInput.CallerAddr, "rcv", vmInput.RecipientAddr, "value", value)
```
iulianpascalau, line 65: sender, receiver instead of snd, rcv

process/smartContract/process.go
```diff
@@ -212,7 +212,7 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
 	executed, err = sc.resolveBuiltInFunctions(txHash, tx, acntSnd, acntDst, vmInput)
 	if err != nil {
 		log.Debug("processed built in functions error", "error", err.Error())
-		return nil
+		return err
```
iulianpascalau, line 215: :1st_place_medal:

process/smartContract/process.go
```diff
@@ -323,14 +323,15 @@ func (sc *scProcessor) resolveBuiltInFunctions(
...
 		return false, nil
 	}
-	if check.IfNil(builtIn) {
-		return true, process.ErrNilBuiltInFunction
-	}
+	// return error here only if acntSnd is not nil - so this is sender shard
 
 	vmOutput, err := builtIn.ProcessBuiltinFunction(acntSnd, acntDst, vmInput)
-	if err != nil {
+	if err != nil && !check.IfNil(acntSnd) {
```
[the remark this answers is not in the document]
sasurobert: done

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -58,8 +58,11 @@ func (e *esdtTransfer) ProcessBuiltinFunction(
 	gasRemaining := uint64(0)
 	esdtTokenKey := append(e.keyPrefix, vmInput.Arguments[0]...)
 	value := big.NewInt(0).SetBytes(vmInput.Arguments[1])
+	if value.Cmp(zero) <= 0 {
```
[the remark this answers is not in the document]
sasurobert: Prefer not to - in order to keep the processing low, and the trie should keep only relevant information.

process/block/preprocess/gasComputation.go
sasurobert: no, in case of cross call it is MOVE balance.

process/smartContract/builtInFunctions/esdtTransfer.go
sasurobert: done

process/smartContract/builtInFunctions/esdtTransfer.go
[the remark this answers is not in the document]
SebastianMarian: ok

PR #1498, opened 2020-04-30
Community address

AdoAdoAdo:
Added rewards for community
At the end of the epoch a percentage defined in config from the protocol rewards will be directed towards a community account


data/block/proto/metaBlock.proto
```diff
@@ -70,9 +70,11 @@ message Economics {
 	bytes  TotalToDistribute      = 2 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
 	bytes  TotalNewlyMinted       = 3 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
 	bytes  RewardsPerBlockPerNode = 4 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
-	bytes  NodePrice              = 5 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
-	uint64 PrevEpochStartRound    = 6;
-	bytes  PrevEpochStartHash     = 7;
+	bytes  RewardsForCommunity    = 5 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
+	bytes  CommunityAddress       = 6;
```
sasurobert, line 74: There is no need for this to appear in the metablock. It is enough to appear in the generated reward transactions

sasurobert: the bigInt of rewards for community can remain as it is super visible, but the address can be read from the config.

epochStart/metachain/economics.go
```diff
@@ -140,6 +146,26 @@ func (e *economics) ComputeEndOfEpochEconomics(
 	return &computedEconomics, nil
 }
 
+// compute the rewards for community - percentage from total rewards
+func (e *economics) computeRewardsForCommunity(totalRewards *big.Int) *big.Int {
+	precentageAdjustment := float64(100)
+	communityPercentage := big.NewInt(0).SetUint64(uint64(e.rewardsHandler.CommunityPercentage() * precentageAdjustment))
+	rewardsForCommunity := big.NewInt(0).Mul(totalRewards, communityPercentage)
```
sasurobert, line 153: you can use core.GetPercentageOfValue instead of this.

epochStart/metachain/rewards.go
```diff
@@ -141,20 +171,49 @@ func (rc *rewardsCreator) CreateRewardsMiniBlocks(metaBlock *block.MetaBlock, va
...
-			finalMiniBlocks = append(finalMiniBlocks, miniBlocks[i])
-		}
+	return nil
+}
+
+func (rc *rewardsCreator) createCommunityRewardTransaction(
+	metaBlock *block.MetaBlock,
+) (*rewardTx.RewardTx, []byte, uint32, error) {
+
+	communityAddress := metaBlock.EpochStart.Economics.CommunityAddress
```
sasurobert, line 195: community address may be saved as an internal variable to the rewardsCreator - and this check can be done, address validity at the time of constructor.

epochStart/metachain/economics_test.go
```diff
@@ -269,6 +301,7 @@ func TestEconomics_ComputeEndOfEpochEconomics(t *testing.T) {
 func TestEconomics_VerifyRewardsPerBlock_DifferentHitRates(t *testing.T) {
 	t.Parallel()
 
+	commAddress := "communityAddress"
```
raduchis, line 304: there should be a problem with the test above, line 288 and 289. It should have 2 shards and 1 metachain,but there are lastFinalizedHeaders on shard 1 and 2. The number of blocks in the previous epoch is wrongly computed.

config/tomlConfig_test.go
```diff
@@ -122,10 +122,13 @@ func TestTomlEconomicsParser(t *testing.T) {
 	maxGasLimitPerBlock := "18446744073709551615"
 	minGasPrice := "18446744073709551615"
 	minGasLimit := "18446744073709551615"
+	communityAddress := "erd1932eft30w753xyvme8d49qejgkjc09n5e49w4mwdjtm0neld797su0dlxp"
 
 	cfgEconomicsExpected := EconomicsConfig{
 		RewardsSettings: RewardsSettings{
-			LeaderPercentage: leaderPercentage,
+			LeaderPercentage:    leaderPercentage,
```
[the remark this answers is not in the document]
AdoAdoAdo: done

config/tomlConfig_test.go
```diff
@@ -138,6 +141,7 @@ func TestTomlEconomicsParser(t *testing.T) {
 [RewardsSettings]
     RewardsValue = "` + rewardsValue + `"
```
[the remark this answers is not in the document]
AdoAdoAdo: done

epochStart/metachain/economics.go
AdoAdoAdo: done

data/block/proto/metaBlock.proto
AdoAdoAdo: done

epochStart/metachain/economics_test.go
AdoAdoAdo: the function computes the nonces from last finalized headers, and indeed the shards are not correctly set.

I think the idea was that the shards did not create any block in the previous epoch, only the metachain the start of epoch block, so the number of blocks = 1, which is correct even with invalid shard IDs (anyway the last notarized blocks from those shards are behind the first block in epoch)

process/economics/economicsData.go
```diff
@@ -179,11 +183,16 @@ func convertValues(economics *config.EconomicsConfig) (*EconomicsData, error) {
 func checkValues(economics *config.EconomicsConfig) error {
 	if isPercentageInvalid(economics.RewardsSettings.LeaderPercentage) ||
```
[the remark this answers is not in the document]
AdoAdoAdo: indeed these are not enough.
But there are different criterias as the leader percentage is applied to the accumulated fees in the proposed block, whereas the developerPercentage is applied to the transaction fee calling the speciffic SC, so also not part of the total. The community percentage is applied to the total.

Due to your comment I checked a bit and found a bug in the accumulated fees and what the Leader receives. The accumulated fees are considered entirely when computing the rewards per block, but then again the leader receives in addition the leader percentage -> leader percentage should have been subtracted before computing the rewards per block, otherwise we are giving that portion twice.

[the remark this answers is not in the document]
sasurobert: great find :+1:

config/tomlConfig_test.go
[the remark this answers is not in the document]
raduchis: ok

epochStart/metachain/economics_test.go
raduchis: yes. it's ok right now

PR #1501, opened 2020-05-01
Detect and fix some race conditions in TxCache

andreibancioiu:
Add a (somehow structural) constraint that `NotifyScoreChange` is only called under an already acquired mutex on the mutating `txListForSender`. Previously, if that were to be called concurrently, data races on `ComputeScore() >> len(linkedList)` would have occurred.

This new constraint also says that a movement of a sender (list of txs) between a score bucket (chunk) to another cannot occur concurrently with another movement of the same sender. Therefore, the atomic flag `ScoreChangeInProgressFlag` is not necessary anymore.

Also, previously, if an eviction primitive (`evictSendersAndTheirTxs`) and mutation of a list (remove tx) were performed concurrently, a data race on `txListBySender.mapChunk` could have occurred. Solved using atomic pointers.


storage/txcache/maps/bucketSortedMapItem.go
```diff
@@ -1,12 +1,32 @@
...
 	GetKey() string
 	ComputeScore() uint32
-	GetScoreChunk() *MapChunk
-	SetScoreChunk(*MapChunk)
-	ScoreChangeInProgressFlag() *atomic.Flag
+	ScoreChunk() *MapChunkPointer
+}
+
+// MapChunkPointer is an atomically-mutable pointer to a MapChunk
+type MapChunkPointer struct {
```
iulianpascalau, line 15: So, you basically created a concurrent "safe" structure just to hold the pointer? Why didn't you simply add a mutex on GetScoreChunk and SetScoreChunk? The project uses a lot the later version.

andreibancioiu: Indeed. Will fix to use a mutex (and also make tiny benchmark for, like, 10 routines with 1k writes each - SetScoreChunk - with mutex vs. atomic mutable pointer).

andreibancioiu: Fixed :)

storage/txcache/testutils_test.go
```diff
@@ -119,3 +121,20 @@ func measureWithStopWatch(b *testing.B, function func()) {
 	duration := sw.GetMeasurementsMap()["time"]
 	b.ReportMetric(duration, "time@stopWatch")
 }
+
+// waitTimeout waits for the waitgroup for the specified max timeout.
+// Returns true if waiting timed out.
+// Reference: https://stackoverflow.com/a/32843750/1475331
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
```
iulianpascalau, line 128: :1st_place_medal: I have used this exact pattern all around the project, never thought of making a function out of it

storage/txcache/sweeping_test.go
```diff
@@ -23,29 +23,30 @@ func TestSweeping_CollectSweepable(t *testing.T) {
...
 	require.Equal(t, 3, len(selection))
 	require.Equal(t, 0, len(cache.sweepingListOfSenders))
 	require.Equal(t, 2, cache.getNumFailedSelectionsOfSender("alice"))
 	require.Equal(t, 2, cache.getNumFailedSelectionsOfSender("bob"))
 	require.Equal(t, 0, cache.getNumFailedSelectionsOfSender("carol"))
 
 	// 3nd fail, collect Alice and Bob as sweepables
-	selection = cache.SelectTransactions(1000, 1000)
+	selection = cache.doSelectTransactions(1000, 1000)
+	// o si rulat gorutina si a sters offfffff. cum pot oare?
```
AdoAdoAdo, line 43: is the comment still valid?
check below as well

andreibancioiu: Thanks! Fixed / removed.

PR #1507, opened 2020-05-04
Prune and cancel prune when snapshot in progress

BeniaminDrasovean:
To be able to enable trie pruning while a trie snapshot is in progress, all pruning operations (prune, cancel prune) must be buffered, and after the snapshotting process has finished, they must executed sequentially and in the same order that they were added to the pruning buffer.


data/trie/trieStorageManager.go
```diff
@@ -177,45 +184,66 @@ func (tsm *trieStorageManager) Prune(rootHash []byte, identifier data.TriePrunin
...
 
-func (tsm *trieStorageManager) prune(oldHashes map[string]struct{}) {
-	for hash := range oldHashes {
-		rootHash := []byte(hash)
-		err := tsm.removeFromDb(rootHash)
-		if err != nil {
-			log.Error("trie storage manager remove from db", "error", err, "rootHash", hex.EncodeToString(rootHash))
+func (tsm *trieStorageManager) resolveBufferedHashes(oldHashes [][]byte) {
+	for _, rootHash := range oldHashes {
+		lastBytePos := len(rootHash) - 1
```
SebastianMarian, line 203: Maybe you should add a check for lastBytePos to be >= 0, otherwise the next code could panic

data/trie/pruningBuffer.go
```diff
@@ -26,29 +26,19 @@ func (pb *pruningBuffer) add(rootHash []byte) {
...
 
-	buffer := make(map[string]struct{})
-	for key := range pb.buffer {
-		buffer[key] = struct{}{}
-	}
-
-	pb.buffer = make(map[string]struct{})
+	buffer := make([][]byte, len(pb.buffer))
+	copy(buffer, pb.buffer)
+	pb.buffer = pb.buffer[:0]
```
bogdan-rosianu, line 41: refactor this as this can lead to hidden bugs because you still can access the old values if you don't overwrite them
```
	sl := []int{1,2,3,4,5,6,7,8}
	sl = sl[:0]

	sl = append(sl, 7)
	fmt.Println(sl[:2])
```
will print [7,2]

data/trie/trieStorageManager.go
bogdan-rosianu: :+1:

data/trie/pruningBuffer.go
bogdan-rosianu: It's ok if you range the buffer, but also the capacity remains set to the old value. Please wait for a second opinion

SebastianMarian: I think it could remain like this, as we range on this buffer, and the garbage remained beyond the buffer len should not be a problem. It is just an allocated memory which is useful for the future appends, if they will be. We use this pattern in many other parts of the project. You can search after [:0]. The alternative, is to replace this approach with make but this would cause to reallocate the memory to each append which would be done afterward.

data/trie/trieStorageManager.go
BeniaminDrasovean: Done.

PR #1513, opened 2020-05-05
EN-6290: change tx data field back to string

bogdan-rosianu:
**DO NOT MERGE**

- changed transaction's data field back to string. (signing + way of receiving via API)
- after this PR will be merged, when signing the transaction, the data field has to be in string format
- also added a custom marshalizer for transactions which does not escape HTML characters in strings.


api/transaction/routes.go
```diff
@@ -43,7 +43,7 @@ type SendTxRequest struct {
 	Sender    string `form:"sender" json:"sender"`
```
ccorcoveanu, line 43: Maybe only change CreateTransaction also?

bogdan-rosianu: done

PR #1514, opened 2020-05-05
Developer fee adjustments

AdoAdoAdo:
This PR allows for propagating the developer fees to metachain as well so that metachain has clear view of all fees in the system and take them into account when computing the rewards at the end of an epoch.

Previously developer fees were managed directly inside the shard, and their quantity was obscured from metachain, which was only aware of the rest of the fees. This introduced also some inconsistency in the amount of new tokens generated by the network at the end of the epoch and also in the percentage taken by the block proposer.


epochStart/genesis/genesisBlockCreator.go
```diff
@@ -82,6 +82,7 @@ func CreateShardGenesisBlockFromInitialBalances(
 		RandSeed:        rootHash,
 		TimeStamp:       genesisTime,
 		AccumulatedFees: big.NewInt(0),
+		DeveloperFees:   big.NewInt(0),
```
iulianpascalau, line 85: Will remember to set this to big.NewInt(0) in feat/genesis2 branch

PR #1518, opened 2020-05-06
Fix edge case on reset probable highest nonce

SebastianMarian:
* Fixed edge case when repetitive timeout errors, or any hind of errors occurred when a node is syncing (requesting next header or missing miniblocks) but before the actual starting of processing/committing block, would not triggered the action of resetting probable highest nonce and the action of cleaning headers pool


process/sync/baseSync.go
```diff
@@ -482,15 +481,22 @@ func (boot *baseBootstrap) syncBlocks() {
...
 }
 
-func (boot *baseBootstrap) doJobOnSyncBlockFail(headerHandler data.HeaderHandler, err error) {
-	if err == process.ErrTimeIsOut {
-		boot.requestsWithTimeout++
-	}
+func (boot *baseBootstrap) doJobOnSyncBlockFail(bodyHandler data.BodyHandler, headerHandler data.HeaderHandler, err error) {
+	nonce := boot.getNonceForNextBlock()
+
+	boot.mutNonceSyncedWithErrors.Lock()
```
iulianpascalau, line 487: Maybe a function for this (L487-L490)?

PR #1523, opened 2020-05-07
Fix putbatch storage closed

AdoAdoAdo:


storage/pruning/pruningStorer.go
```diff
@@ -217,6 +217,9 @@ func (ps *PruningStorer) Put(key, data []byte) error {
 		persisterToUse = persisterInSetEpoch
 	} else {
 		persisterToUse = ps.activePersisters[0]
+		log.Warn("active persister not found",
```
sasurobert, line 220: Should test with prunning storer of, and see if this message appears or not after epoch changes and restarts.

AdoAdoAdo: right, with pruning storer off, the warn comes every time
changed to take that into consideration and only warn if pruning is active but we don't find the specific active persister.

PR #1526, opened 2020-05-08
sync should be critical error in bootstrap

sasurobert:
In case of bootstrap if syncing the epoch start metablock returns with any error, this should be treated as critical, node should not start.


elrondbigscam:
ERD/Elrond Scam Coin. Team ERD is a con man. Don't buy ERD. I send this
 message every day to everyone in the group. Prevent more people from being conned by sharing with you.
 ERD BIG DUMP SELL NOW ...

PR #1725, opened 2020-05-12
Fix minor issue and add the GENESIS_DELAY option

camilbancioiu:
This PR adds the following changes:
* added the `$GENESIS_DELAY` option in the testnet scripts, which specifies how far into the future is the genesis moment for the started testnet, calculated from the moment the testnet is started (with `start.sh`). Expressed in seconds.
* fixed a minor issue with displaying the address(es) of the deployed SmartContracts


process/smartContract/process.go
```diff
@@ -540,7 +540,8 @@ func (sc *scProcessor) DeploySmartContract(
 
 func (sc *scProcessor) printScDeployed(vmOutput *vmcommon.VMOutput, tx data.TransactionHandler) {
 	scGenerated := make([]string, 0, len(vmOutput.OutputAccounts))
-	for addr := range vmOutput.OutputAccounts {
+	for key := range vmOutput.OutputAccounts {
```
iulianpascalau, line 543: suggestion to change to:
```GO
for _, account := range vmOutput.OutputAccounts {
    if check.IfNil(account){
        continue
    }
    ......
```
:1st_place_medal: for finding this

scripts/testnet/include/config.sh
```diff
@@ -76,7 +76,7 @@ updateNodeConfig() {
 
   cp nodesSetup.json nodesSetup_edit.json
   
-  let startTime="$(date +%s) + $NODE_DELAY + 30"
+  let startTime="$(date +%s) + $GENESIS_DELAY"
```
iulianpascalau, line 79: :+1:

process/smartContract/process.go
camilbancioiu: Changed to check `account == nil`.

PR #1728, opened 2020-05-13
Add checks for epoch start config when init bootstrap

bogdan-rosianu:
Added checks for epoch start configuration when starting the bootstrap component. Otherwise, this could occur in late errors.


epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -63,10 +65,10 @@ func NewEpochStartMetaBlockProcessor(
 	if !(consensusPercentage > 0 && consensusPercentage <= 100) {
 		return nil, epochStart.ErrInvalidConsensusThreshold
 	}
-	if minNumConnectedPeers < 1 {
+	if minNumConnectedPeers < minNumConnectedPeers {
```
SebastianMarian, line 68: Could you rename with caps M the constant? They have the same name in left and right side of this compare

epochStart/bootstrap/epochStartMetaBlockProcessor.go
```diff
@@ -63,10 +65,10 @@ func NewEpochStartMetaBlockProcessor(
...
 		return nil, epochStart.ErrNotEnoughNumConnectedPeers
 	}
-	if minNumOfPeersToConsiderBlockValid < 1 {
+	if minNumOfPeersToConsiderBlockValid < minNumOfPeersToConsiderBlockValid {
```
SebastianMarian, line 71: Could you rename with caps M the constant? They have the same name in left and right side of this compare

epochStart/bootstrap/epochStartMetaBlockProcessor.go
bogdan-rosianu: you're right. good find. renamed the received parameter and kept the constant name the same

epochStart/bootstrap/epochStartMetaBlockProcessor.go
bogdan-rosianu: renamed the parameter

PR #1737, opened 2020-05-13
En 6407 trie package unit tests

BeniaminDrasovean:
Added unit tests for trie package.
Removed unused SignalEndOfProcessing() from multiDataInterceptor.
Export trie const in config.


data/trie/extensionNode_test.go
```diff
@@ -1,12 +1,15 @@
...
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
 	"reflect"
 	"testing"
 
+	"github.com/ElrondNetwork/elrond-go/data"
+
```
sasurobert, line 12: goimports

data/trie/trieStorageManager.go
```diff
@@ -376,7 +376,8 @@ func (tsm *trieStorageManager) takeSnapshot(snapshot *snapshotsQueueEntry, msh m
 		return
 	}
 
-	err = newRoot.commit(true, 0, tsm.db, db)
+	maxTrieLevelInMemory := uint(5)
```
sasurobert, line 379: here you need the config values as well.

genesis/process/memoryComponents.go
```diff
@@ -14,7 +14,8 @@ func createAccountAdapter(
 	accountFactory state.AccountFactory,
 	trieStorage data.StorageManager,
 ) (state.AccountsAdapter, error) {
-	tr, err := trie.NewTrie(trieStorage, marshalizer, hasher)
+	maxTrieLevelInMemory := uint(5)
```
sasurobert, line 17: make a constant for this - some component should be able to return the desired level.

integrationTests/testInitializer.go
```diff
@@ -778,7 +779,9 @@ func CreateNewDefaultTrie() data.Trie {
 		MaxSnapshots:       2,
 	}
 	trieStorage, _ := trie.NewTrieStorageManager(CreateMemUnit(), TestMarshalizer, TestHasher, config.DBConfig{}, ewl, generalCfg)
-	tr, _ := trie.NewTrie(trieStorage, TestMarshalizer, TestHasher)
+
+	maxTrieLevelInMemory := uint(5)
```
sasurobert, line 783: make a const.

integrationTests/vm/testInitializer.go
```diff
@@ -118,7 +118,8 @@ func CreateInMemoryShardAccountsDB() *state.AccountsDB {
 		generalCfg,
 	)
 
-	tr, _ := trie.NewTrie(trieStorage, marsh, testHasher)
+	maxTrieLevelInMemory := uint(5)
```
sasurobert, line 121: use a const

update/genesis/import.go
```diff
@@ -238,7 +238,8 @@ func (si *stateImport) getTrie(shardID uint32, accType Type) (data.Trie, error)
 		trieStorageManager = si.trieStorageManagers[triesFactory.PeerAccountTrie]
 	}
 
-	trieForShard, err := trie.NewTrie(trieStorageManager, si.marshalizer, si.hasher)
+	maxTrieLevelInMemory := uint(5)
```
sasurobert, line 241: use the inputted trie level - or make a constant.

data/trie/extensionNode_test.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: Not really, because after the snapshot is done, the pointer to the newRoot will not be used again, so it will not influence the trie in any way.

update/genesis/import.go
BeniaminDrasovean: Done.

genesis/process/memoryComponents.go
BeniaminDrasovean: Done.

integrationTests/testInitializer.go
BeniaminDrasovean: Done.

integrationTests/vm/testInitializer.go
BeniaminDrasovean: Done.

update/genesis/import.go
```diff
@@ -22,6 +22,8 @@ import (
 
 var _ update.ImportHandler = (*stateImport)(nil)
 
+const maxTrieLevelInMemory = uint(5)
```
miiu96, line 25: why you dont use value from cofig file ?

BeniaminDrasovean: Because it is not needed. The tries that are created here are used just to import the data, so there is no need to pass the config.

PR #1739, opened 2020-05-13
En 6420/rewards for smart contracts

sasurobert:
Save accumulated rewards to ELRONDrewards key in the smart contract trie.


process/rewardTransaction/process.go
```diff
@@ -90,9 +95,30 @@ func (rtp *rewardTxProcessor) ProcessRewardTransaction(rTx *rewardTx.RewardTx) e
...
+func (rtp *rewardTxProcessor) saveAccumulatedRewards(
+	rtx *rewardTx.RewardTx,
+	userAccount state.UserAccountHandler,
+) {
+	if !core.IsSmartContractAddress(rtx.RcvAddr) {
+		return
+	}
+
+	existingReward := big.NewInt(0)
+	fullRewardKey := core.ElrondProtectedKeyPrefix + rewardKey
```
iulianpascalau, line 112: shouldn't expose this fullRewardKey through blockchain hook interface so a developer won't bother searching through our source code for constants?

sasurobert: I think it would be enough to explain this in a tutorial and to have it in the docs.

andrei-marinica: It is also easy to expose in the Rust API, no need for additional hook

PR #1742, opened 2020-05-13
Implement go routines closing mechanism

SebastianMarian:
* Added go routines close mechanism through context and io.Close implementation for components: Chronology, Worker, SyncTimer, Bootstrapper, epochStartTrigger, miniBlocksPoolsCleaner, txsPoolsCleaner


consensus/spos/interface.go
```diff
@@ -96,6 +98,7 @@ type SubroundsFactory interface {
 
 //WorkerHandler represents the interface for the SposWorker
 type WorkerHandler interface {
+	io.Closer
```
sasurobert, line 101: shouldn't we use our own interface ? instead of io.Closer

SebastianMarian: yep, removed it from all modified interfaces

PR #1745, opened 2020-05-14
get rewards txs from pool before commitEpochStart method

miiu96:
Bug fix : rewards transactions was not indexed anymore because method that gets rewards transactions from pool was called after transactions was removed from pool.
FIX : call the method that reads the transactions before they are removed from the pool

Also the field 'gasUsed' of structure Transaction that is saved in elasticsearch database  is now calculated in gas units. To convert the field 'gasUsed'  in ERD must by multiply with gas price.

### DIFF
--- a/core/indexer/processTransactions.go
+++ b/core/indexer/processTransactions.go
@@ -69,8 +69,9 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
 		gasUsed := big.NewInt(0).SetUint64(tx.GasPrice)
 		gasUsed.Mul(gasUsed, big.NewInt(0).SetUint64(tx.GasLimit))
 		gasUsed.Sub(gasUsed, rec.Value)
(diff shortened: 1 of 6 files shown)


core/indexer/processTransactions.go
```diff
@@ -69,8 +69,9 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
 		gasUsed := big.NewInt(0).SetUint64(tx.GasPrice)
 		gasUsed.Mul(gasUsed, big.NewInt(0).SetUint64(tx.GasLimit))
 		gasUsed.Sub(gasUsed, rec.Value)
+		gasUsed.Div(gasUsed, big.NewInt(0).SetUint64(tx.GasPrice))
 
-		tx.GasUsed = gasUsed.String()
+		tx.GasUsed = gasUsed.Uint64()
```
sasurobert, line 74: gasUsed for normal transactions can be computed via ComputeGasLimit(tx) txFeeHandler

PR #1747, opened 2020-05-14
Refactor TxPool, improve its robustness

andreibancioiu:
**Do not review - we should wait until `pool++limits` enters development.**

 1. Score computation logic extracted to allow for easier change / algorithm replacements
 2. Some components decoupled from `cacheConfig` etc.
 3. `bucketSortedMap` does not call `computeScore` anymore, but receives the score instead 
 4. `txListForSender` does not compute its score, it only computes the `senderScoreParams`.
 5. 3 and 4 allow us to reason more easily about critical sections within `txListForSender`.
 6. Fixed a possible uint32 overflow in `RemoveTxsBulk` (it would have a benign overflow, the value only used to monitor / diagnose eviction)
 7. Wrapped transaction selection, addition and removal in try / catch blocks.
 8. Removed deprecated cache implementation
 9. Add log with number of senders with failed selection (per selection)
 10. Add log with number of transactions to remove when block is finalized (sharded tx pool)


andreibancioiu:
Diff issues.

PR #1749, opened 2020-05-15
bugFix/change reward address

sasurobert:
Change reward address fix - auction smart contract was calling in the wrong order the staking smart contract.


integrationTests/testProcessorNodeWithMultisigner.go
```diff
@@ -112,6 +113,169 @@ func CreateNodesWithNodesCoordinatorWithCacher(
...
+			)
+		}
+
+		nodesMap[shardId] = append(nodesMap[shardId], nodesList...)
+	}
+
+	return nodesMap
+}
+
+func CreateNodeWithBLSAndTxKeys(
```
iulianpascalau, line 183: Please add comment

integrationTests/testProcessorNodeWithMultisigner.go
```diff
@@ -112,6 +113,169 @@ func CreateNodesWithNodesCoordinatorWithCacher(
 
 }
 
+// CreateNodesWithNodesCoordinatorAndTxKeys
```
iulianpascalau, line 116: please add a " -"

vm/systemSmartContracts/auction.go
```diff
@@ -183,13 +183,15 @@ func (s *stakingAuctionSC) changeRewardAddress(args *vmcommon.ContractCallInput)
 		return vmcommon.UserError
 	}
 
+	txData := "changeRewardAddress@" + hex.EncodeToString(registrationData.RewardAddress)
```
iulianpascalau, line 186: for this, I have used something like this in genesis package
```GO
arguments := make([]string, 0, len(registrationData.BlsPubKeys) + 2)
arguments = append(arguments, changeRewardAddressFunction)
arguments = append(arguments, hex.EncodeToString(registrationData.RewardAddress))
for _, blsKey := range registrationData.BlsPubKeys {
    arguments = append(arguments, hex.EncodeToString(blsKey))
}
txData := strings.Join(arguments, "@")
```
As can be seen, "changeRewardAddress" might be a const for this.

vm/systemSmartContracts/auction.go
```diff
@@ -183,13 +183,15 @@ func (s *stakingAuctionSC) changeRewardAddress(args *vmcommon.ContractCallInput)
...
-		if isError {
-			log.LogIfError(err)
-			return vmcommon.UserError
-		}
+		txData += "@" + hex.EncodeToString(blsKey)
+	}
+	vmOutput, err := s.executeOnStakingSC([]byte(txData))
+	isError := err != nil || vmOutput.ReturnCode != vmcommon.Ok
+	if isError {
+		log.LogIfError(err)
```
iulianpascalau, line 193: maybe log at debug level the original error from vmOutput

vm/systemSmartContracts/staking.go
```diff
@@ -276,8 +276,7 @@ func (r *stakingSC) changeRewardAddress(args *vmcommon.ContractCallInput) vmcomm
 		return vmcommon.UserError
 	}
 
-	for i := 1; i < len(args.Arguments); i++ {
-		blsKey := args.Arguments[i]
+	for _, blsKey := range args.Arguments[1:] {
```
iulianpascalau, line 279: :1st_place_medal:

integrationTests/testProcessorNodeWithMultisigner.go
sasurobert: done

integrationTests/testProcessorNodeWithMultisigner.go
sasurobert: done

vm/systemSmartContracts/auction.go
sasurobert: vmOutput cannot is only userError.

vm/systemSmartContracts/auction.go
sasurobert: i would leave it like this. it is the same.

iulianpascalau: I know, just it was clearer what you add to the arguments list.

vm/systemSmartContracts/auction.go
iulianpascalau: yeah, could have used vmOutput.ReturnMessage but I'm guessing the system SC does not write that.

PR #1753, opened 2020-05-17
removed a mutex from nodes coordinator, removed some weak warnings

iulianpascalau:
- removed the mutex inside `epochNodesConfig` struct
- removed some weak warnings
- added TODOs to further refactor this package


sharding/indexHashedNodesCoordinator.go
```diff
@@ -336,16 +334,13 @@ func (ihgs *indexHashedNodesCoordinator) GetValidatorWithPublicKey(
 		return nil, 0, ErrNilPubKey
 	}
 	ihgs.mutNodesConfig.RLock()
-	nodesConfig, ok := ihgs.nodesConfig[epoch]
-	ihgs.mutNodesConfig.RUnlock()
+	defer ihgs.mutNodesConfig.RUnlock()
```
SebastianMarian, line 337: Add before the return from line 284: ihgs.mutNodesConfig.RUnlock()

iulianpascalau: done

AdoAdoAdo:
I suggest to close this PR as the original branch already solves the issue with the missing mutex unlock and the deadlock because of mutex in end of epoch trigger and channel from peer type provider.

iulianpascalau:
Closed. The change will be postponed. This will be part of a larger refactoring task: https://elrondnetwork.atlassian.net/browse/EN-6445

PR #1757, opened 2020-05-18
Fix maxTrieLevelInMemory invalid value

BeniaminDrasovean:
Added missing mutex to GetAllLeaves() from trie.
Initialize maxTrieLevelInMemory in all the places in which maxTrieLevelInMemory=0
Return error if maxTrieLevelInMemory is initialized with 0.


data/trie/patriciaMerkleTrie.go
```diff
@@ -514,7 +517,9 @@ func (tr *patriciaMerkleTrie) GetSerializedNodes(rootHash []byte, maxBuffToSend
 
 // GetAllLeaves iterates the trie and returns a map that contains all leafNodes information
 func (tr *patriciaMerkleTrie) GetAllLeaves() (map[string][]byte, error) {
-	//TODO: save those leafs into a levelDB struct (cache and storage) and at processing time to get from that structure.
+	tr.mutOperation.Lock()
```
SebastianMarian, line 520: RLock / RUnlock ?

BeniaminDrasovean: Done.

PR #1780, opened 2020-05-21
Add protection for metachain over computation

SebastianMarian:
* Added protection for metachain when it is stuck because of some high load computation
* Fixed panic situation in indexHashedNodesCoordinator when nodesConfig is nil
* Fixed situation when cross mini blocks and cross transactions from shards to meta were not broadcast by proposer at commit time, but only set for delayed broadcast


consensus/spos/bls/subroundEndRound.go
```diff
@@ -376,15 +377,66 @@ func (sr *subroundEndRound) broadcastMiniBlocksAndTransactions() error {
...
+		return sr.broadcast(metaMiniBlocks, metaTransactions)
 	}
 
-	err = sr.BroadcastMessenger().BroadcastMiniBlocks(miniBlocks)
-	if err != nil {
-		return err
+	return sr.broadcast(miniBlocks, transactions)
+}
+
+func (sr *subroundEndRound) extractMetaMiniBlocksAndTransactions(
```
sasurobert, line 393: this became a super special think. might rethink sometime a refactor of consensus, to be more general.

sasurobert: Maybe add a unit test :D

SebastianMarian: Actually the caller method already do things which depend of shardCoordinator.SelfId(), as we wanted to keep only one implementation for all shards

iulianpascalau: Could have become more general by providing the shard ID. For now it will suffice.

PR #1789, opened 2020-05-22
heartbeat peertype fixes

raduchis:
The hash for the validatorStatistics list was the latestRootHash, but the block could still be reverted and inconsistencies could be displayed.
Validators with the list: "eligible (waiting)" could appear.

 - fixed possibility for reverts on validatorStatisticsRootHash by saving the "LastFinalizedRootHash" and when requesting statistics for the heartbeat, this roothash is used.
 - removed possibility with "eligible (waiting)"


process/peer/process.go
```diff
@@ -1075,3 +1076,21 @@ func (vs *validatorStatistics) Process(svi data.ShardValidatorInfoHandler) error
 	pa.SetRating(svi.GetTempRating())
 	return vs.peerAdapter.SaveAccount(pa)
 }
+
+// SetLastFinalizedRootHash - sets the last finalized root hash needed for correct validatorStatistics computations
+func (vs *validatorStatistics) SetLastFinalizedRootHash(lastFinalizedRootHash []byte) {
+	if lastFinalizedRootHash == nil {
```
sasurobert, line 1082: len() == 0

PR #1799, opened 2020-05-25
En 6480 move heartbeat to validator stats

raduchis:
reverted changes from heartbeat and moved to validatorsProvider. This way the heartbeat remains the same for all nodes in the network and the information comes from the nodesCoordinator. The validatorsProvider/validatorStatistics computes differently on the metachain and only there the state is up to date.

- moved query from peerAccountsTrie from heartbeat to validatorStatistics
- added cacheRefreshInterval for the validatorsProvider cache in config


process/errors.go
```diff
@@ -422,17 +422,11 @@ var ErrNilEpochStartTrigger = errors.New("nil start of epoch trigger")
...
 // ErrNilEpochStartNotifier signals that the ErrNilEpochStartNotifier is nil
 var ErrNilEpochStartNotifier = errors.New("nil epochStartNotifier")
 
-// ErrInvalidPeerTypeRefreshIntervalInSec signals that the peerTypeRefreshIntervalInSec is invalid - zero or less
-var ErrInvalidPeerTypeRefreshIntervalInSec = errors.New("nil PeerTypeRefreshIntervalInSec")
-
-// ErrNilContext signals that the context is nil
-var ErrNilContext = errors.New("nil context")
+// ErrInvalidCacheRefreshIntervalInSec signals that the cacheRefreshIntervalInSec is invalid - zero or less
+var ErrInvalidCacheRefreshIntervalInSec = errors.New("nil acheRefreshIntervalInSe")
```
AdoAdoAdo, line 429: fix error message

process/peer/validatorsProvider.go
```diff
@@ -1,69 +1,180 @@
...
+
 	return validatorsProvider, nil
 }
 
 // GetLatestValidators gets the latest configuration of validators from the peerAccountsTrie
 func (vp *validatorsProvider) GetLatestValidators() map[string]*state.ValidatorApiResponse {
-	vp.mutCachedMap.Lock()
-	defer vp.mutCachedMap.Unlock()
+	vp.mutCache.Lock()
+	defer vp.mutCache.Unlock()
```
AdoAdoAdo, line 91: This is an exported function and you are returning a map also used internally without it's mutex, so if the caller accesses the items without locking the same mutex we might get race conditions.
Can we return a map copy instead?

cmd/node/config/config.toml
```diff
@@ -384,6 +383,9 @@
            MaxBatchSize = 100
            MaxOpenFiles = 10
 
+[ValidatorStatistics]
+    CacheRefreshIntervalInSec = 60
```
sasurobert, line 387: why do you need a cache refresh? the data can be taken from the trie at the time of request. If you have another request in a said interval you do return from the cache - only at the next request which is out of the limit yo get the data from the trie again.

process/errors.go
sasurobert: +1

process/peer/peerTypeProvider.go
```diff
@@ -156,70 +110,46 @@ func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
...
 	}
-	aggregatePType(newCache, nodesMapEligible, core.EligibleList)
+	computePeerTypeAndShardId(newCache, nodesMapEligible, core.EligibleList)
 
 	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsPublicKeys(epoch)
 	if err != nil {
 		log.Debug("peerTypeProvider - GetAllWaitingValidatorsPublicKeys failed", "epoch", epoch)
 	}
-	aggregatePType(newCache, nodesMapWaiting, core.WaitingList)
+	computePeerTypeAndShardId(newCache, nodesMapWaiting, core.WaitingList)
```
sasurobert, line 126: update cache only if requests and if request is over the time between requests.

process/errors.go
raduchis: fixed

cmd/node/config/config.toml
raduchis: the idea was not to read from the trie each time, because the validator's trie contains a few KB for each validator and some MB would be read each request.

process/peer/peerTypeProvider.go
raduchis: the update cache method is called here just at the beginning of the epoch so that the heartbeatMonitor has the same values everywhere

cmd/node/config/config.toml
raduchis: fixed. Update only on request refresh, not always on cacheRefreshInterval timout

PR #1802, opened 2020-05-26
Refactor TxPool, improve its robustness

andreibancioiu:
1. Score computation logic extracted to allow for easier change / algorithm replacements
 2. Some components decoupled from `cacheConfig` etc.
 3. `bucketSortedMap` does not call `computeScore` anymore, but receives the score instead 
 4. `txListForSender` does not compute its score, it only computes the `senderScoreParams`.
 5. 3 and 4 allow us to reason more easily about critical sections within `txListForSender`.
 6. Fixed a possible uint32 overflow in `RemoveTxsBulk` (it would have a benign overflow, the value only used to monitor / diagnose eviction)
 7. Wrapped transaction selection, addition and removal in try / catch blocks.
 8. Removed deprecated cache implementation
 9. Add log with number of senders with failed selection (per selection)
 10. Add log with number of transactions to remove when block is finalized (sharded tx pool)


storage/txcache/factory.go
```diff
@@ -0,0 +1,13 @@
+package txcache
+
+// CreateCache creates a (fail-safe) transactions cache
+func CreateCache(config CacheConfig) txCache {
+	cache, err := NewTxCache(config)
+	if err != nil {
+		log.Error("CreateCache()", "err", err)
```
iulianpascalau, line 7: Same question. Why is this a logged error? It can happen during normal operation? If not, why not returning the error? Why does the program runs as nothing bad had happened?

storage/txcache/factory.go
```diff
@@ -0,0 +1,13 @@
+package txcache
+
+// CreateCache creates a (fail-safe) transactions cache
+func CreateCache(config CacheConfig) txCache {
```
iulianpascalau, line 4: Auch. Returning an unexported interface? Why?

storage/txcache/monitoring.go
```diff
@@ -42,10 +42,47 @@ func (cache *TxCache) monitorSelectionStart() *core.StopWatch {
...
+type batchSelectionJournal struct {
+	isFirstBatch  bool
+	copied        int
+	hasInitialGap bool
+	hasMiddleGap  bool
+	isGracePeriod bool
+}
+
+func (cache *TxCache) monitorBatchSelectionEnd(journal batchSelectionJournal) {
+	if journal.isFirstBatch {
```
iulianpascalau, line 70: Can you please refactor this? nested else-ifs inside an if? Why not following the pattern we use in all the rest of the project: if not_ok, return

storage/txcache/score.go
```diff
@@ -2,10 +2,11 @@ package txcache
 
 import (
 	"math"
```
iulianpascalau, line 4: I know I told you to move this in storage rather than core. What was the reason behind this move? The parameters store higher level of info such as gas price and gas limit. Storage is a plugin, **should** not hold elrond business logic. Do we need to reuse this elsewhere?

storage/txcache/score.go
```diff
@@ -93,17 +84,9 @@ func computeSenderScore(params senderScoreParams) float64 {
 	return score
 }
 
-// GetScoreChunk returns the score chunk the sender is currently in
-func (listForSender *txListForSender) GetScoreChunk() *maps.MapChunk {
-	listForSender.scoreChunkMutex.RLock()
-	defer listForSender.scoreChunkMutex.RUnlock()
-
-	return listForSender.scoreChunk
+type disabledScoreComputer struct {
```
iulianpascalau, line 87: When do we use this disabled score computer?

storage/txcache/txCacheFailsafe.go
```diff
@@ -0,0 +1,140 @@
...
+func newTxCacheFailsafe(cache txCache) txCache {
+	return &txCacheFailsafeDecorator{
+		backingCache: cache,
+	}
+}
+
+// AddTx delegates to backing cache, with try / catch
+func (decorator *txCacheFailsafeDecorator) AddTx(tx *WrappedTransaction) (ok bool, added bool) {
+	core.TryCatch(func() {
+		ok, added = decorator.backingCache.AddTx(tx)
```
iulianpascalau, line 25: Why is this allowed to panic?

storage/txcache/txCacheFailsafe.go
```diff
@@ -0,0 +1,140 @@
...
+func (decorator *txCacheFailsafeDecorator) GetByTxHash(txHash []byte) (*WrappedTransaction, bool) {
+	return decorator.backingCache.GetByTxHash(txHash)
+}
+
+// SelectTransactions deledates to backing cache, with try / cache
+func (decorator *txCacheFailsafeDecorator) SelectTransactions(numRequested int, batchSizePerSender int) []*WrappedTransaction {
+	selection := make([]*WrappedTransaction, 0)
+
+	core.TryCatch(func() {
+		selection = decorator.backingCache.SelectTransactions(numRequested, batchSizePerSender)
```
iulianpascalau, line 41: also this? Why this is allowed to panic?

storage/txcache/txCacheFailsafe.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/storage"
+)
+
+var _ storage.Cacher = (*txCacheFailsafeDecorator)(nil)
+var _ txCache = (*txCacheFailsafeDecorator)(nil)
+
+// txCacheFailsafeDecorator is a decorator over a cache, a decorator which wraps some operations in try / catch constructs
+type txCacheFailsafeDecorator struct {
```
iulianpascalau, line 12: Why do we need this fail safe cache? Why the wrapped backing cache is not safe?

storage/txcache/txCacheFailsafe.go
```diff
@@ -0,0 +1,140 @@
...
+	backingCache txCache
```
iulianpascalau, line 13: If this structure is really needed, could have used direct embedding and overload of the required methods. A lot of this code is wrapped code. Please take a look at p2p/libp2p/discover/hostDecorator.go

process/block/preprocess/sortedTransactionsProvider.go
```diff
@@ -0,0 +1,65 @@
+package preprocess
+
+import (
+	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/storage"
+	"github.com/ElrondNetwork/elrond-go/storage/txcache"
+)
+
+// createSortedTransactionsProvider is a "simple factory" for "SortedTransactionsProvider" objects
+func createSortedTransactionsProvider(transactionsPreprocessor *transactions, cache storage.Cacher, cacheKey string) SortedTransactionsProvider {
```
AdoAdoAdo, line 10: transactionsPreprocessor and cacheKey are not used.

storage/txcache/score.go
```diff
@@ -93,17 +84,9 @@ func computeSenderScore(params senderScoreParams) float64 {
...
 }
 
-// SetScoreChunk returns the score chunk the sender is currently in
-func (listForSender *txListForSender) SetScoreChunk(scoreChunk *maps.MapChunk) {
-	listForSender.scoreChunkMutex.Lock()
-	listForSender.scoreChunk = scoreChunk
-	listForSender.scoreChunkMutex.Unlock()
+func (computer *disabledScoreComputer) computeScore(scoreParams senderScoreParams) uint32 {
```
AdoAdoAdo, line 90: unused parameter, maybe use _

storage/txcache/txByHashMap.go
```diff
@@ -58,15 +58,15 @@ func (txMap *txByHashMap) getTx(txHash string) (*WrappedTransaction, bool) {
 
 // RemoveTxsBulk removes transactions, in bulk
 func (txMap *txByHashMap) RemoveTxsBulk(txHashes [][]byte) uint32 {
-	oldCount := uint32(txMap.counter.Get())
+	numRemoved := uint32(0)
 
 	for _, txHash := range txHashes {
-		txMap.removeTx(string(txHash))
+		_, removed := txMap.removeTx(string(txHash))
```
AdoAdoAdo, line 64: the removeTx implementation seems a bit weak as for example if we concurrently try to remove the same txhash we might end up with a wrong count and wrong size (removal succeeds only once but we decrement the counter and the size twice)

storage/txcache/txCache.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 var _ storage.Cacher = (*TxCache)(nil)
+var _ txCache = (*TxCache)(nil)
```
AdoAdoAdo, line 13: Put and HasOrAdd have some unused parameters.

storage/txcache/txCacheFailsafe.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// HasOrAdd delegates to backing cache
+func (decorator *txCacheFailsafeDecorator) HasOrAdd(key []byte, value interface{}) (ok, evicted bool) {
+	return decorator.backingCache.HasOrAdd(key, value)
+}
+
+// Remove delegates to backing cache
+func (decorator *txCacheFailsafeDecorator) Remove(key []byte) {
+	decorator.RemoveTxByHash(key)
```
AdoAdoAdo, line 105: _= unhandled error

storage/txcache/txListBySenderMap.go
```diff
@@ -73,8 +75,10 @@ func (txMap *txListBySenderMap) addSender(sender string) *txListForSender {
 }
 
 // This function should only be called in a critical section managed by a "txListForSender"
-func (txMap *txListBySenderMap) notifyScoreChange(txList *txListForSender) {
-	txMap.backingMap.NotifyScoreChange(txList)
+func (txMap *txListBySenderMap) notifyScoreChange(txList *txListForSender, scoreParams senderScoreParams) {
+	score := txMap.scoreComputer.computeScore(scoreParams)
+	txList.setLastComputedScore(score)
+	txMap.backingMap.NotifyScoreChange(txList, score)
 }
```
AdoAdoAdo, line 82: line 106 removeSender can act incorrectly if called concurrently for same sender, e.g removing the only tx for that sender and removing the sender due to low tx both will call removeSender. The removal is done only once but the counter is decremented twice.

core/tryCatch.go
```diff
@@ -0,0 +1,30 @@
+package core
+
+import (
+	"fmt"
+)
+
+// TryFunction corresponds to the try() part of a try / catch block
+type TryFunction func()
```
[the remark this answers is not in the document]
andreibancioiu: Won't use defer-recover.
Fixed - removed fail-safe decorator and the try-catch constructs.

process/block/preprocess/sortedTransactionsProvider.go
```diff
@@ -0,0 +1,65 @@
+package preprocess
```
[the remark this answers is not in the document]
andreibancioiu: As stated at 8), this PR removes the old, deprecated sorting operations of the old transactions cache.

`preprocess/transactions` handles the `TxCache` using the `Cacher` interface. However, when it selects transactions at proposal time, it requires the **selection** operation, which the `Cacher` interface does not provide. This is why we needed the adapter in the first place. The adapter factory was switching between the new `sortedTxsProvider` and the old one - the old one was used if the cast `cache.(*TxCache)` failed (never the case in real life). Now, since we've removed the old components, we had to use a disabled substitute. Of course, this is suboptimal design. We actually have to fix `transactions.go` to depend on a **well-defined** interface instead, and discard the adapters and the adapter factory altogether. Not doable in this PR though, I've added a TODO now. I will try to figure out another way in the next PR, `pool++cross`.

Furthermore, the factory and the adapters were kept in separate files before this PR. I grouped them together to increase the cohesion of _smelly things that we should remove sometimes soon_. 

Is this reasonable (that is, doing this non-trivial removal / change incrementally to avoid larger changes in the code at once)?

process/block/preprocess/sortedTransactionsProvider.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// createSortedTransactionsProvider is a "simple factory" for "SortedTransactionsProvider" objects
+func createSortedTransactionsProvider(transactionsPreprocessor *transactions, cache storage.Cacher, cacheKey string) SortedTransactionsProvider {
+	txCache, isTxCache := cache.(TxCache)
+	if isTxCache {
+		return newAdapterTxCacheToSortedTransactionsProvider(txCache)
+	}
+
+	log.Error("Could not create a real [SortedTransactionsProvider], will create a disabled one")
```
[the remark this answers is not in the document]
andreibancioiu: Please see comment above.

process/block/preprocess/sortedTransactionsProvider.go
```diff
@@ -0,0 +1,65 @@
...
+	return &disabledSortedTransactionsProvider{}
+}
+
+// adapterTxCacheToSortedTransactionsProvider adapts a "TxCache" to the "SortedTransactionsProvider" interface
+type adapterTxCacheToSortedTransactionsProvider struct {
```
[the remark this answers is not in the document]
andreibancioiu: Please see comment above (we should, indeed, use a better interface in transactions.go).

storage/txcache/txCacheFailsafe.go
andreibancioiu: Removed failsafe decorator. 

Code not allowed to panic, was just an extra safety measure in case of hidden bugs activated by unknown edge cases.

storage/txcache/txCacheFailsafe.go
andreibancioiu: Removed, see above.

storage/txcache/txCacheFailsafe.go
andreibancioiu: Removed, see above.

storage/txcache/txCacheFailsafe.go
andreibancioiu: Removed, see above. Indeed, `hostDecorator` is a good example.

storage/txcache/score.go
andreibancioiu: Scoring computation logic was located in `storage/txcache` since its inception. I've just renamed the file in this PR. The cache itself (storage component) depends on these heuristics to guide the eviction process (initial design, not changed in this PR).

storage/txcache/factory.go
andreibancioiu: Good point! I link here the comments from previous PRs:

[1767](https://github.com/ElrondNetwork/elrond-go/pull/1767#discussion_r429854001)
[1716](https://github.com/ElrondNetwork/elrond-go/pull/1716#discussion_r424249173)

Firstly - it seems that the factory function wasn't used by `shardedTxPool` - possible merge issue, thanks for pointing this out - fixed.

Secondly (comment from previous PR): we should change `shardedTxPool` to create all the needed caches in its constructor, and this way we can return the critical error and stop the node. We can apply this change in the upcoming PR that separates `TxCache` from `CrossTxCache`. And this means we might remove this disabled cache component in the next PR. We have this TODO: _Perhaps create all caches in constructor?_

Is this ok?

storage/txcache/factory.go
andreibancioiu: Actually we often return unexported types from "constructor" functions. This is a particular instance of that practice: we return an unexported type (an interface, not a struct) in a "constructor" function (a simple factory, not an actual constructor in this case).

`txCache` was introduced as an intentionally internal interface - and it was the common denominator for `txCache`, `disabledCache` and `failsafeCache`, just to allow for a return type for `CreateCache`. It it weren't for the `disabledCache`, we would have returned the actual `TxCache` object. 

Anyway, since we plan to eagerly return errors when creating the caches - see comment above - and not have any `disabledCache` anymore, we will also drop this interface.

But in the meantime, can we keep it as unexported / internal interface?

storage/txcache/monitoring.go
andreibancioiu: Missed that, fixed :)

storage/txcache/score.go
andreibancioiu: Missed that - fixed (moved to test utilities).

process/block/preprocess/sortedTransactionsProvider.go
andreibancioiu: Fixed, removed leftovers :)

storage/txcache/score.go
andreibancioiu: Fixed.

storage/txcache/txCacheFailsafe.go
andreibancioiu: The _failSafe_ component was removed in the fix-after-review.

storage/txcache/txCache.go
andreibancioiu: Fixed / `_` (these methods of the `Cacher` interface are not implemented by the `TxCache`).

storage/txcache/txByHashMap.go
andreibancioiu: Good point!

`tx, ok := txMap.getTx(txHash)` could return `ok == true` for both the concurring routines and both will decrement, which, indeed, is not correct.

The issue can appear if transactions are removed during eviction (on high load) and also removed (at the same time) on the "commit / finalize block" event

In `monitoring.go`, in function `diagnose()` we have:

```
numTxsEstimate := int(cache.CountTx()) // this comes from the possibly wrong counter
numTxsInChunks := cache.txByHash.backingMap.Count()
```

If the estimate does not equal `numTxsInChunks`, a `log.Debug()` is emitted and we should see the difference. This `diagnose()` is unfortunately (currently) called only upon transaction selection (on **my turn**).

A possible solution to this issue: in `txMap.backingMap.Remove`, when the map chunk is locked (under mutex), do the "item exists" check and call a `onItemRemovedCallback()` to decrement the atomic counters. Would this be fine?

Another solution would be to apply corrections from time to time - that would work fine for the counter, but the corrections for the _size counter_ would be quite inefficient.

I've added a TODO for this correction (it's an old condition).

storage/txcache/txListBySenderMap.go
andreibancioiu: Good point! As above, added a TODO. Is this ok?

process/block/preprocess/sortedTransactionsProvider.go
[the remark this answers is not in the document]
iulianpascalau: It is reasonable as it is for now.

storage/txcache/factory.go
iulianpascalau: Ok. Will remain as it is for now.

storage/txcache/score.go
iulianpascalau: Now I see all the tx cache components & logic are inside storage package. Something tells me this should be moved maybe in dataRetriever package as it is more strongly coupled to our logic. Along with pruning storer.... Can be left as it is for now.

storage/txcache/txListBySenderMap.go
```diff
@@ -108,6 +108,11 @@ func (txMap *txListBySenderMap) removeSender(sender string) bool {
 		return false
 	}
 
+	// TODO / bugfix: when we concurrently try to remove the same sender,
+	// we might end up decrementing the counter twice.
+	// Possible solution: use an "onItemRemoved" callback in the "backingMap"
```
iulianpascalau, line 113: To prevent double decreasing, might use a mutex or atomic function for that in order to make the decrementation concurrent safe.

storage/txcache/txByHashMap.go
AdoAdoAdo: I was thinking the same, add the check in the Remove under lock.

PR #1804, opened 2020-05-26
Fix trie bugs

BeniaminDrasovean:
- Fixed a bug that wrongfully marked for eviction a node when reduceNode() was called for a branch node.
- The cache for oldHashes and newHashes are now reset every time after a Commit() occurs.


data/trie/patriciaMerkleTrie.go
```diff
@@ -130,6 +132,10 @@ func (tr *patriciaMerkleTrie) Update(key, value []byte) error {
 		}
 		tr.root = newRoot
 		tr.oldHashes = append(tr.oldHashes, oldHashes...)
+
+		for i := range oldHashes {
```
sasurobert, line 136: for _, hash := range oldHashes

data/trie/patriciaMerkleTrie.go
```diff
@@ -145,6 +151,10 @@ func (tr *patriciaMerkleTrie) Update(key, value []byte) error {
...
```
sasurobert, line 155: for _, hash := range oldHashes

data/trie/patriciaMerkleTrie.go
```diff
@@ -369,6 +378,11 @@ func (tr *patriciaMerkleTrie) ResetOldHashes() [][]byte {
 	oldHashes := tr.oldHashes
 	tr.oldHashes = make([][]byte, 0)
 	tr.oldRoot = make([]byte, 0)
+
+	for i := range oldHashes {
```
sasurobert, line 382: for _, hash := range oldHashes

data/trie/patriciaMerkleTrie.go
```diff
@@ -394,6 +408,10 @@ func (tr *patriciaMerkleTrie) GetDirtyHashes() (data.ModifiedHashes, error) {
 		return nil, err
 	}
 
+	for hash := range dirtyHashes {
```
sasurobert, line 411: for _, hash := range oldHashes

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: DirtyHashes is a map[string]struct{}, so it is ok like it is.

PR #1808, opened 2020-05-27
Fix trie memory leak

BeniaminDrasovean:
Execute log.Trace() only if loglevel is `LogTrace`


data/trie/patriciaMerkleTrie.go
```diff
@@ -133,8 +133,10 @@ func (tr *patriciaMerkleTrie) Update(key, value []byte) error {
 		tr.root = newRoot
 		tr.oldHashes = append(tr.oldHashes, oldHashes...)
 
-		for _, hash := range oldHashes {
-			log.Trace("oldHashes after insert", "hash", hash)
+		if log.GetLevel() == logger.LogTrace {
+			for _, hash := range oldHashes {
```
sasurobert, line 137: May take out into a function the whole section

PR #1812, opened 2020-05-27
fix/verify-publkey-validity-all-time

sasurobert:
Currently the message sign verifier is disabled in auction smart contract, even in that case the validity of the bls keys which try to stake should be verified.


genesis/process/disabled/messageSignVerifier.go
```diff
@@ -1,15 +1,39 @@
...
 }
 
-// Verify returns always nil
-func (msv *MessageSignVerifier) Verify(_ []byte, _ []byte, _ []byte) error {
+// Verify returns nil if public key is valid
+func (msv *messageSignVerifier) Verify(_ []byte, _ []byte, pubKey []byte) error {
+	if len(pubKey) == 0 {
+		return genesis.ErrEmptyPubKey
+	}
+	_, err := msv.kg.PublicKeyFromByteArray(pubKey)
```
iulianpascalau, line 28: Surprisingly, this does not work as expected. Please add messageSignVerifier_test.go containing:
```GO
package disabled

import (
    "testing"

    "github.com/ElrondNetwork/elrond-go/crypto/signing"
    "github.com/ElrondNetwork/elrond-go/crypto/signing/mcl"
)

func TestMessageSignVerifier_Verify(t *testing.T) {
    t.Parallel()

    suite := mcl.NewSuiteBLS12()
    keygen := signing.NewKeyGenerator(suite)

    sv, _ := NewMessageSignVerifier(keygen)

    err := sv.Verify(nil, nil, make([]byte, 97))
    assert.NotNil(t, err)
}
```

cmd/node/factory/structs.go
```diff
@@ -918,6 +918,7 @@ func generateGenesisHeadersAndApplyInitialBalances(args *processComponentsFactor
 		TrieStorageManagers:      args.tries.TrieStorageManagers,
```
iulianpascalau, line 918: @LucianMincu did some testing with this branch ans we find out that providing an invalid pub key returns the OK status. Please check auction.go, getVerifiedBLSKeysFromArgs function, L475. I think we should escalate the error in order to stop the staking process.

genesis/process/disabled/messageSignVerifier.go
sasurobert: done

cmd/node/factory/structs.go
sasurobert: fixes on the way.

PR #1821, opened 2020-05-29
Scrips proxy connects meta observers

miiu96:
- Modified testnet scripts to generate setting for proxy application to connects to observer that are on metachain. 
- Also remove log.Debug(...) that is displayed when observers that are connected to `elasticsearch` database try to index a block that does not have miniblocks.

### DIFF
--- a/scripts/testnet/include/config.sh
+++ b/scripts/testnet/include/config.sh
@@ -166,6 +166,17 @@ generateProxyObserverList() {
       let OBSERVER_INDEX++
     done
   done
+  # Start Meta Observers
+  for OBSERVER_IN_SHARD in `seq $META_OBSERVERCOUNT`; do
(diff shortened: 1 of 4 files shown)


scripts/testnet/include/config.sh
```diff
@@ -166,6 +166,17 @@ generateProxyObserverList() {
       let OBSERVER_INDEX++
     done
   done
+  # Start Meta Observers
+  for OBSERVER_IN_SHARD in `seq $META_OBSERVERCOUNT`; do
```
camilbancioiu, line 170: Rename `OBSERVER_IN_SHARD` to `METAOBSERVER`. It doesn't change the function, but it's clearer.

miiu96: done.

PR #1823, opened 2020-05-29
generic API responses

bogdan-rosianu:
Changed the API response structure for all existing routes to include 3 fields:

- Data (which will contain the desired result if the request was successful)
- Error (which will contain the error message or nothing otherwise)
- Code (which will be both human and machine-readable string and will tell the code of the response - example: `bad_request`)


api/address/routes.go
```diff
@@ -39,40 +40,88 @@ func Routes(router *wrapper.RouterWrapper) {
 func GetAccount(c *gin.Context) {
 	ef, ok := c.MustGet("elrondFacade").(FacadeHandler)
 	if !ok {
-		c.JSON(http.StatusInternalServerError, gin.H{"error": errors.ErrInvalidAppContext.Error()})
+		c.JSON(
+			http.StatusInternalServerError,
+			core.GenericAPIResponse{
```
iulianpascalau, line 45: Why not move inside api package? Do we need this elsewhere? Core will get full of unrelated stuff.

api/address/routes.go
```diff
@@ -39,40 +40,88 @@ func Routes(router *wrapper.RouterWrapper) {
...
 func GetBalance(c *gin.Context) {
 	ef, ok := c.MustGet("elrondFacade").(FacadeHandler)
 	if !ok {
-		c.JSON(http.StatusInternalServerError, gin.H{"error": errors.ErrInvalidAppContext.Error()})
+		c.JSON(
+			http.StatusInternalServerError,
+			core.GenericAPIResponse{
+				Data:  nil,
+				Error: errors.ErrInvalidAppContext.Error(),
+				Code:  string(core.ReturnCodeInternalError),
```
iulianpascalau, line 87: return codes also should be inside api package. They are "api" related

api/node/routes.go
```diff
@@ -67,52 +68,115 @@ func Routes(router *wrapper.RouterWrapper) {
...
+		)
 		return
 	}
 
 	details := ef.StatusMetrics().StatusMetricsMapWithoutP2P()
-	c.JSON(http.StatusOK, gin.H{"details": details})
+	c.JSON(
+		http.StatusOK,
+		core.GenericAPIResponse{
+			Data:  gin.H{"details": details},
```
iulianpascalau, line 149: why wrapping it again?

api/node/routes.go
```diff
@@ -67,52 +68,115 @@ func Routes(router *wrapper.RouterWrapper) {
...
+		)
 		return
 	}
 
 	details := ef.StatusMetrics().StatusP2pMetricsMap()
-	c.JSON(http.StatusOK, gin.H{"details": details})
+	c.JSON(
+		http.StatusOK,
+		core.GenericAPIResponse{
+			Data:  gin.H{"details": details},
```
iulianpascalau, line 175: also here?

api/address/routes.go
bogdan-rosianu: first time I placed them in the API package but there were cyclic imports

api/address/routes.go
bogdan-rosianu: first time I placed them in the API package but there were cyclic imports

api/node/routes.go
bogdan-rosianu: in order to group them under a functionality/component. I did the same everywhere. I might rename `details` to `metrics`

api/node/routes.go
bogdan-rosianu: in order to group them under a functionality/component. I did the same everywhere. I might rename details to metrics

api/address/routes.go
bogdan-rosianu: created a sub-package `shared` inside the `api` package

api/address/routes.go
bogdan-rosianu: created a sub-package `shared` inside the `api` package

api/node/routes.go
iulianpascalau: no, the ideea from the meeting was that, undeer the "data" field we should have the raw object. That is what I have understood. @ccorcoveanu ?

ccorcoveanu: A cleaner approach is to have this extra property wrap inside `data`, this way the response is verbose enough so you know what you are getting back.

api/address/routes.go
```diff
@@ -39,40 +40,88 @@ func Routes(router *wrapper.RouterWrapper) {
 func GetAccount(c *gin.Context) {
 	ef, ok := c.MustGet("elrondFacade").(FacadeHandler)
 	if !ok {
-		c.JSON(http.StatusInternalServerError, gin.H{"error": errors.ErrInvalidAppContext.Error()})
+		c.JSON(
+			http.StatusInternalServerError,
+			shared.GenericAPIResponse{
+				Data:  nil,
+				Error: errors.ErrInvalidAppContext.Error(),
+				Code:  string(shared.ReturnCodeInternalError),
```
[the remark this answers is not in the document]
ccorcoveanu: Also, return code suggests integer, so I would use integers for this. If we have the `ReturnCode` type, it could implement `Stringer` interface. We can have a method: `FormatError(code ReturnCode, err error)` that returns `code + ":" + err.Error()`, and this can be our Error for `GenericAPIResponse`.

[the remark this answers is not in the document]
bogdan-rosianu: If I would set it to `ReturnCode` type, why would this help us? You can see the return code of each API call without the need to be included in the response payload. The idea of return code is that we can define custom return codes that can be customized in the future. For now, I only 'copied' the 200, 400 and 500 return codes. 
@andreibancioiu can you please provide an input?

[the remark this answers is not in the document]
ccorcoveanu: I was saying to use the ReturnCode type directly in the structure so you don’t have to do string(ReturnCode)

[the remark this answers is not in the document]
bogdan-rosianu: Done. also, changed return codes messages to snake case

api/node/routes.go
```diff
@@ -279,10 +272,10 @@ func PeerInfo(c *gin.Context) {
 	if !ok {
 		c.JSON(
 			http.StatusInternalServerError,
```
iulianpascalau, line 274: Maybe remove the struct from routes_test.go? The one with the TODO, also the TODO on L352 should be fixed. Thanks.

bogdan-rosianu: done

PR #1825, opened 2020-05-29
revert libp2p libps: reverted to pubsub 0.2.6 libs

iulianpascalau:
revert libp2p libps: reverted to pubsub 0.2.6 libs


iulianpascalau:
closed, problem not originating from libp2p.

PR #1833, opened 2020-06-01
bugFix/stake-stake-claim-auction-sc

sasurobert:
fixing case of stake-stake-claim with nodePrice or greater then nodePrice


vm/systemSmartContracts/auction.go
```diff
@@ -630,6 +630,7 @@ func (s *stakingAuctionSC) stake(args *vmcommon.ContractCallInput) vmcommon.Retu
 		for i := maxNodesToRun*2 + 1; i < uint64(lenArgs); i++ {
 			if len(args.Arguments[i]) == len(args.CallerAddr) {
 				if !isAlreadyRegistered {
+					s.eei.AddReturnMessage("reward address after being registered can be changed only through changeRewardAddress")
```
SebastianMarian, line 633: I think that this message should be placed on else branch. If isAlreadyRegistered

sasurobert: thanks. pushed.

vm/systemSmartContracts/auction.go
```diff
@@ -631,6 +631,8 @@ func (s *stakingAuctionSC) stake(args *vmcommon.ContractCallInput) vmcommon.Retu
 			if len(args.Arguments[i]) == len(args.CallerAddr) {
 				if !isAlreadyRegistered {
 					registrationData.RewardAddress = args.Arguments[i]
```
raduchis, line 633: there is only one registrationData.RewardAddress, but you are iterating through "i" rewardAddresses. Only the last one will be saved, is this the intended behaviour?

PR #1835, opened 2020-06-01
bugFix/account-wrap-is-nil

sasurobert:
the error is from stakingToPeer RemoveAccount call. Still missing a small piece of information - what is the main cause.


process/scToProtocol/stakingToPeer.go
```diff
@@ -165,7 +165,7 @@ func (stp *stakingToPeer) UpdateProtocol(body *block.Body, nonce uint64) error {
 		if len(data) == 0 {
 			err = stp.peerState.RemoveAccount(blsPubKey)
 			if err != nil {
-				return err
+				log.Debug("staking to protocol RemoveAccount error", "err", err, "blsPubKey", blsPubKey)
```
SebastianMarian, line 168: log.Error ? and "staking to protocol RemoveAccount failed" to avoid displaying twice "error" and "err"

process/smartContract/process.go
```diff
@@ -258,7 +258,7 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
 	var results []data.TransactionHandler
 	results, consumedFee, err = sc.processVMOutput(vmOutput, txHash, tx, acntSnd, vmInput.CallType)
 	if err != nil {
-		log.Trace("process vm output error", "error", err.Error())
+		log.Trace("process vm output returned with problem ", "err", err.Error())
```
SebastianMarian, line 261: Should be log.Debug ? and "process vm output failed" ?

vm/systemSmartContracts/auction.go
```diff
@@ -471,7 +467,7 @@ func (s *stakingAuctionSC) registerBLSKeys(
 ) ([][]byte, error) {
 	maxNodesToRun := big.NewInt(0).SetBytes(args[0]).Uint64()
 	if uint64(len(args)) < maxNodesToRun+1 {
-		log.Debug("not enough arguments to process stake function")
+		s.eei.AddReturnMessage("not enough arguments to process stake function")
```
SebastianMarian, line 470: We can use the same pattern used in this file:
s.eei.AddReturnMessage(fmt.Sprintf("not enough arguments to process stake function: expected min %d, got %d", maxNodesToRun+1, len(args)))

process/coordinator/process.go
```diff
@@ -708,8 +708,9 @@ func (tc *transactionCoordinator) CreateMarshalizedData(body *block.Body) map[st
 			continue
 		}
 
+		isPreProcessMiniBlock := miniBlock.Type == block.TxBlock
```
raduchis, line 711: only txBlock miniblocks can be "preprocessed" ?

sasurobert: marshalled data for cross shard can come from preprocessor only in case of transactions.

data/state/accountsDB.go
```diff
@@ -264,6 +264,9 @@ func (adb *AccountsDB) RemoveAccount(address []byte) error {
 	if err != nil {
 		return err
 	}
+	if acnt == nil {
```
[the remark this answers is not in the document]
sasurobert: the getAccount func returns nil, nil. it is used the same way in LoadAccount.

process/smartContract/process.go
sasurobert: no need. for any sc fail you would get a lot of messages. which are not needed.

data/state/journalEntries.go
```diff
@@ -101,7 +103,7 @@ type journalEntryDataTrieUpdates struct {
 // NewJournalEntryDataTrieUpdates outputs a new JournalEntryDataTrieUpdates implementation used to revert an account's data trie
 func NewJournalEntryDataTrieUpdates(trieUpdates map[string][]byte, account baseAccountHandler) (*journalEntryDataTrieUpdates, error) {
 	if check.IfNil(account) {
-		return nil, ErrNilAccountHandler
+		return nil, fmt.Errorf("%w in NewJournalEntryAccount", ErrNilAccountHandler)
```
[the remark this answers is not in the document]
sasurobert: done.

process/scToProtocol/stakingToPeer.go
sasurobert: I will put to log.logiferror.

vm/systemSmartContracts/auction.go
sasurobert: done

process/coordinator/process.go
raduchis: ok

data/state/accountsDB.go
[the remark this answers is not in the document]
raduchis: ok

PR #1841, opened 2020-06-01
[Minor bug / minor fix] TxPool - fix inconsistent counters (estimators)

andreibancioiu:
[Original conversation in PR 1802](https://github.com/ElrondNetwork/elrond-go/pull/1802#discussion_r431027949).

Fixed some inconsistent counters for senders and transactions within the internal structures of the transactions pool. Inconsistencies could occur in case of high concurrency and / or concurrent removals (eviction plus removal due to commit / finalize). The bug was mostly benign, but if it were, say, thousands of such events, the counters would have become slightly off and therefore could have lead - in extreme cases - to OOM (since eviction would have been based on bad, low counters).

Details of the fix:
 - `txListBySenderMap.addSender()`: counter is consistent because the function is called under `txListBySenderMap`'s global mutex
 - `txListBySenderMap.removeSender()`: fixed
 - `txByHashMap.addTx()`: counter is consistent because we use "SetIfAbsent"
 - `txByHashMap.removeTx()`: fixed

### DIFF
--- a/storage/txcache/txByHashMap.go
+++ b/storage/txcache/txByHashMap.go
@@ -34,19 +34,18 @@ func (txMap *txByHashMap) addTx(tx *WrappedTransaction) bool {
 
 // removeTx removes a transaction from the map
 func (txMap *txByHashMap) removeTx(txHash string) (*WrappedTransaction, bool) {
-	tx, ok := txMap.getTx(txHash)
-	if !ok {
+	item, removed := txMap.backingMap.Remove(txHash)
+	if !removed {
 		return nil, false
 	}
 
-	// TODO / bugfix: when we concurrently try to remove the same transaction,
-	// we might end up decrementing the counters twice.
-	// Possible solution: use an "onItemRemoved" callback in the "backingMap"
-	// to decrement the counters.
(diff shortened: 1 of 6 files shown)


storage/txcache/txByHashMap.go
```diff
@@ -34,19 +34,18 @@ func (txMap *txByHashMap) addTx(tx *WrappedTransaction) bool {
...
+	item, removed := txMap.backingMap.Remove(txHash)
+	if !removed {
 		return nil, false
 	}
 
-	// TODO / bugfix: when we concurrently try to remove the same transaction,
-	// we might end up decrementing the counters twice.
-	// Possible solution: use an "onItemRemoved" callback in the "backingMap"
-	// to decrement the counters.
+	tx := item.(*WrappedTransaction)
```
SebastianMarian, line 42: You can check for correct type assertion here, so if will be the case in some strange circumstances you will avoid passing nil to line 46 in estimateTxSize method and also to return nil, true in line 49

```
tx, ok := item.(*WrappedTransaction)
if !ok {
   return nil, false
}
```

PR #1867, opened 2020-06-04
Compute shard id start in epoch false

sasurobert:


epochStart/bootstrap/process_test.go
```diff
@@ -161,6 +162,12 @@ func TestIsStartInEpochZero(t *testing.T) {
 func TestEpochStartBootstrap_BootstrapStartInEpochNotEnabled(t *testing.T) {
 	args := createMockEpochStartBootstrapArgs()
 
+	err := errors.New("localErr")
+	args.LatestStorageDataProvider = &mock.LatestStorageDataProviderStub{
+		GetCalled: func() (storage.LatestDataFromStorage, error) {
+			return storage.LatestDataFromStorage{}, err
```
iulianpascalau, line 168: Is this intended? Returning both useful data and error?

sasurobert: It is not useful. It is a strict not a pointer to a struct. Only error is important for this

iulianpascalau: yup, is not a pointer, missed that

PR #1868, opened 2020-06-04
EN-6612 added ports range capability on network messenger

iulianpascalau:
- added port range capability on the network messenger


p2p/errors.go
```diff
@@ -67,23 +67,29 @@ var ErrTooManyGoroutines = errors.New(" number of goroutines exceeded")
...
 
-// ErrWatchdogAlreadyStarted signals that a peer discovery watchdog is already started
-var ErrWatchdogAlreadyStarted = errors.New("peer discovery watchdog is already started")
+// ErrInvalidEndingPortValue signals that an invalid ending port value has been provided
+var ErrInvalidEndingPortValue = errors.New("invalid ending port value")
 
-// ErrWatchdogNotStarted signals that a peer discovery watchdog is not started
-var ErrWatchdogNotStarted = errors.New("peer discovery watchdog is not started")
+// ErrEndPortSmallerThanStartPort signals that the ending port value is smaller than the starting port value
+var ErrEndPortSmallerThanStartPort = errors.New("ending port value is less than the starting port value")
```
SebastianMarian, line 83: is smaller

p2p/libp2p/ports.go
```diff
@@ -0,0 +1,90 @@
...
+
+		return val, nil
+	}
+
+	ports := strings.Split(port, "-")
+	if len(ports) != 2 {
+		return 0, fmt.Errorf("%w, provided port string `%s` is not in the correct format, expected `start-end`", p2p.ErrInvalidPortsRangeString, port)
+	}
+
+	startPort, err := strconv.Atoi(ports[0])
```
SebastianMarian, line 30: This does not should be checked for negative value?

p2p/libp2p/ports.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return 0, p2p.ErrInvalidStartingPortValue
+	}
+
+	endPort, err := strconv.Atoi(ports[1])
```
SebastianMarian, line 35: This does not should be checked for negative value?

p2p/errors.go
iulianpascalau: done

p2p/libp2p/ports.go
iulianpascalau: it is checked against minRangePortValue which is 1025.

p2p/libp2p/ports.go
iulianpascalau: it is checked against startPort which should be greater than minRangePortValue

p2p/libp2p/ports.go
```diff
@@ -0,0 +1,89 @@
...
+			continue
+		}
+
+		return p, nil
+	}
+
+	return 0, fmt.Errorf("%w, range %d-%d", p2p.ErrNoFreePortInRange, startPort, endPort)
+}
+
+func testFreePort(port int) error {
```
sasurobert, line 75: rename to checkFreePort

p2p/libp2p/ports.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/ElrondNetwork/elrond-go/core/random"
+	"github.com/ElrondNetwork/elrond-go/p2p"
+)
+
+type testPortHandler = func(int) error
```
sasurobert, line 13: this type def is not necessary. if yes rename.

iulianpascalau: removed type definition. However, I kept the function pointer in order to better test the functionality.

p2p/libp2p/ports.go
iulianpascalau: done

PR #1870, opened 2020-06-04
En 6618 p2p rebroadcast message

iulianpascalau:
- added extra message ID cacher in network messenger


p2p/libp2p/netMessenger.go
```diff
@@ -408,7 +412,12 @@ func (netMes *networkMessenger) Close() error {
 	err := netMes.outgoingPLB.Close()
 	log.LogIfError(err)
 
-	return netMes.p2pHost.Close()
+	err = netMes.p2pHost.Close()
+	log.LogIfError(err)
+
+	log.Debug("network messenger closed")
```
SebastianMarian, line 418: I think that this message should be print only if err == nil
You can add in line 417: if err != nil { return err} and in line 420: return nil

iulianpascalau: I think it is better to try close everything. Refactored the code.

SebastianMarian: Of course, but at line 417 already all is closed with error or not

PR #1876, opened 2020-06-05
bug fix get transaction api route

miiu96:
Fix panic bug when try to get a transaction which is in pool using api route /transaction/:txhash

Bug: In pool transactions are kept as objects ( Transaction, RewardTransaction, SmartContracResult, etc..) and when you try to get a transaction from pool there was a wrong cast to []byte which panics.

Fix: Removed the wrong cast and added a new function which do a cast to a specific transaction.


node/nodeTransactions_test.go
```diff
@@ -500,17 +502,42 @@ func TestNode_GetTransaction_ShouldNotFindAndReturnUnknown(t *testing.T) {
 	assert.Error(t, err)
 }
 
-func getCacherHandler(find bool) func() dataRetriever.ShardedDataCacherNotifier {
+func getCacherHandler(find bool, cacherType string) func() dataRetriever.ShardedDataCacherNotifier {
```
andreibancioiu, line 505: You could have created separate functions to create the dummy pools (without the switch). But ok for now.

PR #1894, opened 2020-06-09
TermUI issues fix

bogdan-rosianu:
Fixed the following bugs:

- if the node started with TermUI and the `Ctrl+C` signal was intercepted, the node's process didn't stop as expected. Now the TermUI console will start only after the initial pre-requisites are met (start in epoch, p2p bootstrap, and so on) so the process will be stopped accordingly.
- if the node started with TermUI the log file did not contain all logs. Fixed
- if the TermUI application didn't succeed when connecting to a node, it failed. Now it will keep trying connecting until successful 
- epoch number displayed in TermUI was delayed in comparison with the epoch which was displayed in logs. Fixed by fetching the epoch displayed in TermUI from shard header instead of meta block
- latest stable software version wasn't displayed right in TermUI and an `[invalid_key]` appeared instead. Fixed by setting an initial value for that metric + delayed the start of TermUI, thing which should leave enough time for the software version to be fetched

### DIFF
--- a/cmd/node/factory/structsStatusHandlers.go
+++ b/cmd/node/factory/structsStatusHandlers.go
@@ -1,12 +1,14 @@
 package factory
 
 import (
+	"fmt"
 	"io"
 	"math/big"
 	"os"
 
 	logger "github.com/ElrondNetwork/elrond-go-logger"
 	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/check"
 	"github.com/ElrondNetwork/elrond-go/core/statistics"
 	"github.com/ElrondNetwork/elrond-go/data/metrics"
 	"github.com/ElrondNetwork/elrond-go/data/typeConverters"
@@ -27,8 +29,8 @@ type ArgStatusHandlers struct {
 	Ctx                          *cli.Context
 	Marshalizer                  marshal.Marshalizer
 	Uint64ByteSliceConverter     typeConverters.Uint64ByteSliceConverter
-	ChanStartViews               chan bool
-	ChanLogRewrite               chan bool
+	ChanStartViews               chan struct{}
+	ChanLogRewrite               chan struct{}
 	LogFile                      *os.File
 }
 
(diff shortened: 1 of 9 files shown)


cmd/node/factory/structsStatusHandlers.go
```diff
@@ -63,19 +73,28 @@ func CreateStatusHandlers(arguments *ArgStatusHandlers) (*statusHandlersInfo, er
...
-		writer, ok := presenterStatusHandler.(io.Writer)
-		if ok {
-			logger.ClearLogObservers()
-			err = logger.AddLogObserver(writer, &logger.PlainFormatter{})
-			if err != nil {
-				return nil, err
+		go func() {
+			<-arguments.ChanLogRewrite
+			writer, ok := presenterStatusHandler.(io.Writer)
+			if ok {
```
iulianpascalau, line 84: might invert this as 
```
 if !ok{
         return
 }
```

cmd/node/factory/structsStatusHandlers.go
```diff
@@ -63,19 +73,28 @@ func CreateStatusHandlers(arguments *ArgStatusHandlers) (*statusHandlersInfo, er
...
+				logger.ClearLogObservers()
+				err = logger.AddLogObserver(writer, &logger.PlainFormatter{})
+				if err != nil {
+					log.Warn("cannot add log observer for TermUI", "error", err)
+				}
+				if arguments.LogFile != nil {
```
iulianpascalau, line 90: change the condition and do an early return

cmd/node/factory/structsStatusHandlers.go
```diff
@@ -43,12 +47,18 @@ func NewStatusHandlersFactoryArgs(
...
+	chanStartViews chan bool,
+	chanLogRewrite chan bool,
+	logFile *os.File,
 ) *ArgStatusHandlers {
 	return &ArgStatusHandlers{
 		LogViewName:              logViewName,
 		Ctx:                      ctx,
 		Marshalizer:              marshalizer,
 		Uint64ByteSliceConverter: uint64ByteSliceConverter,
+		ChanStartViews:           chanStartViews,
```
iulianpascalau, line 59: both channels are not checked for nil, code might panic.

cmd/node/main.go
```diff
@@ -878,6 +889,9 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 		return err
 	}
 
+	chanLogRewrite <- true
```
iulianpascalau, line 892: the application will lock here if the node is not started in termui mode because no one will read on those channels. Make them buffered?

cmd/termui/provider/logProvider.go
```diff
@@ -26,9 +27,15 @@ func InitLogHandler(nodeURL string, profile logger.Profile, useWss bool) error {
 	if useWss {
 		scheme = wss
 	}
-	webSocket, err = openWebSocket(scheme, nodeURL, profile)
-	if err != nil {
-		return err
+	for {
```
iulianpascalau, line 30: :+1:

statusHandler/view/termuic/termuiConsole.go
```diff
@@ -47,13 +47,11 @@ func NewTermuiConsole(presenter view.Presenter) (*TermuiConsole, error) {
 }
 
 // Start method - will start termui console
-func (tc *TermuiConsole) Start() error {
-	if err := ui.Init(); err != nil {
-		return err
-	}
-
+func (tc *TermuiConsole) Start(chanStart chan bool) error {
```
iulianpascalau, line 50: should check for nil for  chanStart

cmd/node/factory/structsStatusHandlers.go
bogdan-rosianu: done

cmd/node/factory/structsStatusHandlers.go
bogdan-rosianu: done

cmd/node/factory/structsStatusHandlers.go
bogdan-rosianu: added checks for all parameters

cmd/node/main.go
bogdan-rosianu: thanks. done

statusHandler/view/termuic/termuiConsole.go
bogdan-rosianu: done

PR #1895, opened 2020-06-09
EN-6710 antiflood blocks consensus fix

iulianpascalau:
- changed peerMaxMessagesPerSec to 5 in order to avoid consensus topic being block by the topic flood preventer component
- tweaked antiflood parameters
- added sequence number in antiflood debug lines


consensus/spos/bls/blsWorker.go
```diff
@@ -5,7 +5,17 @@ import (
 	"github.com/ElrondNetwork/elrond-go/consensus/spos"
 )
 
-const peerMaxMessagesPerSec = uint32(2)
+// peerMaxMessagesPerSec defines how many messages can be propagated by a pid in a round. The value was chosen by
```
andreibancioiu, line 8: :+1:

PR #1905, opened 2020-06-10
jailed nodes in epoch zero

raduchis:
after an unstake in epoch 0, the validatorStatus was set to jailed although it should be inactive.

- fixed validatorStatus problem if unstake in epoch zero

### DIFF
--- a/process/peer/process.go
+++ b/process/peer/process.go
@@ -423,7 +423,7 @@ func (vs *validatorStatistics) getValidatorDataFromLeaves(
 
 func getActualList(peerAccount state.PeerAccountHandler) string {
 	savedList := peerAccount.GetList()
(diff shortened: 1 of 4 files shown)


process/peer/process.go
```diff
@@ -423,7 +423,7 @@ func (vs *validatorStatistics) getValidatorDataFromLeaves(
 
 func getActualList(peerAccount state.PeerAccountHandler) string {
 	savedList := peerAccount.GetList()
-	if peerAccount.GetUnStakedEpoch() == 0 {
+	if peerAccount.GetUnStakedEpoch() == math.MaxUint32 {
```
sasurobert, line 426: set unstaked epoch to maxuint when initializing a new node. This can be done in NewPeerAccount

PR #1909, opened 2020-06-10
bugFix/rewards-missmatch-computation

sasurobert:
Fixed the wrong computation of end of epoch rewards. In case the same validator had multiple nodes in multiple shards the actual reward was calculated with NumTotalSelected divided by the last shardID set for this validator. The last shard ID set was from range over a map.

### DIFF
--- a/epochStart/metachain/rewards.go
+++ b/epochStart/metachain/rewards.go
@@ -53,13 +53,10 @@ type rewardsCreator struct {
 }
 
 type rewardInfoData struct {
-	ShardID                    uint32
-	LeaderSuccess              uint32
-	LeaderFailure              uint32
-	ValidatorSuccess           uint32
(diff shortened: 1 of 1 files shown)


epochStart/metachain/rewards.go
```diff
@@ -53,13 +53,10 @@ type rewardsCreator struct {
...
 
 type rewardInfoData struct {
-	ShardID                    uint32
-	LeaderSuccess              uint32
-	LeaderFailure              uint32
-	ValidatorSuccess           uint32
-	ValidatorFailure           uint32
-	NumSelectedInSuccessBlocks uint32
-	AccumulatedFees            *big.Int
+	numSelectedInSuccessBlocks uint32
```
SebastianMarian, line 56: This is unused?

PR #1914, opened 2020-06-10
Check if txs and mbs exists in es

miiu96:
### Bug fix
`Bug`: Status of transactions that are indexed in elasticserach database remains Pending because cross shard transactions was indexed first by the observer from destination shard and after that was indexed again by  the observer from source shard. This happens because observer from source shard was behind. 
`Fix` : Before index a transaction an extra check was added to verify if transaction exists in database.

This bug also occurs when mini-blocks are indexed

### DIFF
--- a/core/indexer/common.go
+++ b/core/indexer/common.go
@@ -326,12 +326,11 @@ func serializeBulkTxs(
 	for _, tx := range bulk {
 		var meta, serializedData []byte
 		if existsInDb[tx.Hash] {
-			if isCrossShardDstMe(tx, selfShardID) && tx.Status != txStatusInvalid {
-				// update transaction
-				meta, serializedData = prepareTxUpdate(tx)
-			} else {
+			if !isCrossShardDstMe(tx, selfShardID) || tx.Status == txStatusInvalid {
 				continue
 			}
+			// update transaction
+			meta, serializedData = prepareTxUpdate(tx)
 		} else {
 			// insert transaction in database
 			meta = []byte(fmt.Sprintf(`{ "index" : { "_id" : "%s", "_type" : "%s" } }%s`, tx.Hash, "_doc", "\n"))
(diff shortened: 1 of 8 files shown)


core/indexer/common.go
```diff
@@ -239,13 +239,29 @@ func canInterpretAsString(bytes []byte) bool {
 	return true
 }
 
-func serializeBulkMiniBlocks(hdrShardID uint32, bulkMbs []*Miniblock) bytes.Buffer {
+func serializeBulkMiniBlocks(
+	hdrShardID uint32,
+	bulkMbs []*Miniblock,
+	foundedObjMap func(hashes []string, index string) (map[string]bool, error),
```
ccorcoveanu, line 245: Should rename to something like `getAlreadyIndexedItems`. Just an idea, but `foundedObjMap` does not help to understand what it does

core/indexer/common.go
```diff
@@ -254,12 +270,18 @@ func serializeBulkMiniBlocks(hdrShardID uint32, bulkMbs []*Miniblock) bytes.Buff
...
 			}
-
 		} else {
 			// update miniblock
 			meta = []byte(fmt.Sprintf(`{ "update" : { "_id" : "%s", "_type" : "%s"  } }%s`, mb.Hash, "_doc", "\n"))
-			serializedData = []byte(fmt.Sprintf(`{ "doc" : { "receiverBlockHash" : "%s" } }`, mb.ReceiverBlockHash))
+			if hdrShardID == mb.SenderShardID {
+				// update sender block hash
+				serializedData = []byte(fmt.Sprintf(`{ "doc" : { "senderBlockHash" : "%s" } }`, mb.SenderBlockHash))
+			} else {
```
ccorcoveanu, line 279: Can we remove the else here so we'd have less indentation?
```go
serializedData = []byte(fmt.Sprintf(`{ "doc" : { "receiverBlockHash" : "%s" } }`, mb.ReceiverBlockHash))
if hdrShardID == mb.SenderShardID {
    // update sender block hash
    serializedData = []byte(fmt.Sprintf(`{ "doc" : { "senderBlockHash" : "%s" } }`, mb.SenderBlockHash))
}
```

core/indexer/common.go
```diff
@@ -282,18 +304,35 @@ func prepareBufferMiniblocks(buff bytes.Buffer, meta, serializedData []byte) byt
...
+		log.Warn("indexer", "error", err.Error())
+		return buff
+	}
+
 	for _, tx := range bulk {
 		var meta, serializedData []byte
-
-		if isCrossShardDstMe(tx, selfShardID) && tx.Status != txStatusInvalid {
-			// update tx
+		if existsInDb[tx.Hash] {
```
ccorcoveanu, line 328: maybe extract this if/else into a function like `getSerializedData`?

core/indexer/common.go
miiu96: good suggestion. renamed.

core/indexer/common.go
miiu96: maybe we can remove the `else` from this `if` but if we do that we will have two assignments in variable `serializedData`  and one of them is pointless

core/indexer/common.go
miiu96: done.

PR #1930, opened 2020-06-11
Confirmation of StartOfEpoch block too fast

SebastianMarian:
* Fixed situation when meta block which confirms the start of epoch meta block was rejected by shards because EpochStartPrepare was not executed yet. Actually the real problem was because of the following things:

1. Broadcast of meta mini blocks (peer mini blocks) to shards was wrongly delayed by leader which created start of epoch meta block (with 2 sec)
2. ExtraDelayForRequestBlockInfo was set to a value too much higher of (4 sec), which created problems on testnets which ran with round duration lower than 6 sec -> 2 sec (from point 1) + 4 sec (from point 2)
3. Whitelist/reject of peer mini blocks should not be done for now (if we want this mechanism in the future we can use then, a delay of 1 sec (at point 1) and afterwards we can put back the whitelist mechanism also for peer mini blocks


consensus/spos/bls/subroundEndRound.go
```diff
@@ -396,7 +396,7 @@ func (sr *subroundEndRound) broadcastMiniBlocksAndTransactions() error {
 		return nil
 	}
 
-	go sr.broadcast(miniBlocks, transactions, core.ExtraDelayForBroadcastBlockInfo)
+	go sr.broadcast(miniBlocks, transactions, 0)
```
sasurobert, line 399: This reintroduces the problem of receiving miniblocks before header. A slight delay is still needed here

SebastianMarian: It is solved as peer mini blocks from meta were not rejected anymore

SebastianMarian: If we want to white list them we should apply what I said in point 3. drom description

sasurobert: All miniblocks should be whitelisted

PR #1935, opened 2020-06-12
fix termui app when shuffled out or node stops

bogdan-rosianu:
Implemented a mechanism that keeps trying to re-connect to the node when log messages are not received anymore. this could have happened when the node was shuffled out or was simply stopped.

### DIFF
--- a/cmd/termui/provider/logProvider.go
+++ b/cmd/termui/provider/logProvider.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"encoding/hex"
+	"fmt"
 	"net/url"
 	"strings"
 	"time"
@@ -21,22 +22,25 @@ const (
 )
 
 // InitLogHandler will open the websocket and set the log level
-func InitLogHandler(nodeURL string, profile logger.Profile, useWss bool) error {
+func InitLogHandler(presenter PresenterHandler, nodeURL string, profile logger.Profile, useWss bool) error {
 	var err error
 	scheme := ws
 	if useWss {
 		scheme = wss
 	}
-	for {
-		webSocket, err = openWebSocket(scheme, nodeURL, profile)
-		if err == nil {
-			break
-		}
-		log.Error("init logger", "error", err)
+	go func() {
+		for {
+			webSocket, err = openWebSocket(scheme, nodeURL, profile)
+			if err != nil {
+				_, _ = presenter.Write([]byte(fmt.Sprintf("termui websocket error, retrying in %v...", retryDuration)))
+				time.Sleep(retryDuration)
+				continue
+			}
 
-		log.Info("connection not available", "duration before retrying", retryDuration)
-		time.Sleep(retryDuration)
-	}
+			startListeningOnWebSocket(presenter)
+			time.Sleep(retryDuration)
+		}
+	}()
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


cmd/termui/provider/logProvider.go
```diff
@@ -21,22 +22,25 @@ const (
...
-		webSocket, err = openWebSocket(scheme, nodeURL, profile)
-		if err == nil {
-			break
-		}
-		log.Error("init logger", "error", err)
+	go func() {
+		for {
+			webSocket, err = openWebSocket(scheme, nodeURL, profile)
+			if err != nil {
+				_, _ = presenter.Write([]byte(fmt.Sprintf("termui websocket error, retrying in %v...", retryDuration)))
```
sasurobert, line 35: check.IfNil for presented. Add context for infinite goRoutine in order to have Close() as well.

cmd/termui/provider/logProvider.go
```diff
@@ -65,29 +69,26 @@ func openWebSocket(scheme string, address string, profile logger.Profile) (*webs
...
-			if !isConnectionClosed {
-				log.Error("termui websocket error", "error", err.Error())
-			} else {
-				log.Debug("termui websocket terminated", "error", err.Error())
-			}
+// startListeningOnWebSocket will listen if a new log message is received and will display it
+func startListeningOnWebSocket(presenter PresenterHandler) {
+	for {
+		msgType, message, err := webSocket.ReadMessage()
+		if msgType == websocket.CloseMessage {
```
sasurobert, line 76: if err != nil do the last part of the code - which has the return as well.

cmd/termui/provider/logProvider.go
bogdan-rosianu: this is a separate application. The goroutine is stopped when closing the application. 
Regarding the check.IfNil, the presenter is initialized before sending it to the method so I don't think there is a need to check it for nil because if it would be nil, the application wouldn't arrive here

cmd/termui/provider/logProvider.go
bogdan-rosianu: It does. if err is not nil, then it will check the error's type, will log the error type and then there is a return

cmd/termui/provider/logProvider.go
```diff
@@ -65,29 +69,26 @@ func openWebSocket(scheme string, address string, profile logger.Profile) (*webs
...
+		}
+		if err == nil {
+			writeMessage(presenter, message)
+			continue
+		}
 
-			time.Sleep(retryDuration)
+		_, isConnectionClosed := err.(*websocket.CloseError)
+		if !isConnectionClosed {
+			log.Error("termui websocket error", "error", err.Error())
```
miiu96, line 86: maybe you can write this error in termui log because

cmd/termui/provider/logProvider.go
```diff
@@ -65,29 +69,26 @@ func openWebSocket(scheme string, address string, profile logger.Profile) (*webs
...
+		} else {
+			log.Debug("termui websocket terminated", "error", err.Error())
```
miiu96, line 88: also this

cmd/termui/provider/logProvider.go
bogdan-rosianu: done

cmd/termui/provider/logProvider.go
bogdan-rosianu: done

PR #1952, opened 2020-06-15
Fix trie snapshot concurrency

BeniaminDrasovean:
_Bug_: When GetSnapshotThatContainsHash() is called, the pointer to the snapshotDb is returned. But concurrently, the same db could have been closed and removed, thus resuting in a concurrency problem.

### DIFF
--- a/data/trie/patriciaMerkleTrie.go
+++ b/data/trie/patriciaMerkleTrie.go
@@ -329,6 +329,7 @@ func (tr *patriciaMerkleTrie) recreateFromSnapshotDb(rootHash []byte) (data.Trie
 	if db == nil {
 		return nil, ErrHashNotFound
 	}
+	defer db.DecreaseNumReferences()
 
 	newTr, newRoot, err := tr.recreateFromDb(rootHash, db, tr.trieStorage)
 	if err != nil {
@@ -340,8 +341,6 @@ func (tr *patriciaMerkleTrie) recreateFromSnapshotDb(rootHash []byte) (data.Trie
 	if err != nil {
 		return nil, err
 	}
-
-	db.DecreaseNumReferences()
 
 	return newTr, nil
 }
@@ -535,6 +534,7 @@ func (tr *patriciaMerkleTrie) GetSerializedNodes(rootHash []byte, maxBuffToSend 
 	if db == nil {
 		return nil, 0, ErrHashNotFound
 	}
+	defer db.DecreaseNumReferences()
 
 	newTsm, err := NewTrieStorageManagerWithoutPruning(db)
 	if err != nil {
@@ -579,8 +579,6 @@ func (tr *patriciaMerkleTrie) GetSerializedNodes(rootHash []byte, maxBuffToSend 
 	}
 
 	remainingSpace := maxBuffToSend - size
-
-	db.DecreaseNumReferences()
 
 	return nodes, remainingSpace, nil
 }
--- a/data/trie/patriciaMerkleTrie.go
+++ b/data/trie/patriciaMerkleTrie.go
@@ -319,6 +319,7 @@ func (tr *patriciaMerkleTrie) recreateFromMainDb(rootHash []byte) data.Trie {
 	newTr, _, err := tr.recreateFromDb(rootHash, tr.Database(), tr.trieStorage)
 	if err != nil {
 		log.Warn("trie recreate error:", "error", err, "root", hex.EncodeToString(rootHash))
+		return nil
 	}
 
 	return newTr
@@ -513,13 +514,13 @@ func getDbThatContainsHash(trieStorage data.StorageManager, rootHash []byte) dat
 	}
 
 	_, err := trieStorage.Database().Get(rootHash)
-	if err == nil {
-		return &snapshotDb{
-			DBWriteCacher: trieStorage.Database(),
-		}
+	if err != nil {
+		return nil
 	}
 
-	return nil
+	return &snapshotDb{
+		DBWriteCacher: trieStorage.Database(),
+	}
 }
 
 // GetSerializedNodes returns a batch of serialized nodes from the trie, starting from the given hash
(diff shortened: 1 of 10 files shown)


data/trie/patriciaMerkleTrie.go
```diff
@@ -302,24 +302,47 @@ func (tr *patriciaMerkleTrie) Recreate(root []byte) (data.Trie, error) {
...
+
+func (tr *patriciaMerkleTrie) recreateFromMainDb(rootHash []byte) data.Trie {
+	_, err := tr.Database().Get(rootHash)
+	if err != nil {
+		return nil
+	}
+
+	newTr, _, err := tr.recreateFromDb(rootHash, tr.Database(), tr.trieStorage)
+	if err != nil {
+		log.Warn("trie recreate error:", "error", err, "root", hex.EncodeToString(rootHash))
```
sasurobert, line 321: return nil

data/trie/patriciaMerkleTrie.go
```diff
@@ -484,13 +507,20 @@ func (tr *patriciaMerkleTrie) ExitSnapshotMode() {
...
-func getDbThatContainsHash(trieStorage data.StorageManager, rootHash []byte) data.DBWriteCacher {
+func getDbThatContainsHash(trieStorage data.StorageManager, rootHash []byte) data.SnapshotDbHandler {
 	db := trieStorage.GetSnapshotThatContainsHash(rootHash)
 	if db != nil {
 		return db
 	}
 
-	return trieStorage.GetDbThatContainsHash(rootHash)
+	_, err := trieStorage.Database().Get(rootHash)
+	if err == nil {
```
sasurobert, line 517: if err != nil {
return err
}

return &snapshotDB ...

data/trie/snapshotDb.go
```diff
@@ -0,0 +1,65 @@
...
+	path            string
+	mutex           sync.RWMutex
+}
+
+// DecreaseNumReferences decreases the num references counter
+func (s *snapshotDb) DecreaseNumReferences() {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	if s.numReferences != 0 {
```
sasurobert, line 22: > 0

data/trie/snapshotDb.go
```diff
@@ -0,0 +1,65 @@
...
+
+	s.path = path
+}
+
+// IsInUse returns true if the numReferences counter is greater than 0
+func (s *snapshotDb) IsInUse() bool {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	if s.numReferences != 0 {
```
sasurobert, line 60: > 0

data/trie/patriciaMerkleTrie.go
SebastianMarian, line 321: You can add return nil here, even if it works now, but if the code will expand below this line there could be a wrong return afterward.

data/trie/patriciaMerkleTrie.go
SebastianMarian: if err != nil { return nil }

data/trie/snapshotDb.go
```diff
@@ -0,0 +1,65 @@
...
+func (s *snapshotDb) SetPath(path string) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	s.path = path
+}
+
+// IsInUse returns true if the numReferences counter is greater than 0
+func (s *snapshotDb) IsInUse() bool {
+	s.mutex.Lock()
```
SebastianMarian, line 57: RLock/RUnlock

data/trie/snapshotDb.go
SebastianMarian, line 60: return s.numReferences != 0 instead lines 60-64

data/trie/trieStorageManager.go
```diff
@@ -435,16 +423,29 @@ func (tsm *trieStorageManager) getSnapshotDb(newDb bool) data.DBWriteCacher {
 func (tsm *trieStorageManager) removeSnapshot() {
 	dbUniqueId := strconv.Itoa(tsm.snapshotId - len(tsm.snapshots))
```
SebastianMarian, line 424: Could be called with 0 snapshots? If yes, I think you should check if len(tsm.snapshots) > 0, and if not, you can return in first line, otherwise I think it could panic in line 427.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Done.

data/trie/snapshotDb.go
BeniaminDrasovean: Done.

data/trie/snapshotDb.go
BeniaminDrasovean: Done.

data/trie/patriciaMerkleTrie.go
BeniaminDrasovean: Done.

data/trie/snapshotDb.go
BeniaminDrasovean: Done.

data/trie/snapshotDb.go
BeniaminDrasovean: Done.

data/trie/trieStorageManager.go
BeniaminDrasovean: As the code is written now, this can not be called with 0 snapshots. Added the check anyway, just to be safe in the future if the code changes.

data/trie/trieStorageManager.go
```diff
@@ -142,6 +143,7 @@ func getSnapshotsAndSnapshotId(snapshotDbCfg config.DBConfig) ([]data.SnapshotDb
 			DBWriteCacher: db,
 		}
 
+		log.Debug("restored snapshot", "snapshot ID", snapshotName)
```
SebastianMarian, line 146: log.Trace?

BeniaminDrasovean: Not needed. This will print just a few lines, and just when the node starts/restarts.

PR #1954, opened 2020-06-15
bugFix/close-app-always

sasurobert:
Added time.Context to the closing method on main.go as there is a known deadlock issue on P2P library when closing the process.

https://github.com/libp2p/go-libp2p/issues/933

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


cmd/node/main.go
```diff
@@ -1350,9 +1368,11 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	err = networkComponents.NetMessenger.Close()
 	log.LogIfError(err)
 
-	handleAppClose(log, sig)
-
-	return nil
+	select {
```
iulianpascalau, line 1371: This doesn't work as expected as this point is never reached (the process will wait to execute L1368)
Please remove the new code, just replace L1368-L1369 with:
```GO
go func(){
    err = networkComponents.NetMessenger.Close()
    log.LogIfError(err)
}()

time.Sleep(time.Second) //or a const here
```

sasurobert: Ok. Will do this.

PR #1958, opened 2020-06-15
Memory footprint tests for tx pool

andreibancioiu:
1) Added `HealthService` to continuously monitor:
 - RAM usage - dumps heap profiles when memory is high (but only keeps the top-most N profiles on disk - for each restart)
 - Registered diagnosable components, such as transaction pools.

2) Added tests to estimate memory footprint of `TxCache` and `CrossTxCache`.

3) Added test to read cache configurations from `config.toml` and compute the "planned memory usage". 

4) Refactored creation of `caches`, so that they require a name now.

5) Added missing tests for `dataPoolFactory`.

6) Cleaned-up common configs used among several tests - moved them to `testscommon`.

7) Added a `deep diagnose` for the TxCache - to check eventual inconsistencies.

8) Improved `GetCounts()` for pools, so that size in MB is also included.

9) The Node can be started with `HealthService` using `--use-health-service` flag.

Example:

```
./node --use-log-view --use-health-service --log-logger-name -log-level="*:INFO,health:TRACE,txcache:DEBUG,txpool:DEBUG,dataretriever/shardeddata:DEBUG,storage/immunitycache:DEBUG"
```


health/healthService.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (h *healthService) setupCancellation() context.Context {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	h.cancelFunction = cancelFunc
+	return ctx
+}
+
+func (h *healthService) monitorContinuously(ctx context.Context) {
+	for i := 0; h.shouldContinueInfiniteLoop(ctx); i++ {
+		shouldMonitorMemory := i%h.config.IntervalVerifyMemoryInSeconds == 0
```
iulianpascalau, line 82: this function can be made more go idiomatic, and we can have other intervals (not necessarily multiples of 1 second):
```GO
    chanMonitorMemory := time.After(IntervalVerifyMemoryInSeconds * time.Second)
    chanDiagnoseComponents := time.After(IntervalDiagnoseComponentsInSeconds * time.Second)
    chanDiagnoseComponentsDeeply := time.After(IntervalDiagnoseComponentsDeeplyInSeconds * time.Second) 
    
    for{
        select{
        case <-chanMonitorMemory:
            h.monitorMemory()
            chanMonitorMemory := time.After(IntervalVerifyMemoryInSeconds * time.Second)
        case <-chanDiagnoseComponents:
            h.diagnoseComponents(false)
            chanDiagnoseComponents := time.After(IntervalDiagnoseComponentsInSeconds * time.Second)
        case <-chanDiagnoseComponentsDeeply:
            h.diagnoseComponents(true)
            chanDiagnoseComponentsDeeply := time.After(IntervalDiagnoseComponentsDeeplyInSeconds * time.Second)
        case <-ctx.Done():
            log.Info("healthService.monitorContinuously() ended")
            return    
        }
    }
```

health/healthService_test.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func TestHealthService_StartThenClose(t *testing.T) {
+	h := NewHealthService(config.HealthServiceConfig{}, "")
+	require.NotNil(t, h)
+
+	h.Start()
+	err := h.Close()
+	require.Nil(t, err)
+}
```
iulianpascalau, line 27: More unit tests?

health/interface.go
```diff
@@ -0,0 +1,13 @@
+package health
+
+// diagnosable is an internal interface, which external components can implement in order to be "diagnosed" by the health service
+type diagnosable interface {
+	Diagnose(deep bool)
```
iulianpascalau, line 5: check if nil?

health/memoryRecord.go
```diff
@@ -0,0 +1,69 @@
...
+	filename := fmt.Sprintf("mem__%s__%s.pprof", timestamp, inUse)
+	return filename
+}
+
+func (record *memoryRecord) getFilename() string {
+	return record.filename
+}
+
+func (record *memoryRecord) save() error {
+	filename := path.Join(record.filename)
```
iulianpascalau, line 46: so, for each save call we open, write an close the file. This is the safest way to persist the data when the process closing is imminent but is slow. Another way to do this is to open once, keep the pointer as long as you need it and call the runtime.SetFinalizer function to close the file if the process exits with OOM. Save will just write in the file. (this is the pattern used in the logger subsytem)

health/memoryRecord.go
```diff
@@ -0,0 +1,69 @@
...
+	"runtime/pprof"
+	"strings"
+	"time"
+
+	"github.com/ElrondNetwork/elrond-go/core"
+)
+
+var _ record = (*memoryRecord)(nil)
+
+type memoryRecord struct {
```
iulianpascalau, line 17: the name is a bit misleading. "Memory" took me thinking that the record would have been kept only in memory. Surprisingly I have found out that it was persisted in a file.

health/healthService_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	logger "github.com/ElrondNetwork/elrond-go-logger"
+	"github.com/ElrondNetwork/elrond-go/config"
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	_ = logger.SetLogLevel("health:TRACE")
```
AdoAdoAdo, line 12: do we need the health module trace logs in tests?

health/records.go
```diff
@@ -0,0 +1,76 @@
+package health
+
+import (
```
iulianpascalau, line 3: Unit tests for memoryRecord.go and records.go?

node/nodeMemoryConfig_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/ElrondNetwork/elrond-go/config"
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/pelletier/go-toml"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMemoryConfig(t *testing.T) {
+	nodeConfig := config.Config{}
+
+	tomlString, err := ioutil.ReadFile("../cmd/node/config/config.toml")
```
iulianpascalau, line 16: loading the real config for this unit test... :thinking: 
this pattern can be used as an early warning for a misconfigured config.toml file but will make tests depend on a physical file. Can be left as it is, will think about this

storage/monitoring.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/atomic"
+)
+
+var log = logger.GetOrCreate("storage")
+
+var cummulatedSizeInBytes atomic.Counter
+
+// Question for review: keep this or remove it (helps us to compute planned memory at runtime)?
+func MonitorNewCache(tag string, sizeInBytes uint64) {
```
iulianpascalau, line 14: :+1: It is useful (at least for the time being). Might move the functionality directly in the NewCache function.

health/records.go
```diff
@@ -0,0 +1,76 @@
...
+		items:    list.New(),
+	}
+}
+
+func (records *records) addRecord(incomingRecord record) {
+	if !records.makeRoomForRecord(incomingRecord) {
+		return
+	}
+
+	insertionPlace := records.findInsertionPlace(incomingRecord)
```
AdoAdoAdo, line 25: is this working properly with concurrent addRecord? I think you would need a mutex locked before finding the place of insertion and unlocked after inserting.

edit: Maybe a mutex lock before making room for record

storage/monitoring.go
```diff
@@ -0,0 +1,17 @@
...
+	logger "github.com/ElrondNetwork/elrond-go-logger"
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/atomic"
+)
+
+var log = logger.GetOrCreate("storage")
+
+var cummulatedSizeInBytes atomic.Counter
+
+// Question for review: keep this or remove it (helps us to compute planned memory at runtime)?
```
AdoAdoAdo, line 13: I didn't quite get how this should work if we keep it, but I only checked the code here in this PR.

cmd/node/main.go
```diff
@@ -1127,12 +1143,7 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	log.Trace("creating time cache for requested items components")
 	requestedItemsHandler := timecache.NewTimeCache(time.Duration(uint64(time.Millisecond) * genesisNodesConfig.RoundDuration))
 
-	whiteListCache, err := storageUnit.NewCache(
-		storageUnit.CacheType(generalConfig.WhiteListPool.Type),
-		generalConfig.WhiteListPool.Capacity,
-		generalConfig.WhiteListPool.Shards,
-		generalConfig.WhiteListPool.SizeInBytes,
-	)
+	whiteListCache, err := storageUnit.NewCache(storageFactory.GetCacherFromConfig(generalConfig.WhiteListPool))
```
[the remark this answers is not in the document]
andreibancioiu: Good point! I've added a JIRA task for this: https://elrondnetwork.atlassian.net/browse/EN-6819

cmd/node/config/config.toml
```diff
@@ -530,6 +561,14 @@
         CacheSize = 10000
         IntervalAutoPrintInSeconds = 20
 
+[Health]
+    IntervalVerifyMemoryInSeconds = 1
+    IntervalDiagnoseComponentsInSeconds = 10
+    IntervalDiagnoseComponentsDeeplyInSeconds = 30
+    MemoryToCreateProfiles = 1073741824 # 1GB
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

cmd/node/main.go
```diff
@@ -1361,12 +1372,16 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 
 func closeAllComponents(
 	log logger.Logger,
+	healthService io.Closer,
 	dataComponents *mainFactory.DataComponents,
 	triesComponents *mainFactory.TriesComponents,
 	networkComponents *mainFactory.NetworkComponents,
 ) {
+	err := healthService.Close()
```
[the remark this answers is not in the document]
andreibancioiu: Added.

health/healthService.go
andreibancioiu: Changed.

health/interface.go
andreibancioiu: Golang strikers again, so yes, I added the check :)

```
package main

import (
	"fmt"
)

type fooInterface interface {
	Foo()
}

type C struct {
}

func (c *C) Foo() {
}

func main() {
	c := (*C)(nil)
	do(c)
}

func do(o interface{}) {
	casted, ok := o.(fooInterface)
	fmt.Println(casted, ok) // <nil> true
}
```
:exploding_head:

health/memoryRecord.go
andreibancioiu: Renamed to `memoryUsageRecord`.

storage/monitoring.go
andreibancioiu: Moved.

core/counting/interface.go
```diff
@@ -1,14 +1,14 @@
...
-}
-
 // Counts is an interface to interact with counts
 type Counts interface {
 	GetTotal() int64
 	String() string
 	IsInterfaceNil() bool
 }
+
+// Counts is an interface to interact with counts
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

core/counting/shardedCountsWithSize.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/ElrondNetwork/elrond-go/core"
+)
+
+var _ CountsWithSize = (*ShardedCountsWithSize)(nil)
+
+// ShardedCounts keeps counts (with sizes) for a sharded data structure
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

cmd/node/config/config.toml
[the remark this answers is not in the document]
andreibancioiu: Fixed.

health/healthService_test.go
andreibancioiu: Not needed, removed.

storage/monitoring.go
andreibancioiu: Was moved to `storageUnit.NewCache()` while working on this fix-after-review.

core/counting/shardedCountsWithSize.go
```diff
@@ -0,0 +1,82 @@
...
+// This implementation is NOT concurrently safe
+type ShardedCountsWithSize struct {
+	byShard map[string]shardedCountsWithSizeItem
```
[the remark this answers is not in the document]
andreibancioiu: Was implemented to be "not concurrently safe" first, but fixed now to be concurrent safe.

core/counting/shardedCountsWithSize.go
```diff
@@ -0,0 +1,82 @@
...
+// NewShardedCountsWithSize creates a new ShardedCountsWithSize
+func NewShardedCountsWithSize() *ShardedCountsWithSize {
+	return &ShardedCountsWithSize{
+		byShard: make(map[string]shardedCountsWithSizeItem),
+	}
+}
+
+// PutCounts registers counts for a shard
+func (counts *ShardedCountsWithSize) PutCounts(shardName string, counter int64, sizeInBytes int64) {
+	counts.byShard[shardName] = shardedCountsWithSizeItem{counter: counter, sizeInBytes: sizeInBytes}
```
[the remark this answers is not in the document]
andreibancioiu: Was implemented to be "not concurrently safe" first, but fixed now to be concurrent safe.

health/memoryRecord.go
andreibancioiu: Nice idea. 

In the case of `memoryUsageRecord` we open the file, write the heap profile, close it, then never open it again. And we write many such files (one every second if when memory is critical). Does the recommendation still apply in this case?

health/records.go
andreibancioiu: Added mutex.

health/records.go
andreibancioiu: Added.

health/healthService_test.go
andreibancioiu: Added. I've sketched one way to mock the time, if it looks good perhaps we can use it in some of our long-running integration tests.

storage/txcache/monitoring.go
```diff
@@ -139,14 +149,101 @@ func (cache *TxCache) diagnose() {
...
+	senders := cache.txListBySender.getSnapshotAscending()
+	if len(senders) == 0 {
+		return
+	}
+
+	var builder strings.Builder
+	builder.WriteString("\n[#index (score)] address [nonce known / nonce vs lowestTxNonce] txs = numTxs, !numFailedSelections\n")
+
+	i := 0
+	for _, sender := range senders {
```
AdoAdoAdo, line 229: for i, sender ?

core/counting/concurrentShardedCountsWithSize.go
```diff
@@ -0,0 +1,94 @@
...
+	defer counts.mutex.RUnlock()
+
+	total := int64(0)
+	for _, item := range counts.byShard {
+		total += item.sizeInBytes
+	}
+	return total
+}
+
+// String returns a string representation of the counts
```
iulianpascalau, line 63: no need for this comment :)

health/healthService_test.go
iulianpascalau: This is the way it can be tested: mock everything that involves timing and, in more general, everything is I/O :+1:

health/interface.go
iulianpascalau: :)))

core/counting/concurrentShardedCountsWithSize.go
andreibancioiu: Removed :)

storage/txcache/monitoring.go
andreibancioiu: Fixed.

PR #1979, opened 2020-06-18
Add feedback for peer honesty in consensus

SebastianMarian:
* Added a consensus rating mechanism to handle the situation when some peers would flood the consensus topic with invalid messages
* Fixed an uncovered situation in which a blacklist should be applied for both, originator and the connected peer, if an invalid message would be received on consensus topic


consensus/interface.go
```diff
@@ -97,3 +97,11 @@ type HeadersPoolSubscriber interface {
 	RegisterHandler(handler func(headerHandler data.HeaderHandler, headerHash []byte))
 	IsInterfaceNil() bool
 }
+
+// ConsensusRating defines the behaivour of a component able to handle/monitor the rating of nodes which are
+// participating in consensus
+type ConsensusRating interface {
```
iulianpascalau, line 103: PeerQualityHandler?

consensus/spos/common.go
```diff
@@ -5,8 +5,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/sharding"
 )
 
-// GetConsensusTopicIDFromShardCoordinator will construct and return the topic ID based on shard coordinator
-func GetConsensusTopicIDFromShardCoordinator(shardCoord sharding.Coordinator) string {
-	return core.ConsensusTopic +
-		shardCoord.CommunicationIdentifier(shardCoord.SelfId())
+// GetConsensusTopicID will construct and return the topic ID based on shard coordinator
+func GetConsensusTopicID(shardCoordinator sharding.Coordinator) string {
```
iulianpascalau, line 9: :+1:

consensus/spos/constants.go
```diff
@@ -2,3 +2,19 @@ package spos
 
 // MaxThresholdPercent specifies the max allocated time percent for doing Job as a percentage of the total time of one round
 const MaxThresholdPercent = 95
+
+// ProposerRatingIncreaseFactor specifies the factor with which the rating of a proposer should be increased if it
+// proposed a block or sent final info, in its correct allocated slot/time-frame/round
+const ProposerRatingIncreaseFactor = 2.0
```
iulianpascalau, line 8: ProposerPeerQualityIncreaseFactor?   rating->PeerQuality as to avoid confusions with peer rating mechanism

SebastianMarian: ProposerHonestyIncreaseFactor?

consensus/interface.go
SebastianMarian: PeerHonestyHandler ?

sasurobert: it should not contain the name Rating at all. In order to not cause misunderstanding.

consensus/mock/consensusRatingStub.go
```diff
@@ -0,0 +1,28 @@
+package mock
+
+// ConsensusRatingStub -
+type ConsensusRatingStub struct {
```
sasurobert, line 4: change the name rating to something else - this is a local view of the node

consensus/spos/bls/subroundBlock.go
```diff
@@ -314,15 +314,27 @@ func (sr *subroundBlock) receivedBlockBodyAndHeader(cnsDta *consensus.Message) b
 		return false
 	}
 
-	if sr.IsBlockBodyAlreadyReceived() {
+	if !sr.IsNodeLeaderInCurrentRound(node) { // is NOT this node leader in current round?
+		sr.ConsensusRating().Decrease(
```
sasurobert, line 318: what will happen in case of forks ? decrease for the other leader which is in the other fork - even if my current node is in the wrong one ?

consensus/spos/bls/subroundBlock.go
```diff
@@ -353,11 +365,23 @@ func (sr *subroundBlock) receivedBlockBodyAndHeader(cnsDta *consensus.Message) b
 func (sr *subroundBlock) receivedBlockBody(cnsDta *consensus.Message) bool {
 	node := string(cnsDta.PubKey)
 
-	if sr.IsBlockBodyAlreadyReceived() {
+	if !sr.IsNodeLeaderInCurrentRound(node) { // is NOT this node leader in current round?
```
sasurobert, line 368: duplicated code with the one above

consensus/spos/bls/subroundBlock.go
```diff
@@ -388,11 +412,23 @@ func (sr *subroundBlock) receivedBlockHeader(cnsDta *consensus.Message) bool {
 		return false
 	}
 
-	if sr.IsHeaderAlreadyReceived() {
+	if !sr.IsNodeLeaderInCurrentRound(node) { // is NOT this node leader in current round?
```
sasurobert, line 415: duplicated code

consensus/spos/worker.go
```diff
@@ -316,22 +317,27 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedP
...
 
-	topic := GetConsensusTopicIDFromShardCoordinator(wrk.shardCoordinator)
+	topic := GetConsensusTopicID(wrk.shardCoordinator)
 	err = wrk.antifloodHandler.CanProcessMessagesOnTopic(message.Peer(), topic, 1, uint64(len(message.Data())), message.SeqNo())
 	if err != nil {
 		return err
 	}
 
+	defer func() {
+		shouldBlacklistPeer := err != nil && !errors.Is(err, ErrMessageForPastRound)
```
sasurobert, line 327: I think it would work without the first check err != nil - only the second one should be enough

consensus/mock/consensusRatingStub.go
SebastianMarian: It would be PeerHonestyStub

consensus/interface.go
SebastianMarian: It would be PeerHonestyHandler

consensus/spos/bls/subroundBlock.go
SebastianMarian: Yes, but this is only in the local node view. It is no problem if that node will send afterward signatures and so on, it honesty will increase in the eyes of this node. Anyway we can set a generous threshold just to be sure that one node will not be banned too easy/soon. The same problem will happen also for nodes which come later in network and will do sync. They will don't now about these kind of local states like: how many times shards were stuck, if there was flooding in network, how many peers have been blacklisted and so on. Anyway I have added a round as a parameter in the call of method Increase and Decrease, just to have a possibility if we want to be more accurate, to inject this interface in BlockProcessor and on rollbacks we can remove the Increase/Decrease actions made in that rollback header round if there exist one of them.

consensus/spos/worker.go
SebastianMarian: Its not working. For err == nil the method will return false and the shouldBlacklistPeer will be set wrongly to true. Added an unit test to check this.

consensus/spos/bls/subroundBlock.go
SebastianMarian: refactored

consensus/spos/bls/subroundBlock.go
SebastianMarian: refactored

consensus/spos/bls/subroundBlock.go
```diff
@@ -589,3 +559,21 @@ func (sr *subroundBlock) getRoundInLastCommittedBlock() int64 {
 
 	return roundInLastCommittedBlock
 }
+
+func (sr *subroundBlock) updateLeaderPeerHonesty(node string) {
```
iulianpascalau, line 563: Duplicate code with subroundEndRound. Suggestion to move this in Subround struct.

consensus/spos/bls/subroundSignature.go
```diff
@@ -279,3 +269,21 @@ func (sr *subroundSignature) remainingTime() time.Duration {
 
 	return remainigTime
 }
+
+func (sr *subroundSignature) updateValidatorPeerHonesty(node string) {
```
iulianpascalau, line 273: This can also be moved in Subround struct

PR #1980, opened 2020-06-19
changed the message displayed when a key is not available in termui

bogdan-rosianu:
changed the message displayed when a key is missing in TermUI from `[key not found]` to `N/A`. Before, that message wasn't displayed correctly and could lead to references to the validator key.

### DIFF
--- a/statusHandler/presenter/common.go
+++ b/statusHandler/presenter/common.go
@@ -6,8 +6,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/core"
 )
 
-const invalidKey = "N/A"
-const invalidType = "[not a string]"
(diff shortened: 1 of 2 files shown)


statusHandler/presenter/common.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/core"
 )
 
-const invalidKey = "[invalid key]"
+const invalidKey = "N/A"
 const invalidType = "[not a string]"
```
iulianpascalau, line 10: can we replace all default strings to N/A or something like this?

bogdan-rosianu: done

PR #1984, opened 2020-06-21
Add consensus originator check

iulianpascalau:
- added peer id in consensus message as to not allow impersonating. Anyone caught sending a consensus message that won't contain its originator peer id will be subject to peer banning.


heartbeat/data/proto/heartbeat.proto
```diff
@@ -31,7 +31,7 @@ message HeartbeatDTO {
 	string  PeerType                     = 11;
 	string  Identity                     = 12;
 	bool    IsValidator                  = 13;
-	int64	LastUptimeDowntime           = 14;
+	int64	  LastUptimeDowntime           = 14;
```
andreibancioiu, line 34: Indentation is not correct.

consensus/spos/consensusMessageValidator.go
```diff
@@ -83,6 +84,12 @@ func (wrk *Worker) checkConsensusMessageValidity(cnsMsg *consensus.Message) erro
 			err.Error())
 	}
 
+	cnsMsgOriginator := core.PeerID(cnsMsg.OriginatorPid)
+	if cnsMsgOriginator != originator {
```
andreibancioiu, line 88: :+1:

heartbeat/data/proto/heartbeat.proto
iulianpascalau: :( will fix that. Gofmt does not run automatically for .proto files

PR #1989, opened 2020-06-22
Compute shard id sp config

sasurobert:


cmd/node/config/config.toml
```diff
@@ -9,7 +9,7 @@
 
    # StartInEpochEnabled represents that the fast bootstrap mechanism from the network is enabled if data is not
    # available in local disk
-   StartInEpochEnabled = true
+   StartInEpochEnabled = false
```
SebastianMarian, line 12: Do we want this change?

cmd/node/config/config.toml
```diff
@@ -426,8 +426,8 @@
 
 # TODO: change this config to real numbers before merging to development
 [EpochStartConfig]
-    MinRoundsBetweenEpochs = 200
-    RoundsPerEpoch         = 200
+    MinRoundsBetweenEpochs = 1200
```
SebastianMarian, line 429: Do we want this change?

cmd/node/config/config.toml
```diff
@@ -426,8 +426,8 @@
...
+    RoundsPerEpoch         = 1200
```
SebastianMarian, line 430: Do we want this change?

PR #1998, opened 2020-06-23
En 6597 state code pruning

BeniaminDrasovean:
When SC code is saved to the trie, now also the number of references to the same code is saved in the same db. If the number of references to a certain code reaches 0, the trie node that contains the code is marked for eviction, and it will be pruned from the db.


data/state/journalEntries_test.go
```diff
@@ -10,61 +10,77 @@ import (
...
 
-func TestJournalEntryCode_RevertUpdatesTheTrie(t *testing.T) {
+func TestJournalEntryCode_OldHashIsNilAndNewHashIsNotNil(t *testing.T) {
 	t.Parallel()
 
+	codeEntry := &state.CodeEntry{
+		Code:          []byte("newCode"),
+		NumReferences: 1,
+	}
+	marsh := &mock.MarshalizerMock{}
```
SebastianMarian, line 71: marsh sounds strange :)

data/state/journalEntries.go
```diff
@@ -1,34 +1,120 @@
...
+		return nil, ErrNilCodeForEvictionMap
+	}
+	if newCode == nil {
+		return nil, ErrNilNewCodeMap
 	}
 
 	return &journalEntryCode{
-		codeHash: codeHash,
-		updater:  updater,
+		oldCodeEntry:    oldCodeEntry,
```
SebastianMarian, line 45: oldCodeEntry could be nil? If not you should check it above.

data/state/journalEntries.go
```diff
@@ -1,34 +1,120 @@
...
+
+func (jea *journalEntryCode) revertOldCodeEntry() error {
+	if len(jea.oldCodeHash) == 0 {
+		return nil
+	}
+
+	if jea.oldCodeEntry.NumReferences == 1 {
+		delete(jea.codeForEviction, string(jea.oldCodeHash))
+	}
+
```
SebastianMarian, line 82: jea.oldCodeEntry.NumReferences-- ?

data/state/journalEntries.go
```diff
@@ -1,34 +1,120 @@
...
+	return nil, nil
+}
+
+func (jea *journalEntryCode) revertOldCodeEntry() error {
+	if len(jea.oldCodeHash) == 0 {
+		return nil
+	}
+
+	if jea.oldCodeEntry.NumReferences == 1 {
```
SebastianMarian, line 79: <= ?

data/state/accountsDB_test.go
```diff
@@ -940,3 +945,294 @@ func TestAccountsDB_GetAllLeaves(t *testing.T) {
 	assert.True(t, recreateCalled)
 	assert.True(t, getAllLeavesCalled)
 }
+
+func getTestAccountsDbAndTrie(marsh marshal.Marshalizer, hsh hashing.Hasher) (*state.AccountsDB, data.Trie) {
```
SebastianMarian, line 949: marsh sounds strange :)

data/state/journalEntries.go
sasurobert: +1

integrationTests/state/stateTrie/stateTrie_test.go
```diff
@@ -1665,3 +1666,146 @@ func testNodeStateCheckpointSnapshotAndPruning(
 		assert.NotNil(t, err)
 	}
 }
+
+func TestContinuouslyAccountCodeChanges(t *testing.T) {
```
sasurobert, line 1670: function too big. make a few smaller ones.

data/state/accountsDB.go
```diff
@@ -20,6 +20,9 @@ type AccountsDB struct {
 	marshalizer    marshal.Marshalizer
 	accountFactory AccountFactory
 
+	codeForEviction map[string]struct{}
```
sasurobert, line 23: codeForEviction cannot be used with current state of smart contract process, as delete function, delete smart contract is not called at all.

data/state/accountsDB.go
```diff
@@ -104,46 +109,150 @@ func (adb *AccountsDB) SaveAccount(account AccountHandler) error {
...
+		return err
+	}
+
+	newCode := accountHandler.GetCode()
+	var newCodeHash []byte
+	if len(newCode) != 0 {
+		newCodeHash = adb.hasher.Compute(string(newCode))
+	}
+
+	if bytes.Equal(oldCodeHash, newCodeHash) {
```
sasurobert, line 138: saveCode should be called only there is a new code. if there is something in old code - delete account / code should be called. If a code is delete the whole account is deleted.

data/state/accountsDB.go
```diff
@@ -90,7 +95,7 @@ func (adb *AccountsDB) SaveAccount(account AccountHandler) error {
 
 	baseAcc, ok := account.(baseAccountHandler)
 	if ok {
-		err = adb.saveCode(baseAcc)
+		err = adb.saveCode(baseAcc, oldAccount)
```
sasurobert, line 98: make a separate function to saveCode and saveDataTrie where you can handle all the cases.

sasurobert: make type assertion for oldAccount as well. for save code you need baseAccountHandler only.

data/state/journalEntries_test.go
BeniaminDrasovean: Renamed.

data/state/journalEntries.go
BeniaminDrasovean: oldCodeEntry could be nil in the following case: If a newCode is set to an account that had no code before, the oldCodeEntry should be nil, in case that a revert occurs.

BeniaminDrasovean: oldCodeEntry could be nil in the following case: If a newCode is set to an account that had no code before, the oldCodeEntry should be nil, in case that a revert occurs.

data/state/journalEntries.go
BeniaminDrasovean: No, because oldCodeEntry already has the numReferences set as before the changes. You can see in the accountDB that updateOldCodeEntry() returns unmodifiedOldCodeEntry that is used when creating a new journal entry.

data/state/journalEntries.go
BeniaminDrasovean: Done.

data/state/accountsDB_test.go
BeniaminDrasovean: Renamed.

data/state/accountsDB.go
```diff
@@ -104,46 +109,150 @@ func (adb *AccountsDB) SaveAccount(account AccountHandler) error {
...
 	}
 
-	codeHash := adb.hasher.Compute(string(code))
-	currentCodeHash := accountHandler.GetCodeHash()
-	noUpdateNeeded := bytes.Equal(codeHash, currentCodeHash)
-	if noUpdateNeeded {
+	return oldAccount.GetCodeHash(), nil
+}
+
+func (adb *AccountsDB) saveCode(accountHandler baseAccountHandler, oldAcc AccountHandler) error {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

data/state/accountsDB.go
BeniaminDrasovean: If a code is updated and the old version is not used anymore, it needs to be evicted.

data/state/accountsDB.go
BeniaminDrasovean: Done.

data/state/accountsDB.go
BeniaminDrasovean: This also sets the codeHash for the account. If a code is updated, isn't the same account used? In this case, there is something in both oldCode and newCode.

integrationTests/state/stateTrie/stateTrie_test.go
BeniaminDrasovean: Done.

data/state/accountsDB.go
```diff
@@ -93,50 +93,46 @@ func (adb *AccountsDB) SaveAccount(account AccountHandler) error {
...
-	if err != nil {
-		return err
+
+	var oldCodeHash []byte
+	if !check.IfNil(oldAcc) {
+		oldCodeHash = oldAcc.GetCodeHash()
 	}
 
-	newCode := accountHandler.GetCode()
+	newCode := newAcc.GetCode()
```
SebastianMarian, line 128: newAcc is always not nil? If not, add a check above

data/state/journalEntries.go
SebastianMarian: ok

data/state/journalEntries.go
SebastianMarian: ok

data/state/journalEntries.go
SebastianMarian: ok

data/state/accountsDB.go
BeniaminDrasovean: newAccount is always not nil. It is checked on L72, when SaveAccount() is called.

PR #1999, opened 2020-06-23
Fix block size throttle

SebastianMarian:
* Optimized mutex Lock/Unlock in txsPoolsCleaner component
* Tweaked some parameters in createAndProcessCrossMiniBlocksDstMe (how many shard/meta headers should be included in one block)
* Reduced number of go routines from 2000 to 100
* Tweaked some parameters in blockSizeThrottle component to speed up adaptivity to the network conditions
* Fixed a situation when block max size, from blockSizeThrottle component,  is increased to slow after a network problem

### DIFF
--- a/process/block/poolsCleaner/txsPoolsCleaner.go
+++ b/process/block/poolsCleaner/txsPoolsCleaner.go
@@ -178,10 +178,11 @@ func (tpc *txsPoolsCleaner) processReceivedTx(
 	receiverShardID uint32,
 	txType int8,
 ) {
-	tpc.mutMapTxsRounds.Lock()
-	defer tpc.mutMapTxsRounds.Unlock()
+	tpc.mutMapTxsRounds.RLock()
+	_, ok := tpc.mapTxsRounds[string(key)]
+	tpc.mutMapTxsRounds.RUnlock()
 
-	if _, ok := tpc.mapTxsRounds[string(key)]; !ok {
+	if !ok {
 		transactionPool := tpc.getTransactionPool(txType)
 		if transactionPool == nil {
 			return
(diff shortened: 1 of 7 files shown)


process/factory/interceptorscontainer/baseInterceptorsContainerFactory.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/sharding"
 )
 
-const numGoRoutines = 2000
+const numGoRoutines = 100
```
andreibancioiu, line 21: Dramatic decrease :)

process/block/poolsCleaner/txsPoolsCleaner.go
```diff
@@ -178,10 +178,11 @@ func (tpc *txsPoolsCleaner) processReceivedTx(
 	receiverShardID uint32,
 	txType int8,
 ) {
-	tpc.mutMapTxsRounds.Lock()
-	defer tpc.mutMapTxsRounds.Unlock()
+	tpc.mutMapTxsRounds.RLock()
```
andreibancioiu, line 181: Now I see that the cleaner removes each tx individually, it doesn't call the bulk operations (e.g. `RemoveSetOfDataFromPool`) but that should be fine nonetheless.

SebastianMarian: Yes, because we don't have such a method implemented, but anyway the cleanTxsPoolsIfNeeded method is ran only once per minute.

process/factory/interceptorscontainer/baseInterceptorsContainerFactory.go
SebastianMarian: Yep,, this was the suggested value by @iulianpascalau

PR #2006, opened 2020-06-24
added denomination in network config

bogdan-rosianu:
added `denomination` metric in the API `/network/config` route. 
plus small refactoring in economics file + wrapper added in the TermUI common


statusHandler/presenter/common.go
```diff
@@ -94,7 +84,17 @@ func (psh *PresenterStatusHandler) computeRoundsPerHourAccordingToHitRate() floa
 
 func (psh *PresenterStatusHandler) computeRewardsInErd() *big.Float {
 	rewardsValue := psh.getBigIntFromStringMetric(core.MetricRewardsValue)
-	denominationCoefficient := psh.getBigFloatFromStringMetric(core.MetricDenominationCoefficient)
+	denomination := psh.getFromCacheAsUint64(core.MetricDenomination)
+	denominationCoefficientStr := "0."
```
SebastianMarian, line 88: I don't think that this works as expected. If denomination would be 0, then the result of denominationCoefficientStr would be 0.1 instead of 1. If denomination would be 1 then the result of denominationCoefficientStr would be 0.01 instead of 0.1. So every time it would be wrong with exactly one order of magnitude. I suggest to replace the code between lines 88-96 with the below proposed which is much simplified and I think it works correct.

```
	denominationCoefficientFloat := 1.0
	if denomination > 0 {
		denominationCoefficientFloat /= float64(denomination)
	}
	
	denominationCoefficient := big.NewFloat(denominationCoefficientFloat)
```

bogdan-rosianu: you're right. fixed

cmd/node/config/economics.toml
```diff
@@ -3,13 +3,13 @@
     TotalSupply = "20000000000000000000000000000" #20BILERD
     MinimumInflation = 0.0
     MaximumInflation = 0.075 #fraction of value 1 - 7.5%
+    Denomination = 18
```
miiu96, line 6: you can add a comment to say what this parameter is

bogdan-rosianu: done

PR #2010, opened 2020-06-24
removed process start function when shuffled out

bogdan-rosianu:
removed `handleAppClose` and `newStartInEpoch` functions from main as they were not working as expected. When running a node using `elrond-go-scripts-v2` it caused additional log files due to multiple processes creation.

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -8,7 +8,6 @@ import (
 	"math"
 	"math/big"
 	"os"
-	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"runtime"
@@ -1349,8 +1348,6 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 		log.Warn("force closing the node", "error", "closeAllComponents did not finished on time")
 	}
 
-	handleAppClose(log, sig)
-
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


cmd/node/main.go
```diff
@@ -1383,47 +1380,6 @@ func closeAllComponents(
 	chanCloseComponents <- struct{}{}
 }
 
-func handleAppClose(log logger.Logger, endProcessArgument endProcess.ArgEndProcess) {
-	log.Debug("closing node")
-
-	switch endProcessArgument.Reason {
-	case core.ShuffledOut:
```
sasurobert, line 1390: leave closing node final log

bogdan-rosianu: done. also added `description` to the internal signal close log

cmd/node/main.go
```diff
@@ -1346,7 +1345,7 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 	case <-sigs:
 		log.Info("terminating at user's signal...")
 	case sig = <-chanStopNodeProcess:
-		log.Info("terminating at internal stop signal", "reason", sig.Reason)
+		log.Info("terminating at internal stop signal", "reason", sig.Reason, "description", sig.Description)
```
iulianpascalau, line 1348: :+1:

PR #2016, opened 2020-06-25
bugFix/shuffle-out-epoch-1-restart

sasurobert:
edge case fix when reshuffling in early epoch 1, the node currently thinks it should start in epoch 0.

### DIFF
--- a/epochStart/bootstrap/process.go
+++ b/epochStart/bootstrap/process.go
@@ -340,9 +340,9 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
 	}
 
 	isStartInEpochZero := e.isStartInEpochZero()
-	isCurrentEpochIsSaved := e.computeIfCurrentEpochIsSaved()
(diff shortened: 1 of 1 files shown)


epochStart/bootstrap/process.go
```diff
@@ -339,7 +339,10 @@ func (e *epochStartBootstrap) Bootstrap() (Parameters, error) {
 		return Parameters{}, err
 	}
 
-	if e.isStartInEpochZero() || e.computeIfCurrentEpochIsSaved() {
+	isStartInEpochZero := e.isStartInEpochZero()
+	isCurrentEpochIsSaved := e.computeIfCurrentEpochIsSaved()
```
bogdan-rosianu, line 343: `isCurrentEpochSaved` *

SebastianMarian, line 343: *isCurrentEpochSaved

bogdan-rosianu: +1

PR #2021, opened 2020-06-25
Reject miniblocks which are not whitelisted

SebastianMarian:
* Implemented a mechanism which will add only whitelisted miniblocks received in pool
* Fixed an edge case in multidata interceptor related to a throttle mechanism

### DIFF
--- a/integrationTests/testProcessorNode.go
+++ b/integrationTests/testProcessorNode.go
@@ -1503,6 +1503,31 @@ func (tpn *TestProcessorNode) BroadcastBlock(body data.BodyHandler, header data.
 	_ = tpn.BroadcastMessenger.BroadcastTransactions(transactions)
 }
 
+// WhiteListBody will whitelist all miniblocks from the given body for all the given nodes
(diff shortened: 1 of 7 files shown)


process/interceptors/multiDataInterceptor.go
```diff
@@ -109,6 +109,7 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
 		message.SeqNo(),
 	)
 	if err != nil {
+		mdi.throttler.EndProcessing()
```
iulianpascalau, line 112: :+1:

integrationTests/testProcessorNode.go
```diff
@@ -1503,6 +1503,31 @@ func (tpn *TestProcessorNode) BroadcastBlock(body data.BodyHandler, header data.
 	_ = tpn.BroadcastMessenger.BroadcastTransactions(transactions)
 }
 
+// WhiteListBody will whitelist all miniblocks from the given body for all the given nodes
+func (tpn *TestProcessorNode) WhiteListBody(nodes []*TestProcessorNode, bodyHandler data.BodyHandler) {
```
sasurobert, line 1507: could have whitelisted from header

PR #2022, opened 2020-06-25
[RC] fix destination shard as observer

bogdan-rosianu:
refactored the way `DestinationShardAsObserver` works inside `epochBootstrapper` so if a shard is set it will "force" the node to start in the specified shard. Before this, the specified information could have been ignored. Now if a value different than `disabled` is set, the node will start in the shard specified in the `prefs.toml` file or using the `--destination-shard-as-observer` flag.

### DIFF
--- a/epochStart/bootstrap/fromLocalStorage.go
+++ b/epochStart/bootstrap/fromLocalStorage.go
@@ -20,18 +20,28 @@ func (e *epochStartBootstrap) initializeFromLocalStorage() {
 	if errNotCritical != nil {
 		e.baseData.storageExists = false
 		log.Debug("no epoch db found in storage", "error", errNotCritical.Error())
-	} else {
-		e.baseData.storageExists = true
-		e.baseData.lastEpoch = latestData.Epoch
-		e.baseData.shardId = latestData.ShardID
-		e.baseData.lastRound = latestData.LastRound
-		e.baseData.epochStartRound = latestData.EpochStartRound
-		log.Debug("got last data from storage",
-			"epoch", e.baseData.lastEpoch,
-			"last round", e.baseData.lastRound,
-			"last shard ID", e.baseData.shardId,
-			"epoch start Round", e.baseData.epochStartRound)
+		return
 	}
+
+	if latestData.ShardID != e.destinationShardAsObserver && e.destinationShardAsObserver != core.DefaultShardIDAsObserver {
+		e.baseData.storageExists = false
+		log.Debug("cannot use epoch db found in storage as a different shard ID than the desired one has been found",
+			"desired shard ID", e.destinationShardAsObserver,
+			"shard ID found in storage", latestData.ShardID)
+		return
+	}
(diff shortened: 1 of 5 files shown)


epochStart/bootstrap/process.go
```diff
@@ -593,6 +597,9 @@ func (e *epochStartBootstrap) processNodesConfig(pubKey []byte) error {
 	}
 
 	e.nodesConfig, e.baseData.shardId, err = e.nodesConfigHandler.NodesConfigFromMetaBlock(e.epochStartMeta, e.prevEpochStartMeta)
+	if e.destinationShardAsObserver != core.DefaultShardIDAsObserver {
+		e.baseData.shardId = e.destinationShardAsObserver
```
sasurobert, line 601: this does not sound good - if the node is validator it shouldn't rewrite it to desired observer state.

epochStart/bootstrap/process.go
```diff
@@ -250,9 +250,13 @@ func (e *epochStartBootstrap) prepareEpochZero() (Parameters, error) {
 		return Parameters{}, err
 	}
 
+	shardIDToReturn := e.genesisShardCoordinator.SelfId()
+	if shardIDToReturn != e.destinationShardAsObserver && e.destinationShardAsObserver != core.DefaultShardIDAsObserver {
```
sasurobert, line 254: same conditions duplicated in several places - make a function for it. This part should count only if node is not validator.

epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -129,6 +139,11 @@ func (e *epochStartBootstrap) prepareEpochFromStorage() (Parameters, error) {
 		}
 	}
 
+	shardIDToReturn := e.shardCoordinator.SelfId()
+	if e.destinationShardAsObserver != core.DefaultShardIDAsObserver && e.destinationShardAsObserver != shardIDToReturn {
+		shardIDToReturn = e.destinationShardAsObserver
```
sasurobert, line 144: this should be taken into consideration if node is not a validator

epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -20,18 +20,28 @@ func (e *epochStartBootstrap) initializeFromLocalStorage() {
...
-		log.Debug("got last data from storage",
-			"epoch", e.baseData.lastEpoch,
-			"last round", e.baseData.lastRound,
-			"last shard ID", e.baseData.shardId,
-			"epoch start Round", e.baseData.epochStartRound)
+		return
 	}
+
+	if latestData.ShardID != e.destinationShardAsObserver && e.destinationShardAsObserver != core.DefaultShardIDAsObserver {
+		e.baseData.storageExists = false
```
sasurobert, line 27: this should be taken into consideration only if node is not a validator.

sasurobert:
it was reopened

PR #2024, opened 2020-06-26
fix destination shard as observer (dev branch)

bogdan-rosianu:
refactored the way DestinationShardAsObserver works inside epochBootstrapper so if a shard is set it will "force" the node to start in the specified shard. Before this, the specified information could have been ignored. Now if a value different than disabled is set, the node will start in the shard specified in the prefs.toml file or using the --destination-shard-as-observer flag.


epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -20,18 +20,26 @@ func (e *epochStartBootstrap) initializeFromLocalStorage() {
...
-		e.baseData.lastRound = latestData.LastRound
-		e.baseData.epochStartRound = latestData.EpochStartRound
-		log.Debug("got last data from storage",
-			"epoch", e.baseData.lastEpoch,
-			"last round", e.baseData.lastRound,
-			"last shard ID", e.baseData.shardId,
-			"epoch start Round", e.baseData.epochStartRound)
+		return
+	}
+	if latestData.ShardID != e.destinationShardAsObserver && e.destinationShardAsObserver != core.DisabledShardIDAsObserver {
```
raduchis, line 25: will this still work if you were a validator, not an observer in the previous epoch, and the destinationShardAsObserver does not match to the shard where the validator validated previously? Will the validator be need to restart a synchronization, not start from his DB?

epochStart/bootstrap/fromLocalStorage.go
```diff
@@ -138,9 +146,10 @@ func (e *epochStartBootstrap) prepareEpochFromStorage() (Parameters, error) {
 		}
 	}
 
+	shardIDToReturn := e.applyShardIDAsObserverIfNeeded(e.shardCoordinator.SelfId())
```
raduchis, line 149: Does an Observer node have anything to do with the shardCoordinator.SelfId? If it is an observer, he is not assigned to a shard by the shardCoordinator

cmd/node/main.go
```diff
@@ -1734,6 +1741,9 @@ func createNodesCoordinator(
 	if err != nil {
 		return nil, err
 	}
+	if shardIDAsObserver == core.DisabledShardIDAsObserver {
```
sasurobert, line 1744: move this if inside the processDestinationShardAsObserver function.

cmd/node/main.go
```diff
@@ -1693,6 +1694,12 @@ func createShardCoordinator(
 		log.Info("starting as observer node")
 
 		selfShardId, err = processDestinationShardAsObserver(prefsConfig)
+		if err != nil {
+			return nil, "", err
+		}
+		if selfShardId == core.DisabledShardIDAsObserver {
```
sasurobert, line 1700: move this if inside the processDestinationShardAsObserver function.

epochStart/bootstrap/fromLocalStorage.go
sasurobert: he is assigned to a shard - shardCoordinator says his shard.

epochStart/bootstrap/fromLocalStorage.go
sasurobert: This condition does not seem right. Save destinationShardAsObserver to the latestData - and return it here. That way you can make a good decision. StorageExists should be left to TRUE as the node will resolve if it is shuffled out in the function prepareEpochFromStorage

epochStart/bootstrap/process.go
```diff
@@ -944,6 +950,16 @@ func (e *epochStartBootstrap) setEpochStartMetrics() {
 	}
 }
 
+func (e *epochStartBootstrap) applyShardIDAsObserverIfNeeded(receivedShardID uint32) uint32 {
```
sasurobert, line 953: this is a nice way to resolve it :)

epochStart/bootstrap/fromLocalStorage.go
bogdan-rosianu: done, removed that part of code

cmd/node/main.go
bogdan-rosianu: I can't place that logic inside processDestinationShardAsObserver because in epoch bootstrapper I need that information, so don't replace `Disabled` with `0` which you cannot tell if it is default or set

cmd/node/main.go
bogdan-rosianu: same as above

cmd/node/main.go
```diff
@@ -1836,6 +1846,11 @@ func processDestinationShardAsObserver(prefsConfig config.PreferencesConfig) (ui
 	if len(destShard) == 0 {
 		return 0, errors.New("option DestinationShardAsObserver is not set in prefs.toml")
 	}
+
+	if destShard == notSetDestinationShardID {
+		return core.DisabledShardIDAsObserver, nil
```
sasurobert, line 1851: instead of this why not return 0 directly here ? after returning from this function you always change to 0 if the result was core.DisabledShardIDAsObserver

PR #2028, opened 2020-06-26
Integrate new arwen iun26

sasurobert:


integrationTests/multiShard/smartContract/scCallingSC_test.go
```diff
@@ -428,9 +428,9 @@ func TestSCCallingDNSUserNames(t *testing.T) {
 		t.Skip("this is not a short test")
 	}
 
-	numOfShards := 2
-	nodesPerShard := 2
-	numMetachainNodes := 2
+	numOfShards := 1
```
iulianpascalau, line 431: this is in multishard package, might revert the values

sasurobert: there is still one limitation - no breaking changes. patch will come later in the weekend.

iulianpascalau: ok

PR #2038, opened 2020-06-29
Refactored p2p blacklist prints

iulianpascalau:
- refactored p2p blacklist prints
- added max limit print in white list data verifier
- fixed immunize/whitelisting of peer changes as they are not hashes but plain peer changes
- fixed internal test scripts


process/interceptors/whiteListDataVerifier.go
```diff
@@ -43,6 +45,12 @@ func (w *whiteListDataVerifier) IsWhiteListed(interceptedData process.Intercepte
 // Add adds all the list to the cache
 func (w *whiteListDataVerifier) Add(keys [][]byte) {
 	for _, key := range keys {
+		if len(key) > maxElementSize {
```
sasurobert, line 48: Do not add elements which are higher than 32

iulianpascalau: you're right, will keep the warn though

PR #2039, opened 2020-06-29
En 6993 prune code on remove account

BeniaminDrasovean:
When an account is removed, also remove  the code (or decrease numReferences).
Removed unneeded code.
Change max trie snapshots to remember from config.
Use a real fork detector in integration tests.

### DIFF
--- a/data/state/accountsDB.go
+++ b/data/state/accountsDB.go
@@ -20,9 +20,6 @@ type AccountsDB struct {
 	marshalizer    marshal.Marshalizer
 	accountFactory AccountFactory
 
-	codeForEviction map[string]struct{}
-	newCode         map[string]struct{}
-
 	lastRootHash []byte
 	dataTries    TriesHolder
 	entries      []JournalEntry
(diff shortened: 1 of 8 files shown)


data/state/accountsDB.go
```diff
@@ -379,13 +369,39 @@ func (adb *AccountsDB) RemoveAccount(address []byte) error {
...
+		return nil
+	}
+
+	oldCodeHash := baseAcc.GetCodeHash()
+	unmodifiedOldCodeEntry, err := adb.updateOldCodeEntry(oldCodeHash)
+	if err != nil {
+		return err
+	}
+
+	codeChangeEntry, err := NewJournalEntryCode(unmodifiedOldCodeEntry, oldCodeHash, nil, adb.mainTrie, adb.marshalizer)
```
raduchis, line 396: Should the newJournalEntryCode have the newCodeHash=nil every time? It does not have necessarily NumReferences<=0 and it is not specific for the current AccountHandler, right?

BeniaminDrasovean: oldCode refers to the code that the account had. After an account is removed, the num references to that code must be decreased, and if it is 0, removed. newCode refers to the code that is added to the account. Since the removed account will not refer to any code, newCodeHash is nil.

raduchis: ok

PR #2042, opened 2020-06-29
bugFix/go-routine-control-on-interceptors

sasurobert:
added go routine control for multi data interceptors. made them launch only one routine per received message.

### DIFF
--- a/process/interceptors/multiDataInterceptor.go
+++ b/process/interceptors/multiDataInterceptor.go
@@ -113,32 +113,26 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P, 
 		return err
 	}
 
-	lastErrEncountered := error(nil)
-	wgProcess := &sync.WaitGroup{}
-	wgProcess.Add(len(multiDataBuff))
-
-	go func() {
-		wgProcess.Wait()
-		mdi.throttler.EndProcessing()
-	}()
-
+	listInterceptedData := make([]process.InterceptedData, len(multiDataBuff))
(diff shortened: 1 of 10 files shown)


process/interceptors/multiDataInterceptor.go
```diff
@@ -113,32 +113,26 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
...
 		}
 
 		isWhiteListed := mdi.whiteListRequest.IsWhiteListed(interceptedData)
 		if !isWhiteListed && errOriginator != nil {
-			lastErrEncountered = errOriginator
-			allWhiteListed = false
-			wgProcess.Done()
-			continue
+			mdi.throttler.EndProcessing()
+			log.Debug("got message from peer on topic only for validators", "originator",
```
iulianpascalau, line 131: Might add a TODO to remove this in the future when we find out that this works correctly

process/interceptors/multiDataInterceptor.go
```diff
@@ -152,28 +146,25 @@ func (mdi *MultiDataInterceptor) ProcessReceivedMessage(message p2p.MessageP2P,
 				"is for this shard", isForCurrentShard,
 				"is white listed", isWhiteListed,
 			)
-			wgProcess.Done()
-			continue
```
iulianpascalau, line 156: Return error here as to not allow broadcast of combined shard data.

process/interceptors/multiDataInterceptor.go
sasurobert: done

process/interceptors/multiDataInterceptor.go
sasurobert: done

PR #2044, opened 2020-06-29
Fix pools cleaner lock-unlock and start from storage edge case

SebastianMarian:
* Fixed edge case on bootstrap from storage when the highest final nonce was higher than highest self notarized nonce committed in the last block (this situation could happen when a node is in sync mode and each committed block become final because the node has in advance info from metachain, which confirms the finality of the committed block, and also that the node is on right chain)
* Fixed/optimized lock/unlock situation in txPoolsCleaner and miniBlockPoolsCleaner
* Optimized computation for PrepareBroadcastBlockDataValidator in subround EndRound

### DIFF
--- a/process/block/poolsCleaner/miniBlocksPoolsCleaner.go
+++ b/process/block/poolsCleaner/miniBlocksPoolsCleaner.go
@@ -82,8 +82,13 @@ func (mbpc *miniBlocksPoolsCleaner) cleanMiniblocksPools(ctx context.Context) {
 		case <-time.After(sleepTime):
 		}
 
+		startTime := time.Now()
 		numMiniblocksInMap := mbpc.cleanMiniblocksPoolsIfNeeded()
-		log.Debug("miniBlocksPoolsCleaner.cleanMiniblocksPools", "num miniblocks in map", numMiniblocksInMap)
+		elapsedTime := time.Since(startTime)
+
+		log.Debug("miniBlocksPoolsCleaner.cleanMiniblocksPools",
+			"num miniblocks in map", numMiniblocksInMap,
+			"elapsed time", elapsedTime)
 	}
 }
 
@@ -123,11 +128,10 @@ func (mbpc *miniBlocksPoolsCleaner) receivedMiniBlock(key []byte, value interfac
 }
 
 func (mbpc *miniBlocksPoolsCleaner) cleanMiniblocksPoolsIfNeeded() int {
-	mbpc.mutMapMiniBlocksRounds.Lock()
-	defer mbpc.mutMapMiniBlocksRounds.Unlock()
-
 	numMbsCleaned := 0
+	hashesToRemove := make(map[string]struct{})
 
+	mbpc.mutMapMiniBlocksRounds.Lock()
 	for hash, mbi := range mbpc.mapMiniBlocksRounds {
 		_, ok := mbpc.miniblocksPool.Get([]byte(hash))
 		if !ok {
@@ -154,7 +158,7 @@ func (mbpc *miniBlocksPoolsCleaner) cleanMiniblocksPoolsIfNeeded() int {
 			continue
 		}
 
-		mbpc.miniblocksPool.Remove([]byte(hash))
+		hashesToRemove[hash] = struct{}{}
 		delete(mbpc.mapMiniBlocksRounds, hash)
 		numMbsCleaned++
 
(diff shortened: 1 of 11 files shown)


process/block/shardblock.go
```diff
@@ -1828,11 +1828,23 @@ func (sp *shardProcessor) getBootstrapHeadersInfo(
 	selfNotarizedHeadersHashes [][]byte,
 ) []bootstrapStorage.BootstrapHeaderInfo {
 
-	if len(selfNotarizedHeaders) == 0 {
+	numSelfNotarizedHeaders := len(selfNotarizedHeaders)
```
iulianpascalau, line 1831: maybe some tests for this?

process/block/poolsCleaner/txsPoolsCleaner.go
```diff
@@ -259,11 +263,22 @@ func (tpc *txsPoolsCleaner) cleanTxsPoolsIfNeeded() int {
...
+
+	startTime := time.Now()
+	for hash, txStore := range hashesToRemove {
+		txStore.Remove([]byte(hash))
+	}
+	elapsedTime := time.Since(startTime)
+
 	if numTxsCleaned > 0 {
-		log.Debug("txsPoolsCleaner.cleanTxsPoolsIfNeeded", "num txs cleaned", numTxsCleaned)
+		log.Debug("txsPoolsCleaner.cleanTxsPoolsIfNeeded",
```
sasurobert, line 276: do we need the same log duplicated ?

process/block/poolsCleaner/miniBlocksPoolsCleaner.go
```diff
@@ -82,8 +82,13 @@ func (mbpc *miniBlocksPoolsCleaner) cleanMiniblocksPools(ctx context.Context) {
 		case <-time.After(sleepTime):
 		}
 
+		startTime := time.Now()
 		numMiniblocksInMap := mbpc.cleanMiniblocksPoolsIfNeeded()
-		log.Debug("miniBlocksPoolsCleaner.cleanMiniblocksPools", "num miniblocks in map", numMiniblocksInMap)
+		elapsedTime := time.Since(startTime)
+
+		log.Debug("miniBlocksPoolsCleaner.cleanMiniblocksPools",
```
sasurobert, line 89: why the same log 2 times ?

process/block/poolsCleaner/txsPoolsCleaner.go
SebastianMarian: Its not duplicated. This one display the number of txs cleaned if they were, and the above one display the number of txs which are in the whole map, and could be cleaned in the future if they will remain unprocessed. Also the elapsed time displayed here refers to the time used to remove the txs from cacher, and the one above refers to the whole time consumed by this method -> cleanTxsPoolsIfNeeded

process/block/poolsCleaner/miniBlocksPoolsCleaner.go
SebastianMarian: Its not duplicated. The below one display the number of mbs cleaned if they were, and tthis one display the number of mbs which are in the whole map, and could be cleaned in the future if they will remain unprocessed. Also the elapsed time displayed below refers to the time used to remove the mbs from cacher, and this one refers to the whole time consumed by this method -> cleanMiniblocksPoolsIfNeeded

process/block/poolsCleaner/txsPoolsCleaner.go
SebastianMarian: Its not duplicated. This one display the number of txs cleaned if they were, and the above one display the number of txs which are in the whole map, and could be cleaned in the future if they will remain unprocessed. Also the elapsed time displayed here refers to the time used to remove the txs from cacher, and the one above refers to the whole time consumed by this method -> cleanTxsPoolsIfNeeded

process/block/poolsCleaner/miniBlocksPoolsCleaner.go
SebastianMarian: Its not the same. The below one display the number of mbs cleaned if they were, and tthis one display the number of mbs which are in the whole map, and could be cleaned in the future if they will remain unprocessed. Also the elapsed time displayed below refers to the time used to remove the mbs from cacher, and this one refers to the whole time consumed by this method -> cleanMiniblocksPoolsIfNeeded

process/block/shardblock.go
SebastianMarian: Added

PR #2058, opened 2020-06-30
Multitiple save key value

sasurobert:
allow saving multiple key values through builtInFunctions with the same tx.


process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -37,58 +37,76 @@ func (k *saveKeyValueStorage) ProcessBuiltinFunction(
...
-		return nil, process.ErrBuiltInFunctionCalledWithValue
+		return process.ErrBuiltInFunctionCalledWithValue
 	}
 	if check.IfNil(acntDst) {
-		return nil, process.ErrNilSCDestAccount
+		return process.ErrNilSCDestAccount
 	}
 	if !bytes.Equal(input.CallerAddr, input.RecipientAddr) {
-		return nil, fmt.Errorf("%w not the owner of the account", process.ErrOperationNotPermitted)
+		return fmt.Errorf("%w not the owner of the account", process.ErrOperationNotPermitted)
```
andreibancioiu, line 101: Is the condition above correct? (`CallerAddr` compared to `RecipientAddr`)

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -37,58 +37,76 @@ func (k *saveKeyValueStorage) ProcessBuiltinFunction(
...
+		key := input.Arguments[i]
+		value := input.Arguments[i+1]
+		length := uint64(len(value))
+		useGas += length * (k.gasConfig.DataCopyPerByte + k.gasConfig.PersistPerByte)
+
+		if !process.IsAllowedToSaveUnderKey(key) {
+			return nil, fmt.Errorf("%w it is not allowed to save under key %s", process.ErrOperationNotPermitted, key)
+		}
+
+		oldValue, _ := acntDst.DataTrieTracker().RetrieveValue(key)
```
andreibancioiu, line 60: We never expect errors here?

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -37,58 +37,76 @@ func (k *saveKeyValueStorage) ProcessBuiltinFunction(
...
+			continue
+		}
+
+		lengthChange := uint64(0)
+		lengthOldValue := uint64(len(oldValue))
+		if lengthOldValue < length {
+			lengthChange = length - lengthOldValue
+		} else {
+			releaseLength := lengthOldValue - length
+			vmOutput.GasRefund = big.NewInt(0).Mul(big.NewInt(0).SetUint64(releaseLength), big.NewInt(0).SetUint64(k.gasConfig.ReleasePerByte))
```
andreibancioiu, line 71: So we do this in a loop. Which means `vmOutput.GasRefund` is overridden - is this correct?

process/smartContract/builtInFunctions/keyValueStorage.go
ccorcoveanu, line 60: Should we treat error?

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -37,58 +37,76 @@ func (k *saveKeyValueStorage) ProcessBuiltinFunction(
...
-
-	value := input.Arguments[1]
-	length := uint64(len(value))
-	useGas := k.funcGasCost + length*(k.gasConfig.DataCopyPerByte+k.gasConfig.PersistPerByte)
-	vmOutput := &vmcommon.VMOutput{GasRemaining: input.GasProvided - useGas}
-
-	key := input.Arguments[0]
-	if !process.IsAllowedToSaveUnderKey(key) {
-		return nil, fmt.Errorf("%w it is not allowed to save under key %s", process.ErrOperationNotPermitted, key)
+	if len(input.Arguments)%2 != 0 {
```
ccorcoveanu, line 106: I would move this to 93 so `Arguments` checks are one after another

process/smartContract/builtInFunctions/keyValueStorage.go
```diff
@@ -37,58 +37,76 @@ func (k *saveKeyValueStorage) ProcessBuiltinFunction(
...
+			releaseLength := lengthOldValue - length
+			vmOutput.GasRefund = big.NewInt(0).Mul(big.NewInt(0).SetUint64(releaseLength), big.NewInt(0).SetUint64(k.gasConfig.ReleasePerByte))
+		}
+
+		useGas += k.gasConfig.StorePerByte * lengthChange
+		if input.GasProvided < useGas {
+			return nil, process.ErrNotEnoughGas
+		}
+
+		acntDst.DataTrieTracker().SaveKeyValue(key, value)
```
andreibancioiu, line 79: This means that we don't set key-values atomically - is this what we want? Perhaps a rollback is actually done in case of `ErrNotEnoughGas` after setting, say, 2 key-values, but not having enough gas for the remaining, say, 5 key-values?

process/smartContract/builtInFunctions/keyValueStorage_test.go
```diff
@@ -67,3 +67,46 @@ func TestSaveKeyValue_ProcessBuiltinFunction(t *testing.T) {
 	_, err = skv.ProcessBuiltinFunction(nil, acc, vmInput)
 	require.True(t, errors.Is(err, process.ErrOperationNotPermitted))
 }
+
+func TestSaveKeyValue_ProcessBuiltinFunctionMultipleKeys(t *testing.T) {
```
andreibancioiu, line 71: Can we perhaps add an extra unit test for the scenario `ErrNotEnoughGas`?

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: yes. only the owner can save under his account some data

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: no error is expected here - anyway it will be rewritten with the new value.

sasurobert: no error is expected here - anyway it will be rewritten with the new value.

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: good find. changed

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: if error is done - rollback is processed for all save key values. you can either save all, or none at all.

process/smartContract/builtInFunctions/keyValueStorage_test.go
sasurobert: done

process/smartContract/builtInFunctions/keyValueStorage.go
sasurobert: done

PR #2061, opened 2020-06-30
Measure Blockchain Hooks duration

andreibancioiu:
- Replace `log.Warn` with `log.Debug` for SC duration measurements.
 - Add duration measurement for Blockchain Hooks.
 - Update txgen scripts under `scripts/testnet`.


process/smartContract/hooks/blockChainHook.go
```diff
@@ -127,6 +133,8 @@ func (bh *BlockChainHookImpl) GetUserAccount(address []byte) (vmcommon.UserAccou
 
 // GetStorageData returns the storage value of a variable held in account's data trie
 func (bh *BlockChainHookImpl) GetStorageData(accountAddress []byte, index []byte) ([]byte, error) {
+	defer stopMeasure(startMeasure("GetStorageData"))
```
sasurobert, line 136: i do not see calling of function startMeasure

core/stopWatch.go
```diff
@@ -91,14 +91,14 @@ func (sw *StopWatch) GetMeasurementsMap() map[string]float64 {
 	return output
 }
 
-// GetMeasurement returns the measurement (duration in seconds) by identifier
-func (sw *StopWatch) GetMeasurement(identifier string) float64 {
+// GetMeasurement returns the measurement by identifier
+func (sw *StopWatch) GetMeasurement(identifier string) time.Duration {
```
ccorcoveanu, line 95: Is it ok(expected) that this function will only return elapsed time from when you called `Stop`?

process/smartContract/hooks/blockChainHook.go
sasurobert: I see right now what it does :)

ccorcoveanu: Was also confused by this. You first see `stopMeasure` and then you also have to know that arguments in a deferred function are evaluated at the defer evaluation. Maybe makes sense for a 2 liner here?

process/smartContract/process.go
```diff
@@ -172,12 +173,12 @@ func (sc *scProcessor) ExecuteSmartContractTransaction(
 	sw.Stop("execute")
 	duration := sw.GetMeasurement("execute")
 
-	logFunc := log.Trace
-	if duration > executeDurationWarnThreshold {
-		logFunc = log.Debug
+	if duration > executeDurationAlarmThreshold {
```
ccorcoveanu, line 176: Not for now, but I think that this:
```go
sw := core.NewStopWatch()
sw.Start("execute")
returnCode, err := sc.doExecuteSmartContractTransaction(tx, acntSnd, acntDst)
sw.Stop("execute")
duration := sw.GetMeasurement("execute")

if duration > executeDurationAlarmThreshold {
	log.Debug(fmt.Sprintf("scProcessor.ExecuteSmartContractTransaction(): execution took > %s", duration), "sc", tx.GetRcvAddr(), "duration", duration, "returnCode", returnCode, "err", err, "data", string(tx.GetData()))
} else {
	log.Trace("scProcessor.ExecuteSmartContractTransaction()", "sc", tx.GetRcvAddr(), "duration", duration, "returnCode", returnCode, "err", err, "data", string(tx.GetData()))
}
```

Could become something like this?

```go
func TestStopwatch(t *testing.T) {
        x := 100
	measure(func() {
		fmt.Println(x)
	})
}

type measuredFuc func()
func measure(myFunc measuredFuc) {
	fmt.Println("initStopwatch")
	myFunc()
	fmt.Println("endStopwatch")
}
```

core/stopWatch.go
andreibancioiu: Indeed, the stop watch was designed that way. `GetMeasurement` returns delta between `Start` and `Stop`, disregarding the moment you call it.

process/smartContract/hooks/blockChainHook.go
andreibancioiu: I intended to make it a one-liner especially to be commonly used (as a pattern). Can we keep it, perhaps adopt it even throughout the project?

process/smartContract/process.go
andreibancioiu: Added a Jira task here: https://elrondnetwork.atlassian.net/browse/EN-7026

PR #2063, opened 2020-07-01
bugFix/sync-empty-trie

sasurobert:
empty trie should not supposed to by synced.


data/trie/sync.go
```diff
@@ -77,7 +77,7 @@ func NewTrieSyncer(
 
 // StartSyncing completes the trie, asking for missing trie nodes on the network
 func (ts *trieSyncer) StartSyncing(rootHash []byte, ctx context.Context) error {
-	if len(rootHash) == 0 {
+	if len(rootHash) == 0 || bytes.Equal(rootHash, EmptyTrieHash) {
```
iulianpascalau, line 80: Question: now we prevented the request of the 000....000 root hash, but it will auto-generate it? I mean, it now finds out it missing the empty trie root hash, tries to ask for it, this function will early exit but it will continue to request the other node hashes?

sasurobert: empty trie roothash happens when someone deploys a smart contract which does not write anything to the data trie.

sasurobert: this function will not request hashes for this data trie - it will do for others.

iulianpascalau: ok, will test this on a faulty node of mine

PR #2070, opened 2020-07-03
fix Dockerfile arwen location

BenWhiteJam:
- removes hardcoded arwen version, instead fetches version from go.mod file
- moves arwen wasm to lib folder in second build stage


docker/elrond/Dockerfile
```diff
@@ -7,12 +7,11 @@ RUN GO111MODULE=on go mod vendor
 #Elrond node
 WORKDIR /go/elrond-go/cmd/node
 RUN go build -i -v -ldflags="-X main.appVersion=$(git describe --tags --long --dirty)"
-
+RUN cp /go/pkg/mod/github.com/!elrond!network/arwen-wasm-vm@$(cat /go/elrond-go/go.mod | grep arwen-wasm-vm | sed 's/.* //')/wasmer/libwasmer_linux_amd64.so /lib/libwasmer_linux_amd64.so
```
iulianpascalau, line 10: :+1:

PR #2077, opened 2020-07-07
Fix possible OOM generated by whitelisted items

SebastianMarian:
* Added late cleaning mechanism for whitelisted mini blocks and transactions to avoid possible out of memory problem


SebastianMarian:
Not needed anymore, the real problem has been found

PR #2084, opened 2020-07-09
renamed fullArchive to cleanOldEpochsData

bogdan-rosianu:
renamed boolean parameter `FullArchive` to `CleanOldEpochsData`

### DIFF
--- a/storage/factory/pruningStorerFactory.go
+++ b/storage/factory/pruningStorerFactory.go
@@ -40,7 +40,7 @@ func NewStorageServiceFactory(
 	if config == nil {
 		return nil, storage.ErrNilConfig
 	}
-	if config.StoragePruning.NumEpochsToKeep < minimumNumberOfEpochsToKeep && !config.StoragePruning.FullArchive {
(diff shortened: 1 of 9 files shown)


storage/factory/pruningStorerFactory.go
```diff
@@ -40,7 +40,7 @@ func NewStorageServiceFactory(
 	if config == nil {
 		return nil, storage.ErrNilConfig
 	}
-	if config.StoragePruning.NumEpochsToKeep < minimumNumberOfEpochsToKeep && !config.StoragePruning.FullArchive {
+	if config.StoragePruning.NumEpochsToKeep < minimumNumberOfEpochsToKeep && !config.StoragePruning.CleanOldEpochsData {
```
AdoAdoAdo, line 43: why is cleanOldEpochsData relevant here?

bogdan-rosianu: If the node doesn't remove old epochs, then there is no use for the `NumEpochsToKeep` so the check shouldn't be blocking. I can remove the condition if needed. Also, thanks for finding this because I have to change the boolean condition

AdoAdoAdo: ok then you can keep the check, but remove negation

PR #2085, opened 2020-07-09
remove get from elastic indexer

miiu96:
Removed all GET requests from elasticsearch observers to elasticsearch database.
Now all observers index data without checking if data for transaction and miniblocks are already in database.


core/indexer/common.go
```diff
@@ -198,7 +198,7 @@ func (cm *commonProcessor) convertScResultInDatabaseScr(sc *smartContractResult.
 		Code:          string(sc.Code),
 		Data:          decodedData,
 		PreTxHash:     hex.EncodeToString(sc.PrevTxHash),
-		CallType:      string(sc.CallType),
+		CallType:      string(rune(sc.CallType)),
```
ccorcoveanu, line 201: Why do we cast to rune?

core/indexer/common.go
```diff
@@ -365,41 +359,38 @@ func serializeTransactions(
...
 	var err error
-	if existsInDb {
-		if !isCrossShardDstMe(tx, selfShardID) || tx.Status == txStatusInvalid {
-			return
-		}
+	if isCrossShardDstMe(tx, selfShardID) && tx.Status != txStatusInvalid {
 		// update transaction
 		meta, serializedData = prepareTxUpdate(tx)
 	} else {
+		// write tx
```
ccorcoveanu, line 391: Remove this extra comment

core/indexer/common.go
miiu96: because for the moment call type is a strange character. also golang suggestion to use rune

core/indexer/common.go
miiu96: done.

core/indexer/common.go
ccorcoveanu: isn't callType integer (0, 1, 2)? if you save it as string (utf8) of course callType will be a strange character. You should save it like this I think :) `strconv.Itoa(sc.CallType)`, this way you will get the correct value into the database (or better save it as integer in the json)

miiu96: refactored.

PR #2092, opened 2020-07-11
Api middlewares refactor

iulianpascalau:
- merged all 3 middlewares in a single one as to avoid calling `c.Next()` on the gin context multiple times


facade/nodeFacade.go
```diff
@@ -198,27 +211,17 @@ func (nf *nodeFacade) startRest() {
...
-
 func (nf *nodeFacade) sourceLimiterReset(reset resetHandler) {
+	betweenResetDuration := time.Second * time.Duration(nf.wsAntifloodConfig.SameSourceResetIntervalInSec)
 	for {
-		time.Sleep(time.Second * time.Duration(nf.wsAntifloodConfig.SameSourceResetIntervalInSec))
-
-		log.Trace("calling reset on WS source limiter")
-		reset.Reset()
+		select {
+		case <-time.After(betweenResetDuration):
```
andreibancioiu, line 218: So this is a "hard" reset, which means from time to time we forgive everyone, which means that an eventual attacker, if starting right before the reset and continuing immediately after a reset, could send `maxNumRequestsPerAddress * 2` requests in a short period of time (just thinking out loud, shouldn't be an issue, sliding mechanism not really required).

facade/nodeFacade.go
```diff
@@ -343,6 +346,14 @@ func (nf *nodeFacade) GetThrottlerForEndpoint(endpoint string) (core.Throttler,
 	return throttlerForEndpoint, isThrottlerOk
 }
 
+// Close will cleanup started go routines
+// TODO use this close method
```
andreibancioiu, line 350: Perhaps add a Jira ticket for this, so that we don't forget?

PR #2093, opened 2020-07-11
[ES] do get requests only for miniblocks

miiu96:


core/indexer/common.go
```diff
@@ -339,14 +354,18 @@ func serializeTransactions(
 func prepareSerializedDataForATransaction(
 	tx *Transaction,
 	selfShardID uint32,
+	isMBOfTxInDB bool,
```
bogdan-rosianu, line 357: you can use full name `isMiniBlockOfTransactionInDatabase`

PR #2095, opened 2020-07-12
bugFix/economics-based-on-genesis-supply-only

sasurobert:
fixed economics computation where the previous total supply was taken in consideration for the total to distribute, instead the genesis total supply has to be taken according to new documents.

### DIFF
--- a/epochStart/metachain/economics.go
+++ b/epochStart/metachain/economics.go
@@ -70,14 +70,15 @@ func NewEndOfEpochEconomicsDataCreator(args ArgsNewEpochEconomics) (*economics, 
 	}
 
 	e := &economics{
-		marshalizer:      args.Marshalizer,
-		hasher:           args.Hasher,
-		store:            args.Store,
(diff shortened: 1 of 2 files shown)


epochStart/metachain/economics.go
```diff
@@ -116,7 +117,7 @@ func (e *economics) ComputeEndOfEpochEconomics(
 	totalNumBlocksInEpoch := e.computeNumOfTotalCreatedBlocks(noncesPerShardPrevEpoch, noncesPerShardCurrEpoch)
 
 	inflationRate := e.computeInflationRate(metaBlock.GetRound())
-	rwdPerBlock := e.computeRewardsPerBlock(prevEpochEconomics.TotalSupply, maxBlocksInEpoch, inflationRate)
+	rwdPerBlock := e.computeRewardsPerBlock(e.genesisTotalSupply, maxBlocksInEpoch, inflationRate)
```
iulianpascalau, line 120: :+1:

PR #2096, opened 2020-07-12
Avoid whitelisting out of range block data

SebastianMarian:
* Rejected whitelisting miniblocks included in headers which are out of range (this will fix the situation when a node is syncing and it will receive headers with higher nonces compared with the last one processed. Because of these headers which are received, it will whitelist its miniblocks and afterwards its transactions. More than this it will also add these transactions into immunized cache. This will fill up these caches (whitelist cache, immunized cache) with a lot of information which is not needed right now and which will also be evicted until its time will come)

### DIFF
--- a/process/track/baseBlockTrack_test.go
+++ b/process/track/baseBlockTrack_test.go
@@ -2434,6 +2434,68 @@ func TestBaseBlockTrack_DoWhitelistWithShardHeaderIfNeededNilShardShouldReturnAn
 	assert.Equal(t, 0, len(cache))
 }
 
+func TestBaseBlockTrack_DoWhitelistWithMetaBlockIfNeededIsHeaderOutOfRangeShouldReturn(t *testing.T) {
+	t.Parallel()
+
+	cache := make(map[string]struct{})
+	mutCache := sync.Mutex{}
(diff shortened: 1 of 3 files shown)


process/track/baseBlockTrack_test.go
```diff
@@ -2495,3 +2557,34 @@ func TestBaseBlockTrack_DoWhitelistWithShardHeaderIfNeededMetaShouldWhitelistCro
 
 	assert.True(t, ok)
 }
+
+func TestBaseBlockTrack_IsHeaderOutOfRangeShouldWork(t *testing.T) {
```
andreibancioiu, line 2561: The tests run under the assumption that `lastCrossNotarizedHeader` has round `0`, nonce `0`. Is this correct?

SebastianMarian: Yes, what I wanted to test is that if the difference between last cross notarized header, no matter if it has nonce 0 (genesis),  and the current header received is out of range ( <= last notarized or  > last notarized + window ) the method works correctly.

SebastianMarian: Anyway, I have refactored the test to be more clear now

PR #2097, opened 2020-07-12
check if node exists in genesis config before setting destination shard

bogdan-rosianu:
fixed the situation where if an entire testnet started with the same `prefs.toml` file and it's value of `DestinationShardAsObserver` would have not been set to `disabled` then all the nodes joined as observers in that shard.

### DIFF
--- a/epochStart/bootstrap/process.go
+++ b/epochStart/bootstrap/process.go
@@ -1,6 +1,7 @@
 package bootstrap
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"time"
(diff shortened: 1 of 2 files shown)


epochStart/bootstrap/process.go
```diff
@@ -264,7 +265,10 @@ func (e *epochStartBootstrap) prepareEpochZero() (Parameters, error) {
 		return Parameters{}, err
 	}
 
-	shardIDToReturn := e.applyShardIDAsObserverIfNeeded(e.genesisShardCoordinator.SelfId())
+	shardIDToReturn := e.genesisShardCoordinator.SelfId()
+	if !e.isNodeInGenesisNodesConfig() {
```
sasurobert, line 269: can you add a unit test for this :) ? good find.

bogdan-rosianu: done. added 2 unit tests

PR #2100, opened 2020-07-13
Fix date in log file name

camilbancioiu:
Previously, the node used to generate log files of the form `elrond-go-YYYY-DD-MM-HH-MM-SS.log` (note the order year-day-month). This PR fixes the order in the date format, changing it to the more readable `YYYY-MM-DD` (year-month-date).

### DIFF
--- a/core/file.go
+++ b/core/file.go
@@ -114,7 +114,8 @@ func CreateFile(prefix string, subfolder string, fileExtension string) (*os.File
 		return nil, err
 	}
 
-	fileName := time.Now().Format("2006-02-01-15-04-05")
+	fileName := time.Now().Format("2006-01-02T15-04-05")
+	fileName = strings.Replace(fileName, "T", "-", 1)
 	if prefix != "" {
 		fileName = prefix + "-" + fileName
 	}


core/file.go
```diff
@@ -114,7 +114,8 @@ func CreateFile(prefix string, subfolder string, fileExtension string) (*os.File
 		return nil, err
 	}
 
-	fileName := time.Now().Format("2006-02-01-15-04-05")
+	fileName := time.Now().Format("2006-01-02T15-04-05")
+	fileName = strings.Replace(fileName, "T", "-", 1)
```
bogdan-rosianu, line 118: why `T` and replace with `-`? it works just fine if you use directly `"2006-01-02-15-04-05"`

camilbancioiu: `"2006-01-02-15-04-05"` doesn't work fine, unfortunately - it was the cause of the issue. The `T` is required as a format specifier, otherwise `time.Format()` will think the `01` in `2006-01-02` is the day, not the month. 

I found no simple way of specifying a custom format in `time.Format()`, and the built-in one requires `T`, which is then replaced with a dash.

bogdan-rosianu: you're right. I've read better the documentation

PR #2104, opened 2020-07-13
ProtocolID instead of randezvous

iulianpascalau:
- renamed the RandezVous option to ProtocolID

### DIFF
--- a/integrationTests/testInitializer.go
+++ b/integrationTests/testInitializer.go
@@ -111,7 +111,7 @@ func createP2PConfig(initialPeerList []string) config.P2PConfig {
 		KadDhtPeerDiscovery: config.KadDhtPeerDiscoveryConfig{
 			Enabled:                          true,
 			RefreshIntervalInSec:             2,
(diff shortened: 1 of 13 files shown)


integrationTests/testInitializer.go
```diff
@@ -143,13 +143,13 @@ func CreateMessengerWithKadDht(initialAddr string) p2p.Messenger {
 }
 
 // CreateMessengerWithKadDhtAndProtocolID creates a new libp2p messenger with kad-dht peer discovery and peer ID
-func CreateMessengerWithKadDhtAndProtocolID(initialAddr string, randezVous string) p2p.Messenger {
+func CreateMessengerWithKadDhtAndProtocolID(initialAddr string, ProtocolID string) p2p.Messenger {
```
SebastianMarian, line 146: why not protocolID ?

iulianpascalau: because of copy paste. Thanks.

iulianpascalau: fixed

PR #2105, opened 2020-07-13
Fix whitelisting in advance for miniblocks

SebastianMarian:
* Fixed a possible problem which could avoid whitelisting miniblocks

* Fixed situation when on headers interceptor and also in block tracker component, some old headers are dropped because they have nonces behind final, but the node still needs them (especially in epoch start headers could exist some old references to headers behind final and they should be accepted, as they are requested and whitelisted)

* Changed some behavior of getting headers from storage with getting headers from pool and afterward from storage

### DIFF
--- a/process/track/baseBlockTrack.go
+++ b/process/track/baseBlockTrack.go
@@ -799,13 +799,6 @@ func (bbt *baseBlockTrack) isHeaderOutOfRange(headerHandler data.HeaderHandler) 
 		return true
 	}
 
-	if headerHandler.GetNonce() <= lastCrossNotarizedHeader.GetNonce() {
-		return true
-	}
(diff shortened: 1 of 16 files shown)


process/track/baseBlockTrack.go
```diff
@@ -462,6 +450,11 @@ func (bbt *baseBlockTrack) getFinalHeader(shardID uint32) (data.HeaderHandler, [
 	return bbt.selfNotarizer.GetFirstNotarizedHeader(shardID)
 }
 
+// CheckBlockAgainstWhiteList returns if the provided intercepted data (blocks) is whitelisted or not
+func (bbt *baseBlockTrack) CheckBlockAgainstWhitelist(interceptedData process.InterceptedData) bool {
```
iulianpascalau, line 454: A little bit over engineered to have a function that has only one simple instruction that already explains what it does

PR #2112, opened 2020-07-14
Fix sync speed problem with attesting finality headers which are not requested in advance

SebastianMarian:
* Fixed sync problem when attesting finality header is not requested in advance


process/block/baseProcess.go
```diff
@@ -228,47 +227,45 @@ func (bp *baseProcessor) requestHeadersIfMissing(
...
+
 	for _, nonce := range missingNonces {
 		go bp.requestHeaderByShardAndNonce(shardId, nonce)
 	}
 
 	return nil
 }
 
+func addMissingNonces(diff int64, lastNonce uint64, missingNonces []uint64) {
+	if diff <= 1 {
```
iulianpascalau, line 253: < 2 ?

PR #2113, opened 2020-07-14
process: fix commit too long, add logs to measure commit time

AdoAdoAdo:
Removing the waitgroup and execution on multiple go routines, as in high goroutines load we might lose more time on scheduling and task switching until the execution is finalized (scheduler might chose other goroutines from the ready pool).

### DIFF
--- a/process/block/shardblock.go
+++ b/process/block/shardblock.go
@@ -741,6 +741,8 @@ func (sp *shardProcessor) CommitBlock(
 		return err
 	}
 
+	timing := make([]time.Time, 0)
+	timing = append(timing, time.Now())
 	sp.store.SetEpochForPutOperation(headerHandler.GetEpoch())
 
 	log.Debug("started committing block",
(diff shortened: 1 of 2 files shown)


process/block/shardblock.go
```diff
@@ -905,16 +928,32 @@ func (sp *shardProcessor) CommitBlock(
...
 	sp.cleanupPools(headerHandler)
+	timing = append(timing, time.Now())
+
+	logTimings(timing)
 
 	return nil
 }
 
+func logTimings(timings []time.Time) {
+	totalDuration := timings[len(timings)-1].UnixNano() - timings[0].UnixNano()
```
iulianpascalau, line 947: same as found in #2214 
we have something for these types of measurements. Might take a look in core/stopWatch.go. It supports identifiers and is optimized to return the values as []interface{} to match the input for a log print.

AdoAdoAdo:
debug patch used to find the issue, which already has a PR.

PR #2116, opened 2020-07-15
Refactored API (again)

iulianpascalau:
- added log prints when REST API is stuck. Should run the node with `api:DEBUG` or `*:DEBUG` to display the info lines
- reverted the changes don in the last refactor round: we are back on three middleware functions from commit 2e37591c031590824cb15c9a116427aee47ccbf0
- refactored throttling on the REST API endpoint by using a special middleware function for this purpose
- fixed API stuck condition that appeared if one handler panic and the queue was not cleaned. Used a defer function instead of executing the cleanup after the call to `c.Next()`.
- added unit tests for endpointThrottler.go
- replaced `c.MustGet()` with `c.Get()` in order to avoid panics in case of bad node configuration. Added unit tests.
- trimmed the API log messages
- removed duplicate code in endpoint handlers
- removed "Elrond" naming inside the package
- switched the api default log level to TRACE
- added constants for the routes names


api/node/routes.go
```diff
@@ -134,7 +160,20 @@ func Statistics(c *gin.Context) {
 
 // StatusMetrics returns the node statistics exported by an StatusMetricsHandler without p2p statistics
 func StatusMetrics(c *gin.Context) {
-	ef, ok := c.MustGet("elrondFacade").(FacadeHandler)
+	efObj, ok := c.Get("elrondFacade")
```
bogdan-rosianu, line 163: you can leave a TODO (no need to do this on this PR) to add commom methods like `RespondWithNilAppContext` or `RespondWithInvalidAppContext`

api/shared/shared.go
```diff
@@ -19,5 +19,5 @@ const ReturnCodeInternalError ReturnCode = "internal_issue"
 // ReturnCodeRequestError defines a request which hasn't been executed successfully due to a bad request received
 const ReturnCodeRequestError ReturnCode = "bad_request"
 
-// ReturnCodeRequestError defines a request which hasn't been executed successfully due to too many requests
+// ReturnCodeSystemBusy defines a request which hasn't been executed successfully due to too many requests
```
bogdan-rosianu, line 22: :+1:

api/middleware/endpointThrottler_test.go
```diff
@@ -0,0 +1,158 @@
...
+	makeRequestGlobalThrottler(ws, &mutResponses, responses)
+
+	mutResponses.Lock()
+	defer mutResponses.Unlock()
+
+	assert.Equal(t, uint32(1), atomic.LoadUint32(&numCalls))
+	assert.Equal(t, 1, responses[http.StatusOK])
+}
+
+func TestCreateEndpointThrottler_ThrottlerCanNotStartShouldNotExecute(t *testing.T) {
```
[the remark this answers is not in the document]
iulianpascalau: done

api/node/routes.go
iulianpascalau: The best TODO you can write is actually the one you don't write. 
Added common function in each package as otherwise would have been an interface-usage nightmare. :)

api/api.go
```diff
@@ -51,7 +52,8 @@ type ginWriter struct {
 }
 
 func (gv *ginWriter) Write(p []byte) (n int, err error) {
-	log.Debug("gin server", "message", string(p))
+	trimmed := bytes.TrimSpace(p)
+	log.Debug("gin server", "message", string(trimmed))
```
sasurobert, line 56: Trace ?

api/transaction/routes.go
```diff
@@ -74,15 +75,42 @@ type TxResponse struct {
 
 // Routes defines transaction related routes
 func Routes(router *wrapper.RouterWrapper) {
-	router.RegisterHandler(http.MethodPost, "/send", SendTransaction)
+	router.RegisterHandler(
+		http.MethodPost,
+		"/send",
```
sasurobert, line 80: some magic strings here

api/api.go
iulianpascalau: :1st_place_medal: done

api/transaction/routes.go
iulianpascalau: added consts for all routes

PR #2120, opened 2020-07-15
Fix mbs and txs requests in advance when node is syncing

SebastianMarian:
* Fixed request of miniblocks in advance without delay when node is syncing
* Added request of transactions in advance for miniblocks which belongs to headers which are in a reasonable window from the last notarized header


epochStart/shardchain/trigger.go
```diff
@@ -458,7 +464,13 @@ func (t *trigger) receivedMetaBlock(headerHandler data.HeaderHandler, metaBlockH
 		t.newEpochHdrReceived = true
 		t.mapEpochStartHdrs[string(metaBlockHash)] = metaHdr
 		// waiting for late broadcast of mini blocks and transactions to be done and received
-		time.Sleep(core.ExtraDelayForRequestBlockInfo)
+		waitTime := core.ExtraDelayForRequestBlockInfo
+		roundDifferences := t.rounder.Index() - int64(headerHandler.GetRound())
+		if roundDifferences > 1 {
+			waitTime = 0
```
iulianpascalau, line 470: :tada:

PR #2121, opened 2020-07-15
TxPool: bulk remove transactions (optimized)

andreibancioiu:
**I still need to write some more tests, but you may have a look if possible.**

In some cases, bulk removal from cache (`removeTxBulk()`) is performed in an unfortunate order.

For example: 

 - for miniblock `0->1`, remove in bulk triggers removal of `[ ... (nonce=3000 -> nonce=7000 ...]` from `txListBySender`.
 - immediately after, for miniblock `0->0` removal applies to the rest of the list (say, 1000 to 2999).
 - and vice-versa, depending on the senders.

In such cases of **tail-first removal**, `removeTxBulk()` took long to compute. This was problematic at sync time.

Applied a number of optimizations. The most important - sort the transactions to be removed - first by sender, then by nonce, then apply a bulk removal - @AdoAdoAdo's idea. Initial implementation has been "picking up" from the last "found to remove" item (efficient, but not very clean). The "pick-up" (recall position hint) algorithm has been implemented at addition time though - where bulk addition isn't yet possible given the current design.

Also optimized a bit `txPoolsCleaner` to remove the transactions in bulks.


storage/txcache/txListForSender.go
```diff
@@ -184,20 +196,81 @@ func (listForSender *txListForSender) findListElementWithTx(txToFind *WrappedTra
...
+	listForSender.mutex.Lock()
+	defer listForSender.mutex.Unlock()
+
+	numRemoved := 0
+	pointerInTxsToRemove := 0
+	pointerInTxs := listForSender.items.Front()
+
+	for {
+		// End of any of the two lists, therefore end of loop
+		if pointerInTxs == nil || pointerInTxsToRemove == len(txsToRemove) {
```
AdoAdoAdo, line 234: can you move the negated condition to the for? then no need for break condition inside loop 
```go
for pointerInTxs != nil && pointerInTxsToRemove != len(txsToRemove) {
    ...
}
```

storage/txcache/txListForSender.go
```diff
@@ -184,20 +196,81 @@ func (listForSender *txListForSender) findListElementWithTx(txToFind *WrappedTra
...
 		}
+	}
+
+	return nil
+}
 
-		// Optimization: stop search at this point, since the list is sorted by nonce
-		if value.Tx.GetNonce() > txToFindNonce {
+// RemoveSortedTransactions removes the provided list of transactions, assuming the list is given sorted by nonce
+// QUESTION FOR REVIEW: should I refactor to have a loop-in-loop when traversing the two lists instead?
```
AdoAdoAdo, line 223: As you wish, I think it is also ok as it is.

storage/txcache/monitoring.go
```diff
@@ -245,3 +245,15 @@ func (cache *TxCache) displaySendersSummary() {
...
+
+func (cache *txListForSender) monitorOnFoundInsertionPlace(iterations *int, incomingTx *WrappedTransaction) {
+	if *iterations > numIterationsSuboptimalListTraversal {
+		log.Trace("txListForSender.monitorOnFoundInsertionPlace: suboptimal", "iterations", *iterations, "nonce", incomingTx.Tx.GetNonce(), "sender", incomingTx.Tx.GetSndAddr())
+	}
+}
+
+func (cache *txListForSender) monitorOnFoundTransaction(iterations *int, txToFind *WrappedTransaction) {
+	if *iterations > numIterationsSuboptimalListTraversal {
+		log.Debug("txListForSender.monitorOnFoundTransaction: suboptimal", "iterations", *iterations, "nonce", txToFind.Tx.GetNonce(), "sender", txToFind.Tx.GetSndAddr())
```
AdoAdoAdo, line 257: wondering why this has a different log level than for monitoring insertion place

storage/txcache/txListForSender.go
andreibancioiu: Fixed, thanks :)

> break and continue "are just the old goto in sheep's clothing" / _Wise man, 20th century_

storage/txcache/txListForSender.go
andreibancioiu: Kept.

storage/txcache/monitoring.go
andreibancioiu: - In our Tx Generator setup, transactions do not arrive in strictly ascending order. Actually, that is leading to **massive** _wall of log_ for `monitorOnFoundInsertionPlace: suboptimal`. Will work a bit more to improve this "bad-enough-case-scenario".

 - `monitorOnFoundTransaction` is called by the "single" version of the removal - `removeTxByHash`, which is originally called from `processAndRemoveBadTransaction`. Might as well change it to **trace**, thanks.

Will fix as well, to have the same log level.

andreibancioiu:
> can be further optimized by calling remove for all current txs block from each txPreprocessor.

process/block/poolsCleaner/txsPoolsCleaner.go
```diff
@@ -267,9 +268,20 @@ func (tpc *txsPoolsCleaner) cleanTxsPoolsIfNeeded() int {
...
-	for hash, txStore := range hashesToRemove {
-		txStore.Remove([]byte(hash))
+
+	for store, hashes := range hashesToRemoveByStore {
+		storeWithRemoveTxBulk, ok := store.(interface{ RemoveTxBulk(keys [][]byte) int })
+		if ok {
+			numRemoved := storeWithRemoveTxBulk.RemoveTxBulk(hashes)
+			log.Debug("RemoveTxBulk()", "num", len(hashes), "numRemoved", numRemoved, "store", fmt.Sprintf("%T", store))
+		} else {
+			log.Debug("bulk removal no supported, will remove one by one", "num", len(hashes), "store", fmt.Sprintf("%T", store))
```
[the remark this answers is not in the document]
andreibancioiu: Fixed :england: 

@sasurobert , I've added this in Jira for your comment: https://elrondnetwork.atlassian.net/browse/EN-7115

PR #2128, opened 2020-07-17
Add validator stats root hash to genesis block

SebastianMarian:
* Called prepare genesis block method only after validator statistics processor has finished its job, to set the correct validator stats root hash in the genesis meta block


cmd/node/factory/structs.go
```diff
@@ -493,10 +506,24 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
...
+	genesisBlock, ok := genesisBlocks[args.shardCoordinator.SelfId()]
+	if !ok {
+		return errors.New("genesis block does not exist")
+	}
+
+	err := args.data.Blkc.SetGenesisHeader(genesisBlock)
+	if err != nil {
+		return err
+	}
+
```
sasurobert, line 519: SetGenessisHeaderHash as well

cmd/node/factory/structs.go
```diff
@@ -339,7 +339,20 @@ func ProcessComponentsFactory(args *processComponentsFactoryArgs) (*Process, err
...
+	log.Debug("Validator stats created", "validatorStatsRootHash", validatorStatsRootHash)
+
+	genesisMetaBlock, ok := genesisBlocks[core.MetachainShardId]
+	if !ok {
+		return nil, errors.New("genesis meta block does not exist")
+	}
+
+	genesisMetaBlock.SetValidatorStatsRootHash(validatorStatsRootHash)
+
```
sasurobert, line 351: Delete empty line

cmd/node/factory/structs.go
SebastianMarian: I have checked the code and It is not needed, only nonce and round is used in newValidatorStatisticsProcessor, until the call of prepareGenesisBlock which will set again the genesis header, genesis hash with the correct validator stats root hash inside, and put them in the storage.

cmd/node/factory/structs.go
SebastianMarian: Done

PR #2132, opened 2020-07-18
Adjust miniblocks parameters

SebastianMarian:
* Adjusted some of the miniblocks parameters


process/constants.go
```diff
@@ -81,7 +81,7 @@ const MaxRoundsWithoutCommittedBlock = 10
 const MinForkRound = uint64(0)
 
 // MaxNumPendingMiniBlocks defines the maximum number of pending miniblocks, after which a shard could be considered stuck
-const MaxNumPendingMiniBlocks = 100
+const MaxNumPendingMiniBlocks = 500
```
iulianpascalau, line 84: As discussed, we might make this dynamically. However, we should use this as a quick fix.

PR #2137, opened 2020-07-21
Use round instead of nonce for synchronization estimation

miiu96:
Change input data for formula that is used to calculate synchronization time (estimation). 
Instead of using current synchronized nonce now is used current synchronized round because if we are using nonce we will have an estimation for current epoch not for  all the epochs that are needed to be synchronized.


statusHandler/presenter/chainInfoGetters.go
```diff
@@ -88,25 +88,25 @@ func (psh *PresenterStatusHandler) CalculateTimeToSynchronize() string {
 // CalculateSynchronizationSpeed will calculate and return speed of synchronization
 // how many block per second are synchronized
```
bogdan-rosianu, line 89: blocks*

miiu96: done.

PR #2153, opened 2020-07-24
Fix tx data

bogdan-rosianu:
Changed the `data` field of transactions to byte array instead of string in order to avoid issues of golang's json marshaler.


AdoAdoAdo:
I think there are still some changes required
For example in file indexer/common.go the transaction structure and the preparation of the transaction to be indexed.

core/indexer/processTransactions.go
```diff
@@ -92,7 +92,7 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
 		if nrScResult < minimumNumberOfSmartContractResults {
 			if len(transactions[hash].SmartContractResults) > 0 {
 				scResultData := transactions[hash].SmartContractResults[0].Data
-				if strings.Contains(scResultData, "@ok") {
+				if bytes.Contains(scResultData, []byte("@ok")) {
```
raduchis, line 95: is this check ok? Could the data contain "@ok" without this being the actual status?

bogdan-rosianu: yes, it should work fine. This check is needed for `esdt` calls which return only one smart contract result. And the check is ok, it will work just as the string comparison before.

raduchis: ok

PR #2157, opened 2020-07-25
Genesis delegation 0.4.0 + DNS contract 0.2.3

andrei-marinica:
New version of
- stripped down delegation
- DNS with `register` function disabled by default


cmd/node/config/genesisSmartContracts.json
```diff
@@ -5,7 +5,7 @@
     "vm-type": "0500",
     "init-parameters": "%auction_sc_address%@03E8@03E8@03E8",
     "type": "delegation",
-    "version": "0.3.*"
+    "version": "0.4.*"
```
iulianpascalau, line 8: :+1: 

Please replace the L87-L88 in /genesis/process/intermediate/baseDeploy.go with
```GO
			dp.pubkeyConv.Encode(scResultingAddressBytes),
			dp.pubkeyConv.Encode(ownerAddress),
```

andrei-marinica:
Looks good to me

PR #2162, opened 2020-07-27
Added reserved fields

iulianpascalau:
- added reserved fields in headers, miniblocks, miniblocks headers, p2p message
- added temporary protections on these fields + unit tests
- removed unwanted log message for account is already staked


data/block/proto/block.proto
```diff
@@ -71,6 +73,7 @@ message Header  {
 	bytes                    SoftwareVersion        = 21;
 	bytes                    AccumulatedFees        = 22 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
 	bytes                    DeveloperFees          = 23 [(gogoproto.casttypewith) = "math/big.Int;github.com/ElrondNetwork/elrond-go/data.BigIntCaster"];
+    bytes                    Reserved               = 24;
```
bogdan-rosianu, line 76: format maybe ?

iulianpascalau: done

PR #2163, opened 2020-07-27
Add num checkpoints to node stats

BeniaminDrasovean:
Add the number of state checkpoints to the /node/statistics response.


data/state/accountsDB.go
```diff
@@ -26,6 +29,9 @@ type AccountsDB struct {
 	dataTries    TriesHolder
 	entries      []JournalEntry
 	mutOp        sync.RWMutex
+
+	numCheckpoints      uint
```
iulianpascalau, line 33: could have used uint32 and for increment function, `atomic.AddUint32()`

data/state/accountsDB.go
```diff
@@ -856,9 +865,19 @@ func (adb *AccountsDB) SetStateCheckpoint(rootHash []byte) {
 		adb.mainTrie.SetCheckpoint(rootHash)
 		adb.snapshotUserAccountDataTrie(rootHash)
 		adb.mainTrie.ExitSnapshotMode()
+
+		adb.increaseNumCheckpoints()
 	}()
 }
 
+func (adb *AccountsDB) increaseNumCheckpoints() {
+	time.Sleep(batchWriteDelay)
```
iulianpascalau, line 874: we depend upon a variable defined in config that can be easily changed without affecting this time.Sleep

data/state/accountsDB.go
```diff
@@ -885,6 +904,14 @@ func (adb *AccountsDB) GetAllLeaves(rootHash []byte) (map[string][]byte, error)
 	return allAccounts, nil
 }
 
+// GetNumCheckpoints returns the total number of state checkpoints
+func (adb *AccountsDB) GetNumCheckpoints() uint {
+	adb.numCheckpointsMutex.RLock()
+	defer adb.numCheckpointsMutex.RUnlock()
+
+	return adb.numCheckpoints
```
iulianpascalau, line 912: atomic.LoadUint32

data/state/accountsDB.go
BeniaminDrasovean: Done.

data/state/accountsDB.go
BeniaminDrasovean: Refactored.

data/state/accountsDB.go
BeniaminDrasovean: Done.

api/node/routes.go
```diff
@@ -219,6 +225,9 @@ func statsFromTpsBenchmark(tpsBenchmark *statistics.TpsBenchmark) statisticsResp
 		}
 	}
 
+	sr.NumAccountStateCheckpoints = facade.GetNumCheckpointsFromAccountState()
```
bogdan-rosianu, line 228: unit test for checking the value?

data/state/accountsDB.go
```diff
@@ -856,9 +863,17 @@ func (adb *AccountsDB) SetStateCheckpoint(rootHash []byte) {
 		adb.mainTrie.SetCheckpoint(rootHash)
 		adb.snapshotUserAccountDataTrie(rootHash)
 		adb.mainTrie.ExitSnapshotMode()
+
+		adb.increaseNumCheckpoints()
 	}()
 }
 
+func (adb *AccountsDB) increaseNumCheckpoints() {
+	time.Sleep(time.Duration(adb.mainTrie.GetSnapshotDbBatchDelay()) * time.Second)
```
bogdan-rosianu, line 872: this metric isn't persistent. Do we need it to be?

bogdan-rosianu: also these metrics might be useful in `/node/status` I guess

api/node/routes.go
BeniaminDrasovean: Done.

data/state/accountsDB.go
BeniaminDrasovean: Yes, will make it persistent. I think that `/node/statistics` is enough for now.

LucianMincu: I think `/node/status` is the right place for this one  🙏

BeniaminDrasovean: Refactored.

data/state/accountsDB.go
```diff
@@ -14,7 +16,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/marshal"
 )
 
-const batchWriteDelay = time.Second * 2
+var numCheckpointsKey = []byte("state checkpoint")
```
bogdan-rosianu, line 19: you can move it to core/constants.go and make it exported and use it everywhere because there are more places / packages where you use this

data/state/accountsDB.go
```diff
@@ -63,10 +67,25 @@ func NewAccountsDB(
...
 	}, nil
 }
 
+func getNumCheckpoints(trie data.Trie) uint32 {
+	val, err := trie.Database().Get(numCheckpointsKey)
+	if err != nil {
+		return 0
+	}
+
+	numCheckpoints, err := strconv.Atoi(string(val))
```
iulianpascalau, line 80: why not use binary.BigEndian.Uint32() and PutUint32 ? When decoding to uint32, should add a len check as to not have less than 4 bytes as it will panic.

api/node/routes.go
```diff
@@ -27,15 +27,21 @@ const (
 	peerInfoPath        = "/peerinfo"
 )
 
+// AccStateCheckpointsKey is used as a key for the number of account state checkpoints in the api response
+const AccStateCheckpointsKey = "num_accounts_state_checkpoints"
```
bogdan-rosianu, line 31: why exported?

api/node/routes.go
```diff
@@ -168,6 +172,8 @@ func StatusMetrics(c *gin.Context) {
 	}
 
 	details := facade.StatusMetrics().StatusMetricsMapWithoutP2P()
+	details[AccStateCheckpointsKey] = facade.GetNumCheckpointsFromAccountState()
```
bogdan-rosianu, line 175: we can think of a different approach because all the metrics excluding these two are fetched from StatusMetrics. Can remain like this for now

data/state/accountsDB.go
BeniaminDrasovean: Actually, this is the only place where it is/will be used, because here is where you save it to db/recover it from db.

data/state/accountsDB.go
BeniaminDrasovean: Done.

api/node/routes.go
BeniaminDrasovean: To be able to test it more easily.

api/node/routes.go
BeniaminDrasovean: Ok.

api/node/routes.go
bogdan-rosianu: at least change it so it will has the `erd_` prefix like the rest of metrics

PR #2166, opened 2020-07-27
Fix tx verification and whitelisting

andreibancioiu:
- Fixed whitelisting for relayed transactions
 - Fixed upper gas limit for transactions


process/transaction/interceptedTransaction.go
```diff
@@ -155,6 +155,8 @@ func (inTx *InterceptedTransaction) CheckValidity() error {
 		if err != nil {
 			return err
 		}
+
+		inTx.whiteListerVerifiedTxs.Add([][]byte{inTx.Hash()})
```
sasurobert, line 159: please add a unit test.

andreibancioiu: Added.

PR #2177, opened 2020-07-28
Fix passing caller to SC upgrade guard

andreibancioiu:
- A cross shard upgrade transaction would have resulted in a `nil` `senderAccount` - failure to get its address
 - Changed to use `vmInput.CallerAddr`


process/smartContract/process.go
```diff
@@ -1289,15 +1289,15 @@ func (sc *scProcessor) processSimpleSCR(
...
+func (sc *scProcessor) checkUpgradePermission(contract state.UserAccountHandler, vmInput *vmcommon.ContractCallInput) error {
 	isUpgradeCalled := vmInput.Function == upgradeFunctionName
 	if !isUpgradeCalled {
 		return nil
 	}
 
 	codeMetadata := vmcommon.CodeMetadataFromBytes(contract.GetCodeMetadata())
 	isUpgradeable := codeMetadata.Upgradeable
-	callerAddress := caller.AddressBytes()
+	callerAddress := vmInput.CallerAddr
```
sasurobert, line 1300: make a unit test for checkUpgradePermission

PR #2182, opened 2020-07-29
fix/all wrong tx consumes gas

sasurobert:
fix transaction processing when a user would send value and no data to the smart contract deployment address (000..000). In that case the transaction would be only deleted from pool and not executed. Changed to treat as failed transaction, thus consuming the provided gas.

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -12,13 +12,13 @@
    StartInEpochEnabled = true
 
    # DisableDeploy represents the flag to enable/disable deployment of smart contracts
-   DisableDeploy = true
+   DisableDeploy = false
 
    # DisableBuiltInFunctions represents the flag to enable/disable built in functions calls
-   DisableBuiltInFunctions = true
+   DisableBuiltInFunctions = false
 
    # DisableRelayedTx represents the flag to enable/disable relayed transaction standard
-   DisableRelayedTx = true
+   DisableRelayedTx = false
 
    # GenesisString represents the encoded string for the genesis block
    GenesisString = "67656E65736973"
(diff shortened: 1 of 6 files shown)


cmd/node/config/config.toml
```diff
@@ -12,13 +12,13 @@
    StartInEpochEnabled = true
 
    # DisableDeploy represents the flag to enable/disable deployment of smart contracts
-   DisableDeploy = false
```
SebastianMarian: Restore all the modified flags in this file

cmd/node/config/systemSmartContractsConfig.toml
```diff
@@ -4,7 +4,7 @@
     UnBondPeriod = 40
     MinStepValue = "100000000000000000000"
     AuctionEnableNonce = 100000000000
-    StakeEnableNonce = 0
```
SebastianMarian: Restore all the modified flags in this file

process/coordinator/transactionType_test.go
```diff
@@ -154,6 +154,26 @@ func TestTxTypeHandler_ComputeTransactionTypeScDeployment(t *testing.T) {
 	assert.Equal(t, process.SCDeployment, txType)
 }
 
+func TestTxTypeHandler_ComputeTransactionTypeRecv0AddressWrongTransaction(t *testing.T) {
```
SebastianMarian, line 157: Recv0Address ? Is the test name ok?

sasurobert: it is receiver 0 address. the special address for deployment.

cmd/node/config/config.toml
sasurobert: done

cmd/node/config/systemSmartContractsConfig.toml
sasurobert: false

process/coordinator/transactionType_test.go
SebastianMarian: ok

PR #2185, opened 2020-07-29
bugFix/relayed invalid nonce

sasurobert:
Fix transactions when invalid relayed transactions is processed. Needed to add prevHash for the created failed smart contract result. Tested manually relayed transactions with invalid user nonce.


process/transaction/shardProcess.go
```diff
@@ -526,13 +532,43 @@ func (txProc *txProcessor) processRelayedTx(
...
 
-func (txProc *txProcessor) computeRelayedTxFees(tx *transaction.Transaction) (*big.Int, *big.Int) {
+func (txProc *txProcessor) computeRelayedTxFees(tx *transaction.Transaction) (*big.Int, *big.Int, uint64) {
 	relayerGasLimit := txProc.economicsFee.ComputeGasLimit(tx)
 	relayerFee := big.NewInt(0).Mul(big.NewInt(0).SetUint64(relayerGasLimit), big.NewInt(0).SetUint64(tx.GetGasPrice()))
 	totalFee := big.NewInt(0).Mul(big.NewInt(0).SetUint64(tx.GetGasLimit()), big.NewInt(0).SetUint64(tx.GetGasPrice()))
 	remainingFee := big.NewInt(0).Sub(totalFee, relayerFee)
 
-	return totalFee, remainingFee
+	return totalFee, remainingFee, tx.GasLimit - relayerGasLimit
```
SebastianMarian, line 541: Could be here an overflow if tx.GasLimit - relayerGasLimit < 0  ? It is better to use int64 and check it outside

sasurobert: no need as it is checked before in checkTxValidity

SebastianMarian: ok

cmd/node/config/economics.toml
```diff
@@ -30,3 +30,4 @@
     MinGasLimit = "50000"
     GasPerDataByte = "1500"
     DataLimitForBaseCalc = "10000"
+    MaxOverPayGasInPercentage = 1.0 #fraction of value 1 - 100%
```
iulianpascalau, line 33: We have to standardize all these "percentage" stuff used in the whole project.

process/transaction/shardProcess.go
```diff
@@ -526,13 +532,43 @@ func (txProc *txProcessor) processRelayedTx(
...
+func (txProc *txProcessor) removeValueAndConsumedFeeFromUser(
+	userTx *transaction.Transaction,
+	relayedTxValue *big.Int,
+) error {
+	userAcnt, err := txProc.getAccountFromAddress(userTx.SndAddr)
+	if err != nil {
+		return err
+	}
+	if check.IfNil(userAcnt) {
+		return process.ErrNilAccountsAdapter
```
iulianpascalau, line 553: ErrNilAccount

cmd/node/config/economics.toml
sasurobert: deleted. float64 is safe and easy.

process/transaction/shardProcess.go
sasurobert: done

process/transaction/shardProcess.go
SebastianMarian: ok

sasurobert:
> Can you please add for each manual testing scenario the following:
> 
> a) Prerequisites
> b) Test steps
> c) Expected results
> d) Actual results
> 
> If there are parameters involved in testing what are the parameters boundaries and the expected results and actual results for testing inside boundaries, on boundaries and outside boundaries.

The test scenario was explained in the pull request header. The problem was treating improperly the failed relayed transaction. The created smart contract result did not have a previous hash, thus it was rejected by interceptor, the shard got stuck. 

The manual testing: make a relayed transaction where the user's tx NONCE is invalid. The transaction has to be processed as failed, smart contract result should be created correctly, shard is not stuck anymore. All the gas is consumed from relayer and from user as well. Nonces for both relayer and user are increased.

PR #2187, opened 2020-07-29
Protection for undeployed delegation

iulianpascalau:
- added protection against delegation to undeployed SC in the genesis process


iulianpascalau:
> Can you please add for each manual testing scenario the following:
> 
> a) Prerequisites
> b) Test steps
> c) Expected results
> d) Actual results
> 
> If there are parameters involved in testing what are the parameters boundaries and the expected results and actual results for testing inside boundaries, on boundaries and outside boundaries.

This has been manually tested by using wrongly constructed .json files and proved to be working as expected.

PR #2191, opened 2020-07-31
Fix wrong genesis block save

iulianpascalau:
- removed a wrong save of the genesis block in hash-nonce storage

Manual testing procedure:
- Start a testnet with nodes that do shuffling between shards. 
- Without the fix, we are able to see an error message containing the string `shard header (nonce-hash)` after a few epochs.
- With the fix, the error message will no longer be logged.


cmd/node/factory/structs.go
```diff
@@ -577,11 +577,6 @@ func prepareGenesisBlock(args *processComponentsFactoryArgs, genesisBlocks map[u
 		if errNotCritical != nil {
 			log.Error("error storing genesis shardblock", "error", errNotCritical.Error())
 		}
-		hdrNonceHashDataUnit := dataRetriever.ShardHdrNonceHashDataUnit + dataRetriever.UnitType(genesisBlock.GetShardID())
```
SebastianMarian, line 580: Actually this storage exists in each shard but only for self shard. In metachain it is created for every shard.

iulianpascalau: But we do not need the shard genesis block stored inside the nonce-hash unit

SebastianMarian: Ok, in this case you can move line 565 inside first if after line 570

iulianpascalau: done

PR #2195, opened 2020-08-01
Update systemSmartContractsConfig.toml

pislaru:
Has someone forgot about this?

### DIFF
--- a/cmd/node/config/systemSmartContractsConfig.toml
+++ b/cmd/node/config/systemSmartContractsConfig.toml
@@ -11,7 +11,7 @@
     MaxNumberOfNodesForStake = 28
     NodesToSelectInAuction = 28
     UnJailValue = "2500000000000000000" #0.1% of genesis node price
-    ActivateBLSPubKeyMessageVerification = false #TODO set to true before mainnet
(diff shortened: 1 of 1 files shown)


cmd/node/config/systemSmartContractsConfig.toml
```diff
@@ -11,7 +11,7 @@
     MaxNumberOfNodesForStake = 28
     NodesToSelectInAuction = 28
     UnJailValue = "2500000000000000000" #0.1% of genesis node price
-    ActivateBLSPubKeyMessageVerification = false #TODO set to true before mainnet
+    ActivateBLSPubKeyMessageVerification = true
```
SebastianMarian, line 14: It will be activated when functionality pf staking (system smart contracts) will be also activated. For now this flag has no importance as staking is also disabled. Anyway, thanks for this reminder.

raduchis:
The configs in this repository are not the ones used in the mainnet. These are the "development" configs.

The actual repository for the configs that is used in the mainnet is 
https://github.com/ElrondNetwork/elrond-config-mainnet/

Once we activate the stake and unstake on the mainnet we are going to change this. Thanks!

PR #2200, opened 2020-08-04
Penalize users who put too much gas

SebastianMarian:
* Fixed a potentially attack when a smart contract call transaction would be sent

 with an intentionally higher gas limit, just to prevent including other transactions
 in one cross miniblock. Now, if gas limit set in one SC tx exceeds the maximum
 higher factor of gas fee compared with the real gas used, all the gas will be
 consumed and nothing will be refunded to the sender

Testing steps:

a) Prerequisites

 Txgen should be able to generate valid system SC txs with different gas limit set
 inside. If this is not possible, then some system SC calls should be manually generated with intentionally higher gas limit set inside (>10x as the normal gas consumption)

b) Test steps

 Start a normal testnet, and use any SC scenario with txgen, system SC for example:
 stake, unstake, unbond, claim, etc., but after a while these txs should be sent with
 higher value set for the gas limit (>10x as the normal gas consumption).

c) Expected results

 We should have all the systems SC calls executed as expected, but for all of them
 which have been set to use a higher gas limit inside (>10x as the normal gas consumption) no gas should be refunded. (all the gas provided would be used)

d) Actual results

 Test process is in pending

### DIFF
--- a/integrationTests/multiShard/relayedTx/relayedTx_test.go
+++ b/integrationTests/multiShard/relayedTx/relayedTx_test.go
@@ -191,7 +191,7 @@ func TestRelayedTransactionInMultiShardEnvironmentWithESDTTX(t *testing.T) {
 	initalSupply := big.NewInt(10000000000)
 	tokenIssuer := nodes[0]
 	txData := "issue" + "@" + hex.EncodeToString([]byte(tokenName)) + "@" + hex.EncodeToString(initalSupply.Bytes())
-	integrationTests.CreateAndSendTransaction(tokenIssuer, issuePrice, factory.ESDTSCAddress, txData)
+	integrationTests.CreateAndSendTransaction(tokenIssuer, issuePrice, factory.ESDTSCAddress, txData, integrationTests.AdditionalGasLimit)
 
 	time.Sleep(time.Second)
 	nrRoundsToPropagateMultiShard := int64(10)
@@ -208,7 +208,7 @@ func TestRelayedTransactionInMultiShardEnvironmentWithESDTTX(t *testing.T) {
 	valueToTopUp := big.NewInt(100000000)
 	txData = core.BuiltInFunctionESDTTransfer + "@" + hex.EncodeToString([]byte(tokenName)) + "@" + hex.EncodeToString(valueToTopUp.Bytes())
 	for _, player := range players {
-		integrationTests.CreateAndSendTransaction(tokenIssuer, big.NewInt(0), player.Address, txData)
+		integrationTests.CreateAndSendTransaction(tokenIssuer, big.NewInt(0), player.Address, txData, integrationTests.AdditionalGasLimit)
 	}
 
 	time.Sleep(time.Second)
--- a/process/smartContract/process.go
+++ b/process/smartContract/process.go
@@ -750,6 +750,23 @@ func (sc *scProcessor) processVMOutput(
 		return nil, nil, process.ErrNilTransaction
 	}
 
+	if sc.isTooMuchGasPutInsideTx(vmOutput.GasRemaining, tx, acntSnd) {
+		log.Warn("scProcessor.processVMOutput: too much gas has been put inside tx",
+			"hash", txHash,
+			"nonce", tx.GetNonce(),
+			"value", tx.GetValue(),
+			"sender", tx.GetSndAddr(),
+			"receiver", tx.GetRcvAddr(),
+			"gas limit", tx.GetGasLimit(),
+			"gas price", tx.GetGasPrice(),
+			"gas remaining", vmOutput.GasRemaining,
+			"return code", vmOutput.ReturnCode.String(),
+			"return message", vmOutput.ReturnMessage,
+		)
+
+		vmOutput.GasRemaining = 0
+	}
+
 	scrForSender, consumedFee := sc.createSCRForSender(
 		vmOutput.GasRefund,
 		vmOutput.GasRemaining,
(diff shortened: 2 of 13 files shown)


process/smartContract/process.go
```diff
@@ -750,6 +750,23 @@ func (sc *scProcessor) processVMOutput(
 		return nil, nil, process.ErrNilTransaction
 	}
 
+	if sc.isTooMuchGasPutInsideTx(vmOutput.GasRemaining, tx, acntSnd) {
+		log.Warn("scProcessor.processVMOutput: too much gas has been put inside tx",
```
sasurobert, line 754: instead of log.Warn - which will fill the validators logs, I suggest to add a new ReturnMessage as well. Through that the user will get a notification in his wallet.

process/smartContract/process.go
```diff
@@ -819,6 +836,40 @@ func (sc *scProcessor) processVMOutput(
...
+	gasPut.Mul(big.NewInt(0).SetUint64(tx.GetGasLimit()), big.NewInt(0).SetUint64(tx.GetGasPrice()))
+
+	if check.IfNil(acntSnd) {
+		// cross shard move balance fee was already consumed at sender shard
+		moveBalanceCost := sc.economicsFee.ComputeFee(tx)
+		gasPut.Sub(gasPut, moveBalanceCost)
+	}
+
+	gasRefunded := big.NewInt(0)
+	gasRefunded.Mul(big.NewInt(0).SetUint64(gasRemaining), big.NewInt(0).SetUint64(tx.GetGasPrice()))
```
sasurobert, line 854: no need to multiplicate both provided gas and remaining gas with gasPrice. just make the difference between those 2. Plus instead of using tx as an input here, take vmInput - as it ca corrected GAS inside

process/smartContract/process.go
```diff
@@ -819,6 +836,40 @@ func (sc *scProcessor) processVMOutput(
...
+
+	gasUsed := big.NewInt(0)
+	gasUsed.Sub(gasPut, gasRefunded)
+
+	gasUsedMultiplied := big.NewInt(0)
+	gasUsedMultiplied.Mul(gasUsed, big.NewInt(0).SetUint64(process.MaxGasFeeHigherFactorAccepted))
+
+	isTooMuchGas := gasPut.Cmp(gasUsedMultiplied) > 0
+	if isTooMuchGas {
+		log.Warn("scProcessor.isTooMuchGasPutInsideTx",
```
sasurobert, line 864: WARN is 2 times already - no need.

process/smartContract/process.go
SebastianMarian: Done

process/smartContract/process.go
SebastianMarian: Done

process/smartContract/process.go
SebastianMarian: Done

process/smartContract/process.go
```diff
@@ -750,6 +751,27 @@ func (sc *scProcessor) processVMOutput(
...
+			"gas limit", tx.GetGasLimit(),
+			"gas price", tx.GetGasPrice(),
+			"gas provided", gasProvided,
+			"gas remained", vmOutput.GasRemaining,
+			"gas used", gasProvided-vmOutput.GasRemaining,
+			"return code", vmOutput.ReturnCode.String(),
+			"return message", vmOutput.ReturnMessage,
+		)
+
+		vmOutput.ReturnMessage = fmt.Sprintf("too much gas has been provided: gas used = %d, gas remained = %d",
```
sasurobert, line 770: do not rewrite return message - just add to the existing return message

process/smartContract/process.go
```diff
@@ -750,6 +751,27 @@ func (sc *scProcessor) processVMOutput(
 		return nil, nil, process.ErrNilTransaction
 	}
 
+	if isTooMuchGasProvided(vmOutput.GasRemaining, gasProvided) {
```
iulianpascalau, line 754: Maybe reuse an integration test for this scenario?

process/smartContract/process.go
SebastianMarian: done

process/smartContract/process.go
SebastianMarian: will do after system tests will pass

integrationTests/multiShard/relayedTx/relayedTx_test.go
```diff
@@ -3,12 +3,13 @@ package relayedTx
 import (
 	"encoding/hex"
 	"fmt"
-	"github.com/ElrondNetwork/elrond-go/process/smartContract/hooks"
-	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 	"math/big"
 	"testing"
 	"time"
 
+	"github.com/ElrondNetwork/elrond-go/process/smartContract/hooks"
```
sasurobert, line 10: goimports

process/smartContract/process.go
```diff
@@ -416,6 +416,27 @@ func (sc *scProcessor) resolveBuiltInFunctions(
 		vmOutput = &vmcommon.VMOutput{ReturnCode: vmcommon.UserError, ReturnMessage: err.Error()}
 	}
 
+	if vmInput.CallType != vmcommon.AsynchronousCall && isTooMuchGasProvided(vmInput.GasProvided, vmOutput.GasRemaining) {
```
sasurobert, line 419: get this if and log trace and output correction in a separate function

process/smartContract/process.go
```diff
@@ -758,6 +775,27 @@ func (sc *scProcessor) processVMOutput(
 		return nil, nil, process.ErrNilTransaction
 	}
 
+	if callType != vmcommon.AsynchronousCall && isTooMuchGasProvided(gasProvided, vmOutput.GasRemaining) {
```
sasurobert, line 778: duplicated code

process/transaction/shardProcess.go
```diff
@@ -275,7 +275,12 @@ func (txProc *txProcessor) executingFailedTransaction(
 	return process.ErrFailedTransaction
 }
 
-func (txProc *txProcessor) createReceiptWithReturnedGas(txHash []byte, tx *transaction.Transaction, acntSnd state.UserAccountHandler) error {
+func (txProc *txProcessor) createReceiptWithReturnedGas(
```
sasurobert, line 278: this will have some conflicts with my PR.

sasurobert (changes requested):

process/smartContract/process.go
iulianpascalau: Might be a separated function for this duplicated code

process/smartContract/process.go
iulianpascalau: Ok

process/smartContract/process.go
SebastianMarian: done

process/smartContract/process.go
SebastianMarian: done

process/transaction/shardProcess.go
SebastianMarian: will solve them later

integrationTests/multiShard/relayedTx/relayedTx_test.go
SebastianMarian: done

iulianpascalau (approved):

sasurobert (approved):

iulianpascalau (approved):

sasurobert (approved):

PR #2201, opened 2020-08-05
integrates the output flag for filegen

BenWhiteJam:
integrates the output directory flag for filegen config pre-processing


.github/workflows/deployment.yml
```diff
@@ -98,7 +98,8 @@ jobs:
             -num-of-observers-in-metachain 1 \
             -chain-id ${{ github.sha }} \
             -hysteresis $HYSTERESIS \
-            -stake-type delegated
+            -stake-type delegated \
+            -output-directory ./
```
bogdan-rosianu, line 102: why not a new directory? like `./output`

BenWhiteJam: as it creates the according files on the deployment server path which then get picked up by the config file sed and the kube config map creation step.
Would push the update to a separate folder into a more general deployment file update branch to parse files more dynamic in be less error prone for future config file creations.

PR #2203, opened 2020-08-05
En 7295 get username route

bogdan-rosianu:
Added an API endpoint for getting the username for a given address.

Testing steps:
a) Prerequisites
-a testnet which has the following routes enabled in `api.toml`:
    - `address/:address/username`
    - `address/:address`
b) Test steps
-make API requests to the routes
c) Expected results
-on the `address/:address/username` route only the username should be returned
-the username should be visible also on the route `address/:address`

-if the given address exists and it has an username set, it should be returned.
-if the given address does not exist an error will be returned.
-if the given address exists, but it doesn't have an username set, an empty username will be returned.

d) Actual results
-tested on a local testnet and the acceptance criteria is met


node/node.go
```diff
@@ -386,29 +386,33 @@ func (n *Node) addCloserInstances(closers ...update.Closer) error {
 
 // GetBalance gets the balance for a specific address
 func (n *Node) GetBalance(address string) (*big.Int, error) {
-	if check.IfNil(n.addressPubkeyConverter) || check.IfNil(n.accounts) {
-		return nil, errors.New("initialize AccountsAdapter and PubkeyConverter first")
-	}
-
-	addr, err := n.addressPubkeyConverter.Decode(address)
+	account, err := n.getAccountHandler(address)
```
iulianpascalau, line 389: GJ! Same level of abstraction

PR #2206, opened 2020-08-06
Tweak config params

iulianpascalau:
- tweaked config.toml values for a better sync process
- removed a bootstorer bug that prevented the node to start correctly from storage if it had incomplete information
- added histogram prints in p2p regarding the peers split on shards 

Manual testing procedures:
- since the new config values accentuates the bootstorer bug, all we need to test is the normal manual testing of this implementation. The problem arrived on random nodes in low value epochs (after a maximum of 4 epochs we had stuck nodes), so testing on the order of tens of epoch should suffice.


cmd/node/config/systemSmartContractsConfig.toml
```diff
@@ -8,8 +8,8 @@
     NumRoundsWithoutBleed = 100
     MaximumPercentageToBleed = 0.5
     BleedPercentagePerRound = 0.00001
-    MaxNumberOfNodesForStake = 28
-    NodesToSelectInAuction = 28
```
sasurobert, line 12: Moved if back to 28

iulianpascalau: @LucianMincu needs to test this in "as close as it can get to mainnet" scenario

iulianpascalau:
> We should test this PR in two different edge cases for which these values have been set to 1. (the most important thing is to keep at least the BatchDelaySeconds = 2). So, the two edge cases are:
> 
> First is when all the nodes (from one shard or all shards) are stopped in almost the same time but before they had the chance to put the batch in the storage and the cacher would not be flushed in. Then the nodes could start behind final and if they would have a majority, and in this case they would certainly have, they would propose another block which is already finalized and broadcasted to meta/shards.
> 
> The second problem is with MetaBlockStorage used by shards when they finalize all the miniblocks from one metablock. Then, and only then, they will put the meta in the storage. If they are roll back or restart from storage, they will load the previous/last info from storage and based on which meta blocks are found in storage they can correctly set which metablocks and miniblocks are finalized and which are not yet. If this info is not found in storage, but only in batch, and getting stuff from batch is not working (now as far as I remember this should be fixed) or the node is stopped unexpectedly and the cacher is not flushed into the persister the shard node will have a problem to start and to sync, as it will not set correctly the already processed miniblocks from finalized metablocks.

Totally agree. The BatchDelaySeconds will be kept to 2 seconds for above mentioned reasons.

process/block/baseProcess.go
```diff
@@ -794,6 +794,15 @@ func (bp *baseProcessor) prepareDataForBootStorer(args bootStorerDataArgs) {
 		EpochStartTriggerConfigKey: args.epochStartTriggerConfigKey,
 	}
 
+	log.Debug("baseProcessor.prepareDataForBootStorer",
```
sasurobert, line 797: this is called in every commit block - is it not too much data for log.Debug ?

process/block/baseProcess.go
```diff
@@ -808,6 +817,21 @@ func (bp *baseProcessor) prepareDataForBootStorer(args bootStorerDataArgs) {
 	}
 }
 
+func (bp *baseProcessor) displayBootstrapHeaders(hdrs []bootstrapStorage.BootstrapHeaderInfo) string {
+	str := "["
+	for _, h := range hdrs {
+		str += bp.displayBoostrapHeaderInfo(h)
+	}
+
```
sasurobert, line 825: delete empty line

process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -161,6 +172,21 @@ func (st *storageBootstrapper) loadBlocks() error {
 	return nil
 }
 
+func (st *storageBootstrapper) displayBootstrapHeaders(hdrs []bootstrapStorage.BootstrapHeaderInfo) string {
+	str := "["
+	for _, h := range hdrs {
+		str += st.displayBoostrapHeaderInfo(h)
+	}
+
```
sasurobert, line 180: delete empty line

process/block/baseProcess.go
iulianpascalau: yeah, was used in debugging

iulianpascalau: removed

process/block/baseProcess.go
iulianpascalau: removed function -> no longer necessary

process/sync/storageBootstrap/baseStorageBootstrapper.go
iulianpascalau: removed function -> no longer necessary

PR #2215, opened 2020-08-10
EN 7229 software version check against epoch

iulianpascalau:
- made software version checks dependent upon the current epoch so the nodes will process everything in a consistent way


cmd/node/factory/interface.go
```diff
@@ -14,9 +14,10 @@ type HeaderSigVerifierHandler interface {
 	IsInterfaceNil() bool
 }
 
-// HeaderIntegrityVerifierHandler is the interface needed to check that a header's integrity is correct
-type HeaderIntegrityVerifierHandler interface {
+// HeaderVersioningHandler encapsulates methods useful to check that a header's versioning is correct
+type HeaderVersioningHandler interface {
```
sasurobert, line 18: it still has the integration verification methods :) the interface name could have remained the same.

process/headerCheck/headerVersioningHandler.go
```diff
@@ -0,0 +1,167 @@
...
+	if ver == wildcard {
+		return hvh.defaultVersion
+	}
+
+	return ver
+}
+
+// GetVersion returns the version by providing the epoch
+func (hvh *headerVersioningHandler) getMatchingVersion(epoch uint32) string {
+	//TODO add a small cache here?
```
sasurobert, line 91: this TODO does not say enough

iulianpascalau: The question was if we should add a small sized cache that will improve the responsiveness of this function since this will be called for each and every header received. Since epochs change with a low frequency, caching will remove the iteration of the provided config array.

iulianpascalau: And yes, I want to remove the TODO before merging this PR :)

cmd/node/factory/interface.go
iulianpascalau: It also provides the correct version now. That is why I have renamed it to handler as it manages the software versions for headers. :)

process/headerCheck/headerVersioningHandler.go
iulianpascalau: removed TODO, added cache

sasurobert: +1

process/headerCheck/headerVersioningHandler.go
```diff
@@ -88,20 +98,48 @@ func (hvh *headerVersioningHandler) GetVersion(epoch uint32) string {
...
 		"epoch", epoch, "default version", hvh.defaultVersion)
 
+	hvh.setInCache(epoch, hvh.defaultVersion)
+
 	return hvh.defaultVersion
 }
 
+func (hvh *headerVersioningHandler) getFromCache(epoch uint32) (string, bool) {
+	key := make([]byte, keySize)
+	binary.BigEndian.PutUint32(key, epoch)
```
sasurobert, line 124: it would be much optimal to use epoch % 256 ?

iulianpascalau: Since the byte slice conversion is quite cheap and we already have a small cache, I wouldn't venture to make this modulo operation since it has the same output for the epochs separated by multiple of 256, leading to wrong value outputs. I want to keep this as straight forward as it can get.

process/headerCheck/headerVersioningHandler.go
```diff
@@ -0,0 +1,205 @@
...
+		return versionsByEpochs[i].StartEpoch < versionsByEpochs[j].StartEpoch
+	})
+
+	currentEpoch := uint32(0)
+	for _, ver := range versionsByEpochs {
+		if ver.StartEpoch >= ver.EndEpoch {
+			return nil, fmt.Errorf("%w, StartEpoch is greater or equal to EndEpoch, version %s",
+				ErrInvalidVersionOnEpochValues, ver.Version)
+		}
+		if currentEpoch != ver.StartEpoch {
```
raduchis, line 74: If the previous EndEpoch must be equal to the current StartEpoch, which one contains the border epoch? 

StartEpoch=1, EndEpoch=10, Version=1
StartEpoch=10, EndEpoch=20, Version=2

Epoch 10 has Version=1?

process/headerCheck/headerVersioningHandler.go
```diff
@@ -0,0 +1,205 @@
...
+			return nil, fmt.Errorf("%w for version %s",
+				ErrInvalidVersionFoundGaps, ver.Version)
+		}
+		if len(ver.Version) > core.MaxSoftwareVersionLengthInBytes {
```
raduchis, line 78: The version should be also test for ascending I think

process/headerCheck/headerVersioningHandler.go
```diff
@@ -0,0 +1,205 @@
...
+
+	for _, ver := range hvh.versions {
+		if ver.StartEpoch <= epoch && epoch < ver.EndEpoch {
+			hvh.setInCache(epoch, ver.Version)
+
+			return ver.Version
+		}
+	}
+
+	log.Debug("headerVersioningHandler.GetVersion version not found",
```
raduchis, line 114: This is the case when the currentEpoch is larger than the last EndEpoch? Or are there any other cases?

process/headerCheck/headerVersioningHandler.go
```diff
@@ -0,0 +1,205 @@
...
+
+// checkSoftwareVersion returns nil if the software version has the correct length
+func (hvh *headerVersioningHandler) checkSoftwareVersion(hdr data.HeaderHandler) error {
+	err := hvh.checkVersionLength(hdr.GetSoftwareVersion())
+	if err != nil {
+		return err
+	}
+
+	version := hvh.getMatchingVersion(hdr.GetEpoch())
+	if version == wildcard {
```
raduchis, line 173: This shouldn't accept anything, it should accept the default version only

cmd/node/factory/interface.go
raduchis: And the chain Id. Could have remained like before

cmd/node/config/config.toml
```diff
@@ -23,6 +23,16 @@
    # GenesisString represents the encoded string for the genesis block
    GenesisString = "67656E65736973"
 
+[Versions]
+   DefaultVersion = "default"
+   VersionsByEpochs = [
+       { StartEpoch = 0, EndEpoch = 1000000, Version = "*" },
```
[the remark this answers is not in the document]
iulianpascalau: Done

cmd/node/factory/structs.go
```diff
@@ -101,7 +101,7 @@ type Process struct {
 	BlackListHandler         process.TimeCacher
 	BootStorer               process.BootStorer
 	HeaderSigVerifier        HeaderSigVerifierHandler
-	HeaderIntegrityVerifier  HeaderIntegrityVerifierHandler
+	HeaderVersionging        HeaderVersioningHandler
```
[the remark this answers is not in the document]
iulianpascalau: Fixed

process/headerCheck/headerVersioningHandler.go
iulianpascalau: Refactored. To be explicit now we have the following scenario. Suppose we have the following config:
```
StartEpoch = 0, Version = "v1"
StartEpoch = 1, Version = "v2"
StartEpoch = 5, Version = "v3"
```
For epoch 0, we'll have "v1", for epoch 1 up to 4 (inclusive), we'll have "v2" and from epoch 5 until max Uint32 we'll have "v3"

process/headerCheck/headerVersioningHandler.go
iulianpascalau: It's not clear how we will encode the version at this step. Nor it is required to know about how it will look.

process/headerCheck/headerVersioningHandler.go
iulianpascalau: Refactored

process/headerCheck/headerVersioningHandler.go
iulianpascalau: we can make use of wildcards, if for example, we want to change the default string. We temporarily put the next versions on wildcard, change what we want (during this change *everything* will be accepted) and then put the desired string.

sasurobert (approved):

cmd/node/config/config.toml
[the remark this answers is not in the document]
raduchis: ok

cmd/node/factory/structs.go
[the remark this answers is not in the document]
raduchis: ok

process/headerCheck/headerVersioningHandler.go
raduchis: ok

process/headerCheck/headerVersioningHandler.go
raduchis: ok

process/headerCheck/headerVersioningHandler.go
raduchis: ok

process/headerCheck/headerVersioningHandler.go
raduchis: ok

raduchis (approved):

raduchis (approved):

sasurobert (approved):

PR #2223, opened 2020-08-13
En 7316 add call value sc query

bogdan-rosianu:
Added two optional parameters to API endpoint `vm-values` (`caller` and `value`) and extended the functionality in order to be able to include the the `callerAddr` and `callValue` inside the SC query service.

Testing steps:
1. start a testnet
2. make a POST request to API endpoint `/vm-values/int` for example in which you will specify the caller address and the call value. Can be a `stake` transaction.
3. you should see the vm output in the response


facade/interface.go
```diff
@@ -69,6 +69,7 @@ type NodeHandler interface {
 // ApiResolver defines a structure capable of resolving REST API requests
 type ApiResolver interface {
 	ExecuteSCQuery(query *process.SCQuery) (*vmcommon.VMOutput, error)
+	ExecuteQueryWithCallerAndValue(query *process.SCQuery, callerAddr []byte, callValue *big.Int) (*vmcommon.VMOutput, error)
```
sasurobert, line 72: change the SCQuery data - add callerAddr and callValue inside - no need to change the interface and add these values

process/smartContract/scQueryService.go
```diff
@@ -42,26 +44,57 @@ func NewSCQueryService(
 
 // ExecuteQuery returns the VMOutput resulted upon running the function on the smart contract
 func (service *SCQueryService) ExecuteQuery(query *process.SCQuery) (*vmcommon.VMOutput, error) {
-	if query.ScAddress == nil {
-		return nil, process.ErrNilScAddress
+	err := checkQueryFields(query)
```
sasurobert, line 47: No need for second function. Use ExecuteQuery and fill the value and callerAddress as they were before if those are empty.

facade/interface.go
bogdan-rosianu: done

process/smartContract/scQueryService.go
bogdan-rosianu: done

PR #2227, opened 2020-08-18
consensus: memory leak debug data

AdoAdoAdo:
memory leak is caused by accumulation of metablocks.
the references to metablocks are stored in an array in the delayed broadcast implementation, where they should get removed also when alarms  are canceled, not only when alarms expire. 
fix is to remove the reference as well on canceling the alarms.

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -453,8 +453,8 @@
    Type = "json"
 
 [EpochStartConfig]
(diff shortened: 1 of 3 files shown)


sasurobert (approved):

SebastianMarian (approved):

valentin-lup (approved):
Tested on AZ-UKS

sasurobert (approved):

PR #2228, opened 2020-08-19
Fix/compute fee only move balance

sasurobert:
consumed fee was wrongly executed in some cases, as of misinterpretation of one interface.


process/block/postprocess/feeHandler.go
```diff
@@ -65,11 +65,17 @@ func (f *feeHandler) ProcessTransactionFee(cost *big.Int, devFee *big.Int, txHas
...
-		devFee: big.NewInt(0).Set(devFee),
+	fee, ok := f.mapHashFee[string(txHash)]
+	if !ok {
+		fee = &feeData{
+			cost:   big.NewInt(0),
+			devFee: big.NewInt(0),
+		}
 	}
 
+	fee.cost.Add(fee.cost, cost)
```
iulianpascalau, line 76: :1st_place_medal:

iulianpascalau (approved):

process/transaction/baseProcess.go
```diff
@@ -137,7 +137,7 @@ func (txProc *baseTxProcessor) checkTxValues(
 		return process.ErrWrongTypeAssertion
 	}
 
-	txFee := txProc.economicsFee.ComputeFee(tx)
+	txFee := core.SafeMul(tx.GasLimit, tx.GasPrice)
```
SebastianMarian, line 140: So, now for this case the txFee will be calculated by directly multiply gasPrice with gasLimit which has been put inside tx and not by applying the below formula for gas limit, as was done before?

```
	gasLimit := ed.minGasLimit
	dataLen := uint64(len(tx.GetData()))
	gasLimit += dataLen * ed.gasPerDataByte

```

process/transaction/interceptedTransaction.go
```diff
@@ -318,7 +318,7 @@ func (inTx *InterceptedTransaction) SenderAddress() []byte {
 
 // Fee returns the estimated cost of the transaction
 func (inTx *InterceptedTransaction) Fee() *big.Int {
-	return inTx.feeHandler.ComputeFee(inTx.tx)
+	return core.SafeMul(inTx.tx.GasPrice, inTx.tx.GasLimit)
```
SebastianMarian, line 321: The same comment as above. Is this calculation OK for any case and any call of this method?

process/transaction/baseProcess.go
SebastianMarian: In this case line 150 could be replaced with cost.Add(cost, txFee) to avoid a multiply operation on big.Int which already has been done?

process/transaction/shardProcess.go
```diff
@@ -236,7 +236,7 @@ func (txProc *txProcessor) executingFailedTransaction(
 		return nil
 	}
 
-	txFee := txProc.economicsFee.ComputeFee(tx)
+	txFee := core.SafeMul(tx.GasPrice, tx.GasLimit)
```
SebastianMarian, line 239: So, when execution will fail we will take all the gas set? Could be a situation when GasLimit is set to 0 and we will not take anything? Is this check, for min gas limit, done on interceptor for every type of transaction?

process/block/postprocess/feeHandler.go
raduchis: could the same txHash appear 2 times? If so, RevertFees does the revert only 1 time. Line 97.

process/transaction/shardProcess.go
```diff
@@ -718,7 +718,12 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
 		return err
 	}
 
-	totalFee := txProc.economicsFee.ComputeFee(userTx)
+	totalFee := core.SafeMul(userTx.GetGasPrice(), userTx.GetGasLimit())
```
SebastianMarian, line 721: So, this will take all the gas set in userTx for in shard relayed txs? Could be a situation when GasLimit is set to 0 and we will not take anything, or worse in cross shard call if totalFee is lower than txProc.economicsFee.ComputeMoveBalanceFee(userTx), then in line 724 we will result in a negative value?

process/transaction/shardProcess.go
```diff
@@ -568,7 +568,7 @@ func (txProc *txProcessor) removeValueAndConsumedFeeFromUser(
 	if err != nil {
 		return err
 	}
-	consumedFee := txProc.economicsFee.ComputeFee(userTx)
+	consumedFee := core.SafeMul(userTx.GasLimit, userTx.GasPrice)
```
SebastianMarian, line 571: So, this will take all the gas set in userTx for relayed txs? Could be a situation when GasLimit is set to 0 and we will not take anything?

process/transaction/interceptedTransaction.go
sasurobert: now it is ok. until now it was wrong.

process/transaction/shardProcess.go
sasurobert: There is no scenario when gasLimit is set to 0 - it will not pass interceptor.

process/transaction/shardProcess.go
sasurobert: No. there is no way to have that. This is the correct way, until now that computation did only movebalance transfer cost which was always smaller than equal to the actual one.

process/transaction/shardProcess.go
sasurobert: No. there is no way to have that. This is the correct way, until now that computation did only movebalance transfer cost which was always smaller than equal to the actual one.

process/transaction/baseProcess.go
sasurobert: done

process/smartContract/process.go
```diff
@@ -927,30 +914,37 @@ func (sc *scProcessor) createSmartContractResult(
...
 }
 
+func (sc *scProcessor) computeActualConsumedGas(
+	tx data.TransactionHandler,
+	vmOutput *vmcommon.VMOutput,
+	acntSnd state.UserAccountHandler,
+) uint64 {
+	consumedGas := tx.GetGasLimit() - vmOutput.GasRemaining
+	if check.IfNil(acntSnd) {
+		consumedGas = consumedGas - sc.economicsFee.ComputeGasLimit(tx)
```
SebastianMarian, line 924: += instead -= ?

iulianpascalau, line 924:
```GO
consumedGas -= sc.economicsFee.ComputeGasLimit(tx)
```

as to be consistent with the rest of the function

process/smartContract/process.go
```diff
@@ -927,30 +914,37 @@ func (sc *scProcessor) createSmartContractResult(
...
+	}
+	for _, outAcc := range vmOutput.OutputAccounts {
+		consumedGas -= outAcc.GasLimit
```
SebastianMarian, line 927: += instead -= ?

process/transaction/interceptedTransaction.go
SebastianMarian: ok

process/transaction/shardProcess.go
SebastianMarian: ok

process/transaction/shardProcess.go
SebastianMarian: ok

process/transaction/shardProcess.go
SebastianMarian: ok

process/smartContract/process.go
sasurobert: it is -=

process/smartContract/process.go
sasurobert: it is -=

process/smartContract/process.go
SebastianMarian: So it refers to what is consumed in destination shard only?

process/smartContract/process.go
SebastianMarian: ok

SebastianMarian (approved):

process/smartContract/process.go
sasurobert: done

sasurobert: yes

iulianpascalau (approved):

valentin-lup (approved):
Tested on AZ-WEU
Upgrade tested on AZ-UKS

PR #2235, opened 2020-08-24
Add receipts hash storer

SebastianMarian:
* Added receipts hash storer which holds cumulative receipt hash as a key and all the receipts hashes list marshalized as a value, for each committed block

Testing steps:

a) Prerequisites

b) Test steps

Start a normal testnet

c) Expected results

We should have another storage/folder/db created in: .../Epoch_X/Shard_Y/**Receipts** which holds cumulative receipt hash as a key and all the receipts hashes list marshalized as a value, for each committed block

d) Actual results

Test process is in pending

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -65,6 +65,19 @@
         MaxBatchSize = 100
         MaxOpenFiles = 10
 
+[ReceiptsStorage]
+    [ReceiptsStorage.Cache]
+        Name = "ReceiptsStorage"
+        Capacity = 1000
+        Type = "SizeLRU"
+        SizeInBytes = 104857600 #100MB
+    [ReceiptsStorage.DB]
+        FilePath = "Receipts"
+        Type = "LvlDBSerial"
+        BatchDelaySeconds = 2
+        MaxBatchSize = 100
+        MaxOpenFiles = 10
+
 [PeerBlockBodyStorage]
     [PeerBlockBodyStorage.Cache]
         Name = "PeerBlockBodyStorage"
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -70,7 +70,7 @@
         Name = "ReceiptsStorage"
         Capacity = 1000
         Type = "SizeLRU"
-        SizeInBytes = 104857600 #100MB
+        SizeInBytes = 10485760 #10MB
     [ReceiptsStorage.DB]
         FilePath = "Receipts"
         Type = "LvlDBSerial"
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -988,14 +988,16 @@ func (bp *baseProcessor) saveBody(body *block.Body, header data.HeaderHandler) {
 		log.Trace("saveBody.Put -> MiniBlockUnit", "time", time.Since(startTime))
 	}
 
-	marshalizedReceiptsHashes, errNotCritical := bp.txCoordinator.CreateMarshalizedReceipts()
-	if errNotCritical != nil {
-		log.Warn("saveBody.CreateMarshalizedReceipts", "error", errNotCritical.Error())
-	}
-
-	errNotCritical = bp.store.Put(dataRetriever.ReceiptsUnit, header.GetReceiptsHash(), marshalizedReceiptsHashes)
-	if errNotCritical != nil {
-		log.Warn("saveBody.Put -> ReceiptsUnit", "error", errNotCritical.Error())
+	if len(header.GetReceiptsHash()) > 0 {
+		marshalizedReceiptsHashes, errNotCritical := bp.txCoordinator.CreateMarshalizedReceipts()
+		if errNotCritical != nil {
+			log.Warn("saveBody.CreateMarshalizedReceipts", "error", errNotCritical.Error())
+		} else {
+			errNotCritical = bp.store.Put(dataRetriever.ReceiptsUnit, header.GetReceiptsHash(), marshalizedReceiptsHashes)
+			if errNotCritical != nil {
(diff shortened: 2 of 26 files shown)


process/block/baseProcess.go
```diff
@@ -988,6 +988,16 @@ func (bp *baseProcessor) saveBody(body *block.Body) {
 		log.Trace("saveBody.Put -> MiniBlockUnit", "time", time.Since(startTime))
 	}
 
+	finalReceiptHash, marshalizedReceiptsHashes, errNotCritical := bp.txCoordinator.CreateMarshalizedReceipts()
+	if errNotCritical != nil {
+		log.Warn("saveBody.CreateMarshalizedReceipts", "error", errNotCritical.Error())
+	}
+
+	errNotCritical = bp.store.Put(dataRetriever.ReceiptsUnit, finalReceiptHash, marshalizedReceiptsHashes)
```
sasurobert, line 996: the problem is not resolved. The saving key for this data structure should come from shardHdr.ReceiptsHash not a new computed hash. otherwise you will not find it,

process/coordinator/process.go
```diff
@@ -952,6 +952,42 @@ func (tc *transactionCoordinator) CreateReceiptsHash() ([]byte, error) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	marshalizedReceiptsHashes, err := tc.marshalizer.Marshal(receiptsBatch)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return finalReceiptHash, marshalizedReceiptsHashes, nil
```
sasurobert, line 988: no need to return finalReceiptHash - use the one from shard Header

sasurobert (changes requested):

process/block/baseProcess.go
SebastianMarian: Actually is not a new computed, it will be exactly the one from shardHdr as this check and the hash calculation is already done in ProcessBlock by calling VerifyCreatedBlockTransactions wich uses the same pattern.

process/coordinator/process.go
SebastianMarian: Yes. I do not have header in caller method but could be added

sasurobert (approved):

process/block/baseProcess.go
```diff
@@ -988,6 +988,16 @@ func (bp *baseProcessor) saveBody(body *block.Body) {
 		log.Trace("saveBody.Put -> MiniBlockUnit", "time", time.Since(startTime))
 	}
 
+	marshalizedReceiptsHashes, errNotCritical := bp.txCoordinator.CreateMarshalizedReceipts()
+	if errNotCritical != nil {
+		log.Warn("saveBody.CreateMarshalizedReceipts", "error", errNotCritical.Error())
+	}
+
+	errNotCritical = bp.store.Put(dataRetriever.ReceiptsUnit, header.GetReceiptsHash(), marshalizedReceiptsHashes)
```
AdoAdoAdo, line 996: if there was an error previously you are storing nil, I think you should skip instead

AdoAdoAdo (changes requested):

cmd/node/config/config.toml
```diff
@@ -65,6 +65,19 @@
...
+    [ReceiptsStorage.Cache]
+        Name = "ReceiptsStorage"
+        Capacity = 1000
+        Type = "SizeLRU"
+        SizeInBytes = 104857600 #100MB
+    [ReceiptsStorage.DB]
+        FilePath = "Receipts"
+        Type = "LvlDBSerial"
+        BatchDelaySeconds = 2
+        MaxBatchSize = 100
```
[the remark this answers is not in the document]
SebastianMarian: it is for the aggregated receipts: key = aggregated receipts hash and value = marshalized individual receipts hashes list. So it is max one record per each block

cmd/node/config/config.toml
```diff
@@ -65,6 +65,19 @@
         MaxBatchSize = 100
         MaxOpenFiles = 10
 
+[ReceiptsStorage]
+    [ReceiptsStorage.Cache]
+        Name = "ReceiptsStorage"
+        Capacity = 1000
+        Type = "SizeLRU"
+        SizeInBytes = 104857600 #100MB
```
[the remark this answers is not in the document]
SebastianMarian: Yes I also thought when I have copied paste this setting from miniblocks but I forgot to adjust it. I think 10 MB is enough for this kind of receipts

update/process/shardBlock.go
```diff
@@ -217,6 +217,16 @@ func (s *shardBlockCreator) saveAllTransactionsToStorageIfSelfShard(
 		}
 	}
 
+	marshalizedReceiptsHashes, errNotCritical := s.txCoordinator.CreateMarshalizedReceipts()
+	if errNotCritical != nil {
+		log.Warn("saveAllTransactionsToStorageIfSelfShard.CreateMarshalizedReceipts", "error", errNotCritical.Error())
+	}
+
+	errNotCritical = s.storage.Put(dataRetriever.ReceiptsUnit, shardHdr.GetReceiptsHash(), marshalizedReceiptsHashes)
```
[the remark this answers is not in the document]
SebastianMarian: Yep, also it should be avoided if len(shardHdr.GetReceiptsHash()) or len(marshalizedReceiptsHashes) is 0 I think

process/block/baseProcess.go
SebastianMarian: Yep, also it should be avoided if len(shardHdr.GetReceiptsHash()) or len(marshalizedReceiptsHashes) is 0 I think

AdoAdoAdo (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2237, opened 2020-08-25
Exchanges integration: return invalid transactions as well, refactor status

andreibancioiu:
- refactor computation of transaction status (encapsulate in `StatusComputer`)
 - add more tests, refactor tests
 - include invalid transactions in response of `/block?withTxs`

 - deprecated field `Status` in `MiniblockMetadata` storage - will not be used anymore. Though, now, `miniblock Type` is required to compute status - thus, a breaking change for the `blockAPI` - but since we will implement replay & index rebuild, this is a non-issue.
 - added new field `Type` in `MiniblockMetadata` storage (see above)
 - also return `miniblock hash` in transaction response, where possible

**How to test:**

 - send both valid and invalid transactions
 - inspect transactions by hash (valid, invalid, rewards, possibly SCs as well)
 - inspect `hyperblock` wrt. to `hyperblockNonce` from step above
 - assert that invalid conditions are included in hyperblocks
 - assert that `miniblock type` is correctly set on `get transaction` responses
 - assert that `miniblock hash`, `source shard`, `destination shard` are present on `get transaction` responses

Get transaction route: `http://proxy:8080/transaction/abracadabra`

Get hyperblock route: `http://proxy:8080/hyperblock/by-nonce/42`


node/nodeTransactions.go
```diff
@@ -31,70 +29,79 @@ func (n *Node) GetTransaction(txHash string) (*transaction.ApiTransactionResult,
...
-		// this should never happen
-		log.Warn("node transaction: cannot unmarshal transaction", "error", err.Error())
-		return putHistoryFieldsInTransaction(nil, miniblockMetadata), nil
+		log.Warn("getFullHistoryTransaction(): unexpected condition, cannot unmarshal transaction")
+		return nil, ErrCannotRetrieveTransaction
 	}
 
-	// merge data about transaction from history storer and transaction storer
-	tx = putHistoryFieldsInTransaction(tx, miniblockMetadata)
+	putHistoryFieldsInTransaction(tx, miniblockMetadata)
```
miiu96, line 87: maybe you can add this method inside structure  `ApiTransactionResult`
`
func (ats *ApiTransactionResult) PutHistoryFieldsInTransaction(miniblockMetadata *fullHistory.MiniblockMetadata)` just a suggestion

data/transaction/status.go
```diff
@@ -0,0 +1,83 @@
...
+	SourceShard      uint32
+	DestinationShard uint32
+	Receiver         []byte
+	TransactionData  []byte
+	SelfShard        uint32
+}
+
+// ComputeStatusWhenInPool computes the transaction status when transaction is in pool
+func (params *StatusComputer) ComputeStatusWhenInPool() TxStatus {
+	if params.isContractDeploy() {
```
sasurobert, line 37: why is this check needed ?

data/transaction/status.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return TxStatusPartiallyExecuted
+}
+
+// ComputeStatusWhenInStorageNotKnowingMiniblock computes the transaction status when transaction is in current epoch's storage
+func (params *StatusComputer) ComputeStatusWhenInStorageNotKnowingMiniblock() TxStatus {
+	// Question for review: here we cannot know if the transaction is, actually, "invalid".
+	// However, when "fullHistory" indexing is enabled, this function is not used.
+
+	if params.isDestinationMe() || params.isContractDeploy() {
```
sasurobert, line 65: we should attach the receipt / smart contract result to the transaction - as we have in elastic search - than it it simple to decide the status of the transaction. It is not a trivial task, this can stay as it is.

node/blockAPI/baseBlock.go
```diff
@@ -39,25 +40,28 @@ func (bap *baseAPIBockProcessor) getTxsByMb(mbHeader *block.MiniBlockHeader, epo
...
 	switch miniBlock.Type {
 	case block.TxBlock:
-		return bap.getTxsFromMiniblock(miniBlock, epoch, "normal", dataRetriever.TransactionUnit)
+		return bap.getTxsFromMiniblock(miniBlock, miniblockHash, epoch, "normal", dataRetriever.TransactionUnit)
 	case block.RewardsBlock:
-		return bap.getTxsFromMiniblock(miniBlock, epoch, "reward", dataRetriever.RewardTransactionUnit)
+		return bap.getTxsFromMiniblock(miniBlock, miniblockHash, epoch, "reward", dataRetriever.RewardTransactionUnit)
 	case block.SmartContractResultBlock:
-		return bap.getTxsFromMiniblock(miniBlock, epoch, "unsigned", dataRetriever.UnsignedTransactionUnit)
+		return bap.getTxsFromMiniblock(miniBlock, miniblockHash, epoch, "unsigned", dataRetriever.UnsignedTransactionUnit)
```
sasurobert, line 54: make some constants for magic strings

sasurobert (changes requested):

node/errors.go
```diff
@@ -43,6 +43,12 @@ var ErrNilPublicKey = errors.New("trying to set nil public key")
 // ErrAccountNotFound signals that an account was not found in trie
 var ErrAccountNotFound = errors.New("account not found")
 
+// ErrTransactionNotFound signals that a transaction was not found
+var ErrTransactionNotFound = errors.New("transaction not found")
+
+// ErrCannotRetrieveTransaction signals that a transaction was not found
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

data/transaction/status.go
```diff
@@ -0,0 +1,83 @@
...
+	if params.isDestinationMe() && params.isCrossShard() {
+		return TxStatusPartiallyExecuted
+	}
+
+	return TxStatusReceived
+}
+
+// ComputeStatusWhenInStorageKnowingMiniblock computes the transaction status for a historical transaction
+func (params *StatusComputer) ComputeStatusWhenInStorageKnowingMiniblock() TxStatus {
+	if params.MiniblockType == block.InvalidBlock {
```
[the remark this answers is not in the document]
andreibancioiu: Added.

node/nodeTransactions.go
andreibancioiu: That is a good suggestion, indeed, to encapsulate this logic in `ApiTransactionResult`. The issue is, though, that ` ApiTransactionResult` is defined in `transaction` package, while `miniblockMetadata` in `fullHistory` package - and that would require an additional interface - we can do in the future.

data/transaction/status.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	// TxStatusReceived = received but not yet executed
+	TxStatusReceived TxStatus = "received"
+	// TxStatusPartiallyExecuted = received and executed on source shard
+	TxStatusPartiallyExecuted TxStatus = "partially-executed"
+	// TxStatusExecuted = received and executed
+	TxStatusExecuted TxStatus = "executed"
+	// TxStatusNotExecuted = received and executed with error
+	// Question for review: not executed means "executed with error", correct?
```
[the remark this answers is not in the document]
andreibancioiu: Currently, also Explorer displays `not executed` for executed transactions (but, say, with user error). Do we keep it like this for the moment or do we change this?

data/transaction/status.go
andreibancioiu: `params.isDestinationMe() && params.isCrossShard()` would be evaluated to `true` for deployments, but still we don't want to return `TxStatusPartiallyExecuted` in that case (seen in previous PR with the status for deployments).

Does it make sense?

node/blockAPI/baseBlock.go
```diff
@@ -83,6 +87,16 @@ func (bap *baseAPIBockProcessor) getTxsFromMiniblock(
 			continue
 		}
 		tx.Hash = hex.EncodeToString([]byte(txHash))
+		tx.MiniBlockType = miniblock.Type
+		tx.MiniBlockHash = hex.EncodeToString([]byte(miniblockHash))
+
+		tx.Status = (&transaction.StatusComputer{
```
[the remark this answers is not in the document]
andreibancioiu: Thanks!

node/blockAPI/baseBlock.go
andreibancioiu: Fixed.

data/transaction/status.go
sasurobert: isCrossShard should return false in case of deployment.

data/transaction/status.go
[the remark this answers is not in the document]
sasurobert: we should change that as well.

data/transaction/status.go
sasurobert: delete the comments question for review

sasurobert (changes requested):

node/nodeTransactions.go
```diff
@@ -150,107 +170,106 @@ func (n *Node) isTxInStorage(hash []byte) bool {
...
 			return nil, err
 		}
-		status := n.computeTransactionStatus(&tx, false)
-		return n.prepareUnsignedTx(&tx, status)
-	default:
-		return &transaction.ApiTransactionResult{Type: string(invalidTx)}, nil // this shouldn't happen
+		return n.prepareUnsignedTx(&tx)
 	}
+
+	return &transaction.ApiTransactionResult{Type: string(transaction.TxTypeInvalid)}, nil // this shouldn't happen
```
AdoAdoAdo, line 266: instead of a comment ca we add a warn here?

AdoAdoAdo (changes requested):

node/nodeTransactions.go
andreibancioiu: Fixed.

node/nodeTransactions.go
```diff
@@ -105,31 +103,53 @@ func putHistoryFieldsInTransaction(tx *transaction.ApiTransactionResult, miniblo
...
+		return nil, ErrTransactionNotFound
+	}
+
+	tx, err := n.unmarshalTransaction(txBytes, txType)
+	if err != nil {
+		return nil, err
+	}
+
+	tx.Status = (&transaction.StatusComputer{
+		// Question for review: should not be long-term (only short-term) problems here in case of adaptive sharding, since we return from "current epoch" storage, correct?
```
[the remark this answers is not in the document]
andreibancioiu: Fixed: renamed & added TODO.

data/transaction/status.go
andreibancioiu: Fixed comment, moved above function.

node/nodeTransactions.go
```diff
@@ -31,70 +20,79 @@ func (n *Node) GetTransaction(txHash string) (*transaction.ApiTransactionResult,
...
-		// transaction is in pool return information directly because in history storer
-		// there are no additional information about the transaction
-		return n.castObjToTransaction(txObj, txType)
-	}
-
 	miniblockMetadata, err := n.historyRepository.GetMiniblockMetadataByTxHash(hash)
 	if err != nil {
-		// transaction is not in history storer
-		return nil, err
+		return nil, ErrTransactionNotFound
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

data/transaction/status.go
andreibancioiu: Currently, `isCrossShard()` is implemented as follows:

```
func (params *StatusComputer) isCrossShard() bool {
	return params.SourceShard != params.DestinationShard
}
```

While `ComputeStatusWhenInPool()` is fed the parameters as follows - in a generic manner, not knowing anything special about the Zero address:

```
SourceShard:      n.shardCoordinator.ComputeId(tx.Tx.GetSndAddr()),
DestinationShard: n.shardCoordinator.ComputeId(tx.Tx.GetRcvAddr()),
Receiver:         tx.Tx.GetRcvAddr(),
TransactionData:  tx.Data,
SelfShard:        n.shardCoordinator.SelfId(),
```

Thus, the `isContractDeploy()` has to be checked first (because the check `isDestinationMe() && isCrossShard()` will be fooled otherwise).

We could change `isCrossShard()` to call `isContractDeploy()`, but perhaps that would be a bit harder to reason about in the future.

Fixed to be more explicit, as before. Is it all right?

data/transaction/status.go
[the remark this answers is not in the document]
andreibancioiu: Only now I see that `TxStatusNotExecuted` is only executed in the context of the transactions simulator:

``
node/txsimulator/txSimulator.go, line 63`
``

And seems to be used well there.

The other occurrence, completely unrelated, is in the indexer component (which, actually, defines a separate set of statuses): `txStatusNotExecuted = "Not Executed"`.

I don't want to change the indexer status in this PR (a bit risky), but we can do that in another PR.

Ok?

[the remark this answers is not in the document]
sasurobert: Ok. that is in another PR changing in the indexer.

data/transaction/status.go
sasurobert: in this case shardCoordinator computeID should have returned selfID. for the empty address.

sasurobert: in computestatuswheninPool you might use another function a wrapper on top of computeID where you add the empty check.

node/errors.go
```diff
@@ -195,3 +190,16 @@ var ErrNilTxSimulatorProcessor = errors.New("nil transaction simulator processor
 
 // ErrNilIntermediateProcessorContainer signals that intermediate processors container is nil
 var ErrNilIntermediateProcessorContainer = errors.New("intermediate processor container is nil")
+
+// NewErrTransactionNotFound creates an error which signals that a transaction was not found
+func NewErrTransactionNotFound(originalErr error) error {
```
sasurobert, line 195: do we really need these ?

node/nodeTransactions.go
```diff
@@ -60,19 +60,19 @@ func (n *Node) optionallyGetTransactionFromPool(hash []byte) (*transaction.ApiTr
...
 	if err != nil {
-		return nil, ErrTransactionNotFound
+		return nil, NewErrTransactionNotFound(err)
 	}
 
 	txBytes, txType, found := n.getTxBytesFromStorageByEpoch(hash, miniblockMetadata.Epoch)
 	if !found {
 		log.Warn("getFullHistoryTransaction(): unexpected condition, cannot find transaction in storage")
-		return nil, ErrCannotRetrieveTransaction
+		return nil, NewErrCannotRetrieveTransaction(err)
```
sasurobert, line 69: error wrapping can happen directly here.

sasurobert (changes requested):

AdoAdoAdo (approved):

node/errors.go
andreibancioiu: Let's have them, please.

node/nodeTransactions.go
andreibancioiu: Let's use the wrappers defined in `errors.go`, please (a bit cleaner).

data/transaction/status.go
andreibancioiu: Let's keep compute ID completely out of the `statusComputer`, please.

Previous location, previous PR (reviewed):
https://github.com/ElrondNetwork/elrond-go/pull/2234/files

sasurobert: Sometimes it happens that these things get approved by a PR and the next one requests changes in the same zone. It would be better to have correct shardId computation at the base. In this case shard coordinator, or computestatus

miiu96 (approved):

node/nodeTransactions.go
andreibancioiu: Fixed.

node/errors.go
andreibancioiu: Fixed.

data/transaction/status.go
andreibancioiu: Left as it is, as discussed separately.

miiu96 (dismissed):

AdoAdoAdo (approved):

miiu96 (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2238, opened 2020-08-25
fixed the situation when empty receipts would have been saved

bogdan-rosianu:
Previously, the function which computed the receipts hash would have returned a hash even if no receipts was present. Added an extra check


SebastianMarian (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed

PR #2240, opened 2020-08-25
Fix save genesis meta to storage

SebastianMarian:
* Fixed situation when node is restarted in epoch 0 with db and with StartInEpochEnabled flag set to false

Testing steps:

a) Prerequisites

b) Test steps

Start a normal testnet with StartInEpochEnabled flag from config.toml set to false. Restart nodes during the epoch 0 with db.

c) Expected results

Nodes should be able to start OK after restart in epoch 0 with db and with StartInEpochEnabled flag set to false.

d) Actual results

Test process is in pending


sasurobert (approved):

iulianpascalau (approved):

iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2241, opened 2020-08-26
Es search order field

ccorcoveanu:
Added new field to be indexed in elastic. This helps the front end using the "search_after" parameter to scroll deeper than 10k elements.


core/indexer/processTransactions.go
```diff
@@ -248,6 +251,33 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
 	return transactions, rewardsTxs
 }
 
+func (tdp *txDatabaseProcessor) setTransactionSearchOrder(transactions map[string]*Transaction, shardId uint32) map[string]*Transaction {
+	currentOrder := 0
+	shardIdentifier := tdp.createShardIdentifier(shardId)
```
miiu96, line 256: shardID

core/indexer/processTransactions.go
```diff
@@ -248,6 +251,33 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
...
+
+	for _, tx := range transactions {
```
SebastianMarian, line 258: The iteration is done over a map. This will not ensure the same order each time. Does it matter?

core/indexer/processTransactions.go
```diff
@@ -248,6 +251,33 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
...
+		stringOrder := fmt.Sprintf("%d%d", shardIdentifier, currentOrder)
```
SebastianMarian, line 259: The 10th tx of metachain will have the order 110 which will be higher than the 9th of the shard 2 whcih will have the order 49. Does it matter for the logic that you needed from this method?

core/indexer/elasticsearchDatabase.go
```diff
@@ -449,3 +451,17 @@ func (esd *elasticSearchDatabase) foundedObjMap(hashes []string, index string) (
 
 	return getDecodedResponseMultiGet(response), nil
 }
+
+func (esd *elasticSearchDatabase) computeBlockSearchOrder(header data.HeaderHandler) uint32 {
+	shardIdentifier := esd.txDatabaseProcessor.createShardIdentifier(header.GetShardID())
+	stringOrder := fmt.Sprintf("%d%d", shardIdentifier, header.GetNonce())
```
SebastianMarian, line 457: The 10th nonce of metachain will have the order 110 which will be higher than the 9th nonce of the shard 2 which will have the order 49. Does it matter for the logic that you needed from this method?

core/indexer/processTransactions.go
ccorcoveanu: ShardID is the parameter passed to the function. I used "identifier" to: 1. avoid name collision and 2. to avoid confusion between an actual shard id in Elrond context and this custom identifier

core/indexer/processTransactions.go
ccorcoveanu: You are right. But we don't care. The only thing we are interested in here is to have a tie breaker for the front end where the shard id is the same and the (round) timestamp is the same

core/indexer/processTransactions.go
ccorcoveanu: That's right but we don't care about that. We already can have a tie breaker and sort by shard id. We only need to have a rank between items in the same shard and the same indexed bulk

core/indexer/elasticsearchDatabase.go
ccorcoveanu: That's right but we don't care about that. We already can have a tie breaker and sort by shard id. We only need to have a rank between items in the same shard and the same indexed bulk

miiu96 (approved):

SebastianMarian (approved):

core/indexer/processTransactions.go
SebastianMarian: ok

core/indexer/processTransactions.go
SebastianMarian: ok

core/indexer/elasticsearchDatabase.go
SebastianMarian: ok

LucianMincu (approved):
System tests passed.

PR #2242, opened 2020-08-26
Add marshalized receipts in storer

SebastianMarian:
* Added in receipts storer as a value a marshalized batch which contains all the marshalized receipts in clear and not only their hashes

Testing steps:

a) Prerequisites

b) Test steps

Start a normal testnet

c) Expected results

We should have another storage/folder/db created in: .../Epoch_X/Shard_Y/Receipts which holds cumulative receipt hash as a key and all the marshalized receipts list marshalized as a value, for each committed block

d) Actual results

Test process is in pending


process/coordinator/process.go
```diff
@@ -952,39 +952,38 @@ func (tc *transactionCoordinator) CreateReceiptsHash() ([]byte, error) {
...
 
-		mb := interProc.GetCreatedInShardMiniBlock()
-		if mb == nil {
+		miniBlock := interProc.GetCreatedInShardMiniBlock()
+		if miniBlock == nil {
 			continue
 		}
 
-		mbHash, err := core.CalculateHash(tc.marshalizer, tc.hasher, mb)
+		marshalizedMiniBlock, err := tc.marshalizer.Marshal(miniBlock)
```
sasurobert, line 969: this is a list of marshalled miniblocks. make sure that these things are not saved 2 times.

sasurobert (commented):
the description does not match the implementation - here you are saving receipt hash - all miniblocks in clear (but those will contain hashes one by one). It is good - maybe we should ensure that receiptshashes are not saved 2 times in storage.

process/coordinator/process.go
```diff
@@ -952,39 +952,38 @@ func (tc *transactionCoordinator) CreateReceiptsHash() ([]byte, error) {
...
 	}
 
-	if len(receiptsHashes) == 0 {
+	if len(receiptsBatch.Data) == 0 {
 		return nil, nil
 	}
 
-	receiptsBatch := &batch.Batch{Data: receiptsHashes}
-	marshalizedReceiptsHashes, err := tc.marshalizer.Marshal(receiptsBatch)
+	marshalizedReceiptsBatch, err := tc.marshalizer.Marshal(receiptsBatch)
```
bogdan-rosianu, line 981: could return directly

bogdan-rosianu (changes requested):

process/coordinator/process.go
SebastianMarian: They are not saved anywhere as these are the ones which are deleted from applyBodyToHeader through newBody := deleteSelfReceiptsMiniBlocks(body) method

bogdan-rosianu (approved):

sasurobert (approved):

process/coordinator/process.go
SebastianMarian: done

LucianMincu (approved):
System tests passed.

PR #2244, opened 2020-08-26
Exchanges integration storers

andreibancioiu:
1) After restarting a node, `GetFromEpoch` could not access persisters older than `NumActivePersisters`.

 - Fix: when `CleanOldEpochsData` is unset, then open all persisters (and close old ones, as before)
 - Fix: call `setIsClosed()` when closing the persisters

2) In the PR [2226](https://github.com/ElrondNetwork/elrond-go/pull/2226), I've mistakenly disabled `epoch by tx hash` indexing (and thus `get tx by hash` API wouldn't work for older epochs). Re-made now the link to epoch, by making `MiniblockHashByTxHashStorage` Static. **Put transactions is fast (1 put per transaction), get 1 transaction is slower (3 reads).**

3) Index genesis block (for hyperblock nonce 1).

4) Fixed path for EpochByHashStorage. `FilePath = "EpochByHash"`, it was empty due to PR [2226](https://github.com/ElrondNetwork/elrond-go/pull/2226).

5) Fixed edge-case for saving hyperblock coordinates of cross-epoch (committed in 42, notarized in 43) transactions. Using `putInEpoch` primitive.

6) Rename "fullHistory" to "db lookup extensions".

**How to test**

1 and 2 - wait a few epochs, stop some observers, start them again, query older transactions and older (hyper)blocks.

3 - Query hyperblock 1 (by nonce, by hash). Query shard blocks 0 (by nonce, by hash). Compare with Explorer.

4 - Assert that DB folders structure in `Static` is nice.

5 - A bit harder to test manually - we have unit tests though.

6 - Inspect `config.toml` and `node/db`'s structure.

**Note that, for tests, I've set**:

```
RoundsPerEpoch = 20
FullHistory.Enabled = true
```


miiu96 (dismissed):

cmd/node/config/config.toml
```diff
@@ -480,8 +480,8 @@
    Type = "json"
 
 [EpochStartConfig]
-    MinRoundsBetweenEpochs = 200
-    RoundsPerEpoch         = 200
+    MinRoundsBetweenEpochs = 20 # UNDO AFTER TESTS
```
iulianpascalau, line 483: Might add a TODO as to signal the line

cmd/node/config/config.toml
```diff
@@ -645,7 +645,7 @@
     PollingIntervalInMinutes = 65
 
 [FullHistory]
-    Enabled = false
+    Enabled = true # DISABLE AFTER TESTS
```
iulianpascalau, line 648: TODO?

AdoAdoAdo (approved):

storage/pruning/pruningStorer.go
```diff
@@ -201,8 +201,12 @@ func initPersistersInEpoch(
 	if oldestEpochActive < 0 {
 		oldestEpochActive = 0
 	}
+	if !args.CleanOldEpochsData {
+		oldestEpochKeep = 0
```
iulianpascalau, line 205: This solution is not quite extensible. If we have thousands of epochs, the node will soon run out of memory. Plus, as seen today opening a storage is a quite costly operation...on the order of magnitude of seconds.

storage/pruning/pruningStorer.go
```diff
@@ -201,8 +201,12 @@ func initPersistersInEpoch(
...
+	}
 
 	for epoch := int64(args.StartingEpoch); epoch >= oldestEpochKeep; epoch-- {
+		log.Debug("initPersistersInEpoch(): createPersisterDataForEpoch", "identifier", args.Identifier, "epoch", epoch, "shardID", shardIDStr)
 		p, err := createPersisterDataForEpoch(args, uint32(epoch), shardIDStr)
```
AdoAdoAdo, line 210: talking with Iulian, seems that opening and closing 1000 DBs can take up to 20 minutes.
Maybe we can change this function createPersisterDataForEpoch, to only create the reference but not open/close. 
We should only open the active persisters at the end.

AdoAdoAdo (changes requested):

cmd/node/config/config.toml
andreibancioiu: Added.

cmd/node/config/config.toml
andreibancioiu: Added.

storage/pruning/pruningStorer.go
andreibancioiu: Sketched a solution.

storage/pruning/pruningStorer.go
andreibancioiu: Sketched a solution.

storage/pruning/pruningStorer.go
```diff
@@ -340,6 +348,7 @@ func (ps *PruningStorer) GetFromEpoch(key []byte, epoch uint32) ([]byte, error)
 	}
 
 	persister, err := ps.persisterFactory.Create(pd.path)
+	// Question for review: save reference on persister data?
```
iulianpascalau, line 351: IMHO, I do not think we will need the old closed reference.

storage/pruning/pruningStorer.go
```diff
@@ -310,6 +316,8 @@ func (ps *PruningStorer) Close() error {
 			log.Error("cannot close persister", err)
 			closedSuccessfully = false
 		}
+		// Question for review - all good here?
+		persister.setIsClosed(true)
```
iulianpascalau, line 320: :+1:

storage/pruning/pruningStorer.go
```diff
@@ -350,6 +359,8 @@ func (ps *PruningStorer) GetFromEpoch(key []byte, epoch uint32) ([]byte, error)
 		if err != nil {
 			log.Debug("persister.Close()", "error", err.Error())
 		}
+		// Question for review: all good here?
+		pd.setIsClosed(true)
```
iulianpascalau, line 363: :+1:

storage/pruning/pruningStorer.go
```diff
@@ -400,6 +412,8 @@ func (ps *PruningStorer) GetBulkFromEpoch(keys [][]byte, epoch uint32) (map[stri
 			if err != nil {
 				log.Debug("persister.Close()", "error", err.Error())
 			}
+			// Question for review - all good here?
+			pd.setIsClosed(true)
```
iulianpascalau, line 416: :+1:

storage/pruning/pruningStorer.go
```diff
@@ -517,6 +532,8 @@ func (ps *PruningStorer) HasInEpoch(key []byte, epoch uint32) error {
...
```
iulianpascalau, line 536: :+1:

storage/pruning/pruningStorer.go
```diff
@@ -882,14 +899,33 @@ func (ps *PruningStorer) IsInterfaceNil() bool {
 	return ps == nil
 }
 
-func createPersisterDataForEpoch(args *StorerArgs, epoch uint32, shardIdStr string) (*persisterData, error) {
-	// TODO: if booting from storage in an epoch > 0, shardId needs to be taken from somewhere else
-	// e.g. determined from directories in persister path or taken from boot storer
+func createShallowPersisterDataForEpoch(args *StorerArgs, epoch uint32, shardIdStr string) *persisterData {
+	filePath := createPersisterPathForEpoch(args, epoch, shardIdStr)
+
+	p := &persisterData{
```
iulianpascalau, line 905: return directly?

bogdan-rosianu (approved):

storage/pruning/pruningStorer.go
andreibancioiu: Fixed.

storage/pruning/pruningStorer.go
```diff
@@ -245,11 +245,16 @@ func (ps *PruningStorer) Put(key, data []byte) error {
 			log.Debug("active persister not found",
 				"epoch", ps.epochForPutOperation,
 				"used", persisterToUse.epoch)
+
+			// Question for review: isn't this a critical error?
```
bogdan-rosianu, line 249: I think it is critical because otherwise you'd save data in a persister which is not for the desired epoch

bogdan-rosianu (approved):

storage/pruning/pruningStorer.go
iulianpascalau: If we return error here, the node won't advance: this will keep the DB as consistent as possible. If we not, the node will advance, but the data might not be stored inside

storage/pruning/pruningStorer.go
```diff
@@ -266,6 +267,60 @@ func (ps *PruningStorer) Put(key, data []byte) error {
...
+	}
+	defer closePersister()
+
+	return ps.doPutInPersister(key, data, persister)
+}
+
+func (ps *PruningStorer) createAndInitPersisterIfClosed(pd *persisterData) (storage.Persister, func(), error) {
+	isOpen := !pd.getIsClosed()
+	if isOpen {
+		noopClose := func() {}
```
iulianpascalau, line 293: noopClose for closing the persister. Smart

node/nodeTransactions_test.go
```diff
@@ -310,13 +310,13 @@ func TestNode_PutHistoryFieldsInTransaction(t *testing.T) {
...
-func createNode(t *testing.T, withFullHistory bool) (*Node, *genericmocks.ChainStorerMock, *testscommon.PoolsHolderMock, *testscommon.HistoryRepositoryStub) {
-	chainStorer := genericmocks.NewChainStorerMock()
+func createNode(t *testing.T, epoch uint32, withDbLookupExt bool) (*Node, *genericmocks.ChainStorerMock, *testscommon.PoolsHolderMock, *testscommon.HistoryRepositoryStub) {
+	chainStorer := genericmocks.NewChainStorerMock(epoch)
 	dataPool := testscommon.NewPoolsHolderMock()
 	marshalizer := &mock.MarshalizerFake{}
 
 	historyRepo := &testscommon.HistoryRepositoryStub{
-		IsEnabledCalled: func() bool { return withFullHistory },
+		IsEnabledCalled: func() bool { return withDbLookupExt },
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

storage/factory/pruningStorerFactory.go
```diff
@@ -481,6 +481,7 @@ func (psf *StorageServiceFactory) createPruningStorerArgs(storageConfig config.S
 		NumOfActivePersisters: numOfActivePersisters,
 		Notifier:              psf.epochStartNotifier,
 		MaxBatchSize:          storageConfig.DB.MaxBatchSize,
+		DbLookupExtensions:    psf.generalConfig.DbLookupExtensions.Enabled,
```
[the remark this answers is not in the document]
andreibancioiu: Fixed.

storage/pruning/pruningStorer.go
andreibancioiu: Understood. Left as it were - to prevent breaking something unforeseen. However, let me know if I should change instead.

iulianpascalau (approved):

cmd/node/config/config.toml
```diff
@@ -480,8 +480,8 @@
    Type = "json"
 
 [EpochStartConfig]
-    MinRoundsBetweenEpochs = 200
-    RoundsPerEpoch         = 200
+    MinRoundsBetweenEpochs = 100 # TODO UNDO AFTER TESTS
```
AdoAdoAdo, line 483: undo ^^

storage/pruning/pruningStorer.go
```diff
@@ -502,6 +517,7 @@ func (ps *PruningStorer) HasInEpoch(key []byte, epoch uint32) error {
 		}
 
 		persister, err := ps.persisterFactory.Create(pd.path)
+		// Question for review: save reference on persister data?
```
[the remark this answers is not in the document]
andreibancioiu: Ok.

cmd/node/config/config.toml
andreibancioiu: Fixed.

storage/pruning/pruningStorer.go
bogdan-rosianu: this case didn't happen because the `epochForPutOperation` is set on `CommitBlock` so there are very small chances that the persister for the specified epoch doesn't exist. can remain like this for now

bogdan-rosianu (approved):

AdoAdoAdo (approved):

cmd/node/factory/structs.go
```diff
@@ -628,20 +628,40 @@ func prepareGenesisBlock(args *processComponentsFactoryArgs, genesisBlocks map[u
...
+		if args.shardCoordinator.SelfId() != shard {
+			continue
+		}
+
+		genesisBlockHash, err := core.CalculateHash(args.coreData.InternalMarshalizer, args.coreData.Hasher, genesisBlockHeader)
+		if err != nil {
+			return err
+		}
+
+		log.Info("indexGenesisBlocks(): historyRepo.RecordBlock", "shard", shard, "hash", genesisBlockHash)
```
miiu96, line 653: is ok this log.Info ?  maybe debug ?

andreibancioiu: Indeed, I wanted it to be more obvious (since it was missing before).

miiu96 (approved):

LucianMincu (approved):
System tests passed.

PR #2253, opened 2020-08-28
Changed log level in scope of debugging

SebastianMarian:


andreibancioiu (approved):

process/block/preprocess/transactions.go
```diff
@@ -907,7 +907,8 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 			}
 
 			numTxsBad++
-			log.Trace("bad tx",
+			//TODO: Change log level back to Trace
```
sasurobert, line 910: too many todo's. after testing this branch may be deleted.

sasurobert (changes requested):

process/block/preprocess/transactions.go
SebastianMarian: This branch should be merged in dev for now, as there are other PRs in waiting list to be merged in dev, and all the tests are done using dev so we need these prints for now in dev to catch the problem if it will be reproduced again

sasurobert (approved):

LucianMincu (approved):
👍

PR #2254, opened 2020-08-28
fix backward compatibility of tx gas consumption in case of invalid txs

sasurobert:
fix backward compatibility of tx gas consumption in case of invalid txs


raduchis (approved):

AdoAdoAdo (approved):

SebastianMarian (approved):

LucianMincu (approved):
System tests passed.

PR #2255, opened 2020-08-28
Fix stop offline syncer

iulianpascalau:
- fixed the stopping process when importing data with the offline syncer

Testing procesures:
- as in PR #2248, the import process should run as expected until a missing data error happens. This time can be any data (tx, miniblock, nonce-hash, receipt, etc.), not just the hash-header info.


SebastianMarian (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2258, opened 2020-08-30
Fix epoch set in manual epoch notifier

iulianpascalau:
- fixed the initial epoch after the instantiation of the manualEpochStartNotifier object
- automatically alter checkpoint modulus value when importing DB

Testing procedures:
- start the import process, let it sync for about 5 epochs
- close the node that does the import
- restart the process
Expected results:
- the node should resume its import process, should not close prematurely stating:
```
terminating at internal stop signal      reason = importComplete description = import ended because data from epochs 0 or 1 does not exist
```

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -95,6 +95,7 @@ const (
 	secondsToWaitForP2PBootstrap = 20
 	maxTimeToClose               = 10 * time.Second
 	maxMachineIDLen              = 10
+	checkpointRoundsModulus      = 14400
 )
 
 var (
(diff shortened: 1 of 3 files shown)


sasurobert (approved):

AdoAdoAdo (approved):

cmd/node/main.go
```diff
@@ -1489,8 +1490,13 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
...
 	importDbDirectoryValue := ctx.GlobalString(importDbDirectory.Name)
 	if len(importDbDirectoryValue) > 0 {
-		log.Info("import DB directory is set, turning off start in epoch", "value", importDbDirectoryValue)
+		log.Info("import DB directory is set, altering config values!",
+			"GeneralSettings.StartInEpochEnabled", "false",
+			"StateTriesConfig.CheckpointRoundsModulus", importCheckpointRoundsModulus,
+			"import DB path", importDbDirectoryValue,
+		)
 		config.GeneralSettings.StartInEpochEnabled = false
+		config.StateTriesConfig.CheckpointRoundsModulus = importCheckpointRoundsModulus
```
SebastianMarian, line 1499: Why not: config.StateTriesConfig.CheckpointRoundsModulus = uint(config.EpochStartConfig.RoundsPerEpoch)
and you can remove importCheckpointRoundsModulus constant

iulianpascalau: done

sasurobert (approved):

SebastianMarian (approved):

AdoAdoAdo (approved):

LucianMincu (approved):
System tests passed.

PR #2259, opened 2020-08-30
Fix block data cleaning from pools

SebastianMarian:
* Canceled the removal of block data at the commit time as this approach was forgotten here and its scope was to test an old issue which was proven later to be from somewhere else. As now, there are situations when block data (miniblocks and transactions) are intentionally broadcast with delay, the call of the removeBlockDataFromPool method at commit time, could remove from pool only partial data. Because of this, when the cleanupPools method is called for the headers behind final, the already removed block data at the commit time for earlier blocks, broke some links between all the components from these blocks and they could not be removed anymore completely from pools.


iulianpascalau (approved):

AdoAdoAdo (approved):

sasurobert (approved):

LucianMincu (changes requested):
System tests failed

cmd/node/config/config.toml
```diff
@@ -541,7 +541,7 @@
 
 [BlockSizeThrottleConfig]
     MinSizeInBytes = 104857 # 104857 is 10% from 1MB
-    MaxSizeInBytes = 943718 # 943718 is 90% from 1MB
+    MaxSizeInBytes = 891289 # 943718 is 85% from 1MB
```
iulianpascalau, line 544: :+1:

iulianpascalau (approved):

sasurobert (approved):

AdoAdoAdo (approved):

process/block/preprocess/transactions.go
```diff
@@ -961,7 +961,12 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 		}
 
 		miniBlock.TxHashes = append(miniBlock.TxHashes, txHash)
-		txs.blockSizeComputation.AddNumTxs(1)
+		numTxs := 1
+		if core.IsSmartContractAddress(tx.RcvAddr) {
```
sasurobert, line 965: this is not the correct way to resolve it. Add blockSizeComputation component to smart contract result post processor and call AddNumTxs there when new cross shard SCRs are generated. It is not ensured that one tx = one scr

sasurobert (changes requested):

process/block/preprocess/transactions.go
iulianpascalau: When we reach the postprocessor it is already too late as the transactions are already executed and we need to include the scr(s) as well. Indeed, it is not ensured that one tx = one scr but at least it will estimate better.

SebastianMarian: Actually you will have nothing in post processor when you create a miniblock from me to meta with 27k stake systems SC calls. In fact at the destination, in this case in meta, could result in 54k txs (from which 27k are SCR). Because you need to process at least one mb completely this could result in a stuck situation. So, you also need to multiply at source, with 2 or more (in terms of size, depending of how many SCR could exist), all the SC calls when you create mbs from me. I think that this situation on normal cases should be solved by gas limit and in case of an invalid SC call, adding only one additional tx which simulates the scr created at destination, should be enough as in this case would be only one SCR for each invalid SC call?

SebastianMarian: Added block size computation taking in consideration also the post process mbs and txs when creating mbs dest me

process/block/preprocess/transactions.go
```diff
@@ -962,7 +961,13 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 		}
 
 		miniBlock.TxHashes = append(miniBlock.TxHashes, txHash)
+		numScCalls := 0
+		if core.IsSmartContractAddress(tx.RcvAddr) {
+			//we need to increment this as to account for the corresponding SCR hash
+			numScCalls++
+		}
 		txs.blockSizeComputation.AddNumTxs(1)
+		txs.blockSizeComputation.AddNumTxs(numScCalls)
```
sasurobert, line 970: numScCalls here is always 1. call txs.blockSizeComputation.AddNumTxs(1) instead of numSCCalls++. Although this protection should be only for cross shard and per shard.

process/coordinator/process.go
```diff
@@ -981,6 +988,27 @@ func (tc *transactionCoordinator) CreateMarshalizedReceipts() ([]byte, error) {
...
+
+	for _, blockType := range tc.keysInterimProcs {
+		interimProc := tc.getInterimProcessor(blockType)
+		if check.IfNil(interimProc) {
+			continue
+		}
+
+		mbs := interimProc.CreateAllInterMiniBlocks()
+		for _, mb := range mbs {
+			numTxs += len(mb.TxHashes)
```
sasurobert, line 1003: do not add intra shard miniblocks.

sasurobert (changes requested):
add unit test

process/block/preprocess/transactions.go
SebastianMarian: Done

process/coordinator/process.go
SebastianMarian: Refactored

process/block/preprocess/transactions.go
```diff
@@ -961,12 +962,12 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
...
 
 		miniBlock.TxHashes = append(miniBlock.TxHashes, txHash)
-		numTxs := 1
-		if core.IsSmartContractAddress(tx.RcvAddr) {
+		txs.blockSizeComputation.AddNumTxs(1)
+		if receiverShardID != txs.shardCoordinator.SelfId() && core.IsSmartContractAddress(tx.RcvAddr) {
+			numCrossShardScCalls++
 			//we need to increment this as to account for the corresponding SCR hash
-			numTxs++
+			txs.blockSizeComputation.AddNumTxs(1)
```
iulianpascalau, line 969: instead of adding one-by-one, we can add directly on line 973
```GO
txs.blockSizeComputation.AddNumTxs(numCrossShardScCalls)
```

process/coordinator/process.go
```diff
@@ -400,6 +400,10 @@ func (tc *transactionCoordinator) ProcessBlockTransaction(
 		return process.ErrNilBlockBody
 	}
 
+	if tc.isMaxBlockSizeReached(body) {
+		return process.ErrMaxBlockSizeReached
```
iulianpascalau, line 404: so this will prevent the validators sign upon a wrongly constructed block that contains too many txhashes?

process/coordinator/process.go
```diff
@@ -981,6 +985,79 @@ func (tc *transactionCoordinator) CreateMarshalizedReceipts() ([]byte, error) {
...
+		for txHash := range mb.TxHashes {
+			tx, ok := allTxs[string(txHash)]
+			if !ok {
+				log.Warn("transactionCoordinator.isMaxBlockSizeReached: tx not found",
+					"mb type", mb.Type,
+					"senderShardID", mb.SenderShardID,
+					"receiverShardID", mb.ReceiverShardID,
+					"numTxHashes", len(mb.TxHashes),
+					"tx hash", txHash)
+				continue
```
iulianpascalau, line 1039: why not escalate the error here, effectively stating that the block body is not ok?

process/coordinator/process.go
```diff
@@ -981,6 +985,79 @@ func (tc *transactionCoordinator) CreateMarshalizedReceipts() ([]byte, error) {
...
+
+	return totalNumMbs, totalNumTxs
+}
+
+func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
+	numMbs := len(body.MiniBlocks)
+	numTxs := 0
+	numCrossShardScCalls := 0
+
+	allTxs := make(map[string]data.TransactionHandler)
```
iulianpascalau, line 1011: some unit tests would have been great

process/coordinator/process.go
```diff
@@ -988,25 +985,77 @@ func (tc *transactionCoordinator) CreateMarshalizedReceipts() ([]byte, error) {
...
+		isCrossShardTxBlockFromSelf := mb.Type == block.TxBlock &&
+			mb.SenderShardID == tc.shardCoordinator.SelfId() &&
+			mb.ReceiverShardID != tc.shardCoordinator.SelfId()
+
+		if !isCrossShardTxBlockFromSelf {
+			continue
 		}
 
-		numMbs += len(mbs)
+		for txHash := range mb.TxHashes {
```
iulianpascalau, line 1030: txHash is the index, not the real hash, as seen in some fresh logs

iulianpascalau (changes requested):

process/coordinator/process.go
```diff
@@ -400,6 +400,10 @@ func (tc *transactionCoordinator) ProcessBlockTransaction(
 		return process.ErrNilBlockBody
 	}
 
+	if tc.isMaxBlockSizeReached(body) {
```
iulianpascalau, line 403: as seen in the logs, calling this here will not work as expected because the transactions are not gathered in preprocessors internal caches (yet). They will be added later, after the actual processing. Figuring out a fix for this, also.

iulianpascalau (changes requested):

process/block/preprocess/transactions.go
SebastianMarian: done

process/coordinator/process.go
SebastianMarian: Exactly

process/coordinator/process.go
SebastianMarian: Actually this error could not happen, as the node can not reach in this point if it has not all the missing txs from the body received. But if there happened something with the source map I just want to signal with an Warn message and to continue the counting procedure even if we can not check if this missing tx is a sc call or not.

process/coordinator/process.go
SebastianMarian: Would be added of course. Just waiting for the system test results

process/coordinator/process.go
SebastianMarian: :+1: f...g cast to string :)

process/coordinator/process.go
SebastianMarian: I don't need them to be created in this case as they are in block body already so I can add them form there as I need only their count. The proposer sent them. This is why inside this method I checked only the mbs with TxBlock type, and inside them I searched only for txs which are cross shard from me, to find some SC calls if they are, to can increase the number of additional txs correctly.

SebastianMarian: Also the receipts and scr from self to self are not included in body. So I don't need to wait for them to be created.

process/coordinator/process.go
iulianpascalau: It did due to programming errors :))

process/coordinator/process.go
iulianpascalau: Bad explaining on my side: the problem lies in the fact that preprocessor fetch the txs (as structures) from cache only during the actual processing of the miniblock. After that, they will be stored inside the internal cache so they could be written in storage. Calling `tc.GetAllCurrentUsedTxs(blockType)` on line 1013 will yield empty maps.

process/coordinator/process.go
```diff
@@ -768,14 +768,22 @@ func (tc *transactionCoordinator) GetAllCurrentUsedTxs(blockType block.Type) map
 
 	preProc := tc.getPreProcessor(blockType)
 	if preProc != nil {
+		log.Debug("transactionCoordinator.GetAllCurrentUsedTxs: preProc is not nil")
```
iulianpascalau, line 771: this can be removed

process/coordinator/process.go
```diff
@@ -768,14 +768,22 @@ func (tc *transactionCoordinator) GetAllCurrentUsedTxs(blockType block.Type) map
...
 		txPool = preProc.GetAllCurrentUsedTxs()
 	}
 
 	interProc := tc.getInterimProcessor(blockType)
 	if interProc != nil {
+		log.Debug("transactionCoordinator.GetAllCurrentUsedTxs: interProc is not nil")
```
iulianpascalau, line 777: this can be removed

process/coordinator/process.go
```diff
@@ -768,14 +768,22 @@ func (tc *transactionCoordinator) GetAllCurrentUsedTxs(blockType block.Type) map
...
 		interTxPool = interProc.GetAllCurrentFinishedTxs()
 	}
 
+	log.Debug("transactionCoordinator.GetAllCurrentUsedTxs",
```
iulianpascalau, line 781: this can be removed

process/coordinator/process.go
```diff
@@ -1011,11 +1019,18 @@ func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
 	allTxs := make(map[string]data.TransactionHandler)
 	for _, blockType := range tc.keysInterimProcs {
 		txs := tc.GetAllCurrentUsedTxs(blockType)
+
+		log.Debug("transactionCoordinator.isMaxBlockSizeReached",
```
iulianpascalau, line 1023: this can be removed

process/coordinator/process.go
```diff
@@ -1011,11 +1019,18 @@ func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
 	allTxs := make(map[string]data.TransactionHandler)
 	for _, blockType := range tc.keysInterimProcs {
```
iulianpascalau, line 1020: so, this will need to be keysTxPreProcs

process/coordinator/process.go
```diff
@@ -1011,11 +1019,18 @@ func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
...
+		log.Debug("transactionCoordinator.isMaxBlockSizeReached",
+			"blockType", blockType,
+			"txs", len(txs))
+
 		for txHash, tx := range txs {
 			allTxs[txHash] = tx
 		}
 	}
 
+	log.Debug("transactionCoordinator.isMaxBlockSizeReached", "allTxs", len(allTxs))
```
iulianpascalau, line 1032: this can be removed

process/coordinator/process.go
iulianpascalau: The maps are already populated because the prepocessors already computed missing and existing txs.

process/coordinator/process.go
SebastianMarian: done

process/coordinator/process.go
SebastianMarian: done

process/coordinator/process.go
SebastianMarian: done

process/coordinator/process.go
SebastianMarian: done

process/coordinator/process.go
SebastianMarian: refactored

process/coordinator/process.go
SebastianMarian: done

iulianpascalau (approved):

iulianpascalau (approved):

sasurobert (approved):

process/coordinator/process.go
```diff
@@ -1008,41 +1008,18 @@ func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
 	numTxs := 0
 	numCrossShardScCalls := 0
 
+	allTxs := make(map[string]data.TransactionHandler)
+
 	preProc := tc.getPreProcessor(block.TxBlock)
-	if preProc == nil {
+	if check.IfNil(preProc) {
```
iulianpascalau, line 1014: :+1:

process/coordinator/process.go
```diff
@@ -1058,6 +1035,40 @@ func (tc *transactionCoordinator) isMaxBlockSizeReached(body *block.Body) bool {
...
+		tx, ok := allTxs[string(txHash)]
+		if !ok {
+			log.Warn("transactionCoordinator.isMaxBlockSizeReached: tx not found",
+				"mb type", mb.Type,
+				"senderShardID", mb.SenderShardID,
+				"receiverShardID", mb.ReceiverShardID,
+				"numTxHashes", len(mb.TxHashes),
+				"tx hash", txHash)
+
+			// If the tx is not found we assume that it is the smart contract call to handle the worst case scenario
```
iulianpascalau, line 1059: :+1:

iulianpascalau (approved):

iulianpascalau (approved):

sasurobert (approved):

core/constants.go
```diff
@@ -505,3 +505,7 @@ const MaxWaitingTimeToReceiveRequestedItem = 5 * time.Second
 // DefaultLogProfileIdentifier represents the default log profile used when the logviewer/termui applications do not
 // need to change the current logging profile
 const DefaultLogProfileIdentifier = "[default log profile]"
+
+// MultiplyFactorForScCall specifies the multiply factor, in terms of number, which should be used for how many sc calls
+// could be included in one block (normal txs -> aprox. 27000, sc calls -> aprox. 2700 with value of 9 set below)
```
iulianpascalau, line 510: comment or value need to be adjusted

process/block/preprocess/blockSizeComputation_test.go
```diff
@@ -134,9 +134,9 @@ func TestBlockSizeComputation_IsMaxBlockSizeReachedShouldWork(t *testing.T) {
 		{numNewMiniBlocks: 1000000, numNewTxs: 0, expected: true, name: "with miniblocks 1000000 and txs 0"},
 		{numNewMiniBlocks: 0, numNewTxs: 1000000, expected: true, name: "with miniblocks 0 and txs 1000000"},
 		{numNewMiniBlocks: 15, numNewTxs: 1000, expected: false, name: "with miniblocks 15 and txs 1000"},
-		{numNewMiniBlocks: 1, numNewTxs: 27756, expected: false, name: "with miniblocks 1 and txs 20077"},
-		{numNewMiniBlocks: 1, numNewTxs: 27757, expected: true, name: "with miniblocks 1 and txs 20078"},
-		{numNewMiniBlocks: 2, numNewTxs: 27756, expected: true, name: "with miniblocks 2 and txs 20077"},
+		{numNewMiniBlocks: 1, numNewTxs: 27756, expected: false, name: "with miniblocks 1 and txs 27756"},
+		{numNewMiniBlocks: 1, numNewTxs: 27757, expected: true, name: "with miniblocks 1 and txs 27757"},
```
iulianpascalau, line 138: :+1:

process/block/preprocess/smartContractResults.go
```diff
@@ -476,6 +477,43 @@ func (scr *smartContractResults) ProcessMiniBlock(miniBlock *block.MiniBlock, ha
 			&totalGasConsumedInSelfShard)
 
 		if err != nil {
+			//TODO: Remove this after testing
```
iulianpascalau, line 480: can't remove this as it is used for the debug print L499

SebastianMarian: This was added to serves the prints and also the prints would been removed now

core/constants.go
SebastianMarian: done

process/block/preprocess/validatorInfoPreProcessor.go
```diff
@@ -145,6 +151,13 @@ func (vip *validatorInfoPreprocessor) ProcessMiniBlock(miniBlock *block.MiniBloc
 		return nil, process.ErrValidatorInfoMiniBlockNotFromMeta
 	}
 
+	if vip.blockSizeComputation.IsMaxBlockSizeWithoutThrottleReached(1, len(miniBlock.TxHashes)) {
```
iulianpascalau, line 154: We need another function in the blocksize estimator that will better handle the peer changes as those are not hashes. cc @sasurobert

iulianpascalau (approved):

process/block/preprocess/validatorInfoPreProcessor.go
sasurobert: Those are taken as it is. It is hard to estimate what is the actual value

sasurobert (dismissed):
CreateSCRIfError is wrongly constructed right now. It is an old code, which is not up to date. Will fix it in separate PR.

process/block/preprocess/validatorInfoPreProcessor.go
iulianpascalau: Please add a TODO to fix the estimator. It is not complicated to do so in a separate PR.

SebastianMarian: The good part is that until then we can reduce if needed the MaxSizeInBytes from [BlockSizeThrottleConfig] in config.toml to 85% from 90% and in this way we should avoid any kind of "message too large" from p2p

sasurobert (approved):

iulianpascalau (dismissed):

process/block/preprocess/transactions.go
```diff
@@ -829,7 +831,17 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
...
 		}
-		if isMaxBlockSizeReached(numNewMiniBlocks, 1) {
+		numNewTxs := 1
+
+		isCrossShardScCall := receiverShardID != txs.shardCoordinator.SelfId() && core.IsSmartContractAddress(tx.RcvAddr)
+		if isCrossShardScCall {
+			if !firstCrossShardScCallFound {
+				numNewMiniBlocks++
+			}
+			numNewTxs += 1 * core.MultiplyFactorForScCall
```
AdoAdoAdo, line 841: 1*core.MultiplyFactorForScCall 
why 1\*x instead of simply x ?

Is it supposed to be changed later to a different value? then use a constant.

process/block/preprocess/transactions.go
```diff
@@ -963,6 +974,15 @@ func (txs *transactions) createAndProcessMiniBlocksFromMe(
 
 		miniBlock.TxHashes = append(miniBlock.TxHashes, txHash)
 		txs.blockSizeComputation.AddNumTxs(1)
+		if isCrossShardScCall {
+			if !firstCrossShardScCallFound {
+				firstCrossShardScCallFound = true
+				txs.blockSizeComputation.AddNumMiniBlocks(1)
+			}
+			//we need to increment this as to account for the corresponding SCR hash
+			txs.blockSizeComputation.AddNumTxs(1 * core.MultiplyFactorForScCall)
```
AdoAdoAdo, line 983: 1\*core.MultiplyFactorForScCall ?

AdoAdoAdo (changes requested):

process/block/preprocess/transactions.go
SebastianMarian: I just kept an old pattern when 1 was a variable named numTxs. I will change.

process/block/preprocess/transactions.go
SebastianMarian: I just kept an old pattern when 1 was a variable named numTxs. I will change.

sasurobert (approved):

AdoAdoAdo (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2261, opened 2020-08-31
Add node watcher to testnet scripts

camilbancioiu:
There is now a variable called `NODE_WATCHER`, set by default to 1. This variable enables the so-called node watcher, a daemon which restarts the node if it dies.

Enabling the node watcher will also make the node a daemon. This means that it cannot be stopped with CTRL+C from tmux anymore. 

A better solution must be implemented. The current node watcher was designed with a different use case in mind, and is somewhat incompatible with the testnet scripts.


scripts/testnet/include/validators.sh
```diff
@@ -0,0 +1,128 @@
...
+
+resumeValidators() {
+  iterateOverValidators resumeSingleValidator
+}
+
+stopValidators() {
+  iterateOverValidators stopSingleValidator
+}
+
+iterateOverValidators() {
```
iulianpascalau, line 22: almost duplicate code with iterateOverObservers

scripts/testnet/include/validators.sh
```diff
@@ -0,0 +1,128 @@
...
+  local source_command="source $ELRONDTESTNETSCRIPTSDIR/include/watcher.sh"
+  local watcher_command="node-start-with-watcher $VALIDATOR_INDEX $PORT &"
+  local node_command=$(assembleCommand_startValidatorNode $VALIDATOR_INDEX)
+  echo "$node_command" > $WORKING_DIR/node-command
+  echo "$PORT" > $WORKING_DIR/node-port
+
+  echo "$source_command ; $watcher_command"
+}
+
+assembleCommand_startValidatorNode() {
```
iulianpascalau, line 105: this doesn't actually start the node but assemble the command line for a node

scripts/testnet/include/validators.sh
sasurobert: +1

scripts/testnet/include/validators.sh
sasurobert: +1

sasurobert (changes requested):
remove duplicated code

scripts/testnet/include/validators.sh
camilbancioiu: Yes, the function is called `assembleCommand_startValidatorNode()`. Can rename to something else if unclear.

iulianpascalau: can remain as it is

scripts/testnet/include/validators.sh
iulianpascalau: can remain as it is

iulianpascalau (approved):

scripts/testnet/include/validators.sh
camilbancioiu: Due to Bash specifics, these two duplicated functions are cleaner than a single deduplicated iterator (at least at my level of knowledge). A future rewrite in Python will allow for cleaner code.

sasurobert (approved):

AdoAdoAdo (approved):

PR #2262, opened 2020-08-31
Fixing buffer too large

iulianpascalau:
- fixed `buffer to large` error by limiting also the miniblock number in case of executing transactions with destination me
- made p2p `buffer too large` error message carry more info

### DIFF
--- a/p2p/libp2p/netMessenger.go
+++ b/p2p/libp2p/netMessenger.go
@@ -718,7 +718,7 @@ func (netMes *networkMessenger) BroadcastOnChannelBlocking(channel string, topic
 
 func (netMes *networkMessenger) checkSendableData(buff []byte) error {
 	if len(buff) > maxSendBuffSize {
-		return p2p.ErrMessageTooLarge
+		return fmt.Errorf("%w, maximum: %d, received %d", p2p.ErrMessageTooLarge, maxSendBuffSize, len(buff))
 	}
 	if len(buff) == 0 {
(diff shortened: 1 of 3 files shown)


p2p/libp2p/netMessenger.go
```diff
@@ -718,7 +718,7 @@ func (netMes *networkMessenger) BroadcastOnChannelBlocking(channel string, topic
 
 func (netMes *networkMessenger) checkSendableData(buff []byte) error {
 	if len(buff) > maxSendBuffSize {
-		return p2p.ErrMessageTooLarge
+		return fmt.Errorf("%w, maximum: %d, received %d", p2p.ErrMessageTooLarge, maxSendBuffSize, len(buff))
```
bogdan-rosianu, line 721: maybe switch order? received and then maximum? also, you can specify the unit (number of bytes). same above

bogdan-rosianu (approved):

p2p/libp2p/netMessenger.go
bogdan-rosianu: also fix failing tests (replace with `errors.Is` and maybe add a `strings.Contains()` for extra check)

iulianpascalau: ok, this PR will close, will take into account your comment :)

iulianpascalau:
changes and fixes are done in #2259

PR #2263, opened 2020-08-31
Handle concurrency issues occurred with building hyperblock coordinates (using the fast-replay method)

andreibancioiu:
By exploring the fastly-rebuilt database lookup extensions, we can confirm that some info is missing from the miniblocks metadata.

```
missing "destination" hyperblock coordinates: 27 miniblocks
missing "source and destination" hyperblock coordinates: 5 miniblocks (around changes of epochs)
```
(data in 3 shards)

Under the assumption of concurrent commits and "notarized" notifications, I've redesigned  the way notifications are processed. All notifications are now added to a pending state, and consumed as soon as possible (both at commit and at notify).

Furthermore, `historyRepository` registers to a brand new, custom handler of the block tracker - by @SebastianMarian, in order to always receive "notarized" notifications, along with the `metaBlock`.


core/dblookupext/historyRepository.go
```diff
@@ -280,72 +280,95 @@ func (hr *historyRepository) onNotarizedMiniblock(metaBlockNonce uint64, metaBlo
...
+		if !ok {
+			continue
 		}
-	}
 
-	err = hr.putMiniblockMetadata(miniblockHash, metadata)
-	if err != nil {
-		log.Warn("onNotarizedMiniblock(): cannot update miniblock metadata", "miniblockHash", miniblockHash, "err", err)
-		return
+		miniblockHash := []byte(key.(string))
```
AdoAdoAdo, line 355: could the type assertions panic?

andreibancioiu: Should never. Added checks though.

cmd/node/config/config.toml
```diff
@@ -645,7 +645,7 @@
     PollingIntervalInMinutes = 65
 
 [DbLookupExtensions]
-    Enabled = false
+    Enabled = true # TODO: Disable after tests
```
AdoAdoAdo, line 648: remember to disable after tests.

core/dblookupext/historyRepository.go
```diff
@@ -126,25 +136,15 @@ func (hr *historyRepository) recordMiniblock(blockHeaderHash []byte, blockHeader
...
-	// Here we need to use queued notifications
-	notification, ok := hr.notarizedAtSourceNotifications.Get(string(miniblockHash))
-	if ok {
-		notificationTyped := notification.(*notarizedAtSourceNotification)
-		miniblockMetadata.NotarizedAtSourceInMetaNonce = notificationTyped.metaNonce
-		miniblockMetadata.NotarizedAtSourceInMetaHash = notificationTyped.metaHash
-
-		hr.notarizedAtSourceNotifications.Remove(string(miniblockHash))
+		hr.pendingNotarizedAtBothNotifications.Set(string(miniblockHash), &notarizedNotification{
+			metaNonce: blockHeader.GetNonce(),
```
SebastianMarian, line 145: blockHeader is always a meta block?

SebastianMarian (approved):

core/dblookupext/historyRepository.go
```diff
@@ -229,16 +229,14 @@ func (hr *historyRepository) RegisterToBlockTracker(blockTracker BlockTracker) {
 	blockTracker.RegisterSelfNotarizedHeadersHandler(hr.onNotarizedBlocks)
 }
 
+// This is called both with "shardID" == "metachain" (regular notarization events),
+// but also with "shardID" IN [0, 1, 2, ...] when blocks produced by the metachain (as a source) are notarized.
 func (hr *historyRepository) onNotarizedBlocks(shardID uint32, headers []data.HeaderHandler, headersHashes [][]byte) {
-	log.Trace("onNotarizedBlocks()", "shardID", shardID, "len(headers)", len(headers))
-
-	if shardID != core.MetachainShardId {
```
AdoAdoAdo, line 235: oNotarizedBlocks will be called:
- (twice) due to line 228 when node receives a header from a different shard than its own
     a) with a list of different shard headers
     b) with a list of own headers
- due to line 229 when node receives a header from own shard
    a) with a list of own headers

I think the condition was added before to not care about shard headers, but only about meta headers, no matter if it was called with a list of own headers or cross headers. 

I am trying to think about the situation when we would want this called also for shard headers, but right now I don't really see any, maybe I'm missing something.

@SebastianMarian please confirm

core/dblookupext/historyRepository.go
```diff
@@ -272,40 +270,42 @@ func (hr *historyRepository) onNotarizedInRegularBlockOfMeta(blockOfMetaNonce ui
 func (hr *historyRepository) onNotarizedMiniblock(metaBlockNonce uint64, metaBlockHash []byte, shardOfContainingBlock uint32, miniblockHeader block.MiniBlockHeader) {
 	miniblockHash := miniblockHeader.Hash
 	isIntra := miniblockHeader.SenderShardID == miniblockHeader.ReceiverShardID
-	notarizedAtSource := miniblockHeader.SenderShardID == shardOfContainingBlock
+	isFromMeta := miniblockHeader.SenderShardID == core.MetachainShardId
 	isToMeta := miniblockHeader.ReceiverShardID == core.MetachainShardId
+	isNotarizedAtBoth := isIntra || isToMeta || isFromMeta
```
AdoAdoAdo, line 275: not sure if isFromMeta is sufficient for the rewards miniblocks. 
maybe notarizedAtDestFromMeta? which should be isNotarizedAtDestination && isFromMeta

AdoAdoAdo (changes requested):

core/dblookupext/historyRepository.go
SebastianMarian: Your description is ok, and this condition was to care about only what meta blocks say (if you are a meta node you only care about your blocks when we talk about setting the final block. And if you are in shard you also care about what meta says about your shard headers when you are trying to set the final block)

core/dblookupext/historyRepository.go
andreibancioiu: Fixed as discussed (now notification for rewards is received correctly). Rewards handled as "normal cross-shard" transactions.

core/dblookupext/historyRepository.go
andreibancioiu: Fixed as discussed internally (removed "simulated notification").

core/dblookupext/historyRepository.go
andreibancioiu: Fixed as discussed, using the new handler (notifier).

core/dblookupext/historyRepository.go
```diff
@@ -227,57 +216,34 @@ func (hr *historyRepository) RegisterToBlockTracker(blockTracker BlockTracker) {
...
-		}
-
-		log.Error("onNotarizedBlocks(): unexpected type of header", "type", fmt.Sprintf("%T", headerHandler))
 	}
 }
 
 func (hr *historyRepository) onNotarizedInMetaBlock(metaBlockNonce uint64, metaBlockHash []byte, shardData *block.ShardData) {
-	for _, miniblockHeader := range shardData.GetShardMiniBlockHeaders() {
-		hr.onNotarizedMiniblock(metaBlockNonce, metaBlockHash, shardData.GetShardID(), miniblockHeader)
+	if metaBlockNonce <= 1 {
```
SebastianMarian, line 241: < 1 instead <= 1. If meta misses some rounds it can include in nonce 1 a lot of shardData if the shards worked good

andreibancioiu: Fixed.

SebastianMarian (approved):

core/dblookupext/historyRepository.go
```diff
@@ -225,127 +216,152 @@ func (hr *historyRepository) RegisterToBlockTracker(blockTracker BlockTracker) {
...
 	miniblockHash := miniblockHeader.Hash
 	isIntra := miniblockHeader.SenderShardID == miniblockHeader.ReceiverShardID
-	notarizedAtSource := miniblockHeader.SenderShardID == shardOfContainingBlock
+	isFromMeta := miniblockHeader.SenderShardID == core.MetachainShardId
 	isToMeta := miniblockHeader.ReceiverShardID == core.MetachainShardId
-
-	iDontCare := miniblockHeader.SenderShardID != hr.selfShardID && miniblockHeader.ReceiverShardID != hr.selfShardID
+	isNotarizedAtSource := miniblockHeader.SenderShardID == shardOfContainingBlock
+	isNotarizedAtDestination := miniblockHeader.ReceiverShardID == shardOfContainingBlock
+	isNotarizedAtBoth := isIntra || isToMeta || (isFromMeta && isNotarizedAtDestination)
```
AdoAdoAdo, line 257: I suggested the last condition, but thinking a bit about it I don't think it is correct.
isNotarizedAtBoth should cover the miniblocks that are notarized in a single metablock for both source and destination. 

- for isIntra this is true, as bot sender and destination solved
- for isToMeta  this is again true as when included in metablock it means that also destination (meta) was executed in adition to source notarization
- for isFromMeta && notarized at destination, this is not true, as the sender notarizationi (metachain) is done in a different metablock, where metachain is the source. For the destination should follow the isNotarizedAtDestination logic.

Was there a problem maybe that the from metachain miniblocks were not reported on source?
Checking the function onNotarizedInMetablock only the shardData.ShardMiniblockHeaders (for every shard) are reported. 
To cover also the from metachain miniblocks at source, also the metablock.MiniBlockHeaders should be reported.

AdoAdoAdo (changes requested):

core/dblookupext/historyRepository.go
andreibancioiu: Makes total sense, fixed & tested.

cmd/node/config/config.toml
andreibancioiu: Disabled.

AdoAdoAdo (approved):

LucianMincu (approved):
System tests passed.

SebastianMarian (approved):

PR #2266, opened 2020-09-02
DNS test scenarios

sasurobert:
a set of DNS test scenarios - all works as expected


cmd/node/config/genesisSmartContracts.json
```diff
@@ -11,8 +11,8 @@
     "owner": "erd16grmckn46ry7fwyvass8e8pz88klazzpc0c5f0pnrv643td4797sgnvjkm",
     "filename": "./config/genesisContracts/dns.wasm",
     "vm-type": "0500",
-    "init-parameters": "056bc75e2d63100000",
+    "init-parameters": "00",
```
camilbancioiu, line 14: What has changed here?

process/smartContract/builtInFunctions/saveUserName.go
```diff
@@ -60,7 +58,7 @@ func (s *saveUserName) ProcessBuiltinFunction(
 	if !ok {
 		return nil, process.ErrCallerIsNotTheDNSAddress
 	}
-	if len(vmInput.Arguments) != 1 || len(vmInput.Arguments[0]) != userNameHashLength {
+	if len(vmInput.Arguments) != 1 {
```
camilbancioiu, line 61: A maximum length would be good here.

bogdan-rosianu (approved):

process/smartContract/builtInFunctions/saveUserName.go
sasurobert: No need. We control the DNS SCs which are only allowed to call this

camilbancioiu (approved):

LucianMincu (approved):
System tests passed.

PR #2268, opened 2020-09-03
Fix set of final block after restart in epoch

SebastianMarian:
* Implemented a mechanism of setting the final checkpoint to the last checkpoint added in list, when a node starts in epoch and it has only one block (start of epoch block) added in fork detector. This should prevent the possible rollbacks behind final block which would been done if the node would not been received the next requested block from the network in some few number of rounds.


sasurobert (approved):

sasurobert (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2269, opened 2020-09-03
Optimize db lookup extensions while sync & fast-reply-database

andreibancioiu:
- Call `RecordBlock` synchronously with `Commit()`. Consume notifications only on "receive notifications" goroutines.
 - ~~Hold LRU caches for recently handled records so that performance at sync is improved~~ not needed anymore
 - Remove dependency from block tracker's notifications. Notification performed at commit time, by `shardBlock` / `metaBlock` processor.

Handle repeated commits of same block (protect against overriding the `miniblockMetadata`) / [EN-7520](https://elrondnetwork.atlassian.net/browse/EN-7520). Edge case was:

 - routine A -  commit → save to storer
 - routine B - receive notification, put it in map at key `Foo`, start processing notifications (in progress), notification is seen, still in map
 - routine A - then commit again same block
 - routine C - receive notification, put in map at same key `Foo`
 - routine B - finish processing notification, remove `Foo` from map
 - routine A - put in storer, override what was written at B.


core/dblookupext/historyRepository.go
```diff
@@ -50,6 +50,7 @@ type historyRepository struct {
 	// that could mistakenly override the "patch()" operations performed when consuming notarization notifications.
 	deduplicationCacheForInsertMiniblockMetadata storage.Cacher
 
+	// Question for review: this can be removed now, using the new approach (without the block tracker), right?
```
SebastianMarian: The only situation when the notifications will be redundant will be when a node will do a rollback. In this situation the same final block/blocks will be notified again, after another new block will be constructed upon it. But this case should not be a regular pattern and could be considered as an exception. If this redundant notifications will not break something in the implementation logic and the below method was used just as an optimization, I would say that you can remove it now, considering the new approach.

core/dblookupext/historyRepository.go
```diff
@@ -246,12 +247,14 @@ func (hr *historyRepository) OnNotarizedBlocks(shardID uint32, headers []data.He
 	for i, headerHandler := range headers {
 		headerHash := headersHashes[i]
 
+		log.Debug("onNotarizedBlocks():", "shardID", shardID, "nonce", headerHandler.GetNonce(), "headerHash", headerHash, "type", fmt.Sprintf("%T", headerHandler))
+
+		// Question for review: this optimization is not required anymore, right?
```
SebastianMarian: The only situation when the notifications will be redundant will be when a node will do a rollback. In this situation the same final block/blocks will be notified again, after another new block will be constructed upon it. But this case should not be a regular pattern and could be considered as an exception. If this redundant notifications will not break something in the implementation logic and the below method was used just as an optimization, I would say that you can remove it now, considering the new approach.

core/dblookupext/historyRepository.go
andreibancioiu: The implementation of `OnNotarizedBlocks()` is all right with being called twice with some redundant data in case of rollbacks.

AdoAdoAdo (changes requested):

core/dblookupext/historyRepository.go
```diff
@@ -158,6 +176,14 @@ func (hr *historyRepository) computeMiniblockHash(miniblock *block.MiniBlock) ([
 	return core.CalculateHash(hr.marshalizer, hr.hasher, miniblock)
 }
 
+func (hr *historyRepository) hasRecentlyInsertedMiniblockMetadata(miniblockHash []byte) bool {
+	return hr.deduplicationCacheForInsertMiniblockMetadata.Has(miniblockHash)
```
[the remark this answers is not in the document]
andreibancioiu: Fixed by using a composite key for the deduplication cache. Is it all right?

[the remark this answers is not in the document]
AdoAdoAdo: looks good

SebastianMarian (approved):

AdoAdoAdo (approved):

SebastianMarian (approved):

LucianMincu (approved):
System tests passed.

PR #2277, opened 2020-09-07
Fixed meta smart contract processor arguments

iulianpascalau:
- passed the enabled epoch configs to the smart contract processor for the metachain nodes
This caused the mainnet import DB folder to cause a stuck situation on a meta node @ nonce 12801

Manual testing procedure:
import a meta node full history from epoch 0. The process should work (should get pass of the nonce 12801)


sasurobert (approved):

raduchis (approved):

LucianMincu (approved):
System tests passed.

PR #2279, opened 2020-09-08
Bug fix panic tps benchmark

miiu96:
Fix concurrent access on map that store shard statistics information for TPS Benchmark. 
When method ShardStatistics() is called will return a copy of map that contains shard statistics.

### DIFF
--- a/core/statistics/tpsBenchmark.go
+++ b/core/statistics/tpsBenchmark.go
@@ -179,13 +179,13 @@ func (s *TpsBenchmark) NrOfShards() uint32 {
 // ShardStatistics returns the current statistical state for a given shard
 func (s *TpsBenchmark) ShardStatistics() map[uint32]ShardStatistic {
 	s.mut.RLock()
-	copyMapShardStatistics := make(map[uint32]ShardStatistic)
+	shardStatisticsMapCopy := make(map[uint32]ShardStatistic)
 	for key, value := range s.shardStatistics {
-		copyMapShardStatistics[key] = value
+		shardStatisticsMapCopy[key] = value
 	}
 	s.mut.RUnlock()
 
-	return copyMapShardStatistics
+	return shardStatisticsMapCopy
 }
 
 // ShardStatistic returns the current statistical state for a given shard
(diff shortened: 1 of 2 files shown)


core/statistics/tpsBenchmark.go
```diff
@@ -179,8 +179,13 @@ func (s *TpsBenchmark) NrOfShards() uint32 {
 // ShardStatistics returns the current statistical state for a given shard
 func (s *TpsBenchmark) ShardStatistics() map[uint32]ShardStatistic {
 	s.mut.RLock()
-	defer s.mut.RUnlock()
-	return s.shardStatistics
+	copyMapShardStatistics := make(map[uint32]ShardStatistic)
```
andreibancioiu, line 182: Seems fine, I see that `updateStatistics` is protected by the same lock.

andreibancioiu (approved):

core/statistics/tpsBenchmark.go
bogdan-rosianu: rename to `shardStatisticsMapCopy`

core/statistics/tpsBenchmark_test.go
```diff
@@ -636,3 +636,47 @@ func TestTpsBenchmark_ZeroTxMetaBlockAndEmptyShardHeader(t *testing.T) {
 	bigTxCount := big.NewInt(0)
 	assert.Equal(t, bigTxCount, tpsBenchmark.TotalProcessedTxCount())
 }
+
+func TestTpsBenchmark_ShardStatisticConcurrentAccess(t *testing.T) {
+	t.Parallel()
+
+	tpsBenchmark, _ := statistics.NewTPSBenchmark(12, 4)
```
bogdan-rosianu, line 643: use recover() as you should check that no panic has encountered during the test

bogdan-rosianu (changes requested):

core/statistics/tpsBenchmark_test.go
miiu96: done.

core/statistics/tpsBenchmark.go
miiu96: done.

bogdan-rosianu (approved):

andreibancioiu (approved):

LucianMincu (approved):
System tests passed

PR #2287, opened 2020-09-14
wrapped vm output api response

bogdan-rosianu:
created a wrapped structure which will be returned when calling a `vm-values` endpoint. Now it will display addresses in bech32 format when needed and added custom json tags (with lowercase)

Testing procedure:
- run a testnet
- make a POST request to `vm-values/query` for example and the output should follow json format and not having unreadable characters

### DIFF
--- a/data/vm/vmOutputApi.go
+++ b/data/vm/vmOutputApi.go
@@ -1,0 +1,72 @@
+package vm
+
+import (
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
+)
+
+// VMOutputApi is a wrapper over the vmcommon's VMOutput
+type VMOutputApi struct {
+	ReturnData      [][]byte                     `json:"returnData"`
+	ReturnCode      string                       `json:"returnCode"`
+	ReturnMessage   string                       `json:"returnMessage"`
+	GasRemaining    uint64                       `json:"gasRemaining"`
+	GasRefund       *big.Int                     `json:"gasRefund"`
+	OutputAccounts  map[string]*OutputAccountApi `json:"outputAccounts"`
+	DeletedAccounts [][]byte                     `json:"deletedAccounts"`
+	TouchedAccounts [][]byte                     `json:"touchedAccounts"`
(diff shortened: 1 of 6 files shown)


data/vm/vmOutputApi.go
```diff
@@ -0,0 +1,72 @@
...
+// LogEntryApi is a wrapper over vmcommon's LogEntry
+type LogEntryApi struct {
+	Identifier []byte   `json:"identifier"`
+	Address    string   `json:"address"`
+	Topics     [][]byte `json:"topics"`
+	Data       []byte   `json:"data"`
+}
+
+// GetFirstReturnData is a helper function that returns the first ReturnData of VMOutput, interpreted as specified.
+func (vmOutput *VMOutputApi) GetFirstReturnData(asType vmcommon.ReturnDataKind) (interface{}, error) {
```
iulianpascalau, line 53: unit test?

facade/nodeFacade.go
```diff
@@ -411,6 +418,65 @@ func (nf *nodeFacade) GetNumCheckpointsFromPeerState() uint32 {
 	return nf.peerState.GetNumCheckpoints()
 }
 
+func (nf *nodeFacade) convertVmOutputToApiResponse(input *vmcommon.VMOutput) *vm.VMOutputApi {
+	outputAccounts := make(map[string]*vm.OutputAccountApi)
+	for key, acc := range input.OutputAccounts {
+		outPutAddress, err := nf.node.EncodeAddressPubkey(acc.Address)
```
iulianpascalau, line 424: outputAddress*

SebastianMarian (approved):

data/vm/vmOutputApi.go
bogdan-rosianu: added unit tests for all cases (as big int, as big int string, as hex and as string)

facade/nodeFacade.go
bogdan-rosianu: done

iulianpascalau (approved):

SebastianMarian (approved):

LucianMincu (approved):
System tests passed.

PR #2288, opened 2020-09-14
Unit test elastic indexer 

miiu96:
Unit test for package core/indexer


core/indexer/elasticProcessor.go
```diff
@@ -237,7 +237,7 @@ func (ei *elasticProcessor) RemoveMiniblocks(header data.HeaderHandler, body *bl
 
 		isDstMe := selfShardID == miniblock.ReceiverShardID
 		isCrossShard := miniblock.ReceiverShardID != miniblock.SenderShardID
-		if !(isDstMe && isCrossShard) {
+		if isDstMe && isCrossShard {
```
SebastianMarian, line 240: So the condition is completely reversed. The bulk remove is not done anymore for cross miniblocks wit dest me.

miiu96: Yes. Now is right because if a miniblock is indexed on source shard and if a rollback appears cross shard miniblocks should not be removed because I have to keep information that was indexed on source shard. When miniblock will be executed again on destination shard information will be rewritten.

SebastianMarian (approved):

sasurobert (approved):

PR #2289, opened 2020-09-15
standard tx status responses

bogdan-rosianu:
Changed transactions' statuses for indexer and API transaction status fetcher. Now they use the same messages from the same place. 

Changes:
**Elastic Search** responses:

|    Before        |     After      |
|------------------ | -------------- |
|  Invalid          | invalid        |  
|  Pending       | pending      |
|Not Executed|unsuccessful|
|  Success       | successful |

**API transaction status** responses:

|    Before        |     After      |
|------------------ | -------------- |
|invalid|invalid|
|  received       | pending |
|  partially-executed       | pending      |
|  not-executed          | unsuccessful        |
|executed|successful|

Meanings:
_Invalid/invalid_: the transaction is invalid and won't be processed. will be changed to **invalid**
_partially-executed_: executed only in sender shard. in order to abstract sharding terminology, it will be changed to **pending**
_received_: in the pool on sender shard, but not executed yet. will be changed to **pending**
_Not Executed/not-executed_: processing of the transaction failed. will be changed to **unsuccessful**
_Success/executed_: fully-executed transaction. will be changed to **successful** for consistency


SebastianMarian (approved):

data/transaction/status.go
```diff
@@ -9,18 +9,21 @@ import (
...
-	// TxStatusPartiallyExecuted = received and executed on source shard
-	TxStatusPartiallyExecuted TxStatus = "partially-executed"
-	// TxStatusExecuted = received and executed
-	TxStatusExecuted TxStatus = "executed"
-	// TxStatusNotExecuted = received and executed with error
-	TxStatusNotExecuted TxStatus = "not-executed"
+	// TxStatusPending = received and maybe executed on source shard, but not on destination shard
+	TxStatusPending TxStatus = "pending"
+	// TxStatusSuccessful = received and executed
+	TxStatusSuccessful TxStatus = "successful"
```
andreibancioiu, line 15: Breaking change for the exchanges consuming the API. But we will deal with in on the Proxy side somwhow :partying_face: 

Perhaps we can introduce a `v2` prefix in the URL on the Proxy (and re-map to old names in v1 URLs).

node/txsimulator/txSimulator.go
```diff
@@ -55,16 +55,16 @@ func NewTransactionSimulator(args ArgsTxSimulator) (*transactionSimulator, error
...
 // ProcessTx will process the transaction in a special environment, where state-writing is not allowed
 func (ts *transactionSimulator) ProcessTx(tx *transaction.Transaction) (*transaction.SimulationResults, error) {
-	txStatus := transaction.TxStatusReceived
+	txStatus := transaction.TxStatusPending
 	failReason := ""
 	retCode, err := ts.txProcessor.ProcessTransaction(tx)
 	if err != nil {
 		failReason = err.Error()
-		txStatus = transaction.TxStatusNotExecuted
+		txStatus = transaction.TxStatusUnsuccessful
```
andreibancioiu, line 63: Actually, here it is really `non-executed`. But `Unsuccessful` should be all right for most purposes, correct?

andreibancioiu (approved):

node/txsimulator/txSimulator.go
bogdan-rosianu: yes, it can remain like this. this type of status would have been returned also on the normal transaction sending

data/transaction/status.go
bogdan-rosianu: added a jira ticket with ID `EN-7718`

LucianMincu (approved):
System tests passed.

PR #2292, opened 2020-09-16
EN-6936: log route on seednode

bogdan-rosianu:
Added the `/log` route inside the seednode. If enabled (Rest API interface is not set to "off") then the logs can be fetched via the `logviewer` app. 

Testing procedure:
-run the seednode. Default API interface is `localhost:8080` and can be replaced with the `--rest-api-interface X` flag 
-try to connect via the `logviewer` app: `./logviewer -address 127.0.0.1:8080` . If seednode's Rest API interface is not "off", it should work and fetch logs real time. If seednode's Rest API interface was set to "off" (`./seednode --rest-api-interface off`), then no connection should be made.

### DIFF
--- a/cmd/seednode/api/api.go
+++ b/cmd/seednode/api/api.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"fmt"
 	"net/http"
 
 	logger "github.com/ElrondNetwork/elrond-go-logger"
@@ -15,13 +14,13 @@ import (
 var log = logger.GetOrCreate("seednode/api")
 
 // Start will boot up the api and appropriate routes, handlers and validators
-func Start(port int, marshalizer marshal.Marshalizer) error {
+func Start(restApiInterface string, marshalizer marshal.Marshalizer) error {
 	ws := gin.Default()
 	ws.Use(cors.Default())
 
 	registerRoutes(ws, marshalizer)
 
-	return ws.Run(fmt.Sprintf(":%d", port))
+	return ws.Run(restApiInterface)
 }
 
 func registerRoutes(ws *gin.Engine, marshalizer marshal.Marshalizer) {
--- a/cmd/seednode/api/api.go
+++ b/cmd/seednode/api/api.go
@@ -1,0 +1,53 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	logger "github.com/ElrondNetwork/elrond-go-logger"
+	"github.com/ElrondNetwork/elrond-go/api/logs"
+	"github.com/ElrondNetwork/elrond-go/marshal"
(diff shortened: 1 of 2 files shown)


cmd/seednode/api/api.go
```diff
@@ -0,0 +1,53 @@
...
+var log = logger.GetOrCreate("seednode/api")
+
+// Start will boot up the api and appropriate routes, handlers and validators
+func Start(port int, marshalizer marshal.Marshalizer) error {
+	ws := gin.Default()
+	ws.Use(cors.Default())
+
+	registerRoutes(ws, marshalizer)
+
+	return ws.Run(fmt.Sprintf(":%d", port))
```
iulianpascalau, line 24: We need to specify the interface, just like in elrond-go node

cmd/seednode/main.go
```diff
@@ -49,6 +50,12 @@ VERSION:
 			"`0, 10230, 15670` or range of ports such as `5000-10000`",
 		Value: "10000",
 	}
+	restApiPortFlag = cli.IntFlag{
```
iulianpascalau, line 53: we should be able to define the interface as well

cmd/seednode/main.go
```diff
@@ -147,6 +155,16 @@ func startNode(ctx *cli.Context) error {
 		}
 	}
 
+	restApiPort := ctx.GlobalInt(restApiPortFlag.Name)
```
iulianpascalau, line 158: check if set

cmd/seednode/main.go
```diff
@@ -147,6 +155,16 @@ func startNode(ctx *cli.Context) error {
...
+	if restApiPort > 0 {
+		go func() {
```
iulianpascalau, line 160: define a function instead of the anonymous one?

iulianpascalau (changes requested):

cmd/seednode/api/api.go
bogdan-rosianu: done

cmd/seednode/main.go
bogdan-rosianu: done

cmd/seednode/main.go
bogdan-rosianu: no need for this. it was an int flag (after review it is a string flag) so we cannot tell that the flag was set or the default value was used.

cmd/seednode/main.go
bogdan-rosianu: done. extracted functions

iulianpascalau (approved):

SebastianMarian (approved):

LucianMincu (approved):
System tests passed.

PR #2301, opened 2020-09-17
Limit the number of the same consensus message type from the same pk

SebastianMarian:
* Implemented a mechanism which limits the number of the same consensus message type which could be received/accepted in one round from the same public key


sasurobert (dismissed):

consensus/spos/bls/subroundStartRound.go
```diff
@@ -27,6 +28,7 @@ func NewSubroundStartRound(
 	extend func(subroundId int),
 	processingThresholdPercentage int,
 	executeStoredMessages func(),
```
iulianpascalau, line 30: Maybe we can aggregate these function handlers arguments in an interface (for better code readability) somewhere in the future :)

consensus/spos/consensusMessageValidator.go
```diff
@@ -290,3 +304,33 @@ func (wrk *Worker) checkMessageWithFinalInfoValidity(cnsMsg *consensus.Message)
 
 	return nil
 }
+
+func (wrk *Worker) isMessageTypeLimitReached(pk []byte, msgType consensus.MessageType) bool {
+	wrk.mutPkConsensusMessages.RLock()
+	defer wrk.mutPkConsensusMessages.RUnlock()
+
+	mapMsgType, ok := wrk.mapPkConsensusMessages[string(pk)]
```
iulianpascalau, line 312: Maybe we should have the key as a concatenation between pk and round. This will take care of the edgecases in which a node receives 2 messages of the same type but in consecutive rounds caused by the small differences between synced clocks.

consensus/spos/worker.go
```diff
@@ -73,6 +73,9 @@ type Worker struct {
 	publicKeySize       int
 	publicKeyBitmapSize int
 	cancelFunc          func()
+
+	mutPkConsensusMessages sync.RWMutex
```
iulianpascalau, line 77: the way the code looks now, a better approach (IMHO) would have been to extract these 2 fields & the methods `Worker.addMessageTypeToPublicKey`, `Worker.isMessageTypeLimitReached` and `Worker.ResetConsensusMessages` into a separate component (in this way Worker struct wouldn't grow so much)

consensus/spos/bls/subroundStartRound.go
SebastianMarian: Yep, probably in the next task which will touch this part

consensus/spos/consensusMessageValidator.go
SebastianMarian: We can do this just to be super safe, but if the node receives a consensus message with the previous round inside, it will drop this message in checkConsensusMessageValidity method in line 66 so it will not be added in this map

consensus/spos/consensusMessageValidator.go
```diff
@@ -305,11 +361,13 @@ func (wrk *Worker) checkMessageWithFinalInfoValidity(cnsMsg *consensus.Message)
...
 
-func (wrk *Worker) isMessageTypeLimitReached(pk []byte, msgType consensus.MessageType) bool {
-	wrk.mutPkConsensusMessages.RLock()
-	defer wrk.mutPkConsensusMessages.RUnlock()
+func (cmv *consensusMessageValidator) isMessageTypeLimitReached(pk []byte, round int64, msgType consensus.MessageType) bool {
+	cmv.mutPkConsensusMessages.RLock()
+	defer cmv.mutPkConsensusMessages.RUnlock()
 
-	mapMsgType, ok := wrk.mapPkConsensusMessages[string(pk)]
+	key := fmt.Sprintf("%s_%d", string(pk), round)
```
iulianpascalau, line 368: :+1:

iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2310, opened 2020-09-21
Sync tracker pool with headers pool

SebastianMarian:
* Sync tracker pool with headers pool LRU cacher when missing headers from tracker are requested (in this situation a delete will be done in headers pool, just to be sure that the notifiy event will be called for the requested/received header, even if in some edge case situations the requested header exists only in headers pool but not in tracker pool)

This situation could appear only in some rare events of restarting the node after more than 1000 rounds from its last commit

### DIFF
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -255,7 +255,9 @@ func (bp *baseProcessor) requestHeadersIfMissing(
 		missingNonces = append(missingNonces, nonces...)
 	}
 
+	headersPool := bp.dataPool.Headers()
 	for _, nonce := range missingNonces {
+		headersPool.RemoveHeaderByNonceAndShardId(nonce, shardId)
 		go bp.requestHeaderByShardAndNonce(shardId, nonce)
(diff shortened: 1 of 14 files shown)


sasurobert (dismissed):

process/block/baseProcess.go
```diff
@@ -255,7 +255,9 @@ func (bp *baseProcessor) requestHeadersIfMissing(
 		missingNonces = append(missingNonces, nonces...)
 	}
 
+	headersPool := bp.dataPool.Headers()
```
iulianpascalau, line 258: some tests for the added code would be nice.

SebastianMarian: would be added until the system tests are done

iulianpascalau: great

sasurobert (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2313, opened 2020-09-22
bugFix/failed-relayed-with-value

sasurobert:
fix relayed tx when transaction fails and the value has to be sent back to the relayer. With old code the shard would stuck.

### DIFF
--- a/process/smartContract/process.go
+++ b/process/smartContract/process.go
@@ -518,7 +518,7 @@ func (sc *scProcessor) ProcessIfError(
 		return err
 	}
 
-	err = sc.processForRelayerWhenError(tx, returnMessage)
+	err = sc.processForRelayerWhenError(tx, txHash, returnMessage)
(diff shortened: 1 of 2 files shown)


SebastianMarian (approved):

process/smartContract/process.go
```diff
@@ -561,6 +562,7 @@ func (sc *scProcessor) processForRelayerWhenError(
 		RcvAddr:        relayedSCR.RelayerAddr,
 		SndAddr:        relayedSCR.RcvAddr,
 		OriginalTxHash: relayedSCR.OriginalTxHash,
+		PrevTxHash:     txHash,
```
iulianpascalau, line 565: Good we do not have this activated on the mainnet :D

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2316, opened 2020-09-22
Add missed headers from headers pool into tracker pool

SebastianMarian:
* Implemented a mechanism which adds into tracker pool the needed headers which are not found there but they exist in headers pool

### DIFF
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -966,7 +966,6 @@ func (mp *metaProcessor) requestShardHeadersIfNeeded(
 	hdrsAddedForShard map[uint32]uint32,
 	lastShardHdr map[uint32]data.HeaderHandler,
 ) {
-	headersPool := mp.dataPool.Headers()
 	for shardID := uint32(0); shardID < mp.shardCoordinator.NumberOfShards(); shardID++ {
 		log.Debug("shard headers added",
 			"shard", shardID,
@@ -979,7 +978,7 @@ func (mp *metaProcessor) requestShardHeadersIfNeeded(
 			fromNonce := lastShardHdr[shardID].GetNonce() + 1
 			toNonce := fromNonce + uint64(mp.shardBlockFinality)
 			for nonce := fromNonce; nonce <= toNonce; nonce++ {
-				headersPool.RemoveHeaderByNonceAndShardId(nonce, shardID)
+				mp.addHeaderIntoTrackerPool(nonce, shardID)
 				go mp.requestHandler.RequestShardHeaderByNonce(shardID, nonce)
 			}
 		}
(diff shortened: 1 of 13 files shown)


iulianpascalau (dismissed):
Tested also in import-db mode. Works as expected.

iulianpascalau (approved):

process/block/shardblock.go
```diff
@@ -1611,7 +1611,7 @@ func (sp *shardProcessor) createAndProcessMiniBlocksDstMe(
 	}
 	sp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 
-	sp.requestMetaHeadersIfNeeded(hdrsAdded, lastMetaHdr)
+	go sp.requestMetaHeadersIfNeeded(hdrsAdded, lastMetaHdr)
```
sasurobert, line 1614: no need for go routine

process/track/blockProcessor.go
```diff
@@ -231,7 +231,7 @@ func (bp *blockProcessor) ComputeLongestChain(shardID uint32, header data.Header
 	var sortedHeadersHashes [][]byte
 
 	defer func() {
-		bp.requestHeadersIfNeeded(header, sortedHeaders, headers)
+		go bp.requestHeadersIfNeeded(header, sortedHeaders, headers)
```
sasurobert, line 234: I think there is a go routine in that function - thus there is no need for this. or delete go routine in the underlying funciton

sasurobert (changes requested):

process/block/metablock.go
```diff
@@ -957,7 +957,7 @@ func (mp *metaProcessor) createAndProcessCrossMiniBlocksDstMe(
 	}
 	mp.hdrsForCurrBlock.mutHdrsForBlock.Unlock()
 
-	mp.requestShardHeadersIfNeeded(hdrsAddedForShard, lastShardHdr)
+	go mp.requestShardHeadersIfNeeded(hdrsAddedForShard, lastShardHdr)
```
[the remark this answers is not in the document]
SebastianMarian: Actually I have added because the call of the new method addHeaderIntoTrackerPool which is inside of underline method. I will remove go from there

process/block/shardblock.go
SebastianMarian: Actually I have added because the call of the new method addHeaderIntoTrackerPool which is inside of underline method. I will remove go from there

process/track/blockProcessor.go
SebastianMarian: Actually I have added because the call of the new method AddHeaderFromPool which has been added in underlying method. I will remove go from there

iulianpascalau (commented):
Looks ok, pending test result....

sasurobert (approved):
much better.

iulianpascalau (approved):
initial import-db test passed. Sync time ~7-8h on shard 0

LucianMincu (approved):
System tests passed.

PR #2317, opened 2020-09-23
Feat/full history nodes

iulianpascalau:
- added the possibility to run nodes in "full history nodes" as to provide all the ElrondNetwork blockchain history. This feature will allow other nodes to become full history nodes just by connecting them to the network. We expect this to be a lengthy process but is a full automatic process able to completely re-check all previous history and re-validate the chain.


gabi-vuls (approved):
System test passed

sasurobert (approved):

raduchis (approved):

PR #2322, opened 2020-09-25
set-to-jail-from-inactive

sasurobert:
set to jailed when account is inactive and with infinite unstaked epoch after jailing is enabled.

### DIFF
--- a/process/peer/process.go
+++ b/process/peer/process.go
@@ -645,11 +645,7 @@ func (vs *validatorStatistics) ResetValidatorStatisticsAtNewEpoch(vInfos map[uin
 				return process.ErrWrongTypeAssertion
 			}
 			peerAccount.ResetAtNewEpoch()
-
-			shouldBeJailed := vs.flagJailedEnabled.IsSet() && validator.List == string(core.JailedList) && peerAccount.GetList() != string(core.JailedList)
-			if shouldBeJailed {
-				peerAccount.SetListAndIndex(validator.ShardId, validator.List, validator.Index)
-			}
+			vs.setToJailedIfNeeded(peerAccount, validator)
 
 			err = vs.peerAdapter.SaveAccount(peerAccount)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


process/peer/process.go
```diff
@@ -661,6 +657,24 @@ func (vs *validatorStatistics) ResetValidatorStatisticsAtNewEpoch(vInfos map[uin
...
+		return
+	}
+
+	if validator.List == string(core.JailedList) && peerAccount.GetList() != string(core.JailedList) {
+		peerAccount.SetListAndIndex(validator.ShardId, string(core.JailedList), validator.Index)
+		return
+	}
+
+	if peerAccount.GetUnStakedEpoch() == core.DefaultUnstakedEpoch && peerAccount.GetList() == string(core.InactiveList) {
+		peerAccount.SetListAndIndex(validator.ShardId, string(core.JailedList), validator.Index)
```
SebastianMarian, line 674: The same set here and in line 669. Why not extract both conditions in shouldBeJailed variable and reduce the code

process/peer/process.go
```diff
@@ -661,6 +657,24 @@ func (vs *validatorStatistics) ResetValidatorStatisticsAtNewEpoch(vInfos map[uin
...
+	if !vs.flagJailedEnabled.IsSet() {
+		return
+	}
+
+	if validator.List == string(core.JailedList) && peerAccount.GetList() != string(core.JailedList) {
+		peerAccount.SetListAndIndex(validator.ShardId, string(core.JailedList), validator.Index)
+		return
+	}
+
+	if peerAccount.GetUnStakedEpoch() == core.DefaultUnstakedEpoch && peerAccount.GetList() == string(core.InactiveList) {
```
raduchis, line 673: no need for the validator.List == string (core.JailedList) ? Could it be not jailed in validator.List and still want to set it to jailed based on the new condition?

raduchis (approved):

process/peer/process.go
sasurobert: This is for the scenario when in old code it was set to inactive, and should have been set to jailed

process/peer/process.go
sasurobert: Not exactly the same. Would leave as it is

sasurobert: It is only 1 line of code.

SebastianMarian (approved):

LucianMincu (approved):
System tests passed.

PR #2334, opened 2020-09-30
Multi interceptors on topic

iulianpascalau:
- added multi-interceptor capability on network messenger. Each topic now supports n interceptors that will be called sequentially if a message is received


p2p/libp2p/topicProcessors_test.go
```diff
@@ -0,0 +1,138 @@
...
+	require.Equal(t, 1, len(tp.identifiers))
+
+	err = tp.removeTopicProcessor(identifier1)
+
+	assert.Nil(t, err)
+	require.Equal(t, 0, len(tp.processors))
+	require.Equal(t, 0, len(tp.identifiers))
+}
+
+func TestTopicProcessorsGetListShouldWorkAndPreserveOrder(t *testing.T) {
```
bogdan-rosianu, line 89: :+1:

bogdan-rosianu: if possible, please add a concurrent test for topicProcessor or directly in netMessenger where you continuously add and remove interceptors for a topic and then send a message to that topic. Code should not panic.

bogdan-rosianu (changes requested):

p2p/libp2p/netMessenger.go
```diff
@@ -748,31 +739,33 @@ func (netMes *networkMessenger) Broadcast(topic string, buff []byte) {
...
-		return fmt.Errorf("%w, operation RegisterMessageProcessor, topic %s",
-			p2p.ErrTopicValidatorOperationNotSupported,
-			topic,
-		)
+	topicProcs := netMes.processors[topic]
+	if topicProcs == nil {
+		topicProcs = newTopicProcessors()
+		netMes.processors[topic] = topicProcs
+
+		err := netMes.pb.RegisterTopicValidator(topic, netMes.pubsubCallback(topicProcs, topic))
```
AdoAdoAdo, line 754: Just to make sure I understand the changes correctly.
Previously: the topicMessageProcessor acts also as a validator for the messages, so one message will not be propagated to the network if the validator returned an error.

Now: consider the following scenario where we have two processors on the same topic, one of them does in-depth validation and returns an error on one message in case e.g aggregated signature is invalid/cannot be validated, but the other doesn't care about signatures and only does some statistics on the received messages. 

1. Will the node propagate the message to the network or not if there is one processor returning an error and another returning "no error"? 

2. In case both processors return "no error" will the node propagate the message twice or only once?

Can we also add some tests for the above?

AdoAdoAdo: I saw below that you do not propagate the message if either of the processors return an error, so that answers my first question.

Also for my second question, I see you only return from pubsub callback when both processors have finished, so if both returned no error then message is propagated only once

:+1:

p2p/libp2p/topicProcessors.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/ElrondNetwork/elrond-go/p2p"
+)
+
+type topicProcessors struct {
+	processors    map[string]p2p.MessageProcessor
+	identifiers   []string
```
AdoAdoAdo, line 12: why do we need a list of identifiers since we already have the map keys?
I think it can be safely removed, as for getList() you can create the identifiers list together with the processors list from the map.

AdoAdoAdo (changes requested):

p2p/libp2p/topicProcessors.go
iulianpascalau: I wanted to call the processors in the exact same order as the processors were registered. Of course this is not mandatory, it just gives a more deterministic flavor to the process. The list was introduced to not sort each time the map keys. Not having a strong opinion on keeping the list. :)

p2p/libp2p/netMessenger.go
iulianpascalau: Right, all registered processors are called. If at least one of them returns an error, the message won't propagate to the other peers. If all processors return nil errors, the message will propagate only once to the other peers.

p2p/libp2p/topicProcessors_test.go
iulianpascalau: Will try to assemble a test

iulianpascalau: done

p2p/errors.go
```diff
@@ -143,3 +143,9 @@ var ErrUnsupportedMessageVersion = errors.New("unsupported message version")
 
 // ErrNilSyncTimer signals that a nil sync timer was provided
 var ErrNilSyncTimer = errors.New("nil sync timer")
+
+// ErrMessageProcessorAlreadyDefined signals that a message processor was already defined on the provided topic and identifier
+var ErrMessageProcessorAlreadyDefined = errors.New("message processor already defined")
+
+// ErrMessageProcessorDoesNotExists signals that a message processor does not exists on the provided topic and identifier
```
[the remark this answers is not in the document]
iulianpascalau: done

core/constants.go
```diff
@@ -517,3 +517,20 @@ const DefaultLogProfileIdentifier = "[default log profile]"
 // Ex.: normal txs -> aprox. 27000, sc calls -> aprox. 6250 = 27000 / (MultiplyFactorForScCall + 1),
 // considering that constant below is set to 3
 const MultiplyFactorForScCall = 3
+
+// DefaultResolversIdentifier represents the identifier that is used in conjunction with regular resolvers
+//(that make the node run properly)
```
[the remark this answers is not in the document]
iulianpascalau: done

p2p/libp2p/topicProcessors.go
iulianpascalau: removed identifiers, added comment on networkMessenger.RegisterMessageProcessor that states the order is not guaranteed.

bogdan-rosianu (approved):

AdoAdoAdo (approved):

PR #2339, opened 2020-10-05
Integration test on staking process with real BLS signed values

iulianpascalau:
- added integration test that proves that a staking transaction with real BLS signature will work. Test data will be provided from the frontend testing process.

### DIFF
--- a/integrationTests/testInitializer.go
+++ b/integrationTests/testInitializer.go
@@ -75,6 +75,7 @@ var P2pBootstrapDelay = 5 * time.Second
 // InitialRating is used to initiate a node's info
 var InitialRating = uint32(50)
 
+// AdditionalGasLimit is the value that can be added on a transaction in the GasLimit
 var AdditionalGasLimit = uint64(999000)
 
 var log = logger.GetOrCreate("integrationtests")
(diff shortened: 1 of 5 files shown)


integrationTests/testProcessorNode.go
```diff
@@ -74,6 +76,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/testscommon"
 	"github.com/ElrondNetwork/elrond-go/update"
 	"github.com/ElrondNetwork/elrond-go/vm"
+	process2 "github.com/ElrondNetwork/elrond-go/vm/process"
```
sasurobert, line 79: change process2

integrationTests/testInitializer.go
```diff
@@ -1358,6 +1385,38 @@ func CreateAndSendTransaction(
 	node.OwnAccount.Nonce++
 }
 
+// CreateAndSendTransactionFromWallet will generate a transaction with provided parameters, sign it with the provided
+// wallet's tx sign private key and send it on the transaction topic
+func CreateAndSendTransactionFromWallet(
```
sasurobert, line 1390: I think there is a function which does the same already. PlayerSendsTransactoin ?

sasurobert (approved):

sasurobert (changes requested):
small stuff

integrationTests/testInitializer.go
iulianpascalau: PlayerSendsTransactoin is sent from the node's address. Wanted to used that but couldn't

sasurobert (approved):

raduchis (approved):

LucianMincu (approved):
👍

PR #2343, opened 2020-10-07
EN 7439 transaction routing fixes

iulianpascalau:
- removed the possibility to broadcast a transaction from any node in the network. Each transaction will get broadcast only if it will be routed to the sender's shard ID node.
- added unit and integration tests that prove that the transaction validation works as expected

### DIFF
--- a/integrationTests/testInitializer.go
+++ b/integrationTests/testInitializer.go
@@ -75,6 +75,7 @@ var P2pBootstrapDelay = 5 * time.Second
 // InitialRating is used to initiate a node's info
 var InitialRating = uint32(50)
 
+// AdditionalGasLimit is the value that can be added on a transaction in the GasLimit
 var AdditionalGasLimit = uint64(999000)
 
 var log = logger.GetOrCreate("integrationtests")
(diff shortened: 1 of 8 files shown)


sasurobert (approved):

integrationTests/testInitializer.go
```diff
@@ -1353,7 +1354,53 @@ func CreateAndSendTransaction(
...
+
+	wasSend := false
+	for _, senderNode := range nodes {
+		if senderNode.ShardCoordinator.SelfId() != senderShardID {
+			continue
+		}
+
+		_, err := senderNode.SendTransaction(tx)
+		if err != nil {
+			log.Error("could not create transaction", "address", node.OwnAccount.Address, "error", err)
```
bogdan-rosianu, line 1396: could not send?

node/node.go
```diff
@@ -824,7 +829,7 @@ func (n *Node) ValidateTransaction(tx *transaction.Transaction) error {
 		core.MaxTxNonceDeltaAllowed,
 	)
 	if err != nil {
-		return nil
+		return fmt.Errorf("%w in Node.ValidateTransaction", err)
```
bogdan-rosianu, line 832: this will arrive in the API response. I don't think a regular user would be interested to know the method where the validation failed. Please add a log for the wrapped error and return it normally

bogdan-rosianu (changes requested):

integrationTests/testInitializer.go
iulianpascalau: done

node/node.go
iulianpascalau: done. The error is now logged as warn as it is a programming error (non critical).

bogdan-rosianu (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2345, opened 2020-10-07
Tx Simulator - sc calls accounts

bogdan-rosianu:
Used the read only accounts db also in the smart contract processor. If not, in some cases, when simulating more than one transaction in the same round, the nonce would have been affected until the beginning of the next round and simulation with the same nonce would have returned error.

Testing procedure:
1. run a testnet with this branch
2. send 2 invalid SC calls (not enough gas or wrong function name) one after the other one with the same nonce. Both simulations should work and not throw an error regarding lower nonce. 

Alternative using `erdjs` (also the ways this issue was reproduced and checked that is has been fixed):
1. run a testnet with this branch
2. update the following test with the proxy URL and also send tokens to Alice:
https://github.com/ElrondNetwork/elrond-sdk/blob/2a142ffe6590046dfe243c589c4b6a10b00bcf3f/erdjs/src/smartcontracts/smartContract.spec.ts#L125
2. run the test. it should display the `lower nonce` message
3. comment the `fake increment` part (which was a temporary fix) and run again -> should work


iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2346, opened 2020-10-07
add hysteresis to minimum number of nodes

AdoAdoAdo:
Currently the minimum number of nodes in the network is given to the system staking SC at genesis and is equal to 
nbShards * nbEligiblePerShard

The waiting list does not have a minimum number of nodes, which means  it can reach 0 nodes.
No nodes in waiting list will prevent shuffling between shards, so we need to ensure a minimum number of nodes for shuffling.

Decision is to use the hysteresis config for this, and when enabling adaptivity to add also another parameter targetHysteresis > hysteresis. Hysteresis will then be renamed to minHysteresis and used to decide merging of shards and targetHysteresis will be used to decide splitting shards.

This PR uses an enabling epoch for considering also the hysteresis in the SystemStakingSC when deciding the minimum number of nodes. For this reason a StakingSC Function was defined for updating the staking SC storage for minimum number of nodes and a SystemSC method was defined for calling the staking SC function on epochStart with the right credentials when the feature becomes active.


vm/systemSmartContracts/staking.go
```diff
@@ -1168,6 +1170,31 @@ func (r *stakingSC) switchJailedWithWaiting(args *vmcommon.ContractCallInput) vm
...
+	}
+
+	stakeConfig := r.getConfig()
+	if len(args.Arguments) != 1 {
+		r.eei.AddReturnMessage("number of arguments must be 1")
+		return vmcommon.UserError
+	}
+
+	newMinNodes := big.NewInt(0).SetBytes(args.Arguments[0]).Int64()
+	// TODO: newMinNodes extra validation?
```
sasurobert, line 1186: there is no other access here to check. Can remain like this.

sasurobert (approved):

epochStart/metachain/systemSCs.go
```diff
@@ -412,5 +468,10 @@ func (s *systemSCProcessor) IsInterfaceNil() bool {
 // EpochConfirmed is called whenever a new epoch is confirmed
 func (s *systemSCProcessor) EpochConfirmed(epoch uint32) {
 	s.flagSwitchEnabled.Toggle(epoch >= s.switchEnableEpoch)
+
+	// only toggle on exact epoch. In future epochs the config should have already been synchronized from peers
```
iulianpascalau, line 472: :+1:

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2350, opened 2020-10-08
EN-7874: Rest API endpoint for economics data

bogdan-rosianu:
Added a new endpoint `/node/economics` that will return the economics data for the last epoch. It is available for both meta nodes and shard nodes.

_Note that the values are returned in string format and the denomination is included._

Example output:
``` JSON
{
  "data": {
    "metrics": {
      "erd_dev_rewards": "0",
      "erd_epoch_number": 2,
      "erd_inflation": "14030640106544902492",
      "erd_total_fees": "350000000000000",
      "erd_total_supply": "20000027511395306950789200"
    }
  },
  "error": "",
  "code": "successful"
}
```

Testing procedure:
1. run a testnet with this branch and make sure that `api.toml` has this endpoint enabled
2. open `/node/economics` endpoint (of type GET) on multiple nodes.
 2.1. in epoch 0, only the total supply value should be populated
 2.2. after passing in a new epoch, the values will update and should be the same for all the nodes.


iulianpascalau (dismissed):

sasurobert (dismissed):

sasurobert (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2351, opened 2020-10-08
Feat/elastic search new impl

miiu96:
Modified how the mechanism that indexes data in elastic search database works. Now all the methods that want to write data in elastic search use a buffered chan and a goroutine writes data that is stored in the chan sequentially.

Testing procedure:
1. run a testnet
2. verify that all the data was indexed in the elastic search server => improvements would be visible only on a very high load


iulianpascalau (approved):

sasurobert (dismissed):

iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):
Initial tests passed. More testing with development branch will follow.

PR #2367, opened 2020-10-12
Bugfix import db testnet version T1.1.3.0 compatibility issue

iulianpascalau:
- fixed testnet compatibility issue that happened when trying to import a DB on the testnet version T1.1.3.0

Manual testing procedure: import-db on shard 0 with the T1.1.3.0 version configs & resulted DB. The process will work up until epoch 6 where the new "bad" code kicked-in. (by using a small hack that reverted this fix for epochs>=5, I was able to fully sync the 12 epochs)


sasurobert (approved):

raduchis (approved):

LucianMincu (approved):
System tests passed.

PR #2368, opened 2020-10-12
Fix multiple broadcast for cross txs

SebastianMarian:
* Added / changed some prints and their log level in broadcast component
* Added a missed return


sasurobert (approved):

consensus/broadcast/shardChainMessenger.go
```diff
@@ -231,6 +232,7 @@ func (scm *shardChainMessenger) PrepareBroadcastBlockDataValidator(
 	err = scm.delayedBlockBroadcaster.SetValidatorData(broadcastData)
 	if err != nil {
 		log.Error("shardChainMessenger.PrepareBroadcastBlockDataValidator", "error", err)
+		return
```
iulianpascalau, line 235: :+1:

iulianpascalau (approved):

consensus/broadcast/shardChainMessenger.go
AdoAdoAdo, line 235: this return is redundant

consensus/broadcast/shardChainMessenger.go
```diff
@@ -196,6 +196,7 @@ func (scm *shardChainMessenger) PrepareBroadcastHeaderValidator(
 	err = scm.delayedBlockBroadcaster.SetHeaderForValidator(vData)
 	if err != nil {
 		log.Error("shardChainMessenger.PrepareBroadcastHeaderValidator", "error", err)
+		return
```
AdoAdoAdo, line 199: this return is redundant

consensus/broadcast/metaChainMessenger.go
```diff
@@ -160,6 +161,7 @@ func (mcm *metaChainMessenger) PrepareBroadcastHeaderValidator(
 	err = mcm.delayedBlockBroadcaster.SetHeaderForValidator(vData)
 	if err != nil {
 		log.Error("metaChainMessenger.PrepareBroadcastHeaderValidator", "error", err)
+		return
```
AdoAdoAdo, line 164: this return is redundant

consensus/broadcast/metaChainMessenger.go
```diff
@@ -141,6 +141,7 @@ func (mcm *metaChainMessenger) PrepareBroadcastHeaderValidator(
 ) {
 	if check.IfNil(header) {
 		log.Error("metaChainMessenger.PrepareBroadcastHeaderValidator", "error", spos.ErrNilHeader)
+		return
```
AdoAdoAdo, line 144: :+1:

consensus/broadcast/shardChainMessenger.go
SebastianMarian: I would keep it just for the case when the method would be extended by someone who would forgot to added back

consensus/broadcast/shardChainMessenger.go
SebastianMarian: I would keep it just for the case when the method would be extended by someone who would forgot to added back

consensus/broadcast/metaChainMessenger.go
SebastianMarian: I would keep it just for the case when the method would be extended by someone who would forgot to added back

AdoAdoAdo (approved):

LucianMincu (approved):
System tests passed.

PR #2376, opened 2020-10-14
Import DB further speed improvements

iulianpascalau:
- altered some config storage values in import-db mode to speed-up storage syncing


cmd/node/main.go
```diff
@@ -1530,9 +1530,27 @@ func applyCompatibleConfigs(isInImportMode bool, importDbNoSigCheckFlag bool, lo
 		config.Heartbeat.DurationToConsiderUnresponsiveInSec = math.MaxInt32
 		config.Heartbeat.MinTimeToWaitBetweenBroadcastsInSec = math.MaxInt32 - 2
 		config.Heartbeat.MaxTimeToWaitBetweenBroadcastsInSec = math.MaxInt32 - 1
+
+		changeStorageConfigs(config)
 	}
 }
 
+func changeStorageConfigs(config *config.Config) {
```
bogdan-rosianu, line 1538: `alterStorageConfigsForDBImport` ?
also, you could send value instead of pointer

bogdan-rosianu (changes requested):

cmd/node/main.go
iulianpascalau: I will change the name but I need to keep the pointer to be modified by the function

bogdan-rosianu (approved):

sasurobert (approved):

sasurobert (dismissed):

config/config.go
```diff
@@ -234,7 +234,7 @@ type StateTriesConfig struct {
 type TrieStorageManagerConfig struct {
 	PruningBufferLen   uint32
 	SnapshotsBufferLen uint32
-	MaxSnapshots       uint8
+	MaxSnapshots       uint32
```
bogdan-rosianu, line 237: could this cause comapatibility issues?

bogdan-rosianu (dismissed):

config/config.go
iulianpascalau: No, this is strictly a processing component. The max snapshot value is strictly used when deciding to keep (or not) an older trie snapshot. This is saved nowhere.

LucianMincu (approved):
conflicts

sasurobert (approved):

stefanandy (approved):
System test passed. succesfull import
--doubled checked with developer--

raduchis (approved):

PR #2385, opened 2020-10-19
saveKeyValue-protection

sasurobert:
implemented protection at save key value level - which keeps leafs size under control.


camilbancioiu (approved):

data/state/trackableDataTrie.go
```diff
@@ -67,13 +68,19 @@ func trimValue(value []byte, tailLength int) ([]byte, error) {
 
 // SaveKeyValue stores in dirtyData the data keys "touched"
 // It does not care if the data is really dirty as calling this check here will be sub-optimal
-func (tdaw *TrackableDataTrie) SaveKeyValue(key []byte, value []byte) {
+func (tdaw *TrackableDataTrie) SaveKeyValue(key []byte, value []byte) error {
 	var identifier []byte
-	if len(value) != 0 {
+	lenValue := uint64(len(value))
```
iulianpascalau, line 73: unit test?

sasurobert: done

camilbancioiu (dismissed):

iulianpascalau (approved):

data/state/trackableDataTrie_test.go
```diff
@@ -191,3 +193,14 @@ func TestTrackableDataTrie_SetAndGetDataTrie(t *testing.T) {
...
+
+func TestTrackableDataTrie_SaveKeyValueTooBig(t *testing.T) {
+	t.Parallel()
+
+	identifier := []byte("identifier")
+	trie := &mock.TrieStub{}
+	tdaw := state.NewTrackableDataTrie(identifier, trie)
+
+	err := tdaw.SaveKeyValue([]byte("key"), make([]byte, core.MaxLeafSize+1))
+	assert.Error(t, err, data.ErrLeafSizeTooBig)
```
bogdan-rosianu, line 205: `assert.Equal(t, data.ErrLeafSizeTooBig, err)`

process/smartContract/builtInFunctions/esdtFreezeWipe.go
```diff
@@ -66,7 +66,10 @@ func (e *esdtFreezeWipe) ProcessBuiltinFunction(
 	log.Trace(vmInput.Function, "sender", vmInput.CallerAddr, "receiver", vmInput.RecipientAddr, "token", esdtTokenKey)
 
 	if e.wipe {
-		acntDst.DataTrieTracker().SaveKeyValue(esdtTokenKey, nil)
+		err := acntDst.DataTrieTracker().SaveKeyValue(esdtTokenKey, nil)
```
bogdan-rosianu, line 69: the error check here is useless because the function will only throw error if the value length is too big, but in your case the value will have the length 0 so no error will be returned. However, can remain like this for future updates in the `SaveKeyValue` function

sasurobert: i know, but it is better to check the error here. As you do not know how the underlying implementation can change.

data/state/trackableDataTrie_test.go
sasurobert: thx.

bogdan-rosianu (approved):

camilbancioiu (approved):

iulianpascalau (approved):

stefanandy (approved):
Passed system test

PR #2388, opened 2020-10-20
EN-7780-delegation unit tests 2

BeniaminDrasovean:
Small bugfixes and unit tests for delegation system SC


vm/systemSmartContracts/delegation.go
```diff
@@ -1023,6 +1038,7 @@ func (d *delegation) withdraw(args *vmcommon.ContractCallInput) vmcommon.ReturnC
 				break
 			}
 		}
+		delegator.UnStakedFunds = append(delegator.UnStakedFunds[:fundIndex], delegator.UnStakedFunds[fundIndex+1:]...)
```
sasurobert, line 1041: this is not safe. good find. However take the unbonded fundIndex into a list and add a delete function after.

sasurobert (changes requested):

sasurobert (approved):

vm/systemSmartContracts/delegation.go
```diff
@@ -872,6 +873,20 @@ func (d *delegation) addValueToFund(key []byte, value *big.Int) error {
 	return err
 }
 
+func (d *delegation) addDelegatorIfNew(dStatus *DelegationContractStatus, address []byte) {
```
iulianpascalau, line 876: I think this can be optimized to something like:
```GO
func (d *delegation) addDelegatorIfNew(dStatus *DelegationContractStatus, address []byte) {
	for i := range dStatus.Delegators {
		if bytes.Equal(address, dStatus.Delegators[i]) {
			return
		}
	}

	dStatus.Delegators = append(dStatus.Delegators, address)
}
```

vm/systemSmartContracts/delegation.go
```diff
@@ -1024,6 +1041,7 @@ func (d *delegation) withdraw(args *vmcommon.ContractCallInput) vmcommon.ReturnC
 			}
 		}
 	}
+	delegator.UnStakedFunds = tempUnStakedFunds
```
iulianpascalau, line 1044: :+1:

vm/systemSmartContracts/delegation_test.go
```diff
@@ -20,7 +20,7 @@ import (
 
 func createMockArgumentsForDelegation() ArgsNewDelegation {
 	return ArgsNewDelegation{
-		DelegationSCConfig:     config.DelegationSystemSCConfig{MinStakeAmount: "10"},
+		DelegationSCConfig:     config.DelegationSystemSCConfig{MinStakeAmount: "10", MinServiceFee: 10, MaxServiceFee: 200},
```
iulianpascalau, line 23: multiple lines?

vm/systemSmartContracts/delegation_test.go
```diff
@@ -1147,3 +1146,795 @@ func TestDelegationSystemSC_ExecuteUnBondNodes(t *testing.T) {
 	dGlobalFund, _ := d.getGlobalFundData()
 	assert.Equal(t, big.NewInt(200), dGlobalFund.TotalUnBondedFromNodes)
 }
+
+func TestDelegationSystemSC_ExecuteUnJailNodesUserErrors(t *testing.T) {
+	t.Parallel()
```
iulianpascalau, line 1151: all in one test :)

vm/systemSmartContracts/delegation_test.go
```diff
@@ -367,7 +366,7 @@ func TestDelegationSystemSC_ExecuteInitShouldWork(t *testing.T) {
 
 	ok, delegator, err := d.getOrCreateDelegatorData(ownerAddr)
 	assert.Nil(t, err)
-	assert.True(t, ok)
+	assert.False(t, ok)
```
iulianpascalau, line 369: Please resolve the weak warning at L835.
the function `getSuccessAndUnSuccessKeys` needs more testing

iulianpascalau (commented):
GJ

vm/systemSmartContracts/delegation.go
sasurobert: no need for for here. _, dData, err := d.getOrCreateDelegatorData(args.CallerAddr) the first return parameter is called isNew :). plus this function is called only if it was new. No need for these checks.

sasurobert: the simple append is good enough, while checking the isNew return param.

sasurobert (changes requested):

vm/systemSmartContracts/delegation.go
BeniaminDrasovean: Done.

vm/systemSmartContracts/delegation_test.go
BeniaminDrasovean: Done.

vm/systemSmartContracts/delegation_test.go
BeniaminDrasovean: Well, it tests for all the user errors.

vm/systemSmartContracts/delegation_test.go
BeniaminDrasovean: Done.

vm/systemSmartContracts/delegation.go
sasurobert: addDelegatorIfNew(dStatus *DelegationContractStatus, address []byte). This is called after  d.getOrCreateDelegatorData(address), thus you do not need to get it again. 
 addDelegatorIfNew function is not needed. in the delegate function you need if isNew -> append

sasurobert (changes requested):

vm/systemSmartContracts/delegation.go
```diff
@@ -827,7 +827,9 @@ func (d *delegation) delegate(args *vmcommon.ContractCallInput) vmcommon.ReturnC
 			return vmcommon.UserError
 		}
 
-		d.addDelegatorIfNew(dStatus, args.CallerAddr)
+		if isNew {
```
sasurobert, line 830: make a function out of lines 823-837 addNewDelegatorToList which does the getDelegationStatus and setDelegationStatus and the append. The function should be called under the if isNew

sasurobert (changes requested):

vm/systemSmartContracts/delegation.go
BeniaminDrasovean: Done.

sasurobert (approved):

iulianpascalau (approved):

PR #2389, opened 2020-10-20
Change max nodes waiting list

AdoAdoAdo:
This PR enables the maximum nodes in the network to be changed through a config change.
Also the maximum number of nodes that can be swapped between eligible and waiting lists is also configurable.
Both config options have been added in the same config structure as they will come together and are functionally connected.

Backwards compatibility was implemented.
For the maximum number of nodes allowed, systemSCs.go was changed to permit the change when an epoch is confirmed.
In order to cover any epoch start block revert edge case, the config is set with the right number of nodes in each new epoch, even if the same number was set in the previous epoch.

For the maximum nodes allowed to be shuffled/swapped per shard, changes were done in shuffler.
The shuffler is only called upon start of epoch events, so there was no need to use an epochNotifier, the swap limit was integrated into the update nodes process. The backwards compatibility is ensured by selecting the right configuration depending on the epoch.


sasurobert (dismissed):

epochStart/metachain/systemSCs.go
```diff
@@ -127,14 +133,27 @@ func NewSystemSCProcessor(args ArgsNewEpochStartSystemSCProcessing) (*systemSCPr
 		hystNodesEnableEpoch:     args.SwitchHysteresisForMinNodesEnableEpoch,
 	}
 
+	s.maxNodesEnableConfig = make([]config.MaxNodesChangeConfig, len(args.MaxNodesEnableConfig))
+	copy(s.maxNodesEnableConfig, args.MaxNodesEnableConfig)
+	sort.Slice(s.maxNodesEnableConfig, func(i, j int) bool {
```
raduchis, line 138: do we need a check that there are no 2 equal epochEnableNonces?

epochStart/metachain/systemSCs.go
```diff
@@ -471,7 +530,21 @@ func (s *systemSCProcessor) EpochConfirmed(epoch uint32) {
 
 	// only toggle on exact epoch. In future epochs the config should have already been synchronized from peers
 	s.flagHystNodesEnabled.Toggle(epoch == s.hystNodesEnableEpoch)
+
+	for _, maxNodesConfig := range s.maxNodesEnableConfig {
+		if epoch >= maxNodesConfig.EpochEnable {
+			// to cover also rollbacks, we always set the maxNodes and set the Enabled flag
+			s.flagChangeMaxNodesEnabled.Toggle(true)
```
raduchis, line 537: this will be called for all epochEnable, not just for the last one. If there is no way to process in parallel this and ProcessSystemSmartContract, there should be no problem, because the last set will be considered.

raduchis (approved):

sasurobert (approved):

raduchis (approved):

PR #2391, opened 2020-10-21
EN-7932: add storer for epoch-root hash in import db mode

bogdan-rosianu:
Added a new storer for `import-db` mode that will hold the `epoch` - `root hash` pairs related to tries snapshots. Based on this, we will be able to re-load the trie state at each end of epoch

### DIFF
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -2,8 +2,10 @@ package block
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"math/big"
 	"sort"
 	"time"
 
@@ -25,6 +27,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/process/block/bootstrapStorage"
 	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/ElrondNetwork/elrond-go/storage/storageUnit"
 )
 
 var log = logger.GetOrCreate("process/block")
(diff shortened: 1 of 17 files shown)


process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
 		bp.blockTracker.AddTrackedHeader(headers[i], hashes[i])
 	}
 }
+
+func (bp *baseProcessor) commitTrieEpochRootHashIfNeeded(metaBlock *block.MetaBlock) error {
+	trieEpochRootHashStorageUnit := bp.store.GetStorer(dataRetriever.TrieEpochRootHashUnit)
+	if check.IfNil(trieEpochRootHashStorageUnit) {
+		return nil
+	}
+	_, isStorerDisabled := trieEpochRootHashStorageUnit.(*storageUnit.NilStorer)
```
iulianpascalau, line 1251: I think we can get rid of this check. If the storer is nil, it will return on line 1249.

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
...
+	if isStorerDisabled {
+		return nil
+	}
+
+	userAccountsDb := bp.accountsDB[state.UserAccountsState]
+	if userAccountsDb == nil {
+		return nil
```
iulianpascalau, line 1258: this is a non recoverable error, we might return err here

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
...
+	}
+
+	currentRootHash, err := userAccountsDb.RootHash()
+	if err != nil {
+		return err
```
iulianpascalau, line 1263: this is a non recoverable error, we might return err here

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
...
+	}
+	epochBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(epochBytes, metaBlock.Epoch)
```
iulianpascalau, line 1266: please use bp.uint64Converter instead of big endian (to be consistent with the rest of the project)

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
...
+	allLeaves, err := userAccountsDb.GetAllLeaves(currentRootHash)
+	if err != nil {
+		return err
+	}
+
+	balanceSum := big.NewInt(0)
+	for addressKey, userAccountsBytes := range allLeaves {
+		userAccount, errUnmarshal := unmarshalUserAccount([]byte(addressKey), userAccountsBytes, bp.marshalizer)
+		if errUnmarshal != nil {
+			log.Debug("cannot unmarshal user account. it may be a code leaf", "error", errUnmarshal)
```
iulianpascalau, line 1282: suggestion to move this to log.Trace

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
...
+			continue
+		}
+		balanceSum.Add(balanceSum, userAccount.GetBalance())
+	}
+
+	log.Debug("sum of addresses in shard at epoch start",
```
iulianpascalau, line 1288: since you compute the total balance, why not saving it in trieEpochRootHashStorageUnit (might rename it).

process/block/baseProcess.go
```diff
@@ -1239,3 +1242,66 @@ func (bp *baseProcessor) addHeaderIntoTrackerPool(nonce uint64, shardID uint32)
 		bp.blockTracker.AddTrackedHeader(headers[i], hashes[i])
 	}
 }
+
+func (bp *baseProcessor) commitTrieEpochRootHashIfNeeded(metaBlock *block.MetaBlock) error {
```
iulianpascalau, line 1246: Can we have some unit tests for this?

storage/factory/pruningStorerFactory.go
```diff
@@ -486,3 +503,23 @@ func (psf *StorageServiceFactory) createPruningStorerArgs(storageConfig config.S
...
 }
+
+func (psf *StorageServiceFactory) createTrieEpochRootHashStorerIfNeeded() (storage.Storer, error) {
+	if !psf.createTrieEpochRootHashStorer {
+		return storageUnit.NewNilStorer(), nil
+	}
+
+	trieEpochRootHashDbConfig := GetDBFromConfig(psf.generalConfig.TrieEpochRootHashStorage.DB)
+	shardId := core.GetShardIDString(psf.shardCoordinator.SelfId())
+	dbPath := psf.pathManager.PathForStatic(shardId, psf.generalConfig.TrieEpochRootHashStorage.DB.FilePath)
```
iulianpascalau, line 514: :+1:

process/block/baseProcess.go
bogdan-rosianu: no.. line 1249 returns a nil/disabled storer so this check is mandatory. An alternative would have been adding a function `IsNilStorer() bool` inside the `Storer` interface but the changeset would have been too big

process/block/baseProcess.go
bogdan-rosianu: done

process/block/baseProcess.go
bogdan-rosianu: already returned

process/block/baseProcess.go
bogdan-rosianu: done

process/block/baseProcess.go
bogdan-rosianu: I had this in mind.. I let it to `Debug` for system tests but forgot to move it to `Trace`. :D 
done

process/block/baseProcess.go
bogdan-rosianu: i would leave it like this for the moment. after further discussions over the feature branch, we can rename the storer and use it in multiple ways

process/block/baseProcess.go
bogdan-rosianu: added some unit tests

storage/factory/pruningStorerFactory.go
bogdan-rosianu: :+1:

SebastianMarian (approved):

iulianpascalau (approved):

PR #2394, opened 2020-10-22
En 7780 delegation manager unit test

BeniaminDrasovean:
Unit tests an small fixes for delegation manager.


vm/systemSmartContracts/eei.go
```diff
@@ -465,6 +465,14 @@ func (host *vmContext) CreateVMOutput() *vmcommon.VMOutput {
 		}
 		if outAcc.Nonce > 0 {
 			outAccs[addr].Nonce = outAcc.Nonce
+
+			if len(outAcc.CodeDeployerAddress) > 0 {
```
sasurobert, line 469: not like this :). what you need to check if that blockchainHook.CurrentNonce() > 0 and set these.

sasurobert (changes requested):

vm/systemSmartContracts/eei.go
BeniaminDrasovean: Done.

sasurobert (approved):

iulianpascalau (approved):

PR #2395, opened 2020-10-22
smallFix/epoch-confirmation-block-request

sasurobert:
Scenario: metachain creates epoch start block, then creates 2 epoch start attesting metablock (different rounds), Finally metachain creates the metachain block + 2. 
The node in the shard receives the epochstart metablock and the first attesting metablock (the one with lower round). The actual finality attesting block is the one with the higher round.
The node receives the metachain block + 2, but it does not have the one with the previous hash, thus will not modify the FinalityAttestingRound value.
From that moment onwards, the node will not be able to process the first valid block from the shard, as it will say epoch does not match.

The modified code fixes the aforementioned scenario


iulianpascalau (commented):
:+1:  Please add description

AdoAdoAdo (dismissed):

raduchis (approved):

iulianpascalau (dismissed):

raduchis (approved):

AdoAdoAdo (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2396, opened 2020-10-23
Fix import-DB for the testnet edge case

iulianpascalau:
This PR fixes the edge case found in testnet when a shard stopped including metaheaders for 2 epochs

- made the storage resolvers open 2 epochs in advance. Also, the maximum active persisters value is set to the maximum value it can hold (config.StoragePruning.NumEpochsToKeep) in import-db mode

### DIFF
--- a/dataRetriever/storageResolvers/headerResolver.go
+++ b/dataRetriever/storageResolvers/headerResolver.go
@@ -79,7 +79,7 @@ func (hdrRes *headerResolver) RequestDataFromHash(hash []byte, _ uint32) error {
 	metaEpoch := hdrRes.epochHandler.MetaEpoch()
 	hdrRes.mutEpochHandler.RUnlock()
 
-	hdrRes.manualEpochStartNotifier.NewEpoch(metaEpoch + 1)
+	hdrRes.manualEpochStartNotifier.NewEpoch(metaEpoch + 2)
 
 	buff, err := hdrRes.hdrStorage.SearchFirst(hash)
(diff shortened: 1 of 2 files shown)


dataRetriever/storageResolvers/headerResolver.go
```diff
@@ -79,7 +79,7 @@ func (hdrRes *headerResolver) RequestDataFromHash(hash []byte, _ uint32) error {
 	metaEpoch := hdrRes.epochHandler.MetaEpoch()
 	hdrRes.mutEpochHandler.RUnlock()
 
-	hdrRes.manualEpochStartNotifier.NewEpoch(metaEpoch + 1)
+	hdrRes.manualEpochStartNotifier.NewEpoch(metaEpoch + 2)
```
sasurobert, line 82: will this work always ?

sasurobert (approved):

dataRetriever/storageResolvers/headerResolver.go
iulianpascalau: yes

iulianpascalau: it opens in advance 2 persisters rather than 1. Works on the 0->1 and x->x+1 edge cases as tested with a fetched import-db archive

sasurobert (approved):

AdoAdoAdo (approved):

LucianMincu (approved):
System tests passed.

PR #2398, opened 2020-10-23
use elastic upsert instead of insert/update for transactions

bogdan-rosianu:
Use `upsert` instead of different logic for `insert` and `update` for elastic search transactions.
This should solve cases when reimporting data would have caused a sender shard process a transaction after the receiver shard and update the status to 'pending'

Testing procedure:
1. run several nodes with import-db
2. transactions' statuses should not be 'pending'


core/indexer/common.go
```diff
@@ -367,62 +367,26 @@ func serializeAccountsHistory(accounts map[string]*AccountBalanceHistory) []byte
...
-	}
-	return
-}
+	meta = []byte(fmt.Sprintf(`{"update":{"_id":"%s", "_type": "_doc"}}%s`, tx.Hash, "\n"))
 
-func prepareSerializedAccountInfo(address string, account *AccountInfo) (meta []byte, serializedData []byte) {
-	var err error
-	meta = []byte(fmt.Sprintf(`{ "index" : { "_id" : "%s" } }%s`, address, "\n"))
-	serializedData, err = json.Marshal(account)
+	marshalledTx, err := json.Marshal(tx)
```
iulianpascalau, line 374: marshaledTx *

core/indexer/common.go
```diff
@@ -367,62 +367,26 @@ func serializeAccountsHistory(accounts map[string]*AccountBalanceHistory) []byte
...
 	if err != nil {
 		log.Debug("indexer: marshal",
```
iulianpascalau, line 376: shouldn't this be an unrecoverable error? If something goes terrible wrong, the node will just move forward, no one will read the logs but then we will start wondering why the accounts balances do not match.

core/indexer/common.go
```diff
@@ -367,62 +367,26 @@ func serializeAccountsHistory(accounts map[string]*AccountBalanceHistory) []byte
...
-	if err != nil {
-		log.Debug("indexer: marshal",
-			"error", "could not serialize account history entry, will skip indexing",
-			"address", address)
+	if !isCrossShardDstMe(tx, selfShardID) || tx.Status == transaction.TxStatusInvalid.String() {
+		serializedData =
+			[]byte(fmt.Sprintf(`{"script":{"source":""},"upsert":%s}`,
+				string(marshalledTx)))
+		log.Warn("log meta - insert", "data", string(meta))
+		log.Warn("log srld data - insert", "data", string(serializedData))
```
iulianpascalau, line 386: what is srld? why do we have 2 log.Warn calls?

core/indexer/common.go
```diff
@@ -367,62 +367,26 @@ func serializeAccountsHistory(accounts map[string]*AccountBalanceHistory) []byte
...
 
-	return
-}
-
-func prepareTxUpdate(tx *Transaction) ([]byte, []byte) {
-	var meta, serializedData []byte
-
-	meta = []byte(fmt.Sprintf(`{ "update" : { "_id" : "%s", "_type" : "%s"  } }%s`, tx.Hash, "_doc", "\n"))
-
 	marshalizedLog, err := json.Marshal(tx.Log)
```
iulianpascalau, line 390: marshaledLog *

core/indexer/common.go
```diff
@@ -448,15 +412,62 @@ func prepareTxUpdate(tx *Transaction) ([]byte, []byte) {
 
 	if tx.GasUsed == tx.GasLimit {
 		// do not update gasUsed because it is the same with gasUsed when transaction was saved first time in database
-		serializedData = []byte(fmt.Sprintf(`{ "doc" : { "log" : %s, "scResults" : %s, "status": "%s", "timestamp": %s } }`,
-			string(marshalizedLog), string(scResults), tx.Status, string(marshalizedTimestamp)))
+		serializedData =
```
iulianpascalau, line 415: a lot of duplicated code here.

core/indexer/common.go
```diff
@@ -448,15 +412,62 @@ func prepareTxUpdate(tx *Transaction) ([]byte, []byte) {
...
+				`ctx._source.log = params.log;`+
+				`ctx._source.scResults = params.scResults;`+
+				`ctx._source.timestamp = params.timestamp;`+
+				`ctx._source.gasUsed = params.gasUsed;`+
+				`","lang": "painless","params":`+
+				`{"status": "%s", "log": %s, "scResults": %s, "timestamp": %s, "gasUsed": %d}},"upsert":%s}`,
+				tx.Status, string(marshalizedLog), string(scResults), string(marshalizedTimestamp), tx.GasUsed, string(marshalledTx)))
+	}
+
+	log.Warn("log meta - update", "data", string(meta))
```
iulianpascalau, line 438: why do we need 2 log.Warn calls here? Also, srld hits again and you have commented out code.

core/indexer/common.go
```diff
@@ -448,15 +412,62 @@ func prepareTxUpdate(tx *Transaction) ([]byte, []byte) {
...
+
+	return
+}
+
+func prepareSerializedAccountInfo(address string, account *AccountInfo) (meta []byte, serializedData []byte) {
+	var err error
+	meta = []byte(fmt.Sprintf(`{ "index" : { "_id" : "%s" } }%s`, address, "\n"))
+	serializedData, err = json.Marshal(account)
+	if err != nil {
+		log.Debug("indexer: marshal",
```
iulianpascalau, line 450: we should not skip indexing

core/indexer/common.go
```diff
@@ -448,15 +412,62 @@ func prepareTxUpdate(tx *Transaction) ([]byte, []byte) {
...
-	return meta, serializedData
+	return
+}
+
+func prepareSerializedAccountBalanceHistory(address string, account *AccountBalanceHistory) (meta []byte, serializedData []byte) {
+	var err error
+	meta = []byte(fmt.Sprintf(`{ "index" : { "_id" : "%s" } }%s`, address, "\n"))
+	serializedData, err = json.Marshal(account)
+	if err != nil {
+		log.Debug("indexer: marshal",
```
iulianpascalau, line 464: here also, do not skip indexing when an error occurs.

iulianpascalau (changes requested):

core/indexer/common.go
bogdan-rosianu: done

core/indexer/common.go
bogdan-rosianu: refactored return parameters and added an error also. I have also refactored the components calling this in order to return errors instead of just logging

core/indexer/common.go
bogdan-rosianu: removed

core/indexer/common.go
bogdan-rosianu: done

core/indexer/common.go
bogdan-rosianu: I know but I couldn't find another approach to include an additional field by using the same base string (a substring has to be added inside `script` part and inside `params` as well)

core/indexer/common.go
bogdan-rosianu: removed

core/indexer/common.go
bogdan-rosianu: done

core/indexer/common.go
bogdan-rosianu: done. with a note: the errors for specified areas will now arrive in the `workItem` part that won't stop the node's execution as I know.. so we will need another task maybe for treating differently the errors there

core/indexer/common.go
```diff
@@ -246,16 +246,17 @@ func serializeTransactions(
...
 	var buff bytes.Buffer
 	buffSlice := make([]bytes.Buffer, 0)
 	for _, tx := range transactions {
 		isMBOfTxInDB := mbsHashInDB[tx.MBHash]
-		meta, serializedData := prepareSerializedDataForATransaction(tx, selfShardID, isMBOfTxInDB)
-		if len(meta) == 0 {
-			continue
+		meta, serializedData, errPrepareTx := prepareSerializedDataForATransaction(tx, selfShardID, isMBOfTxInDB)
+		if errPrepareTx != nil {
+			log.Warn("error preparing transaction for indexing", "tx hash", tx.Hash, "error", err)
```
iulianpascalau, line 258: We need a task (that will be done in a next PR) to further refactor and remove all these warns as for this error we will get 3 log.Warn prints.

iulianpascalau (commented):
please add a description

core/indexer/common.go
iulianpascalau: it will eventually stop the node as the queue will not be consumed (if the item.Save errors - by any mean - it will get stuck in  a retry mode) and that queue will fill up, preventing syncBlock to finish executing.

core/indexer/common.go
bogdan-rosianu: ok.. good catch. added jira ID EN-7977

iulianpascalau (approved):

core/indexer/common.go
```diff
@@ -361,102 +371,106 @@ func serializeAccountsHistory(accounts map[string]*AccountBalanceHistory) []byte
...
-	meta = []byte(fmt.Sprintf(`{ "index" : { "_id" : "%s" } }%s`, address, "\n"))
-	serializedData, err = json.Marshal(account)
-	if err != nil {
-		log.Debug("indexer: marshal",
-			"error", "could not serialize account history entry, will skip indexing",
-			"address", address)
-		return
+	if !isCrossShardDstMe(tx, selfShardID) || tx.Status == transaction.TxStatusInvalid.String() {
+		serializedData :=
+			[]byte(fmt.Sprintf(`{"script":{"source":""},"upsert":%s}`,
```
ccorcoveanu, line 393: I think there are some fields (@miiu96 can add to this) that only the source shard has, right? Those should be added to the source field.

core/indexer/common.go
ccorcoveanu: I think we should have a separate task here. This json encoding for transaction serialisation became complicated enough that it deserves a component. Could be a simple structure with some helper methods that can receive a list of fields to update, a list of fields for insert, etc. It can have json tags on it's fields and then we can just marshal it

core/indexer/common.go
ccorcoveanu: We should definitely decide how to treat this kind of errors.
In this specific case we should treat it as an invalid document and skip it. I am saying this because if it's a `json.Marshal` error, the node will crash anyway.

ccorcoveanu (changes requested):

core/indexer/common.go
bogdan-rosianu: no. based on the old implementation, on the `insert` part it only marshaled the transaction and indexed it. the same thing happens here.

ccorcoveanu: Discussed in call, it's ok like this 👍

ccorcoveanu (approved):

iulianpascalau (changes requested):
PR is ok, we are not ready to merge it yet.

iulianpascalau (approved):

PR #2399, opened 2020-10-23
Elastic search will not index sender transactions in import db mode

iulianpascalau:
- made the elastic search not index transactions if the node is node in the destination shard when running in import-db

### DIFF
--- a/core/indexer/processTransactions.go
+++ b/core/indexer/processTransactions.go
@@ -19,6 +19,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/hashing"
 	"github.com/ElrondNetwork/elrond-go/marshal"
 	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/sharding"
 	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 )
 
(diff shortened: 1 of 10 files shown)


sasurobert (approved):

SebastianMarian (approved):

core/indexer/processTransactions.go
```diff
@@ -254,6 +266,15 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
 	return transactions, rewardsTxs
 }
 
+func (tdp *txDatabaseProcessor) shouldIndex(txHandler data.TransactionHandler) bool {
+	if !tdp.isInImportMode {
+		return true
+	}
+
+	destShardId := tdp.shardCoordinator.ComputeId(txHandler.GetRcvAddr())
```
miiu96, line 274: you dont need shardCoordinator here because you have destination shard from miniblock. no need to insert shardCoordinator here.

LucianMincu (approved):
👍 system tests passed

core/indexer/processTransactions.go
iulianpascalau: will be fixed on the new elastic search implementation feature branch :+1:

miiu96 (approved):

PR #2401, opened 2020-10-23
Fix indexing pending problem when running in import-DB mode

iulianpascalau:
- made the elastic search not index transactions if the node is node in the destination shard when running in import-db

### DIFF
--- a/core/indexer/elasticProcessor.go
+++ b/core/indexer/elasticProcessor.go
@@ -65,8 +65,8 @@ func NewElasticProcessor(arguments ArgElasticProcessor) (ElasticProcessor, error
 	)
 
 	if arguments.IsInImportDBMode {
-		log.Warn("the node is in import mode! Elastic search will not send cross shard transactions and rewards " +
-			"where destination shard is not the current node's shard")
+		log.Warn("the node is in import mode! Cross shard transactions and rewards where destination shard is " +
+			"not the current node's shard won't be indexed in Elastic Search")
 	}
 
 	if arguments.Options.UseKibana {
--- a/core/indexer/elasticProcessor.go
+++ b/core/indexer/elasticProcessor.go
@@ -60,7 +60,14 @@ func NewElasticProcessor(arguments ArgElasticProcessor) (ElasticProcessor, error
 		arguments.AddressPubkeyConverter,
 		arguments.ValidatorPubkeyConverter,
 		arguments.FeeConfig,
+		arguments.ShardCoordinator,
+		arguments.IsInImportDBMode,
 	)
+
+	if arguments.IsInImportDBMode {
+		log.Warn("the node is in import mode! Elastic search will not send cross shard transactions and rewards " +
+			"where destination shard is not the current node's shard")
+	}
 
 	if arguments.Options.UseKibana {
 		err = ei.initWithKibana(arguments.IndexTemplates, arguments.IndexPolicies)
@@ -98,6 +105,9 @@ func checkArgElasticProcessor(arguments ArgElasticProcessor) error {
 	}
 	if arguments.Options == nil {
 		return ErrNilOptions
+	}
+	if check.IfNil(arguments.ShardCoordinator) {
+		return ErrNilShardCoordinator
 	}
 
 	return nil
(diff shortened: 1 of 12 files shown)


statusHandler/factory/viewsFactory.go
```diff
@@ -35,7 +38,7 @@ func (wf *viewsFactory) Create() ([]Viewer, error) {
 }
 
 func (wf *viewsFactory) createTermuiConsole() (*termuic.TermuiConsole, error) {
```
SebastianMarian, line 40: vf instead wf

iulianpascalau: Done in the update branch PR

SebastianMarian (dismissed):

SebastianMarian (dismissed):

core/indexer/elasticProcessor.go
```diff
@@ -60,8 +60,15 @@ func NewElasticProcessor(arguments ArgElasticProcessor) (ElasticProcessor, error
 		arguments.AddressPubkeyConverter,
 		arguments.ValidatorPubkeyConverter,
 		arguments.FeeConfig,
+		arguments.ShardCoordinator,
+		arguments.IsInImportDBMode,
 	)
 
+	if arguments.IsInImportDBMode {
+		log.Warn("the node is in import mode! Elastic search will not send cross shard transactions and rewards " +
```
bogdan-rosianu, line 68: rephrase to `cross shard transactions and rewards where destination shard is not the current node's shard won't be indexed in Elastic Search` as the es is the receiver here

core/indexer/processTransactions.go
```diff
@@ -280,6 +291,15 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
 	return transactions, rewardsTxs, alteredAddresses
 }
 
+func (tdp *txDatabaseProcessor) shouldIndex(txHandler data.TransactionHandler) bool {
+	if !tdp.isInImportMode {
+		return true
+	}
+
+	destShardId := tdp.shardCoordinator.ComputeId(txHandler.GetRcvAddr())
+	return destShardId == tdp.shardCoordinator.SelfId()
```
bogdan-rosianu, line 300: what about invalid transactions? they won't be indexed. 
also, a transaction could be `pending` for a long time due to small gas limit and validators avoiding it. I know it doesn't happen that often but it's something to think about

bogdan-rosianu (changes requested):

core/indexer/elasticProcessor.go
iulianpascalau: rephrased

core/indexer/processTransactions.go
iulianpascalau: Invalid transactions are detected and processed only in sender shard (for example: insufficient funds for transfer value but enough balance for the transaction fee)
Gas limit does not impact the transaction processing delay since the excess is returned to the sender. Gas price does. Since we are high throughput system, this is not of a major importance since, in the end, the destination shard will eventually process the transaction.

bogdan-rosianu (approved):

bogdan-rosianu (approved):

core/indexer/processTransactions.go
```diff
@@ -275,10 +275,10 @@ func (tdp *txDatabaseProcessor) groupNormalTxsAndRewards(
 		case block.RewardsBlock:
 			rTxs := getRewardsTransaction(txPool, mb.TxHashes)
 			for hash, rtx := range rTxs {
+				dbTx := tdp.commonProcessor.buildRewardTransaction(rtx, []byte(hash), mbHash, mb, header, mbTxStatus)
+				addToAlteredAddresses(dbTx, alteredAddresses, mb, selfShardID, true)
+				alteredAddresses[dbTx.Receiver] = struct{}{}
```
SebastianMarian, line 280: This set is ok? It is done in the method addToAlteredAddresses called above only under some certain conditions. For case block.TxBlock: above this set is not done, this is why I thought that could be something wrong here or something missed above.

iulianpascalau: removed, thanks

SebastianMarian (approved):

bogdan-rosianu (approved):

PR #2412, opened 2020-10-28
EN-7957/Gas lock in system vm

sasurobert:
Implemented gas lock mechanism in system vm

### DIFF
--- a/vm/process/systemVM.go
+++ b/vm/process/systemVM.go
@@ -8,33 +8,53 @@ import (
 )
 
 type systemVM struct {
-	systemEI        vm.ContextHandler
-	vmType          []byte
-	systemContracts vm.SystemSCContainer
+	systemEI             vm.ContextHandler
+	vmType               []byte
+	systemContracts      vm.SystemSCContainer
+	asyncCallbackGasLock uint64
+	asyncCallStepCost    uint64
+}
+
+// ArgsNewSystemVM defines the needed arguments to create a new system vm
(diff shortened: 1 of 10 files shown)


vm/process/systemVM.go
```diff
@@ -91,10 +111,43 @@ func (s *systemVM) RunSmartContractCall(input *vmcommon.ContractCallInput) (*vmc
...
+	// being used by executing built-in functions; this amount will be restored
+	// to the caller, so that there is sufficient gas for the async callback
+	gasToLock := s.asyncCallStepCost + s.asyncCallbackGasLock
+
+	// gasToDeduct also contains an extra asyncCallStepCost, apart from
+	// gasToLock; asyncCallStepCost will be deducted, but not refunded, just as
+	// Arwen does when executing an async call
+	gasToDeduct := s.asyncCallStepCost + gasToLock
+	if input.GasProvided <= gasToDeduct {
+		return 0, vm.ErrNotEnoughGas
```
raduchis, line 147: the lockedgas would be refunded to the user. In this case, although the asyncCall will not happen, the gas for the gasToLock is taken?

raduchis (approved):

vm/process/systemVM.go
sasurobert: it would have failed in the sender part first. this is only extra protection.

BeniaminDrasovean (approved):

PR #2413, opened 2020-10-28
added log for elastic search response body parse

bogdan-rosianu:
- added an additional log for displaying the body content if the parsing fails
- made each indexer error be a not shippable one, so the indexer will try to re-index the failed data
- treated error in elasticDefaultErrorResponseHandler function
- added unit tests
- changed the indexer so it can better output errors when failing to index a header
- refactored the way receipts and smart contract results are indexed


core/indexer/elasticClientCommon.go
```diff
@@ -55,9 +55,16 @@ func loadResponseBody(body io.ReadCloser, dest interface{}) error {
...
-	decodeErr := loadResponseBody(res.Body, &responseBody)
-	if decodeErr != nil {
-		return decodeErr
+	bodyBytes, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		log.Warn("cannot read elastic response body bytes", "error", err)
+		return err
+	}
+
+	notJson := json.Unmarshal(bodyBytes, &responseBody)
```
iulianpascalau, line 64: err, please use err

core/indexer/elasticClientCommon.go
```diff
@@ -55,9 +55,16 @@ func loadResponseBody(body io.ReadCloser, dest interface{}) error {
...
+	if notJson != nil {
+		log.Warn("cannot unmarshal elastic response body to map[string]interface{}", "decode error", notJson, "body response", string(bodyBytes))
+		return ErrBackOff
```
iulianpascalau, line 67: we should return ErrBackOff only if the err string contains something like `throttled due to too many requests`, otherwise we should return the exact same error. Failing to do this will lead in incorrect masking of returned error when we will drop the extra log.Warns.

core/indexer/elasticClientCommon.go
```diff
@@ -55,9 +55,16 @@ func loadResponseBody(body io.ReadCloser, dest interface{}) error {
...
-		return decodeErr
+	bodyBytes, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		log.Warn("cannot read elastic response body bytes", "error", err)
+		return err
+	}
+
+	notJson := json.Unmarshal(bodyBytes, &responseBody)
+	if notJson != nil {
+		log.Warn("cannot unmarshal elastic response body to map[string]interface{}", "decode error", notJson, "body response", string(bodyBytes))
```
iulianpascalau, line 66: let's return here a wrapped error instead of keeping the wrong pattern of re-logging the same thing over and over again.

iulianpascalau (changes requested):
Good catch, code will need to be changed a little bit.

core/indexer/elasticClientCommon.go
ccorcoveanu: This is a special case that can be easily treated. Totally agree with the re-logging. First, it should be a general rule, that **we don't log an error if we return it**. Special cases where you return another error than what is received can be logged in my opinion.

Here I think is safe to say something like: `received error from elastic service {error}, returning back off to caller`

iulianpascalau: done

core/indexer/elasticClientCommon.go
iulianpascalau: done

core/indexer/elasticClientCommon.go
iulianpascalau: done

sasurobert (dismissed):

core/indexer/processTransactions.go
```diff
@@ -136,6 +141,16 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
 	return append(convertMapTxsToSlice(transactions), rewardsTxs...), alteredAddresses
 }
 
+func (tdp *txDatabaseProcessor) addScrsReceiverToAlteredAccounts(alteredAddress map[string]struct{}, scrs map[string]*smartContractResult.SmartContractResult) {
```
iulianpascalau, line 144: no need to inject shardCoordinator, could have used the current shard ID and a range on the provided body. Miniblocks already contain destination/sender shard ID

core/indexer/processTransactions_test.go
```diff
@@ -328,6 +336,103 @@ func TestGetGasUsedFromReceipt_DataError(t *testing.T) {
...
+	selfShardID := uint32(0)
+
+	expectedAlteredAccounts := make(map[string]struct{})
+	// addresses marked with a comment should be added to the altered addresses map
+
+	// normal txs
+	addressSh0NormalTx1 := []byte("sender in shard 0 - tx 1") // should be added
+	addressSh1NormalTx1 := []byte("receiver in shard 1 - tx 1")
+	expectedAlteredAccounts[hex.EncodeToString(addressSh0NormalTx1)] = struct{}{}
+	tx1 := &transaction.Transaction{SndAddr: addressSh0NormalTx1, RcvAddr: addressSh1NormalTx1}
```
iulianpascalau, line 349: maybe split on multiple lines

core/indexer/processTransactions_test.go
```diff
@@ -328,6 +336,103 @@ func TestGetGasUsedFromReceipt_DataError(t *testing.T) {
...
 }
 
+func TestAlteredAddresses(t *testing.T) {
+	selfShardID := uint32(0)
+
+	expectedAlteredAccounts := make(map[string]struct{})
+	// addresses marked with a comment should be added to the altered addresses map
+
+	// normal txs
+	addressSh0NormalTx1 := []byte("sender in shard 0 - tx 1") // should be added
```
iulianpascalau, line 346: no need for this complex scenario if we drop the shard coordinator injection

core/indexer/processTransactions_test.go
```diff
@@ -328,6 +336,103 @@ func TestGetGasUsedFromReceipt_DataError(t *testing.T) {
...
+	tx1 := &transaction.Transaction{SndAddr: addressSh0NormalTx1, RcvAddr: addressSh1NormalTx1}
+	tx1Hash := []byte("tx1Hash")
+
+	addressSh0NormalTx2 := []byte("receiver in shard 0 - tx 2") // should be added
+	addressSh1NormalTx2 := []byte("sender in shard 1 - tx 2")
+	expectedAlteredAccounts[hex.EncodeToString(addressSh0NormalTx2)] = struct{}{}
+	tx2 := &transaction.Transaction{SndAddr: addressSh1NormalTx2, RcvAddr: addressSh0NormalTx2}
+	tx2Hash := []byte("tx2hash")
+
+	txMiniBlock1 := &block.MiniBlock{Type: block.TxBlock, TxHashes: [][]byte{tx1Hash}, SenderShardID: 0, ReceiverShardID: 1}
```
iulianpascalau, line 358: multiline on all miniblocks definition

core/indexer/processTransactions_test.go
```diff
@@ -328,6 +336,103 @@ func TestGetGasUsedFromReceipt_DataError(t *testing.T) {
...
+	for addrActual := range alteredAddresses {
+		for addrExpected := range expectedAlteredAccounts {
+			if addrActual == addrExpected {
+				count++
+				break
+			}
+		}
+	}
+
+	require.Equal(t, len(expectedAlteredAccounts), count)
```
iulianpascalau, line 433: I think we should change this a little bit:
```GO
for addrActual := range alteredAddresses {
    _, found := expectedAlteredAccounts[addrActual]
    if !found{
        assert.Fail(t, fmt.Sprintf("address %s not found", string(addrActual))
    }
}

require.Equal(t, len(expectedAlteredAccounts), len(alteredAddresses))
```

iulianpascalau (approved):

iulianpascalau (changes requested):
Not done yet

core/indexer/processTransactions.go
```diff
@@ -141,10 +137,27 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
...
+			continue
+		}
+		if mb.ReceiverShardID != selfShardID {
+			continue
+		}
+
+		for _, txHash := range mb.TxHashes {
+			scr, ok := scrs[string(txHash)]
+			if !ok {
+				log.Warn("internal error: missing already computed scr when altering address",
```
bogdan-rosianu, line 157: add `continue` ?

core/mock/shardCoordinatorStub.go
```diff
@@ -0,0 +1,49 @@
+package mock
```
miiu96, line 1: I think this file can be removed.

core/indexer/processTransactions_test.go
```diff
@@ -343,50 +339,99 @@ func TestAlteredAddresses(t *testing.T) {
...
+	}
 	tx1Hash := []byte("tx1Hash")
 
-	addressSh0NormalTx2 := []byte("receiver in shard 0 - tx 2") // should be added
-	addressSh1NormalTx2 := []byte("sender in shard 1 - tx 2")
-	expectedAlteredAccounts[hex.EncodeToString(addressSh0NormalTx2)] = struct{}{}
-	tx2 := &transaction.Transaction{SndAddr: addressSh1NormalTx2, RcvAddr: addressSh0NormalTx2}
+	address3 := []byte("address3") // should be added
+	address4 := []byte("address4")
+	expectedAlteredAccounts[hex.EncodeToString(address4)] = struct{}{}
```
bogdan-rosianu, line 353: the comment is added to `address3` but you add `address4`

core/indexer/processTransactions_test.go
```diff
@@ -420,17 +456,14 @@ func TestAlteredAddresses(t *testing.T) {
...
-				break
-			}
+		_, found := expectedAlteredAccounts[addrActual]
+		if !found {
+			assert.Fail(t, fmt.Sprintf("address %s not found", addrActual))
 		}
 	}
 
-	require.Equal(t, len(expectedAlteredAccounts), count)
+	require.Equal(t, len(expectedAlteredAccounts), len(alteredAddresses))
```
bogdan-rosianu, line 466: after refactoring, this check if duplicated to line 457

core/mock/shardCoordinatorStub.go
iulianpascalau: removed :+1:

core/indexer/processTransactions.go
iulianpascalau: thanks!

core/indexer/processTransactions_test.go
iulianpascalau: done

core/indexer/processTransactions_test.go
iulianpascalau: removed

ccorcoveanu (approved):

iulianpascalau (approved):
Approve after self implementation and self review

sasurobert (approved):

core/indexer/interface.go
```diff
@@ -22,7 +22,8 @@ type DataIndexerFactory interface {
 // This could be an elastic search index, a MySql database or any other external services.
 type Indexer interface {
 	SetTxLogsProcessor(txLogsProc process.TransactionLogProcessorDatabase)
-	SaveBlock(body data.BodyHandler, header data.HeaderHandler, txPool map[string]data.TransactionHandler, signersIndexes []uint64, notarizedHeadersHashes []string)
+	SaveBlock(body data.BodyHandler, header data.HeaderHandler, txPool map[string]data.TransactionHandler,
```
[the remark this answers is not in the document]
iulianpascalau: Indeed, added TODO

core/indexer/processTransactions.go
iulianpascalau: It does not work as the intrashard miniblock holding smart contract results is not added in the body.

sasurobert (dismissed):

iulianpascalau (approved):

core/indexer/processTransactions.go
```diff
@@ -70,9 +74,11 @@ func (tdp *txDatabaseProcessor) prepareTransactionsForDatabase(
 	selfShardID uint32,
 ) ([]*Transaction, map[string]struct{}) {
 	transactions, rewardsTxs, alteredAddresses := tdp.groupNormalTxsAndRewards(body, txPool, header, selfShardID)
-	receipts := groupReceipts(body, txPool)
-	scResults := groupSmartContractResults(body, txPool)
-	tdp.addScrsReceiverToAlteredAccounts(body, alteredAddresses, scResults, selfShardID)
+	//we can not iterate smart contracts results directly on the miniblocks contained in the block body
```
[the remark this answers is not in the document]
iulianpascalau: done

iulianpascalau (dismissed):

sasurobert (dismissed):

iulianpascalau (approved):

andreibancioiu (approved):

raduchis (approved):

PR #2415, opened 2020-10-28
Update production policy setup

ccorcoveanu:
Changed the policy setup for a production environment:
1. All indexes have a number of shards depending on the expected load (more = better for indexing)
2. Rollover should happen at about 20GB of storage/shard (a bit lower for transactions and account history to give them more breathing)
3. Indexes that receive data only once/epoch have one shard
4. Indexes that have a predefined amount of data (tps) should have no rollover policies, so I removed them


sasurobert (approved):
wasn't the feat branch merged to dev ?

iulianpascalau (approved):

LucianMincu (approved):
👍

PR #2434, opened 2020-11-05
Added mapping for rating and group json files for indexer

miiu96:
Fixed problem that appears when rating is saved as a **long** instead of **float**. When information about rating are saved first time in elasticsearch database is a chance that rating of a validator to be seen by database like type  long instead of float.

Problem was fixed. Now when rating index is created is specified also the type of rating. 

Issue: https://github.com/ElrondNetwork/elrond-go/issues/2426


iulianpascalau (approved):

sasurobert (approved):

ccorcoveanu (approved):

LucianMincu (approved):
System tests passed.

PR #2437, opened 2020-11-05
epochStart: rewards computation edgecase fix

AdoAdoAdo:
there is an edge case in rewards distribution which is not treated correctly:
if an eligible node is not selected at all the entire epoch as block producer but participates in signing all the blocks it is assigned in consensus group and all those blocks end up in the canonical blockchain, then the node will not receive any rewards for that particular epoch, although it should have received.

this PR fixes this.

### DIFF
--- a/epochStart/metachain/rewards.go
+++ b/epochStart/metachain/rewards.go
@@ -39,6 +39,8 @@ type ArgsNewRewardsCreator struct {
 	DataPool                      dataRetriever.PoolsHolder
 	ProtocolSustainabilityAddress string
 	NodesConfigProvider           epochStart.NodesConfigProvider
+	// this fixes the situation where a node has 0 proposed blocks and 0 validator failures
+	// but does not receive rewards.
 	RewardsFix1EpochEnable        uint32
 }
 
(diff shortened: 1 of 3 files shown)


sasurobert (dismissed):

epochStart/metachain/rewards.go
```diff
@@ -264,7 +264,7 @@ func (rc *rewardsCreator) computeValidatorInfoPerRewardAddress(
 			rewardsPerBlockPerNodeForShard := rc.mapRewardsPerBlockPerValidator[validatorInfo.ShardId]
 			protocolRewardValue := big.NewInt(0).Mul(rewardsPerBlockPerNodeForShard, big.NewInt(0).SetUint64(uint64(validatorInfo.NumSelectedInSuccessBlocks)))
 
-			if validatorInfo.LeaderSuccess == 0 && validatorInfo.ValidatorFailure == 0 {
+			if validatorInfo.LeaderSuccess == 0 && validatorInfo.ValidatorSuccess == 0 {
```
iulianpascalau, line 267: good catch! :+1:

iulianpascalau (approved):

iulianpascalau (approved):

LucianMincu (approved):
import-db passed on mainnet. 💯

epochStart/metachain/rewards.go
```diff
@@ -264,7 +270,12 @@ func (rc *rewardsCreator) computeValidatorInfoPerRewardAddress(
 			rewardsPerBlockPerNodeForShard := rc.mapRewardsPerBlockPerValidator[validatorInfo.ShardId]
 			protocolRewardValue := big.NewInt(0).Mul(rewardsPerBlockPerNodeForShard, big.NewInt(0).SetUint64(uint64(validatorInfo.NumSelectedInSuccessBlocks)))
 
-			if validatorInfo.LeaderSuccess == 0 && validatorInfo.ValidatorFailure == 0 {
+			isFix1Enabled := rc.isRewardsFix1Enabled(epoch)
```
sasurobert, line 273: May have taken out into a separate function

sasurobert (approved):

raduchis (approved):

PR #2458, opened 2020-11-11
Fix waiting all signatures edge case after NTP sync

SebastianMarian:
* Fixed an edge case in signature round when proposer is waiting after all sigs and meantime a NTP sync update is done with a negative offset
* Added warn log if NTP sync offset is out of normal bounds
* Updated beevik/ntp library to version v0.3.0

**How to be tested:**

**Start a normal system test with #allin and set in config.toml the value of the SyncPeriodSeconds parameter from [NTPConfig] section to a lower value. Ex: from 3600 to 60 seconds. This will help to have more NTP syncs after 1-2 hours of testing and afterwards we can analyze if the behavior of the updated NTP library is good and also if the edge case found is fixed or not.**

### DIFF
--- a/consensus/spos/bls/subroundSignature.go
+++ b/consensus/spos/bls/subroundSignature.go
@@ -216,9 +216,8 @@ func (sr *subroundSignature) doSignatureConsensusCheck() bool {
 
 	areSignaturesCollected, numSigs := sr.areSignaturesCollected(threshold)
 	areAllSignaturesCollected := numSigs == sr.ConsensusGroupSize()
-	isTimeOut := sr.remainingTime() <= 0
 
-	isJobDoneByLeader := isSelfLeader && (areAllSignaturesCollected || (areSignaturesCollected && isTimeOut))
+	isJobDoneByLeader := isSelfLeader && (areAllSignaturesCollected || (areSignaturesCollected && sr.WaitingAllSignaturesTimeOut))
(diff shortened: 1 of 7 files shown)


consensus/spos/bls/subroundSignature.go
```diff
@@ -280,6 +279,8 @@ func (sr *subroundSignature) waitAllSignatures() {
 		return
 	}
 
+	sr.WaitingAllSignaturesTimeOut = true
```
iulianpascalau, line 282: can we have a race condition here? If yes, you can use atomic.Flag

SebastianMarian: Actually not. As the reset is done on startRound  initialization and this go routine stays live only 55% of the current round and this set on true is done in the middle of the current round

sasurobert: +1

sasurobert (approved):

consensus/spos/bls/subroundSignature.go
SebastianMarian: But we can use atomic.Flag anyway

iulianpascalau (approved):

consensus/spos/bls/subroundSignature.go
iulianpascalau: Not necessarily if we can not have a race condition. Asked just to double-check. Green from me

LucianMincu (approved):

PR #2460, opened 2020-11-11
Gas schedule tests

sasurobert:


integrationTests/vm/arwen/arwenVM/gasSchedule_test.go
```diff
@@ -0,0 +1,236 @@
...
+func Benchmark_TestCryptoSecp256k1Compressed(b *testing.B) {
+	runWASMVMBenchmark(b, "../testdata/c-api-tests/crypto/output/cryptoTest.wasm", 0, "verifySecp256k1CompressedKeyTest", nil, b.N, nil)
+}
+
+func Benchmark_TestCryptoDoNothing(b *testing.B) {
+	runWASMVMBenchmark(b, "../testdata/c-api-tests/crypto/output/cryptoTest.wasm", 0, "doNothing", nil, b.N, nil)
+}
+
+func Benchmark_TestStorageRust(b *testing.B) {
+	gasSchedule, _ := core.LoadGasScheduleConfig("../gasSchedule.toml")
```
camilbancioiu, line 117: Should this benchmark be run with `gasScheduleV2.toml`? Or using the old gas schedule was intentional, for comparison?

sasurobert: StorageTest had good values with V1 as well so did not change it as the difference was small.

camilbancioiu (dismissed):

cmd/node/config/gasSchedules/gasScheduleV2.toml
```diff
@@ -0,0 +1,607 @@
+[BuiltInCost]
```
iulianpascalau, line 1: Synthesis of the changes:
1. AoTPreparePerByte increased 4x
2. Elrond API costs increased some operations 10x-50x
3. Eth API costs increased some operations 10x-1000x
4. Big int API costs increased all operations 10x-60x except load & store
5. Crypto API cost increased the VerifyEd25519 operation 200x

cmd/node/config/gasSchedules/gasScheduleV2.toml
```diff
@@ -0,0 +1,607 @@
...
+    BigIntGetUnsignedArgument   = 1000
+    BigIntGetSignedArgument     = 1000
+    BigIntGetCallValue          = 1000
+    BigIntGetExternalBalance    = 10000
+
+[CryptoAPICost]
+    SHA256          = 600
+    Keccak256       = 600
+    Ripemd160       = 600
+    VerifyBLS       = 1000
```
iulianpascalau, line 154: Why is this so cheap? It is actually the hardest computation stuff we have in the protocol. Can we have a test that checks a couple of BLS signatures in the VM? Also, the rest of the crypto operations are calibrated?

integrationTests/vm/arwen/gasScheduleV2.toml
```diff
@@ -0,0 +1,607 @@
+[BuiltInCost]
```
iulianpascalau, line 1: Suggestion to not have this file but instead link it to .../elrond-go/node/cmd/config/gasSchedules as to avoid the situation when we push modifications only in the node config file without changing this file

iulianpascalau (changes requested):

cmd/node/config/gasSchedules/gasScheduleV2.toml
sasurobert: We have. Check the benchmarks and the corresponding C file. Actually this is executed rather fast.

cmd/node/config/gasSchedules/gasScheduleV2.toml
```diff
@@ -0,0 +1,607 @@
...
+    DelegateExecution    = 160000
+    AsyncCallStep        = 200000
+    AsyncCallbackGasLock = 20000000
+    ExecuteReadOnly      = 160000
+    CreateContract       = 300000
+    GetReturnData        = 100
+    GetNumReturnData     = 100
+    GetReturnDataSize    = 100
+
+[EthAPICost]
```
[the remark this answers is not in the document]
sasurobert: It is still not completely removed from Arwen side and would crash at run time if no costs are provided for all APIs and all opcode's. First we need to remove it from Arwen. It is planned.

integrationTests/vm/arwen/gasScheduleV2.toml
sasurobert: First we change this file. Only after the one on node side. In the past it did not work to link from config. Will try it now.

iulianpascalau: It will work and will comply with the rest of the project when the align factories to specs will be merged (that already uses the original files from node/cmd/config).

cmd/node/config/gasSchedules/gasScheduleV2.toml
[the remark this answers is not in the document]
iulianpascalau: Great!

iulianpascalau (approved):

camilbancioiu (approved):

PR #2468, opened 2020-11-13
view-func-inactive-with-bad-rating

sasurobert:
fixed edge case on view function when a node can be inactive, still  lose some rating and have a bad rating, thus will not be able to unbond before another unjail.


sasurobert:
this is not the fix

PR #2469, opened 2020-11-13
jail inactive validator with bad rating

sasurobert:


iulianpascalau (approved):
GJ! Will need to figure out how to test this

iulianpascalau (dismissed):

raduchis (dismissed):

raduchis (approved):

process/peer/process.go
```diff
@@ -680,7 +700,7 @@ func (vs *validatorStatistics) setToJailedIfNeeded(
 		return
 	}
 
-	if peerAccount.GetUnStakedEpoch() == core.DefaultUnstakedEpoch && peerAccount.GetList() == string(core.InactiveList) {
+	if vs.isValidatorWithLowRating(peerAccount) {
```
SebastianMarian, line 703: && peerAccount.GetList() == string(core.InactiveList), as it could be at this point already jailed

sasurobert: That set does not matter. It is ok right now. In case of waiting or eligible there is an early return

SebastianMarian (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2471, opened 2020-11-13
Hardfork: execute pending txs in correct order

SebastianMarian:
* Implemented a fix in Hardfork mechanism, which would execute all pending transactions in the correct order
* Exported/Imported unFinished metaBlocks as well, as they are needed for execution of pending transactions in the correct order
* Extracted common used code in a new file
* Refactored some prints and some methods names

### DIFF
--- a/update/common.go
+++ b/update/common.go
@@ -1,0 +1,124 @@
+package update
+
+import (
+	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/data/block"
+)
+
+// TxInfo defines the structure which hold the tx info
+type TxInfo struct {
+	TxHash []byte
+	Tx     data.TransactionHandler
+}
+
+// GetPendingMiniBlocks get all the pending miniBlocks from epoch start metaBlock and unFinished metaBlocks
+func GetPendingMiniBlocks(
+	epochStartMetaBlock *block.MetaBlock,
+	unFinishedMetaBlocksMap map[string]*block.MetaBlock,
+) ([]block.MiniBlockHeader, error) {
+
+	if epochStartMetaBlock == nil {
+		return nil, ErrNilEpochStartMetaBlock
+	}
+	if unFinishedMetaBlocksMap == nil {
+		return nil, ErrNilUnFinishedMetaBlocksMap
+	}
+
+	pendingMiniBlocks := make([]block.MiniBlockHeader, 0)
+	nonceToHashMap := CreateNonceToHashMap(unFinishedMetaBlocksMap)
+
+	for _, shardData := range epochStartMetaBlock.EpochStart.LastFinalizedHeaders {
+		computedPending, err := ComputePendingMiniBlocksFromUnFinished(
+			shardData,
+			unFinishedMetaBlocksMap,
+			nonceToHashMap,
+			epochStartMetaBlock.GetNonce(),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		pendingMiniBlocks = append(pendingMiniBlocks, computedPending...)
+	}
+
+	//TODO: These pendingMiniBlocks are already added in method ComputePendingMiniBlocksFromUnFinished called above
+	//when nonce == epochStartMetaBlockNonce in "for" done there, by calling method GetAllMiniBlocksWithDst inside it.
+	//The below commented code should be removed, otherwise it would duplicate these miniblocks
+	//(especially the rewards miniblocks) which would be executed twice!
+	//for _, mbHdr := range epochStartMetaBlock.MiniBlockHeaders {
+	//	if mbHdr.SenderShardID != core.MetachainShardId || mbHdr.Type == block.PeerBlock {
+	//		continue
+	//	}
+	//	pendingMiniBlocks = append(pendingMiniBlocks, mbHdr)
+	//}
+
+	return pendingMiniBlocks, nil
+}
+
+// CreateNonceToHashMap creates a map of nonce to hash from all the given metaBlocks
+func CreateNonceToHashMap(unFinishedMetaBlocks map[string]*block.MetaBlock) map[uint64]string {
+	nonceToHashMap := make(map[uint64]string, len(unFinishedMetaBlocks))
+	for metaBlockHash, metaBlock := range unFinishedMetaBlocks {
+		nonceToHashMap[metaBlock.GetNonce()] = metaBlockHash
+	}
+
+	return nonceToHashMap
+}
+
+// ComputePendingMiniBlocksFromUnFinished computes all the pending miniBlocks from unFinished metaBlocks
+func ComputePendingMiniBlocksFromUnFinished(
+	epochStartShardData block.EpochStartShardData,
+	unFinishedMetaBlocks map[string]*block.MetaBlock,
+	nonceToHashMap map[uint64]string,
+	epochStartMetaBlockNonce uint64,
+) ([]block.MiniBlockHeader, error) {
+	pendingMiniBlocks := make([]block.MiniBlockHeader, 0)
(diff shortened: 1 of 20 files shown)


update/common.go
```diff
@@ -0,0 +1,124 @@
...
+	//		continue
+	//	}
+	//	pendingMiniBlocks = append(pendingMiniBlocks, mbHdr)
+	//}
+
+	return pendingMiniBlocks, nil
+}
+
+// CreateNonceToHashMap creates a map of nonce to hash from all the given metaBlocks
+func CreateNonceToHashMap(unFinishedMetaBlocks map[string]*block.MetaBlock) map[uint64]string {
```
iulianpascalau, line 59: do we need all these functions exported?

update/common.go
```diff
@@ -0,0 +1,124 @@
...
+
+		for _, mbHdr := range metaBlock.ShardInfo[i].ShardMiniBlockHeaders {
+			if mbHdr.ReceiverShardID == destShardID && mbHdr.SenderShardID != destShardID {
+				mbHdrs = append(mbHdrs, mbHdr)
+			}
+		}
+	}
+
+	for _, mbHdr := range metaBlock.MiniBlockHeaders {
+		if mbHdr.ReceiverShardID == destShardID && mbHdr.SenderShardID != destShardID {
```
iulianpascalau, line 118:
```GO
isForMe := (mbHdr.ReceiverShardID == destShardID || mbHdr.ReceiverShardID == core.AllShardId) && mbHdr.SenderShardID != destShardID
````
as we have to take into account all shard ID value

File: update/process/processPending.go [hunk 3]
```diff
@@ -76,29 +70,27 @@ func NewPendingTransactionProcessor(args ArgsPendingTransactionProcessor) (*pend
 }
 
 // ProcessTransactionsDstMe processes all the transactions in which destination is the current shard
-func (p *pendingProcessor) ProcessTransactionsDstMe(mapTxs map[string]data.TransactionHandler) (block.MiniBlockSlice, error) {
-	sortedTxs := getSortedSliceFromTxsMap(mapTxs)
-
+func (p *pendingProcessor) ProcessTransactionsDstMe(txsInfo []*update.TxInfo) (block.MiniBlockSlice, error) {
```
iulianpascalau, line 73: :+1:

update/process/shardBlock.go
```diff
@@ -161,6 +164,45 @@ func (s *shardBlockCreator) createBody() (*block.Body, error) {
...
+			return nil, update.ErrMiniBlockNotFoundInImportedMap
+		}
+
+		for _, txHash := range miniBlock.TxHashes {
+			tx, transactionFound := importedTransactionsMap[string(txHash)]
+			if !transactionFound {
+				return nil, update.ErrTransactionNotFoundInImportedMap
+			}
+
+			txsInfo = append(txsInfo, &update.TxInfo{TxHash: txHash, Tx: tx})
```
iulianpascalau, line 195: multiline?

sasurobert (changes requested):

update/common.go
SebastianMarian: No, only GetPendingMiniBlocks, I will make them unexported

update/common.go
SebastianMarian: Actually core.AllShardId refers to PeerBlock right now, and PendingTransactionProcessor does not know to process this type of miniblock. More than this, in the commented code above from line 49 this type of miniblock was also avoided in the original code. Actually, the whole file is just a copy-paste of the original one with slightly renaming here and there.  We should discuss/analyze if this type of miniblock which could appear only in the start of epoch metablock (hardfork meta block) should be executed or not in the import phase or it would be executed in a normal way afterwards.

update/process/shardBlock.go
SebastianMarian: Done

update/common.go
```diff
@@ -0,0 +1,124 @@
...
+
+		pendingMiniBlocksFromMetaBlock := GetAllMiniBlocksWithDst(metaBlock, epochStartShardData.ShardID)
+		pendingMiniBlocks = append(pendingMiniBlocks, pendingMiniBlocksFromMetaBlock...)
+	}
+
+	return pendingMiniBlocks, nil
+}
+
+// GetAllMiniBlocksWithDst returns all miniBlock headers with the given destination from the given metaBlock
+func GetAllMiniBlocksWithDst(metaBlock *block.MetaBlock, destShardID uint32) []block.MiniBlockHeader {
```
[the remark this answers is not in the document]
SebastianMarian: It is one method in data/metablock.go -> GetOrderedCrossMiniblocksWithDst but It's not the same, it returns a structure called MiniBlockInfo which not contains the miniblock, but only its Hash, SenderShardID and Round. Also it returns PeerMiniBlocks as well.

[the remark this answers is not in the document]
SebastianMarian: Refactored the original method anyway, and deleted this one. We should discuss anyway about PeerBlock miniblocks, as right now PendingTransactionProcessor does not know to process this type of miniblock, and for now they should be skipped.

update/common.go
SebastianMarian: Anyway I have deleted this method and I refactored the one from data/metablock.go -> GetOrderedCrossMiniblocksWithDst to be useful also here. This method takes in count also the PeerBlock miniblocks which should be skipped for now here, as PendingTransactionProcessor does not know to process them.

update/genesis/export.go
```diff
@@ -180,31 +186,71 @@ func (se *stateExport) exportAllTries() error {
...
+	}
+
+	log.Debug("Exported metaBlock",
+		"identifier", identifier,
+		"version key", versionKey,
+		"hash", metaHash,
+		"epoch", metaBlock.Epoch,
+		"round", metaBlock.Round,
+		"nonce", metaBlock.Nonce,
+		"start of epoch block", metaBlock.Nonce == 0 || metaBlock.IsStartOfEpochBlock(),
```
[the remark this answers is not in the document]
SebastianMarian: I took this condition from update/sync/syncHeaders.go method below:

```
// GetEpochStartMetaBlock returns the synced epoch start metaBlock
func (h *headersToSync) GetEpochStartMetaBlock() (*block.MetaBlock, error) {
	h.mutMeta.Lock()
	meta := h.epochStartMetaBlock
	h.mutMeta.Unlock()

	if meta.IsStartOfEpochBlock() || meta.Nonce == 0 {
		return meta, nil
	}

	return nil, update.ErrNotSynced
}

```
Also IsStartOfEpochBlock has this definition:

```
// IsStartOfEpochBlock verifies if the block is of type start of epoch
func (m *MetaBlock) IsStartOfEpochBlock() bool {
	return len(m.EpochStart.LastFinalizedHeaders) > 0
}

```
If GenesisBlock with nonce 0 has len(LastFinalizedHeaders) > 0 this method returns false

update/process/shardBlock.go
```diff
@@ -161,6 +164,45 @@ func (s *shardBlockCreator) createBody() (*block.Body, error) {
 	}, nil
 }
 
+func (s *shardBlockCreator) getPendingTxsInCorrectOrder() ([]*update.TxInfo, error) {
```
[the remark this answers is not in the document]
SebastianMarian: Done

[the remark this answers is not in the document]
sasurobert: OOM in for all when you have millions of transactions.

sasurobert (approved):

update/common.go
iulianpascalau: I know, more mental mapping to have as to not use core.AllShardId on miniblocks other than peer miniblocks because there are some edgecases.

iulianpascalau (approved):

LucianMincu (approved):

PR #2503, opened 2020-11-24
EN 8164 new kad dht peer discoverer implementation

iulianpascalau:
- added new kad dht peer discoverer instance. The old (legacy) instance and the new one can be switched from p2p.toml config file

Specifications for the new peer discoverer implementation:
- the initial connections for seeders attempt, if failed, will not stop the node starting its bootstrapping process. This will improve the initial connections for the seeders as they can be cross-linked in the p2p.toml file with ease. 
- added automatic reconnection to the seeders as to allow network un-split if the seeder(s) recover from unexpected shutdown
- the reconnection to seeders interval is now adaptive (2 levels) as to retry more often to reconnect to the seeders in case no connection succeeded


p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
```diff
@@ -0,0 +1,231 @@
...
+		if err != nil {
+			log.Debug("error connecting to seeder", "seeder", seederAddress, "error", err.Error())
+		} else {
+			connectedToOneSeeder = true
+		}
+
+		select {
+		case <-ctx.Done():
+			return true
+		default:
```
AdoAdoAdo, line 190: maybe add a sleep here? or use instead a case <- time.after

AdoAdoAdo (changes requested):

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, err
+	}
+
+	go okdd.processLoop(arg.Context)
+
+	return okdd, nil
+}
+
+// Bootstrap will start the bootstrapping new peers process
+func (okdd *optimizedKadDhtDiscoverer) Bootstrap() error {
```
ccorcoveanu, line 77: Is there any logic flow that prevents this implementation in having the same function calls as `continuousKadDhtDiscoverer`? In the first implementation `Bootstrap` calls `startDHT` and will actually start the process loop, whereas in the new implementation the process loop is started in the constructor, `Bootstrap` only gives the go signal.

Having the same flow and will make it easier for anyone looking at the difference between the two implementations in parallel.

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
```diff
@@ -0,0 +1,231 @@
...
+		bucketSize:                  arg.BucketSize,
+		routingTableRefresh:         arg.RoutingTableRefresh,
+		status:                      statNotInitialized,
+		chanInit:                    make(chan struct{}),
+		errChanInit:                 make(chan error),
+		chanConnectToSeeders:        make(chan struct{}),
+		chanDoneConnectToSeeders:    make(chan struct{}),
+	}
+
+	okdd.initKadDhtHandler = okdd.initKadDht
```
ccorcoveanu, line 65: Why do it like this? Do we plan to allow to inject this handler from other places? Otherwise I don't see the point of a function pointer here.

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
```diff
@@ -0,0 +1,231 @@
...
+		return err
+	}
+
+	okdd.kadDHT = kadDhtHandler
+	okdd.status = statInitialized
+
+	return nil
+}
+
+func (okdd *optimizedKadDhtDiscoverer) initKadDht(ctx context.Context) (KadDhtHandler, error) {
```
ccorcoveanu, line 140: This is more like a create instead of init?

ccorcoveanu (changes requested):
Just some small suggestions

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
iulianpascalau: Was impossible to test otherwise

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
iulianpascalau: can be renamed

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
iulianpascalau: The `continuousKadDhtDiscoverer` will be pending for removal (after this new implementation will be tested thoroughly) because it is harder to maintain and there are a lot of pitfalls there. 
Regarding the starting of the go routine I have decided to implement it in this way as to avoid mutexes (we need some sort of mutex usage and flags when starting the go routine as to not start it twice).

p2p/libp2p/discovery/optimizedKadDhtDiscoverer.go
iulianpascalau: done

AdoAdoAdo (dismissed):

ccorcoveanu (approved):

AdoAdoAdo (approved):

ccorcoveanu (approved):

PR #2515, opened 2020-11-26
En 7444 sync from genesis resolvers

iulianpascalau:
- changed the resolvers so they can be used by the full history nodes by taking into account the provided epoch value

### DIFF
--- a/dataRetriever/resolvers/baseSimpleResolver.go
+++ b/dataRetriever/resolvers/baseSimpleResolver.go
@@ -1,0 +1,17 @@
+package resolvers
+
+import (
+	"github.com/ElrondNetwork/elrond-go/storage"
+)
+
+type baseSimpleResolver struct {
+	storer storage.Storer
+}
+
+func (bsr *baseSimpleResolver) getFromStorage(key []byte, _ uint32) ([]byte, error) {
+	return bsr.storer.SearchFirst(key)
+}
+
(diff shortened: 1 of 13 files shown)


dataRetriever/resolvers/baseSimpleResolver.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/ElrondNetwork/elrond-go/storage"
+)
+
+type baseSimpleResolver struct {
+	storer storage.Storer
+}
+
+func (bsr *baseSimpleResolver) getFromStorage(key []byte, _ uint32) ([]byte, error) {
+	return bsr.storer.SearchFirst(key)
```
SebastianMarian, line 12: You can call bsr.searchFirst(key), as it does the same thing, and this way will avoid to have two identical methods

iulianpascalau: I need those 2 identical methods as to satisfy the newly introduced interface that has a different implementation on `baseFullHistoryResolver` struct. For simplicity, I would keep this code as it is.

SebastianMarian: ok

SebastianMarian (approved):

raduchis (dismissed):

raduchis (approved):

SebastianMarian (approved):

PR #2516, opened 2020-11-26
Dev to staking 26nov

sasurobert:


vm/systemSmartContracts/delegationManager.go
```diff
@@ -380,6 +382,13 @@ func (d *delegationManager) saveDelegationContractList(list *DelegationContractL
 	return nil
 }
 
+// SetNewGasCost is called whenever a gas cost was changed
+func (d *delegationManager) SetNewGasCost(gasCost vm.GasCost) {
+	d.mutExecution.Lock()
```
iulianpascalau, line 387: We are only locking the mutex for setting the gas lock but we are not setting the mutex when actually using the gas lock in `Execute` method.

sasurobert: done

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #2517, opened 2020-11-27
EN-8343-optimize-trie-sync

BeniaminDrasovean:
Instead of keeping the whole trie in memory during the syncing process, and Commit() the trie when it is fully synced, only keep the collapsed root and commit each node individually as it is synced.

### DIFF
--- a/data/trie/sync.go
+++ b/data/trie/sync.go
@@ -100,11 +100,6 @@ func (ts *trieSyncer) StartSyncing(rootHash []byte, ctx context.Context) error {
 
 		numUnResolved := ts.requestNodes()
 		if !shouldRetryAfterRequest && numUnResolved == 0 {
-			err = ts.trie.Commit()
-			if err != nil {
-				return err
-			}
-
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


data/trie/sync.go
```diff
@@ -100,11 +100,6 @@ func (ts *trieSyncer) StartSyncing(rootHash []byte, ctx context.Context) error {
 
 		numUnResolved := ts.requestNodes()
 		if !shouldRetryAfterRequest && numUnResolved == 0 {
-			err = ts.trie.Commit()
```
sasurobert, line 103: why delete this ? isn't it better to have the final commit done ?

data/trie/sync.go
```diff
@@ -186,6 +176,22 @@ func (ts *trieSyncer) checkIfSynced() (bool, error) {
...
+
+			if !ts.rootFound && bytes.Equal([]byte(nodeHash), ts.rootHash) {
+				var collapsedRoot node
+				collapsedRoot, err = currentNode.getCollapsed()
+				if err != nil {
+					return false, err
+				}
+
+				ts.trie.root = collapsedRoot
+				ts.rootFound = true
```
sasurobert, line 193: so we set root only after finding all its childs ?

data/trie/sync.go
BeniaminDrasovean: If we commit here, the whole trie will be loaded again from storage. The current approach is to commit each node as it is synced.

data/trie/sync.go
BeniaminDrasovean: Yes, it should have no impact upon the functioning of the algorithm.

sasurobert (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #2520, opened 2020-11-27
fix epoch-root hash storer config

bogdan-rosianu:
Fixed the configuration for the `TrieEpochRootHashStorage`. The previous config would have caused the node to stop with the error `cache size is lower than batch size`

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -214,7 +214,8 @@
     [TrieEpochRootHashStorage.Cache]
         Name = "TrieEpochRootHashCache"
         Capacity = 1000
-        Type = "LRU"
+        Type = "SizeLRU"
+        SizeInBytes = 314572800 #300MB
     [TrieEpochRootHashStorage.DB]
         FilePath = "TrieEpochRootHashStorageDB"
         Type = "LvlDBSerial"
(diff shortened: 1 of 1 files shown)


iulianpascalau (commented):
Please provide a description

iulianpascalau (dismissed):

BeniaminDrasovean (approved):

cmd/node/config/config.toml
```diff
@@ -213,14 +213,13 @@
 [TrieEpochRootHashStorage]
     [TrieEpochRootHashStorage.Cache]
         Name = "TrieEpochRootHashCache"
-        Capacity = 10
-        Type = "SizeLRU"
-        SizeInBytes = 5242 #5MB
+        Capacity = 1000
+        Type = "LRU"
```
AdoAdoAdo, line 217: don't we want to add also limit in bytes?
I would suggest to add 500MB

bogdan-rosianu: I don't think that much is needed because we only write in that persister. The read is done by an external tool. Changes cache type to `SizeLRU` of 300 MB

AdoAdoAdo (approved):

iulianpascalau (approved):

BeniaminDrasovean (approved):

PR #2521, opened 2020-11-27
max number of shards for genesis config

AdoAdoAdo:
Add maximum number of shards for genesis nodes config

This will allow more flexible initial test-net nodes config.
Should be backwards compatible, as long as the set maximum number of shards >=  with the legacy end result (3 shards without MetaChain)

Legacy implementation would have split the nodes into maximum number of shards allowed by the constraints (number of eligible nodes per shard + minimum hysteresis for waiting nodes), now even if more shards could be created, the given maximum number of shards is kept as a hard limit, the remaining nodes being distributed to the shards waiting lists.

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -46,6 +46,9 @@
    # GenesisString represents the encoded string for the genesis block
    GenesisString = "67656E65736973"
 
+   # GenesisMaxNumberOfShards represents the maximum number of shards to be created at genesis (excluding metaChain shard)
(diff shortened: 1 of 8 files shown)


sasurobert (approved):

BeniaminDrasovean (dismissed):

sasurobert (approved):

BeniaminDrasovean (approved):

LucianMincu (approved):
Initial tests passed. More stress-testing follows. 👍

PR #2522, opened 2020-11-27
Full history lru persisters

raduchis:
Added functionality for full history nodes' storers to keep a list of oldActivePersisters in order to server other nodes that are synchronizing older epochs.

* added a LRUCache for older epoch persisters


cmd/node/flags.go
```diff
@@ -382,6 +382,26 @@ func applyFlags(ctx *cli.Context, cfgs *config.Configs, log logger.Logger) {
 	applyCompatibleConfigs(isInImportMode, importDbNoSigCheckFlag, log, cfgs.GeneralConfig, cfgs.P2pConfig)
 	flagsConfig.IsInImportMode = isInImportMode
 	flagsConfig.ImportDbNoSigCheckFlag = importDbNoSigCheckFlag
+	flagsConfig.ImportDbDirectory = importDbDirectoryValue
+
+	// if FullArchive is enabled, we override the conflicting StoragePruning settings and StartInEpoch as well
```
iulianpascalau, line 387: please remove L387-L404 as they are done in the applyCompatibleConfigs function

storage/lrucache/lrucache.go
```diff
@@ -29,6 +29,24 @@ func NewCache(size int) (*lruCache, error) {
 		return nil, err
 	}
 
+	c := createLRUCache(size, cache)
+
+	return c, nil
+}
+
+// NewCacheWithEviction creates a new sized LRU cache instance with eviction function
+func NewCacheWithEviction(size int, onEvicted func(key interface{}, value interface{})) (*lruCache, error) {
```
iulianpascalau, line 38: :+1:

storage/pruning/fullHistoryPruningStorer.go
```diff
@@ -0,0 +1,133 @@
...
+	pdata, exists := fhps.oldEpochsActivePersisters.Get([]byte(epochString))
+	fhps.lock.RUnlock()
+	if !exists {
+		newPdata, errPersisterData := createPersisterDataForEpoch(fhps.args, epoch, fhps.shardId)
+		if errPersisterData != nil {
+			return nil, errPersisterData
+		}
+
+		fhps.oldEpochsActivePersisters.Put([]byte(epochString), newPdata, 0)
+		fhps.persistersMapByEpoch[epoch] = newPdata
```
iulianpascalau, line 111: shouldn't this be protected by mutex?
I think get-and-open-if-missing should be done in a single Lock-Unlock area as to avoid overwriting persisters that will never get closed (L101-L119). Performance is still desired but not mandatory here.

storage/pruning/pruningStorer.go
```diff
@@ -866,21 +870,22 @@ func (ps *PruningStorer) closeAndDestroyPersisters(epoch uint32) error {
 	}
 	ps.lock.Unlock()
 
-	for _, p := range persistersToClose {
-		err := p.persister.Close()
+	for _, pd := range persistersToClose {
+		//TODO: check if this should return the error or try to close all other persisters first like other methods
```
iulianpascalau, line 874: should try to close as much as it can (we can remove the TODO) + keep log.Error id closing fails.

iulianpascalau (changes requested):

cmd/node/flags.go
raduchis: yes, I think one of the merges dupplicated the code

storage/pruning/pruningStorer.go
raduchis: changed to log.warn and removed //TODO:

cmd/node/flags.go
```diff
@@ -382,26 +382,6 @@ func applyFlags(ctx *cli.Context, cfgs *config.Configs, log logger.Logger) {
 	applyCompatibleConfigs(isInImportMode, importDbNoSigCheckFlag, log, cfgs.GeneralConfig, cfgs.P2pConfig)
 	flagsConfig.IsInImportMode = isInImportMode
 	flagsConfig.ImportDbNoSigCheckFlag = importDbNoSigCheckFlag
-	flagsConfig.ImportDbDirectory = importDbDirectoryValue
```
iulianpascalau: I think we still need this line

storage/pruning/fullHistoryPruningStorer.go
```diff
@@ -96,11 +96,46 @@ func (fhps *FullHistoryPruningStorer) searchInEpoch(key []byte, epoch uint32) ([
...
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := persister.Get(key)
+	if err == nil {
+		return res, nil
+	}
+
+	log.Warn("GetFromEpoch persister",
```
iulianpascalau, line 109: I would keep this as maximum to log.Debug (maybe log.Trace?), otherwise there is an attack vector: I connect to a history node and request data from a very-large epoch

iulianpascalau (dismissed):

storage/pruning/fullHistoryPruningStorer.go
```diff
@@ -0,0 +1,156 @@
...
+
+	pdata, exists = fhps.oldEpochsActivePersisters.Get([]byte(epochString))
+	if !exists {
+		newPdata, errPersisterData := createPersisterDataForEpoch(fhps.args, epoch, fhps.shardId)
+		if errPersisterData != nil {
+			return nil, errPersisterData
+		}
+
+		fhps.oldEpochsActivePersisters.Put([]byte(epochString), newPdata, 0)
+		fhps.persistersMapByEpoch[epoch] = newPdata
```
AdoAdoAdo, line 146: mutex protection?

AdoAdoAdo (changes requested):

AdoAdoAdo (approved):

iulianpascalau (approved):

PR #2531, opened 2020-11-30
En 8339/validator unstake unbond

sasurobert:
New functions for unStake and unBond with value, with nodes and with all.
Deleted slashing from staking SC - it needs to be implemented to validator SC later on.


vm/systemSmartContracts/validator.go
```diff
@@ -896,9 +900,38 @@ func (v *validatorSC) unStake(args *vmcommon.ContractCallInput) vmcommon.ReturnC
...
+
+	numSuccess := v.unStakeNodesFromStakingSC(args.Arguments, registrationData)
+	if !v.flagEnableTopUp.IsSet() {
+		return vmcommon.Ok
+	}
+
+	// continue by unstaking tokens as well
+	validatorConfig := v.getConfig(v.eei.BlockChainHook().CurrentEpoch())
+	unStakeFromNodes := big.NewInt(0).Mul(validatorConfig.NodePrice, big.NewInt(0).SetUint64(numSuccess))
+	if unStakeFromNodes.Cmp(registrationData.TotalStakeValue) > 0 {
```
iulianpascalau, line 925: shouldn't this be an error? In what case can we have unStakeFromNodes > TotalStakeValue ? (changed NodePrice ?)

vm/systemSmartContracts/validator.go
```diff
@@ -1176,13 +1183,19 @@ func (v *validatorSC) unStakeTokens(args *vmcommon.ContractCallInput) vmcommon.R
...
 }
 
-func (v *validatorSC) unStakeValue(registrationData *ValidatorDataV2, unStakeValue *big.Int) vmcommon.ReturnCode {
-	registrationData.TotalStakeValue.Sub(registrationData.TotalStakeValue, unStakeValue)
-	if registrationData.TotalStakeValue.Cmp(zero) < 0 {
-		v.eei.AddReturnMessage("contract error on unStakeTokens function, total stake < unstake value")
+func (v *validatorSC) processUnStakeValue(registrationData *ValidatorDataV2, unStakeValue *big.Int) vmcommon.ReturnCode {
+	unstakeValueIsOk := unStakeValue.Cmp(v.minUnstakeTokensValue) >= 0 || unStakeValue.Cmp(registrationData.TotalStakeValue) == 0
+	if !unstakeValueIsOk {
+		v.eei.AddReturnMessage("can not unstake the provided value either because is under the minimum threshold or " +
```
iulianpascalau, line 1189: I think this will break the backwards compatibility if it can happen in a situation that can now happen on the mainnet.
Strongly suggest as to run the new code if the topupV2 flag is set, otherwise run the exact same old code. We should have different .go files containing old/legacy code in order to keep the logic in hand.

vm/systemSmartContracts/validator.go
```diff
@@ -1311,7 +1324,7 @@ func (v *validatorSC) unBondTokensFromRegistrationData(
 	registrationData.UnstakedInfo = registrationData.UnstakedInfo[index:]
 	registrationData.TotalUnstaked.Sub(registrationData.TotalUnstaked, totalUnBond)
 	if registrationData.TotalUnstaked.Cmp(zero) < 0 {
-		v.eei.AddReturnMessage("contract error on unBondTokens function, total unstaked < total unbond")
+		v.eei.AddReturnMessage("too much requested to unBond")
```
iulianpascalau, line 1327: compatibility issues?

iulianpascalau (changes requested):
I think we should have a different approach (before is too late) maintaining old code vs new code. Strongly suggest to have different .go files

vm/systemSmartContracts/validator.go
sasurobert: UnStakeTokens before and unStake after.

vm/systemSmartContracts/validator.go
sasurobert: This is exactly the same as before. This does not have any backward compatibility issues.

vm/systemSmartContracts/validator.go
sasurobert: Will revert.

vm/systemSmartContracts/validator.go
sasurobert: This means a lot of copy pasted code.

vm/systemSmartContracts/validator.go
raduchis: Is the totalStakeValue changed by slashing?

vm/systemSmartContracts/validator.go
```diff
@@ -907,15 +940,22 @@ func (v *validatorSC) unStake(args *vmcommon.ContractCallInput) vmcommon.ReturnC
 	return vmcommon.Ok
 }
 
-func (v *validatorSC) unStakeNodeOneNodeFromStakingSC(blsKey []byte, rewardAddress []byte) vmcommon.ReturnCode {
-	vmOutput, errExec := v.executeOnStakingSC([]byte("unStake@" + hex.EncodeToString(blsKey) + "@" + hex.EncodeToString(rewardAddress)))
-	if errExec != nil {
+func (v *validatorSC) unStakeNodes(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	if !v.flagEnableTopUp.IsSet() {
```
raduchis, line 944: flagEnableTopUp because it will be enabled with the Topup?

vm/systemSmartContracts/validator.go
```diff
@@ -1176,13 +1183,19 @@ func (v *validatorSC) unStakeTokens(args *vmcommon.ContractCallInput) vmcommon.R
...
+		v.eei.AddReturnMessage("can not unstake the provided value either because is under the minimum threshold or " +
+			"is not the value left to be unStaked")
+		return vmcommon.UserError
+	}
+	if unStakeValue.Cmp(registrationData.TotalStakeValue) > 0 {
+		v.eei.AddReturnMessage("can not unstake a bigger value than the possible allowed value which is " + registrationData.TotalStakeValue.String())
 		return vmcommon.UserError
 	}
 
+	registrationData.TotalStakeValue.Sub(registrationData.TotalStakeValue, unStakeValue)
```
raduchis, line 1198: also here no backwardscompatibility issue?

vm/systemSmartContracts/validator.go
sasurobert: yes. slashing has to be moved to the validatorSC. But as there is not yet used, there is no need for it.

vm/systemSmartContracts/staking.go
```diff
@@ -670,119 +670,82 @@ func (r *stakingSC) unBond(args *vmcommon.ContractCallInput) vmcommon.ReturnCode
...
-	err = r.saveStakingData(args.Arguments[0], registrationData)
-	if err != nil {
-		r.eei.AddReturnMessage("cannot save staking data: error " + err.Error())
-		return vmcommon.UserError
-	}
-
-	return vmcommon.Ok
+// backward compatibility
+func (s *stakingSC) slash(_ *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	s.eei.AddReturnMessage("slash function called by not the owners address")
```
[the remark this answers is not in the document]
sasurobert: it has to stay the same as on mainnet.

vm/systemSmartContracts/validator.go
sasurobert: yes. new functions on validatorSC will be enabled with topup

vm/systemSmartContracts/validator.go
sasurobert: this is ok. flag check is before.

vm/systemSmartContracts/validator.go
sasurobert: unStakeValue can be called only if flag topup is enabled.

vm/systemSmartContracts/staking.go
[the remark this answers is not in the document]
raduchis: ok

vm/systemSmartContracts/validator.go
raduchis: ok

vm/systemSmartContracts/validator.go
raduchis: ok

raduchis (approved):

vm/systemSmartContracts/validator.go
iulianpascalau: ok

iulianpascalau (approved):

PR #2533, opened 2020-12-01
Resolver fixes

iulianpascalau:
- fixed the following resolver-requester issues:
    - fixed reporting of missing transactions and miniblocks in resolvers to the debugging component
    - replaced function that gets the data from persister in miniblock resolver from `Get` to `SearchFirst` as to align with the rest of the resolvers.
- removed unused flags in main.go
- fixed a wrong iteration on an int slice

Testing:
Launch a normal testnet with #AllIn, things should go pretty smooth and no node should be left behind especially at the epoch change event.

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -353,20 +353,6 @@ VERSION:
 			"and will have a full history over epochs.",
 	}
 
-	numEpochsToSave = cli.Uint64Flag{
-		Name: "num-epochs-to-keep",
-		Usage: "This flag represents the number of epochs which will kept in the databases. It is relevant only if " +
-			"the full archive flag is not set.",
-		Value: uint64(2),
-	}
-
-	numActivePersisters = cli.Uint64Flag{
-		Name: "num-active-persisters",
-		Usage: "This flag represents the number of databases (1 database = 1 epoch) which are kept open at a moment. " +
-			"It is relevant even if the node is full archive or not.",
-		Value: uint64(2),
-	}
-
 	startInEpoch = cli.BoolFlag{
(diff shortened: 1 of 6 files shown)


sasurobert (approved):

SebastianMarian (dismissed):

cmd/node/main.go
```diff
@@ -353,20 +353,6 @@ VERSION:
 			"and will have a full history over epochs.",
 	}
 
-	numEpochsToSave = cli.Uint64Flag{
```
bogdan-rosianu, line 356: why did you remove the flags?

iulianpascalau: no longer required, they are defined in the config.toml and observed that we never used them as flags. Initially I thought we are mistakenly changed those when running the node. Didn't turned out that we specify them in the testnet scripts.

bogdan-rosianu: Ok, just wondering if one would have used these flags in a custom node script. It is easier to configure these settings via flags.

iulianpascalau: It is very easy to mess things out. I was even wondering if we can remove some configs values from config.toml (regarding this area) as to further decrease the chances of breaking things up.

bogdan-rosianu (dismissed):

SebastianMarian (approved):

SebastianMarian (approved):

sasurobert (approved):

bogdan-rosianu (approved):

LucianMincu (approved):
System tests passed.

PR #2534, opened 2020-12-01
Fix epoch start trigger when epoch start previous header is missing

SebastianMarian:
* Fixed situation when previous epoch start meta block is missing and trigger was not activated
* Fixed a sort situation when more headers have the same nonce in block tracker component

### DIFF
--- a/epochStart/shardchain/trigger.go
+++ b/epochStart/shardchain/trigger.go
@@ -456,11 +456,14 @@ func (t *trigger) receivedMetaBlock(headerHandler data.HeaderHandler, metaBlockH
 		return
 	}
 
-	_, ok = t.mapFinalizedEpochs[metaHdr.Epoch]
-	if t.metaEpoch == headerHandler.GetEpoch() && ok {
-		t.changeEpochFinalityAttestingRoundIfNeeded(metaHdr, metaBlockHash)
-		return
+	if !t.isPreviousEpochStartMetaBlock(metaHdr, metaBlockHash) {
+		_, ok = t.mapFinalizedEpochs[metaHdr.Epoch]
+		if t.metaEpoch == headerHandler.GetEpoch() && ok {
+			t.changeEpochFinalityAttestingRoundIfNeeded(metaHdr, metaBlockHash)
+			return
+		}
 	}
+
 	if !t.newEpochHdrReceived && !metaHdr.IsStartOfEpochBlock() {
 		return
 	}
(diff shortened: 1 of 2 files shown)


epochStart/shardchain/trigger.go
```diff
@@ -496,6 +499,24 @@ func (t *trigger) receivedMetaBlock(headerHandler data.HeaderHandler, metaBlockH
...
 
+// call only if mutex is locked before
+func (t *trigger) isPreviousEpochStartMetaBlock(metaBlock *block.MetaBlock, metaBlockHash []byte) bool {
+	metaHdrHashesWithNonce := t.mapNonceHashes[metaBlock.Nonce+1]
+	for _, hash := range metaHdrHashesWithNonce {
+		epochStartMetaBlock, ok := t.mapEpochStartHdrs[hash]
+		if !ok {
+			continue
+		}
+		if !bytes.Equal(metaBlockHash, epochStartMetaBlock.PrevHash) {
```
sasurobert, line 510: I think this can be done without range with the following logic. When previous block is missing and request is done, we should set a flag to true with the requested hash, if the new received one is with the same hash, the processing must be done

sasurobert (changes requested):

epochStart/shardchain/trigger.go
iulianpascalau: Since the iteration is done only on the headers with the same nonce, IMHO I do not think this impacts too much the performance.

SebastianMarian: The request is already done in method isMetaBlockValid -> getHeaderWithNonceAndHash. The main problem is that the previous header will never pass from that line from receivedMetaBlock:
```
		_, ok = t.mapFinalizedEpochs[metaHdr.Epoch]
		if t.metaEpoch == headerHandler.GetEpoch() && ok {
			t.changeEpochFinalityAttestingRoundIfNeeded(metaHdr, metaBlockHash)
			return
		}

```

iulianpascalau (approved):

epochStart/shardchain/trigger.go
SebastianMarian: I think we do not need here the flag mechanism used for attesting headers. How many forks could be on the epoch start meta block? The range will be done 99.99% only for one item.

SebastianMarian: More than this this chance is allowed only if the start of epoch is already received, otherwise it will be dropped in the next line after the code above which is:
```
	if !t.newEpochHdrReceived && !metaHdr.IsStartOfEpochBlock() {
		return
	}

```

sasurobert (approved):

LucianMincu (approved):
System test passed.

PR #2535, opened 2020-12-02
Try to write all the items from  the queue when method Close is called

miiu96:
Try to index all the items that are in the `dataDispatcher` queue when method `Close` is called.

### DIFF
--- a/core/indexer/dataDispatcher.go
+++ b/core/indexer/dataDispatcher.go
@@ -24,11 +24,13 @@ const (
 )
 
 type dataDispatcher struct {
-	backOffTime   time.Duration
-	chanWorkItems chan workItems.WorkItemHandler
-	cancelFunc    func()
-	wasClosed     atomic.Flag
-	writeMutex    sync.Mutex
+	backOffTime         time.Duration
+	chanWorkItems       chan workItems.WorkItemHandler
+	cancelFunc          func()
+	wasClosed           *atomic.Flag
+	currentWriteDone    chan struct{}
+	closeStartTime      time.Time
+	mutexCloseStartTime sync.RWMutex
 }
 
 // NewDataDispatcher creates a new dataDispatcher instance, capable of saving sequentially data in elasticsearch database
@@ -38,9 +40,10 @@ func NewDataDispatcher(cacheSize int) (*dataDispatcher, error) {
 	}
 
 	dd := &dataDispatcher{
-		chanWorkItems: make(chan workItems.WorkItemHandler, cacheSize),
-		wasClosed:     atomic.Flag{},
-		writeMutex:    sync.Mutex{},
+		chanWorkItems:       make(chan workItems.WorkItemHandler, cacheSize),
+		wasClosed:           &atomic.Flag{},
+		currentWriteDone:    make(chan struct{}),
+		mutexCloseStartTime: sync.RWMutex{},
 	}
 
 	return dd, nil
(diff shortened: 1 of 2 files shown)


core/indexer/dataDispatcher.go
```diff
@@ -74,12 +93,24 @@ func (d *dataDispatcher) Add(item workItems.WorkItemHandler) {
 		log.Warn("dataDispatcher.Add nil item: will do nothing")
 		return
 	}
+	if d.wasClosed.IsSet() {
+		return
```
iulianpascalau, line 97: add log print?

core/indexer/dataDispatcher.go
```diff
@@ -54,17 +58,32 @@ func (d *dataDispatcher) startWorker(ctx context.Context) {
...
+	close(d.chanWorkItems)
+	for wi := range d.chanWorkItems {
+		if time.Since(start) > closeTimeout {
+			log.Warn("cannot write all items from the queue",
+				"error", "timeout",
+			)
+			return nil
+		}
+
+		remainingTime := closeTimeout - time.Since(start)
```
iulianpascalau, line 83: what happens if remainingTime will drop to 0 because the remainder jobs did not end? It will make the doWork function wait indefinitely. Do we have protection for negative values?

iulianpascalau (changes requested):

core/indexer/dataDispatcher.go
miiu96: added

core/indexer/dataDispatcher.go
miiu96: Added extra check for the negative values. Also the case when the remaining time can be 0 was treated.

bogdan-rosianu (approved):
fix tests

core/indexer/dataDispatcher.go
```diff
@@ -27,6 +28,7 @@ type dataDispatcher struct {
 	chanWorkItems chan workItems.WorkItemHandler
 	cancelFunc    func()
 	wasClosed     atomic.Flag
+	writeMutex    sync.Mutex
```
iulianpascalau, line 31: no need for this mutex

core/indexer/dataDispatcher.go
```diff
@@ -58,29 +61,40 @@ func (d *dataDispatcher) startWorker(ctx context.Context) {
...
+			d.writeMutex.Lock()
 			d.doWork(wi, 0)
+			d.writeMutex.Unlock()
 		}
 	}
 }
 
 // Close will close the endless running go routine
 func (d *dataDispatcher) Close() error {
+	if d.wasClosed.IsSet() {
```
iulianpascalau, line 73: concurrency problems here: check for set and the actual set should be done atomically. Worst case: use an uint32 value and functions found in `atomic` package

iulianpascalau (changes requested):

core/indexer/dataDispatcher.go
```diff
@@ -74,12 +118,37 @@ func (d *dataDispatcher) Add(item workItems.WorkItemHandler) {
...
 	d.chanWorkItems <- item
+
+	if d.wasClosed.IsSet() {
+		close(d.chanWorkItems)
+	}
 }
 
-func (d *dataDispatcher) doWork(wi workItems.WorkItemHandler) {
+func (d *dataDispatcher) doWork(wi workItems.WorkItemHandler, timeout time.Duration) {
+	if wi == nil {
```
iulianpascalau, line 134: since we won't close the channel, this check is useless now

core/indexer/dataDispatcher.go
```diff
@@ -54,18 +61,55 @@ func (d *dataDispatcher) startWorker(ctx context.Context) {
...
+
+	start := time.Now()
+	d.wasClosed.Set()
+
 	if d.cancelFunc != nil {
 		d.cancelFunc()
 	}
 
-	return nil
+	d.writeMutex.Lock()
```
iulianpascalau, line 84: we can have a deadlock here: suppose the mutex was captured by L64, this will never Close

iulianpascalau (changes requested):

core/indexer/dataDispatcher.go
```diff
@@ -51,35 +61,82 @@ func (d *dataDispatcher) startWorker(ctx context.Context) {
...
+	d.currentWriteDone <- struct{}{}
+}
+
 // Close will close the endless running go routine
 func (d *dataDispatcher) Close() error {
+	if d.wasClosed.Set() {
+		return nil
+	}
+
+	d.mutexCloseStartTime.Lock()
```
iulianpascalau, line 87: :+1:

core/indexer/dataDispatcher.go
iulianpascalau: do we still need this nil check?

core/indexer/dataDispatcher.go
```diff
@@ -16,14 +18,19 @@ var log = logger.GetOrCreate("core/indexer")
 const durationBetweenErrorRetry = time.Second * 3
 
 const (
-	backOffTime = time.Second * 10
-	maxBackOff  = time.Minute * 5
+	closeTimeout = time.Second * 5
```
iulianpascalau, line 21: maybe increase this to ~20seconds?

core/indexer/dataDispatcher.go
miiu96: removed.

core/indexer/dataDispatcher.go
miiu96: done.

core/indexer/dataDispatcher.go
```diff
@@ -51,35 +61,78 @@ func (d *dataDispatcher) startWorker(ctx context.Context) {
...
+			timeout := d.doWork(wi)
+			if timeout {
+				d.stopWorked()
+				return
+			}
 		}
 	}
 }
 
+func (d *dataDispatcher) stopWorked() {
```
bogdan-rosianu, line 76: rename to `stopWork` or something similar.

bogdan-rosianu (changes requested):

core/indexer/dataDispatcher.go
miiu96: done.

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #2549, opened 2020-12-04
Merge soft restart full history 2020.12.04

iulianpascalau:


node/external/nodeApiResolver.go
```diff
@@ -73,7 +73,7 @@ func (nar *NodeApiResolver) Close() error {
 		err2 = nar.vmFactory.Close()
 	}
 	if err1 != nil || err2 != nil {
-		return fmt.Errorf("err closing vmContainer: %w, err closing vmFactory: %w", err1, err2)
+		return fmt.Errorf("err closing vmContainer: %v, err closing vmFactory: %v", err1, err2)
```
bogdan-rosianu, line 76: is this ok? at least use `%s` and `errx.String()`

sharding/networksharding/peerShardMapper.go
```diff
@@ -313,7 +313,7 @@ func (psm *PeerShardMapper) UpdatePeerIdShardId(pid core.PeerID, shardId uint32)
 
 // UpdatePeerIdSubType updates the peerIdSubType search map containing peer IDs and peer subtypes
 func (psm *PeerShardMapper) UpdatePeerIdSubType(pid core.PeerID, peerSubType core.P2PPeerSubType) {
-	psm.peerIdSubType.HasOrAdd([]byte(pid), uint32(peerSubType), uint32Size)
+	psm.peerIdSubTypeCache.HasOrAdd([]byte(pid), peerSubType, uint32Size)
```
bogdan-rosianu, line 316: why `HasOrAdd` if returned values are ignored? Or is it because of recent-ness?

bogdan-rosianu (changes requested):

process/block/baseProcess.go
```diff
@@ -1265,7 +1265,7 @@ func (bp *baseProcessor) Close() error {
 		err2 = bp.vmContainerFactory.Close()
 	}
 	if err1 != nil || err2 != nil {
-		return fmt.Errorf("vmContainer close error: %w, vmContainerFactory close error: %w", err1, err2)
+		return fmt.Errorf("vmContainer close error: %v, vmContainerFactory close error: %v", err1, err2)
```
bogdan-rosianu, line 1268: same here.

node/external/nodeApiResolver.go
iulianpascalau: no, will panic if err is nil. Works as expected as it is

process/block/baseProcess.go
iulianpascalau: as stated, works as it is. The reason behind this is that Goland 2020.3 reports this as an error

sharding/networksharding/peerShardMapper.go
iulianpascalau: HasOrAdd adds the value if that value exists. It does the checking and adding atomically because this method can be called in a concurrent manner.

node/external/nodeApiResolver.go
bogdan-rosianu: ok. tested locally and the error printing is the same.

bogdan-rosianu (approved):

sasurobert (approved):

PR #2561, opened 2020-12-07
Fix gas used value for SC call transactions

miiu96:
Fixed the problem with the `gasUsed` field of transactions that are indexed in elasticsearch.


core/indexer/processTransactions.go
```diff
@@ -196,19 +196,41 @@ func (tdp *txDatabaseProcessor) addScResultInfoInTx(scHash string, scr *smartCon
...
 	return tx
 }
 
+func computeTxGasUsedField(dbScResult ScResult, tx *Transaction) uint64 {
+	fee := big.NewInt(0).SetUint64(tx.GasPrice)
+	fee.Mul(fee, big.NewInt(0).SetUint64(tx.GasLimit))
+
+	refundValue, ok := big.NewInt(0).SetString(dbScResult.Value, 10)
+	if !ok {
+		return fee.Uint64()
```
bogdan-rosianu, line 211: is this ok?

sasurobert (approved):

core/indexer/processTransactions_test.go
```diff
@@ -481,3 +481,40 @@ func txPoolHasSearchOrder(txPool map[string]*Transaction, searchOrder uint32) bo
...
+	nonce := uint64(10)
+	sender := "sender"
+
+	tx := &Transaction{
+		Hash:   txHash,
+		Nonce:  nonce,
+		Sender: sender,
+	}
+	sc := ScResult{
+		Data:      []byte("@" + hex.EncodeToString([]byte("ok")) + "@something"),
```
bogdan-rosianu, line 498: you can use directly `6f6b` instead of hex encoded `ok`. This way you simulate better the real scenario.

core/indexer/processTransactions.go
bogdan-rosianu: is there a case that the value is invalid and the fee must be this? or is it just not to return error?

bogdan-rosianu (changes requested):

core/indexer/processTransactions.go
miiu96: No. This case will never happen because `dbScResult.Value` is set from a `big.Int` so this `SetString` will never fail.

core/indexer/processTransactions_test.go
miiu96: small fix

core/indexer/processTransactions.go
bogdan-rosianu: ok. At least set a log warn. Who knows?

miiu96: done.

bogdan-rosianu (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #2562, opened 2020-12-07
bugFix/accumulated fees for built in functions

sasurobert:
built in functions consumes gas at sender only. Otherwise all gas should be used - gas refund was already resolved at the sender shard. The gas at the destination should not be added to devfees / accumulated fees


process/smartContract/process.go
```diff
@@ -571,6 +586,7 @@ func (sc *scProcessor) ExecuteBuiltInFunction(
 		log.Debug("processed built in functions error", "error", err.Error())
 		return 0, err
 	}
+	builtInFuncGasUsed := vmInput.GasProvided - vmOutput.GasRemaining
```
andreibancioiu, line 589: Above, we use `safeSubUint64`.

```
consumedGas := safeSubUint64(tx.GetGasLimit(), vmOutput.GasRemaining)
```

though here we do not. Not needed (because the VM will not output incorrect values, such as a gas remaining higher than gas provided)?

process/smartContract/process.go
```diff
@@ -464,14 +464,27 @@ func (sc *scProcessor) isSelfShard(address []byte) bool {
 	return sc.shardCoordinator.ComputeId(address) == sc.shardCoordinator.SelfId()
 }
 
+func safeSubUint64(a, b uint64) uint64 {
+	if a < b {
+		return 0
```
andreibancioiu, line 469: Now, due to the refactoring below, is this still possible or is it an error / critical condition?

process/smartContract/builtInFunctions/changeOwnerAddress.go
```diff
@@ -67,7 +68,18 @@ func (c *changeOwnerAddress) ProcessBuiltinFunction(
...
 
-	return &vmcommon.VMOutput{GasRemaining: vmInput.GasProvided - c.gasCost, ReturnCode: vmcommon.Ok}, nil
+	return &vmcommon.VMOutput{GasRemaining: gasRemaining, ReturnCode: vmcommon.Ok}, nil
+}
+
+func computeGasRemaining(snd state.UserAccountHandler, gasProvided uint64, gasToUse uint64) uint64 {
+	if gasProvided < gasToUse {
+		return 0
+	}
+	if check.IfNil(snd) {
```
andreibancioiu, line 78: Perhaps add a comment? `cross-shard call, therefore ...`

process/smartContract/process.go
sasurobert: this should be not more possible - however it is better to be safe than sorry, thus have this kind of protection.

process/smartContract/builtInFunctions/changeOwnerAddress.go
sasurobert: done

process/smartContract/process.go
sasurobert: this should be impossible

process/smartContract/process.go
```diff
@@ -480,13 +493,15 @@ func (sc *scProcessor) computeTotalConsumedFeeAndDevRwd(
...
-			consumedGas -= sentGas
+			consumedGas = safeSubUint64(consumedGas, sentGas)
 		}
 	}
 
 	totalFee := core.SafeMul(tx.GetGasPrice(), consumedGas)
 
-	consumedGas -= builtInFuncGasUsed
+	if !senderInAnotherShard {
+		consumedGas = safeSubUint64(consumedGas, builtInFuncGasUsed)
```
AdoAdoAdo, line 503: 1. If I understand correctly, here you remove the builtin function gas cost, to compute the devFees since the devFees apply only to gas consumed by the SC call not also the builtin functions.
Can you use a different variable name here e.g consumedGasWithoutBuiltin otherwise it is confusing.

2. Why is the builtin function gas cost removed for senderInShard only? 
I assume if builtInFuncGasUsed > 0 then it means something has been spent here in this shard, so why don't we subtract it if sender is in different shard?
Or if we have a cross shard SC TX, is it never possible to spend gas on builtin functions on the destination shard?

process/smartContract/process.go
AdoAdoAdo: is it ok here not to use the "safe" function? 
I would use the same logic here as well even if it seems impossible, to keep it consistent.

AdoAdoAdo (changes requested):

process/smartContract/process.go
sasurobert: Done

process/smartContract/process.go
```diff
@@ -464,14 +464,27 @@ func (sc *scProcessor) isSelfShard(address []byte) bool {
...
-		consumedGas -= sc.economicsFee.ComputeGasLimit(tx)
+	if tx.GetGasLimit() == 0 {
+		return big.NewInt(0), big.NewInt(0)
+	}
+
+	senderInAnotherShard := !sc.isSelfShard(tx.GetSndAddr())
+	consumedGas := safeSubUint64(tx.GetGasLimit(), vmOutput.GasRemaining)
+	if senderInAnotherShard {
+		consumedGas = safeSubUint64(consumedGas, sc.economicsFee.ComputeGasLimit(tx))
+		consumedGas = safeSubUint64(consumedGas, builtInFuncGasUsed)
```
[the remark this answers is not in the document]
sasurobert: Built in function gas is consumed fully at sender shard

process/smartContract/process.go
sasurobert: Yes for 1.

sasurobert: It is never possible to use gas on destination SC. Only in sender sc for built in functions

process/smartContract/process.go
SebastianMarian, line 487: It is correct to do this sub here? Now the totalFee from line 500 will be the same with totalFeeMinusBuiltIn from line 505 if senderInAnotherShard is true. Until now the functionality was different. totalFee was always the fee without the builtInFuncGasUsed substracted

process/smartContract/builtInFunctions/claimDeveloperRewards.go
```diff
@@ -62,7 +63,7 @@ func (c *claimDeveloperRewards) ProcessBuiltinFunction(
 		return nil, err
 	}
 
-	vmOutput := &vmcommon.VMOutput{GasRemaining: vmInput.GasProvided - c.gasCost}
+	vmOutput := &vmcommon.VMOutput{GasRemaining: gasRemaining}
```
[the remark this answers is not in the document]
sasurobert: by default it is OK.

process/smartContract/builtInFunctions/esdtBurn.go
```diff
@@ -94,7 +94,8 @@ func (e *esdtBurn) ProcessBuiltinFunction(
 		return nil, err
 	}
 
-	vmOutput := &vmcommon.VMOutput{GasRemaining: vmInput.GasProvided - e.funcGasCost}
+	gasRemaining := computeGasRemaining(acntSnd, vmInput.GasProvided, e.funcGasCost)
+	vmOutput := &vmcommon.VMOutput{GasRemaining: gasRemaining}
```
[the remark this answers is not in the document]
sasurobert: done

process/smartContract/process.go
sasurobert: it is correct. as all gas in case of builtin is consumed in sender shard only.

process/smartContract/process.go
```diff
@@ -499,10 +499,11 @@ func (sc *scProcessor) computeTotalConsumedFeeAndDevRwd(
...
 	totalFee := core.SafeMul(tx.GetGasPrice(), consumedGas)
 
-	if !senderInAnotherShard {
-		consumedGas = safeSubUint64(consumedGas, builtInFuncGasUsed)
+	consumedGasWithoutBuiltin := consumedGas
+	if senderInSelfShard {
+		consumedGasWithoutBuiltin = safeSubUint64(consumedGasWithoutBuiltin, builtInFuncGasUsed)
 	}
-	totalFeeMinusBuiltIn := core.SafeMul(tx.GetGasPrice(), consumedGas)
+	totalFeeMinusBuiltIn := core.SafeMul(tx.GetGasPrice(), consumedGasWithoutBuiltin)
```
SebastianMarian, line 506: totalFeeWithoutBuiltIn

SebastianMarian (approved):

AdoAdoAdo (approved):

andreibancioiu (approved):

LucianMincu (approved):
System tests passed.

PR #2575, opened 2020-12-10
Relayed tx audit

raduchis:
code review/audit on the shardProcessor

* optimized the accountsLoading
* added more unit tests
* all other fixes are already in the feature branch


core/address.go
```diff
@@ -40,6 +40,7 @@ func IsSmartContractAddress(rcvAddress []byte) bool {
 		return true
 	}
 
+	//TODO: check how a normal address with 8 leading zeroes influences everything
```
sasurobert, line 43: it doesn't. you can't send a transaction to it.

process/economics/economicsData.go
```diff
@@ -223,6 +223,7 @@ func (ed *EconomicsData) ComputeMoveBalanceFee(tx process.TransactionWithFeeHand
 
 // ComputeTxFee computes the provided transaction's fee using enable from epoch approach
 func (ed *EconomicsData) ComputeTxFee(tx process.TransactionWithFeeHandler) *big.Int {
+	//TODO: if flag is set, always take all the fee?
```
sasurobert, line 226: yes. as otherwise it is an error.

sasurobert: in other components this function may be wrongly used.

process/smartContract/process.go
```diff
@@ -1172,6 +1172,7 @@ func (sc *scProcessor) processSCPayment(tx data.TransactionHandler, acntSnd stat
 	cost = cost.Mul(big.NewInt(0).SetUint64(tx.GetGasPrice()), big.NewInt(0).SetUint64(tx.GetGasLimit()))
 	cost = cost.Add(cost, tx.GetValue())
 
+	// TODO: check why and how the cost could be zero
```
sasurobert, line 1175: only in unit tests.

process/transaction/interceptedTransaction.go
```diff
@@ -262,6 +262,7 @@ func (inTx *InterceptedTransaction) integrity(tx *transaction.Transaction) error
 	if tx.Value.Sign() < 0 {
 		return process.ErrNegativeValue
 	}
+	//TODO: check these checks - why len(RecvUserName) != hasher.Size() ?
```
sasurobert, line 265: good find - this was changed and forgot it here. user names will be in clear. in the first implementation it was hashed.

process/transaction/shardProcess.go
```diff
@@ -169,35 +169,38 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction) (vmco
...
 			if receiptErr != nil {
+				// TODO: check if OK and receiptError is correct
 				return 0, receiptErr
 			}
 		}
 
 		if errors.Is(err, process.ErrUserNameDoesNotMatchInCrossShardTx) {
 			errProcessIfErr := txProc.processIfTxErrorCrossShard(tx, err.Error())
 			if errProcessIfErr != nil {
+				// TODO: check if OK and errProcessIfErr is correct
```
sasurobert, line 180: it doesn't matter - these are critical errors.

process/transaction/shardProcess.go
```diff
@@ -169,35 +169,38 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction) (vmco
...
 				return 0, errProcessIfErr
 			}
+			// TODO: check if UserError and nil is correct
```
sasurobert, line 183: yes. correct.

process/transaction/shardProcess.go
```diff
@@ -169,35 +169,38 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction) (vmco
 		if errors.Is(err, process.ErrInsufficientFunds) {
 			receiptErr := txProc.executingFailedTransaction(tx, acntSnd, err)
 			if receiptErr != nil {
+				// TODO: check if OK and receiptError is correct
```
sasurobert, line 172: it doesn't matter - these are critical errors.

process/transaction/shardProcess.go
```diff
@@ -248,6 +251,7 @@ func (txProc *txProcessor) executingFailedTransaction(
 		return nil
 	}
 
+	//TODO: for a failed transaction take all the gas fee always?
```
sasurobert, line 254: yes

process/transaction/shardProcess.go
```diff
@@ -293,6 +297,7 @@ func (txProc *txProcessor) createReceiptWithReturnedGas(
 	acntSnd state.UserAccountHandler,
 	cost *big.Int,
 ) error {
+	// TODO: check why no receiptWithReturnedGas if not in sender shard
```
sasurobert, line 300: because it is a move balance and all cost is taken out exactly at the sender shard.

process/transaction/shardProcess.go
```diff
@@ -301,6 +306,7 @@ func (txProc *txProcessor) createReceiptWithReturnedGas(
 	}
 
 	totalProvided := big.NewInt(0)
+	//TODO: check if always gasPrice*gasLimit is correct. Could it be that not all was deducted like for a moveBalance?
```
sasurobert, line 309: no. it is not possible.

process/transaction/shardProcess.go
```diff
@@ -334,6 +340,7 @@ func (txProc *txProcessor) processTxFee(
 		return big.NewInt(0), nil
 	}
 
+	//TODO: just for crossShardSCCall have a different cost? What about builtInFunctions and Deploys?
```
sasurobert, line 343: deploy is always intra shard. built in function cost is always taken in self shard.

process/transaction/shardProcess.go
```diff
@@ -439,6 +439,8 @@ func (txProc *txProcessor) processMoveBalance(
 		return err
 	}
 
+	// TODO: check if correct that gas is refunded to sender instead of relayer for relayed Txs
+	// TODO: check 2 - create receiptWithReturnedGas even if we are in the senderShard?
```
sasurobert, line 443: this was taken out in gasPrice-modifier branch - there will be no receipt in case of relayed.

process/transaction/shardProcess.go
```diff
@@ -451,62 +453,39 @@ func (txProc *txProcessor) processMoveBalance(
...
 ) (vmcommon.ReturnCode, error) {
 
 	_, args, err := txProc.argsParser.ParseCallData(string(tx.GetData()))
 	if err != nil {
-		return 0, err
-	}
-
-	relayerAcnt, acntDst, err := txProc.getAccounts(adrSrc, adrDst)
-	if err != nil {
+		//TODO: check if return code 0 is ok here
```
sasurobert, line 482: yes it is. otherwise this is a critical error - these transactions were already verified by interceptor.

process/transaction/shardProcess.go
```diff
@@ -451,62 +453,39 @@ func (txProc *txProcessor) processMoveBalance(
...
 		return 0, err
 	}
 
 	if len(args) != 1 {
+		//TODO: check if ok if the relayerAcnt is nil, not in the current shard, the transaction will fail with UserError and nil error
```
sasurobert, line 487: the relayed transaction MUST always start at the sender shjard-  thus relayer. all is consumed there. in case it fails on destination execution failed is called and value is sent back to relayer.

sasurobert: executingFailedTransaction can only happen in relayer shard - as it is executed here FIRST.

process/transaction/shardProcess.go
```diff
@@ -522,6 +501,7 @@ func (txProc *txProcessor) processRelayedTx(
 	if !bytes.Equal(userTx.SndAddr, tx.RcvAddr) {
 		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxBeneficiaryDoesNotMatchReceiver)
 	}
+	//TODO: check why this check? Why can't the relayer send a bigger value?
```
sasurobert, line 504: why would you want to send more value to user ? if you want to give value to user you can simply transfer to him.

process/transaction/shardProcess.go
```diff
@@ -530,6 +510,7 @@ func (txProc *txProcessor) processRelayedTx(
 	}
 
 	totalFee, remainingFee, relayerFee, remainingGasLimit := txProc.computeRelayedTxFees(tx)
+	//TODO: is this check really necessary? Why?
```
sasurobert, line 513: yes it is, in order to correctly send and use gas all over the protocol. to have correct refunds. If this invariant does not stay it is hard to manage the rest. There is no reason for a relayer to pay more gas than this.

process/transaction/shardProcess.go
```diff
@@ -556,6 +537,7 @@ func (txProc *txProcessor) processRelayedTx(
 			return 0, err
 		}
 
+		//TODO: check if the relayerFee has the movebalance inside, or when is the -movebalance in the relayer shard
```
sasurobert, line 540: it has.

process/transaction/shardProcess.go
```diff
@@ -587,6 +569,8 @@ func (txProc *txProcessor) computeRelayedTxFees(tx *transaction.Transaction) (*b
 	totalFee := core.SafeMul(tx.GetGasLimit(), tx.GetGasPrice())
 	remainingFee := big.NewInt(0).Sub(totalFee, relayerFee)
 
+	//TODO: check tx.GasLimit - relayerGasLimit to be positive
```
sasurobert, line 572: otherwise will fail the previous check of gasLimit.

process/transaction/shardProcess.go
```diff
@@ -605,6 +589,7 @@ func (txProc *txProcessor) removeValueAndConsumedFeeFromUser(
 	if err != nil {
 		return err
 	}
+	//TODO: check if always the consumedFee was gasLimit*GasPrice - movebalance for example
```
sasurobert, line 592: yes it is. this is somewhat changed in gas price modifier.

process/transaction/shardProcess.go
```diff
@@ -649,18 +635,19 @@ func (txProc *txProcessor) processUserTx(
 			txHash,
 			err.Error())
 	}
-
+	// TODO: check if there is no problem with hashing a new SCR and not the previous transaction. Are the hashes equal?
```
sasurobert, line 638: the SCR is where the execution continues.

process/transaction/shardProcess.go
```diff
@@ -649,18 +635,19 @@ func (txProc *txProcessor) processUserTx(
...
 	switch txType {
 	case process.MoveBalance:
-		err = txProc.processMoveBalance(userTx, userTx.SndAddr, userTx.RcvAddr, txProc.getUserTxCost(userTx, txHash, txType))
+		err = txProc.processMoveBalance(userTx, acntSnd, acntDst, txProc.getUserTxCost(userTx, txHash, txType))
 	case process.SCDeployment:
 		returnCode, err = txProc.scProcessor.DeploySmartContract(scrFromTx, acntSnd)
 	case process.SCInvoking:
 		returnCode, err = txProc.scProcessor.ExecuteSmartContractTransaction(scrFromTx, acntSnd, acntDst)
 	case process.BuiltInFunctionCall:
+		//TODO: where does the userTxValue get back to the sender/relayer?
```
sasurobert, line 650: ProcessIfError

process/transaction/shardProcess.go
```diff
@@ -678,6 +665,8 @@ func (txProc *txProcessor) processUserTx(
 			err.Error())
 	}
 
+	// TODO: check why executeFailedRelay just for these 2 cases? What happens if moveBalance fails in the senderShard
```
sasurobert, line 668: these 2 cases are when execution fails in sender shard.
In destination shard not Payable SC is resolved in another way through smart contract processing in case of relayed TX. smartContractProcessor.ProcessSmartContractResult -> moveBalance case.

process/transaction/shardProcess.go
```diff
@@ -730,6 +719,7 @@ func (txProc *txProcessor) getUserTxCost(
 				"gas used", gasUsed,
 			)
 
+			// TODO: always use all the gas?
```
sasurobert, line 722: this is changed in gasPrice-modifier - all gas should be consumed in case of any error.

process/transaction/shardProcess.go
```diff
@@ -800,12 +790,15 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
 		return err
 	}
 
+	//TODO: check if the totalFee can be smaller, in case of a moveBalance for example
```
sasurobert, line 793: not possible.

process/transaction/shardProcess.go
```diff
@@ -800,12 +790,15 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
...
 	totalFee := core.SafeMul(userTx.GetGasPrice(), userTx.GetGasLimit())
 	senderShardID := txProc.shardCoordinator.ComputeId(userTx.SndAddr)
 	if senderShardID != txProc.shardCoordinator.SelfId() {
+		// TODO: check what this does, if the sender is not here, process everything without the moveBalance here?
 		totalFee.Sub(totalFee, txProc.economicsFee.ComputeMoveBalanceFee(userTx))
 	}
 
+	//TODO: check if this was not processed again in the processRelayedTx
```
sasurobert, line 801: it was not.

process/transaction/shardProcess.go
```diff
@@ -800,12 +790,15 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
 		return err
 	}
 
+	//TODO: check if the totalFee can be smaller, in case of a moveBalance for example
 	totalFee := core.SafeMul(userTx.GetGasPrice(), userTx.GetGasLimit())
 	senderShardID := txProc.shardCoordinator.ComputeId(userTx.SndAddr)
 	if senderShardID != txProc.shardCoordinator.SelfId() {
+		// TODO: check what this does, if the sender is not here, process everything without the moveBalance here?
```
sasurobert, line 797: yes. as move balance was done in the sender shard.

sasurobert (changes requested):

core/address.go
raduchis: ok

process/smartContract/process.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: in most of the cases just the error is verified, not the retunCode. Is it ok that the error is nil?

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/economics/economicsData.go
raduchis: this changed, now also SCs are taken in to consideration... 

is line 258 correct? When can this happen? The transaction has a lower limit than the moveBalance?
if tx.GetGasLimit() <= gasLimitForMoveBalance {
        return moveBalanceFee
}

process/transaction/interceptedTransaction.go
raduchis: this changed: can we have only up to 32-byte usernames?

process/transaction/shardProcess.go
raduchis: ok... I think this changed

process/transaction/shardProcess.go
raduchis: this also changed

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

raduchis: I think right now it will happend, because the GasLimit is multiplied by the gasModifier here. So it is divided by 100.

process/transaction/shardProcess.go
raduchis: here the totalFee could be smaller than the computeMoveBalanceFee I think

process/transaction/shardProcess.go
raduchis: Ok and nonNilError should work, but this could be different, because most of the times the error is checked.

process/transaction/shardProcess.go
raduchis: ok

raduchis: does it impact something if the executingFailedTransaction return nil?

process/economics/economicsData.go
sasurobert: this was a bug initially - as computetxfee used only movebalance fee.
this introduced errors in how the function was used. now the code is correct. It is not possible for tx to have a gaslimit lower than moveBalance gas cost. It is checked by interceptor.

process/transaction/interceptedTransaction.go
sasurobert: yes. changed. only up to 32 bytes

process/transaction/shardProcess.go
sasurobert: userError is returned back mostly for tests. if error is returned and the error is not ErrFailedTransaction -> than the transaction is of TYPE invalid and gets into the INVALID BLOCK. If other error - state is reverted to before the tx and the transaction is deleted from the pool - not added to the block.

process/transaction/shardProcess.go
sasurobert: in that case it is protocol error - the transaction is removed from pool, state reverted to the one before the tx. 
The only way to return nil is if acntSND == nil - which cannot happen.

process/transaction/shardProcess.go
sasurobert: totalFee cannot be lower than moveBalanceFee. it is always moveBalance + extra fee.

process/transaction/shardProcess.go
sasurobert: totalFee cannot be lower than moveBalanceFee. it is always moveBalance + extra fee.

process/economics/economicsData.go
raduchis: ok

process/transaction/interceptedTransaction.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: exactly, but with the case here, the returned error is nil, so the transaction is valid?

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok not here, but in the next if

process/transaction/shardProcess.go
```diff
@@ -827,9 +806,11 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
 		return err
 	}
 
+	//TODO: check if the totalFee can be smaller, in case of a moveBalance for example
 	totalFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, userTx.GasLimit)
 	senderShardID := txProc.shardCoordinator.ComputeId(userTx.SndAddr)
 	if senderShardID != txProc.shardCoordinator.SelfId() {
+		// TODO: check if totalFee can be smaller than the computeMoveBalanceFee
```
raduchis, line 813: I think here there could be a negative balance.

process/transaction/shardProcess.go
raduchis: ok!

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: ok

process/transaction/shardProcess.go
raduchis: fixed by robert

sasurobert (dismissed):

sasurobert (approved):

mihaib79 (approved):

SebastianMarian (approved):

PR #2589, opened 2020-12-14
Fix/merging of output accs

sasurobert:
fix the merge of output accounts in case of system VM eei.
fixed the order and creation of scr in case of staking SC

### DIFF
--- a/core/vmcommon/output.go
+++ b/core/vmcommon/output.go
@@ -180,3 +180,53 @@ func (vmOutput *VMOutput) GetFirstReturnData(asType ReturnDataKind) (interface{}
 
 	return nil, fmt.Errorf("can't interpret return data")
 }
+
+// MergeOutputAccounts merges the given account into the current one
+func (o *OutputAccount) MergeOutputAccounts(outAcc *OutputAccount) {
+	if len(outAcc.Address) != 0 {
+		o.Address = outAcc.Address
+	}
+
+	o.MergeStorageUpdates(outAcc)
+
+	if outAcc.Balance != nil {
+		o.Balance = outAcc.Balance
+	}
+	if o.BalanceDelta == nil {
+		o.BalanceDelta = big.NewInt(0)
+	}
+	if outAcc.BalanceDelta != nil {
+		o.BalanceDelta = outAcc.BalanceDelta
+	}
+	if len(outAcc.Code) > 0 {
+		o.Code = outAcc.Code
+	}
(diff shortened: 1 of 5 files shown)


process/smartContract/process.go
```diff
@@ -1457,6 +1457,14 @@ func (sc *scProcessor) createSmartContractResults(
 	txHash []byte,
 ) (bool, []data.TransactionHandler) {
 
+	if bytes.Equal(outAcc.Address, vm.StakingSCAddress) {
+		storageUpdates := process.GetSortedStorageUpdates(outAcc)
```
iulianpascalau, line 1461: why do we need to sort the output accounts only in the case of staking address? Should we sort them always?

vm/systemSmartContracts/eei.go
```diff
@@ -268,10 +268,71 @@ func (host *vmContext) copyFromContext(currContext *vmContext) {
...
+		if !exist {
+			leftAccount = &vmcommon.OutputAccount{}
+			host.outputAccounts[string(outAcc.Address)] = leftAccount
+		}
+		mergeOutputAccounts(leftAccount, outAcc)
+	}
 	host.scAddress = currContext.scAddress
 }
 
+func mergeOutputAccounts(
```
iulianpascalau, line 282: why not move this functionality in vmcommon.OutputAccount.Merge function? It can be used elsewhere and most important, will be kept up-to-date more easily. Otherwise we will add a field in OutputAccount struct and forget to search in the whole project if we have a kind of clone functionality

process/smartContract/process.go
sasurobert: storage updates / not output accounts - we need this in order for all validators to create the same smart contract result. The created SCR is used in stakingToPeer

vm/systemSmartContracts/eei.go
sasurobert: done

core/vmcommon/output.go
```diff
@@ -180,3 +180,53 @@ func (vmOutput *VMOutput) GetFirstReturnData(asType ReturnDataKind) (interface{}
 
 	return nil, fmt.Errorf("can't interpret return data")
 }
+
+// MergeOutputAccounts merges the given account into the current one
```
iulianpascalau, line 184: :+1:

iulianpascalau (approved):

SebastianMarian (approved):

PR #2592, opened 2020-12-16
fix/compute-move-balance-fee

sasurobert:
small little fixes of gas used in scr and user tx
Make moveBalanceCost take into consideration SCR vs TX. SCR does not consume moveBalance

### DIFF
--- a/process/block/postprocess/feeHandler.go
+++ b/process/block/postprocess/feeHandler.go
@@ -14,6 +14,8 @@ type feeData struct {
 	devFee *big.Int
 }
 
+var zero = big.NewInt(0)
+
 type feeHandler struct {
 	mut             sync.RWMutex
 	mapHashFee      map[string]*feeData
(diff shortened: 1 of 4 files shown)


process/transaction/shardProcess.go
```diff
@@ -830,7 +830,9 @@ func (txProc *txProcessor) executeFailedRelayedTransaction(
 	totalFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, userTx.GasLimit)
 	senderShardID := txProc.shardCoordinator.ComputeId(userTx.SndAddr)
 	if senderShardID != txProc.shardCoordinator.SelfId() {
-		totalFee.Sub(totalFee, txProc.economicsFee.ComputeMoveBalanceFee(userTx))
+		moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(userTx)
+		moveBalanceUserFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, moveBalanceGasLimit)
+		totalFee.Sub(totalFee, moveBalanceUserFee)
```
SebastianMarian, line 835: SafeSub ?

sasurobert: this is safe :D it is a big.Int.

SebastianMarian (approved):

process/block/postprocess/feeHandler.go
```diff
@@ -14,6 +14,8 @@ type feeData struct {
 	devFee *big.Int
 }
 
+var zero = big.NewInt(0)
```
iulianpascalau, line 17: :+1:

iulianpascalau (approved):

PR #2597, opened 2020-12-16
Merge dev staking v2 2020.12.16

iulianpascalau:


sasurobert (dismissed):

bogdan-rosianu (approved):

sasurobert (approved):

node/txsimulator/txSimulator.go
```diff
@@ -57,14 +57,15 @@ func NewTransactionSimulator(args ArgsTxSimulator) (*transactionSimulator, error
...
 	retCode, err := ts.txProcessor.ProcessTransaction(tx)
 	if err != nil {
 		failReason = err.Error()
 		txStatus = transaction.TxStatusFail
-	}
-
-	if retCode == vmcommon.Ok {
-		txStatus = transaction.TxStatusSuccess
+	} else {
+		if retCode == vmcommon.Ok {
```
bogdan-rosianu, line 66: was there an issue here? Is there any vm call that returns `Ok` but also an error? @raduchis

bogdan-rosianu (approved):

node/txsimulator/txSimulator.go
raduchis: not with vm call but the ProcessTransaction in the txProcessor returns something like this:

if errors.Is(err, process.ErrInsufficientFunds) {
			receiptErr := txProc.executingFailedTransaction(tx, acntSnd, err)
			if receiptErr != nil {
				return 0, receiptErr
			}
		}

		if errors.Is(err, process.ErrUserNameDoesNotMatchInCrossShardTx) {
			errProcessIfErr := txProc.processIfTxErrorCrossShard(tx, err.Error())
			if errProcessIfErr != nil {
				return 0, errProcessIfErr
			}
			return vmcommon.UserError, nil
		}

PR #2605, opened 2020-12-17
Arithmetic difference economics useful print

bogdan-rosianu:
[low prio]
Useful when parsing logs with `grep` or other tools.

### DIFF
--- a/epochStart/metachain/rewards.go
+++ b/epochStart/metachain/rewards.go
@@ -41,7 +41,7 @@ type ArgsNewRewardsCreator struct {
 	NodesConfigProvider           epochStart.NodesConfigProvider
 	// this fixes the situation where a node has 0 proposed blocks and 0 validator failures
 	// but does not receive rewards.
-	RewardsFix1EpochEnable        uint32
+	RewardsFix1EpochEnable uint32
 }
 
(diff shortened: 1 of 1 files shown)


SebastianMarian (approved):

epochStart/metachain/rewards.go
```diff
@@ -41,7 +41,7 @@ type ArgsNewRewardsCreator struct {
 	NodesConfigProvider           epochStart.NodesConfigProvider
 	// this fixes the situation where a node has 0 proposed blocks and 0 validator failures
 	// but does not receive rewards.
-	RewardsFix1EpochEnable        uint32
+	RewardsFix1EpochEnable uint32
```
iulianpascalau, line 44: where do we use this field?

bogdan-rosianu: it is set to the struct member `rewardsFix1EnableEpoch` and is used here:
```
func (rc *rewardsCreator) isRewardsFix1Enabled(epoch uint32) bool {
	return epoch > rc.rewardsFix1EnableEpoch
}
```
this function is called and proceed accrodingly in `computeValidatorInfoPerRewardAddress`

iulianpascalau (approved):

PR #2608, opened 2020-12-18
Optimize set owner in system sc

iulianpascalau:
- optimized the process of setting the owners of the bls public keys by calling the functionality on the staking SC directly. The update should be done in "bulk mode", improving by 10 times the performance of the update process.


vm/systemSmartContracts/staking.go
```diff
@@ -1380,30 +1380,33 @@ func (s *stakingSC) getWaitingListRegisterNonceAndRewardAddress(args *vmcommon.C
 	return vmcommon.Ok
 }
 
-func (s *stakingSC) setOwner(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+func (s *stakingSC) setOwnersOnAddresses(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
 	if !s.flagStakingV2.IsSet() {
 		s.eei.AddReturnMessage("invalid method to call")
 		return vmcommon.UserError
 	}
 	if !bytes.Equal(args.CallerAddr, s.stakeAccessAddr) {
```
sasurobert, line 1388: change s.stakeAccessAddr to endOfEpochAddr

vm/systemSmartContracts/staking.go
```diff
@@ -1380,30 +1380,33 @@ func (s *stakingSC) getWaitingListRegisterNonceAndRewardAddress(args *vmcommon.C
...
+		s.eei.AddReturnMessage("invalid number of arguments: expected an even number of arguments")
 		return vmcommon.UserError
 	}
-
-	stakedData.OwnerAddress = args.Arguments[1]
-	err = s.saveStakingData(args.Arguments[0], stakedData)
-	if err != nil {
-		s.eei.AddReturnMessage(err.Error())
-		return vmcommon.UserError
+	for i := 0; i < len(args.Arguments); i += 2 {
```
AdoAdoAdo, line 1396: i<len() -1 ?

vm/systemSmartContracts/staking.go
```diff
@@ -1380,30 +1380,33 @@ func (s *stakingSC) getWaitingListRegisterNonceAndRewardAddress(args *vmcommon.C
...
+		stakedData, err := s.getOrCreateRegisteredData(args.Arguments[i])
+		if err != nil {
+			s.eei.AddReturnMessage(err.Error())
+			s.eei.AddReturnMessage(fmt.Sprintf("process stopped at index %d, bls key %s", i, hex.EncodeToString(args.Arguments[i])))
+			return vmcommon.UserError
+		}
```
sasurobert, line 1402: I think we have getExistingRegisteredData. if not you may check that the data actually exists -> thus check if len(stakeData.RewardAddress) > 0

sasurobert (changes requested):

vm/systemSmartContracts/staking.go
iulianpascalau: no need, there is a check that the len of arguments is an even number. So, supposing the len is 2, the loop instructions will be executed only once as the i == 0 (satisfy < 2 condition). After that iteration i becomes 2 and will no longer satisfy the condition i < 2 as 2 < 2 is false. Another example, suppose that len is 8. It will iterate exactly 4 times for i == 0, i == 2, i == 4 and i == 6. We have 3 tests (1 unit test and 2 semi-integrated tests) that check this iteration is done correctly,

vm/systemSmartContracts/staking.go
iulianpascalau: done

vm/systemSmartContracts/staking.go
iulianpascalau: added check

sasurobert (approved):

AdoAdoAdo (approved):

LucianMincu (approved):

PR #2617, opened 2020-12-22
Fixed backwards compatibility new issue

iulianpascalau:
- fixed a backwards compatibility issue when computing transaction fee

### DIFF
--- a/process/smartContract/testUtils.go
+++ b/process/smartContract/testUtils.go
@@ -6,10 +6,12 @@ import (
 	"strings"
 
 	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/data/smartContractResult"
 )
 
 // GetLatestTestError should only be used in tests!
 // It locates the latest error in the collection of smart contracts results
+// TODO remove this file as it is a horrible hack to test some conditions
 func GetLatestTestError(scProcessorAsInterface interface{}) error {
 	scProcessor, ok := scProcessorAsInterface.(*scProcessor)
 	if !ok {
@@ -26,6 +28,11 @@ func GetLatestTestError(scProcessorAsInterface interface{}) error {
 	scResults := scrProvider.GetIntermediateTransactions()
 
 	for i := len(scResults) - 1; i >= 0; i-- {
+		_, isScr := scResults[i].(*smartContractResult.SmartContractResult)
+		if !isScr {
+			continue
+		}
+
 		tx := scResults[i]
 		txData := string(tx.GetData())
 		dataTrimmed := strings.Trim(txData, "@")
--- a/process/transaction/shardProcess.go
+++ b/process/transaction/shardProcess.go
@@ -373,7 +373,8 @@ func (txProc *txProcessor) processTxFee(
 		return nil, nil, err
 	}
 
-	if !txProc.flagPenalizedTooMuchGas.IsSet() {
+	if !txProc.flagMetaProtection.IsSet() {
+		//backwards compatibility return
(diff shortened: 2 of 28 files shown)


sasurobert (approved):

process/transaction/shardProcess.go
```diff
@@ -373,7 +373,8 @@ func (txProc *txProcessor) processTxFee(
 		return nil, nil, err
 	}
 
-	if !txProc.flagPenalizedTooMuchGas.IsSet() {
+	if !txProc.flagMetaProtection.IsSet() {
+		//backwards compatibility return
 		totalCost := core.SafeMul(tx.GasLimit, tx.GasPrice)
 		return moveBalanceFee, totalCost, nil
 	}
```
sasurobert, line 380: there is an actual bug in the new implementation, because of which receipts will not be created. line 382 should be return moveBalanceFee, totalGiven, nil

process/transaction/shardProcess.go
```diff
@@ -353,32 +353,27 @@ func (txProc *txProcessor) processTxFee(
...
 
 	moveBalanceFee := txProc.economicsFee.ComputeMoveBalanceFee(tx)
+	totalCost := txProc.economicsFee.ComputeTxFee(tx)
+	if !txProc.flagPenalizedTooMuchGas.IsSet() {
```
iulianpascalau, line 357: I think this should be txProc.flagMetaProtection. Otherwise the sync will stop at epoch 132

process/transaction/shardProcess.go
```diff
@@ -353,32 +353,27 @@ func (txProc *txProcessor) processTxFee(
 	}
 
 	moveBalanceFee := txProc.economicsFee.ComputeMoveBalanceFee(tx)
+	totalCost := txProc.economicsFee.ComputeTxFee(tx)
```
iulianpascalau, line 356: Are we taking into account not to produce an invalid receipt for the following situation?
Account balance: 1 eGLD, move balance tx with gasLimit * gasPrice = 150eGLD? We should have a test for this scenario checking the receipt's value not to be around 135eGLD.

SebastianMarian (dismissed):

process/transaction/shardProcess.go
sasurobert: it is the same in both cases. total cost in computeTxFee is gaslimit * gasPrice

process/transaction/shardProcess.go
sasurobert: the code will not reach here if there is not enough balance in the account. it will stop at checktxvalues

sasurobert (approved):

process/transaction/shardProcess.go
iulianpascalau: Added test, works as expected :+1:

process/transaction/shardProcess.go
iulianpascalau: Added test, the code performs as designed :+1:

sasurobert (approved):

process/smartContract/testUtils.go
```diff
@@ -6,10 +6,12 @@ import (
 	"strings"
 
 	"github.com/ElrondNetwork/elrond-go/data"
+	"github.com/ElrondNetwork/elrond-go/data/smartContractResult"
 )
 
 // GetLatestTestError should only be used in tests!
 // It locates the latest error in the collection of smart contracts results
+// TODO remove this file as it is a horrible hack to test some conditions
```
SebastianMarian, line 14: :)

SebastianMarian (approved):

LucianMincu (approved):
Initial tests passed.

PR #2618, opened 2020-12-22
Fix token name length condition

SebastianMarian:
* Fixed token name length condition on ESDT issuance


vm/systemSmartContracts/esdt.go
```diff
@@ -191,7 +191,7 @@ func (e *esdt) issue(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
 		e.eei.AddReturnMessage(err.Error())
 		return vmcommon.UserError
 	}
-	if len(args.Arguments[0]) < minLengthForTickerName ||
+	if len(args.Arguments[0]) < int(esdtConfig.MinTokenNameLength) ||
```
sasurobert, line 194: There is no need for no min length check.

sasurobert (changes requested):

vm/systemSmartContracts/esdt.go
iulianpascalau, line 194: do we still use the const/var `minLengthForTickerName` ?

SebastianMarian: It is used for ticker name

sasurobert (approved):

iulianpascalau (approved):

LucianMincu (approved):

PR #2626, opened 2020-12-29
reDelegate-rewards

sasurobert:
added function to redelegate rewards
added function to call unStake at end of epoch in case of system delegation smart contracts.
added function reStakeUnStakedNodes

### DIFF
--- a/epochStart/metachain/stakingDataProvider.go
+++ b/epochStart/metachain/stakingDataProvider.go
@@ -297,11 +297,11 @@ func (sdp *stakingDataProvider) getValidatorInfoFromSC(validatorAddress string) 
 }
 
 // ComputeUnQualifiedNodes will compute which nodes are not qualified - do not have enough tokens to be validators
-func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint32][]*state.ValidatorInfo) ([][]byte, [][]byte, error) {
+func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint32][]*state.ValidatorInfo) ([][]byte, map[string][][]byte, error) {
 	sdp.mutStakingData.Lock()
 	defer sdp.mutStakingData.Unlock()
 
-	ownersWithNotEnoughFunds := make([][]byte, 0)
+	mapOwnersKeys := make(map[string][][]byte, 0)
 	keysToUnStake := make([][]byte, 0)
 	mapBLSKeyStatus := createMapBLSKeyStatus(validatorInfos)
 	for ownerAddress, stakingInfo := range sdp.cache {
@@ -323,10 +323,12 @@ func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint3
 		}
 
 		keysToUnStake = append(keysToUnStake, selectedKeys...)
-		ownersWithNotEnoughFunds = append(ownersWithNotEnoughFunds, []byte(ownerAddress))
+
+		mapOwnersKeys[ownerAddress] = make([][]byte, len(selectedKeys))
+		copy(mapOwnersKeys[ownerAddress], selectedKeys)
 	}
 
-	return keysToUnStake, ownersWithNotEnoughFunds, nil
+	return keysToUnStake, mapOwnersKeys, nil
 }
 
 func createMapBLSKeyStatus(validatorInfos map[uint32][]*state.ValidatorInfo) map[string]string {
--- a/epochStart/metachain/stakingDataProvider.go
+++ b/epochStart/metachain/stakingDataProvider.go
@@ -301,7 +301,7 @@ func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint3
 	sdp.mutStakingData.Lock()
 	defer sdp.mutStakingData.Unlock()
 
-	mapOwnersKeys := make(map[string][][]byte, 0)
+	mapOwnersKeys := make(map[string][][]byte)
 	keysToUnStake := make([][]byte, 0)
 	mapBLSKeyStatus := createMapBLSKeyStatus(validatorInfos)
 	for ownerAddress, stakingInfo := range sdp.cache {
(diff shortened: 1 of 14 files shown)


epochStart/metachain/stakingDataProvider.go
```diff
@@ -297,11 +297,11 @@ func (sdp *stakingDataProvider) getValidatorInfoFromSC(validatorAddress string)
 }
 
 // ComputeUnQualifiedNodes will compute which nodes are not qualified - do not have enough tokens to be validators
-func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint32][]*state.ValidatorInfo) ([][]byte, [][]byte, error) {
+func (sdp *stakingDataProvider) ComputeUnQualifiedNodes(validatorInfos map[uint32][]*state.ValidatorInfo) ([][]byte, map[string][][]byte, error) {
 	sdp.mutStakingData.Lock()
 	defer sdp.mutStakingData.Unlock()
 
-	ownersWithNotEnoughFunds := make([][]byte, 0)
+	mapOwnersKeys := make(map[string][][]byte, 0)
```
SebastianMarian, line 304: Without , 0

epochStart/metachain/systemSCs.go
```diff
@@ -343,6 +354,48 @@ func (s *systemSCProcessor) unStakeOneNode(blsKey []byte, epoch uint32) error {
...
+			RecipientAddr: []byte(address),
+			Function:      "unStakeAtEndOfEpoch",
+		}
+
+		vmOutput, err := s.systemVM.RunSmartContractCall(vmInput)
+		if err != nil {
+			return err
+		}
+		if vmOutput.ReturnCode != vmcommon.Ok {
+			log.Debug("unStakeOneNode", "returnMessage", vmOutput.ReturnMessage, "returnCode", vmOutput.ReturnCode.String())
```
SebastianMarian, line 386: "updateDelegationContracts" instead "unStakeOneNode"

epochStart/metachain/systemSCs.go
```diff
@@ -343,6 +354,48 @@ func (s *systemSCProcessor) unStakeOneNode(blsKey []byte, epoch uint32) error {
...
+			continue
+		}
+		sortedDelegationsSCs = append(sortedDelegationsSCs, address)
+	}
+
+	sort.Slice(sortedDelegationsSCs, func(i, j int) bool {
+		return sortedDelegationsSCs[i] < sortedDelegationsSCs[j]
+	})
+
+	for _, address := range sortedDelegationsSCs {
```
iulianpascalau, line 370: I think we should have a limit or a sort of limit because multiple calls in a for loop at the end of epoch might result in a metachain stuck situation when it does not have enough time to iterate all delegation SCs

epochStart/metachain/systemSCs.go
```diff
@@ -343,6 +354,48 @@ func (s *systemSCProcessor) unStakeOneNode(blsKey []byte, epoch uint32) error {
...
+			},
+			RecipientAddr: []byte(address),
+			Function:      "unStakeAtEndOfEpoch",
+		}
+
+		vmOutput, err := s.systemVM.RunSmartContractCall(vmInput)
+		if err != nil {
+			return err
+		}
+		if vmOutput.ReturnCode != vmcommon.Ok {
```
iulianpascalau, line 385: this process will stop if a delegator does not have enough funds to re-delegate because he/she received too few rewards from the past epoch.

iulianpascalau (changes requested):

epochStart/metachain/systemSCs.go
sasurobert: this is unStakeAtEndOfEpoch. no redelegation.

sasurobert: this will not error.

epochStart/metachain/systemSCs.go
sasurobert: I think there should be no big deal here and we have enough processing time.

iulianpascalau: Initial tests when we had similar process proved otherwise

epochStart/metachain/systemSCs.go
iulianpascalau: right

epochStart/metachain/stakingDataProvider.go
sasurobert: done

epochStart/metachain/systemSCs.go
sasurobert: done

vm/systemSmartContracts/validator.go
```diff
@@ -696,6 +698,118 @@ func (v *validatorSC) cleanRegisteredData(args *vmcommon.ContractCallInput) vmco
...
+
+func (v *validatorSC) getNumStakedAndWaitingNodes(registrationData *ValidatorDataV2, mapCheckedKeys map[string]struct{}) (uint64, error) {
+	numActiveNodes := uint64(0)
+	for _, blsKey := range registrationData.BlsPubKeys {
+		_, exists := mapCheckedKeys[string(blsKey)]
+		if exists {
+			continue
+		}
+
+		stakedData, err := v.getStakedData(blsKey)
```
SebastianMarian, line 763: I think that lines between 763-774 should be replace with numActiveNodes++, as you need to calculate all the other blsKeys beside mapCheckedKeys to know if there are sufficient funds to stake all the node from arguments no matter if some of my nodes are in jail or not. 

With this code I am not be able to restake my unstaked nodes if one of my other staked nodes is in jail

sasurobert: no. that is not true. active nodes are only those which are staked or which are in waiting list.

sasurobert: yes. First you need to resolve your jailed nodes and only after can restake.

epochStart/metachain/systemSCs.go
sasurobert: this is somewhat hard to reach. to have 1000 calls here. you need 1000 delegation contracts where you have unDelegate in the same epoch for more values.

iulianpascalau (approved):

SebastianMarian (approved):

LucianMincu (approved):

PR #2666, opened 2021-01-09
bug fix gas used field

miiu96:
bug-fixes : 
- wrong gas used value for the fail cross-shard transaction (invalid sc call or sc deploy)
- formula that calculates gas used was wrong
- relayed transaction gas used is not equal with gas limit
- invalid transaction receiver shard id is not equal with the receiver address shard id 

new data: 
- added the sender and receiver user name in elasticsearch transaction


SebastianMarian (dismissed):

sasurobert (approved):

sasurobert (approved):

SebastianMarian (approved):

sasurobert (approved):

SebastianMarian (approved):

stefanandy (approved):
System Test passed

PR #2692, opened 2021-01-13
sharding: make the math right for ComputeIdFromBytes

zhiqiangxu:
The correct `numberOfShards` -> `bytesNeed` mapping should be:

```
1~2**8 : 1
2**8+1~2**16 : 2
2**16+1~2**24 : 3
2**24+1~2**32 : 4
```

(`numberOfShards` is guaranteed `>=1` in constructor)

For example, if `numberOfShards` is 256*4, the `bytesNeed` should be 2 instead of 5.


AdoAdoAdo (dismissed):

AdoAdoAdo (commented):
could you do a merge from "external" first?

AdoAdoAdo (approved):

AdoAdoAdo:
good find!

raduchis (approved):

PR #2695, opened 2021-01-13
Fixes related to transactions with username fields

andreibancioiu:
- On Node's API, added fields for sender & receiver username.
 - In the processing components, mark transactions with wrong username as invalid.


sasurobert (approved):

AdoAdoAdo (dismissed):

bogdan-rosianu (dismissed):

node/node.go
```diff
@@ -13,7 +13,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/ElrondNetwork/elrond-go-logger"
+	logger "github.com/ElrondNetwork/elrond-go-logger"
```
SebastianMarian, line 16: Not need for alias

process/transaction/shardProcess.go
```diff
@@ -173,6 +173,13 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction) (vmco
 			}
 		}
 
+		if errors.Is(err, process.ErrUserNameDoesNotMatch) && txProc.flagRelayedTx.IsSet() {
+			receiptErr := txProc.executingFailedTransaction(tx, acntSnd, err)
+			if receiptErr != nil {
+				return vmcommon.UserError, receiptErr
```
SebastianMarian, line 179: For ErrInsufficientFunds error above, if executingFailedTransaction fails it is returned (0, receiptErr) and here (vmcommon.UserError, receiptErr), it is ok?

process/transaction/shardProcess.go
```diff
@@ -173,6 +173,13 @@ func (txProc *txProcessor) ProcessTransaction(tx *transaction.Transaction) (vmco
...
+			}
+		}
```
SebastianMarian, line 181: Should be retuned (vmcommon.UserError, err) or (vmcommon.UserError, nil) like in L188?

process/transaction/shardProcess.go
andreibancioiu: CC: @sasurobert

sasurobert: Yes it is. User code return is not used on the return, only in tests.

process/transaction/shardProcess.go
sasurobert: It is correct now. ReceiptErr must be returned as it signals ErrFailedTransaction

SebastianMarian: I want to say if it is missed a return between line 180-181: return vmcommon.UserError, nil, or it is correct to use the final return from line 190: return vmcommon.UserError, err ?

sasurobert: Yes. It is correct. That tx will not be taken into consideration when building the block.

sasurobert (approved):

sasurobert (approved):

bogdan-rosianu (approved):

AdoAdoAdo (approved):

SebastianMarian (approved):

LucianMincu (approved):

PR #2698, opened 2021-01-13
- fix: let echo command to add the new line character

claudiulataretu:
go toml parser will fail to parse proxy config file because of the
incompatible right value in Observers config parameters

Signed-off-by: Claudiu Ion Lataretu <claudiu.lataretu@gmail.com>


andreibancioiu:
Hello @claudiulataretu!

Do you think you can enable commit signing in your git client (sorry for not mentioning this earlier) - then re-do the PR, with the signed commit?

https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/signing-commits

PR #2699, opened 2021-01-14
- fix: let echo command to add the new line character

claudiulataretu:
go toml parser will fail to parse proxy config file because of the
incompatible right value in Observers config parameters

Signed-off-by: Claudiu Ion Lataretu <claudiu.lataretu@gmail.com>


iulianpascalau (approved):

AdoAdoAdo (approved):

iulianpascalau (approved):
GJ! Thanks :+1:

PR #2702, opened 2021-01-14
added protection for maximum tx value length

bogdan-rosianu:
Added a check when creating a transaction. This way, a transaction is accepted only if the length of the transaction's value (represented as string) is lower than the length of the genesis total supply


cmd/node/main.go
```diff
@@ -1356,6 +1356,8 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 		processComponents.TxLogsProcessor.EnableLogToBeSavedInCache()
 	}
 
+	maxTxValueStringLength := len(economicsConfig.GlobalSettings.GenesisTotalSupply)
```
SebastianMarian, line 1359: Actually the theoretically max supply will be bigger than genesis supply but actually it will have the same number of digits (31 mil vs 20 mil)

node/errors.go
```diff
@@ -178,9 +178,12 @@ var ErrNilFallbackHeaderValidator = errors.New("nil fallback header validator")
 // ErrNilWatchdog signals that a nil watchdog has been provided
 var ErrNilWatchdog = errors.New("nil watchdog")
 
-// ErrInvalidTransactionVersion signals  that an invalid transaction version has been provided
+// ErrInvalidTransactionVersion signals that an invalid transaction version has been provided
 var ErrInvalidTransactionVersion = errors.New("invalid transaction version")
 
+// ErrInvalidMaxTransactionValueLength signals that an invalid maximum transaction value length has been provided
+var ErrInvalidMaxTransactionValueLength = errors.New("invalid transaction version")
```
SebastianMarian, line 185: Change the error message according to its name

node/node.go
```diff
@@ -1079,6 +1080,10 @@ func (n *Node) CreateTransaction(
 		return nil, nil, errors.New("could not fetch signature bytes")
 	}
 
+	if len(value) > n.maxTransactionValueLength {
+		return nil, nil, errors.New("value is too large")
```
SebastianMarian, line 1084: Why it is not used the error defined in node/errors.go ?

node/node_test.go
```diff
@@ -745,6 +751,37 @@ func TestCreateTransaction_SenderShardIdIsInDifferentShardShouldNotValidate(t *t
...
+	gasPrice := uint64(10)
+	gasLimit := uint64(20)
+	txData := []byte("-")
+	signature := "617eff4f"
+
+	tx, txHash, err := n.CreateTransaction(nonce, value, receiver, sender, gasPrice, gasLimit, txData, signature, "chId", 1, 0)
+	assert.Nil(t, tx)
+	assert.Empty(t, txHash)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "too large")
```
SebastianMarian, line 782: Use the error defined in node/errors.go

node/options.go
```diff
@@ -542,6 +542,18 @@ func WithMinTransactionVersion(minTransactionVersion uint32) Option {
 	}
 }
 
+// WithMaxTransactionValueLength sets up the maximum length for the transaction's value field
+func WithMaxTransactionValueLength(maxLength int) Option {
+	return func(n *Node) error {
+		if maxLength == 0 {
+			return ErrInvalidTransactionVersion
```
SebastianMarian, line 549: Change the error according to this method. Anyway I'm not sure if this check is necessary or if it is, we should check with <= as the maxLength is of int type, so it could have a negative value

cmd/node/main.go
bogdan-rosianu: I know, I thought of this and the number of characters is the same. Maybe I can increase it with `1` in case we want to start a testnet with a total supply of let's say `999` and after a few epochs there will be a different number of characters

node/errors.go
bogdan-rosianu: done

node/node.go
bogdan-rosianu: defined a new one

node/node_test.go
bogdan-rosianu: defined a new one

node/options.go
bogdan-rosianu: this check is useful in case of default values (0). extended the protection also for negative values

SebastianMarian (approved):

valentincraciun99 (approved):

node/node.go
```diff
@@ -1079,6 +1080,10 @@ func (n *Node) CreateTransaction(
 		return nil, nil, errors.New("could not fetch signature bytes")
 	}
 
+	if len(value) > n.maxTransactionValueLength {
```
iulianpascalau, line 1083: What about the other fields? Example: receiver, sender, dataField, signatureHex and chainID?

iulianpascalau (changes requested):

node/node.go
```diff
@@ -1057,15 +1059,33 @@ func (n *Node) CreateTransaction(
 	if version == 0 {
 		return nil, nil, ErrInvalidTransactionVersion
 	}
-	if chainID == "" {
-		return nil, nil, ErrInvalidChainID
+	if chainID == "" || len(chainID) != len(string(n.chainID)) {
```
iulianpascalau, line 1062: I think chainID == "" condition is useless here, n.chainID can not be empty

iulianpascalau (approved):

iulianpascalau (approved):

SebastianMarian (approved):

stefanandy (approved):
System Test Passed

PR #2713, opened 2021-01-18
epochStart, core, process: fix economics v2 delta in distributed rewards

AdoAdoAdo:
1. This pull request fixes for economics v2 some differences in the end of epoch computed rewards and the distributed rewards.
This differences appear when unstake is done, due to eligible-leaving validators not being considered in the distribution of rewards.

2. Another fix is regarding the dust from integer division applied on the leader fees for each block proposer 10% of the fees are accumulated for each produced block:

2.1 fees proposer_i = sum(fees_block_j/10), where j is the nonce of a block produced by proposer i
2.2 fees proposers = sum(fees_proposer_i), where i iterates over all eligible validators in shard
2.3 end of epoch computed leaders fees = allFees/10 in epoch

there are losses due to integer division in 2.1 & 2.2 compared to the integer division in 2.3

x1/10 + x2/10 + ... + xn/10 <= (x1+x2+...+xn)/10, where '/' represents integer division.


process/peer/process.go
```diff
@@ -581,8 +586,14 @@ func (vs *validatorStatistics) ProcessRatingsEndOfEpoch(
...
 	for shardId, validators := range validatorInfos {
 		for _, validator := range validators {
-			if validator.List != string(core.EligibleList) {
-				continue
+			if !vs.flagStakingV2Enabled.IsSet() {
+				if validator.List != string(core.EligibleList) {
+					continue
+				}
+			} else {
+				if validator.List != string(core.EligibleList) || !validatorInfo.IsLeavingEligible(validator) {
```
SebastianMarian, line 594: && instead ||  ?

AdoAdoAdo: right thanks

SebastianMarian (dismissed):

core/validatorInfo/validatorInfoUtils.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// IsLeavingEligible returns true if the validator was eligible in the epoch but has done an unstake.
+func IsLeavingEligible(valInfo *state.ValidatorInfo) bool {
+	if valInfo == nil {
+		return false
+	}
+
+	return valInfo.List == string(core.LeavingList) &&
+		(valInfo.LeaderFailure > 0 ||
```
sasurobert, line 15: Maybe you could have summed those numbers and check the sum to be higher than 0

core/validatorInfo/validatorInfoUtils.go
```diff
@@ -0,0 +1,19 @@
+package validatorInfo
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/data/state"
+)
+
+// IsLeavingEligible returns true if the validator was eligible in the epoch but has done an unstake.
+func IsLeavingEligible(valInfo *state.ValidatorInfo) bool {
+	if valInfo == nil {
```
sasurobert, line 10: Maybe this could be renamed to WasEligibleInCurrentEpoch and would return true even if validatorInfo.List == string(core.eligible)

sasurobert: What happens with the validators who are jailed but kept in the system as eligible?

process/peer/process.go
```diff
@@ -591,7 +591,7 @@ func (vs *validatorStatistics) ProcessRatingsEndOfEpoch(
 					continue
 				}
 			} else {
-				if validator.List != string(core.EligibleList) && !validatorInfo.IsLeavingEligible(validator) {
+				if validator.List != string(core.EligibleList) && !validatorInfo.WasLeavingEligibleInCurrentEpoch(validator) {
```
SebastianMarian, line 594: WasJailedEligibleInCurrentEpoch should be taken into consideration? if yes, i think the condition should be
if !vInfo.WasEligibleInCurrentEpoch(validatorInfo)

sasurobert: +1 for Sebi's idea. one function

AdoAdoAdo: the condition is required as we want to process the rating change for the validators that were active either in the Eligible list, or in the Leaving list (did unstake but while active in the last epoch). 
@SebastianMarian The reason why WasEligibleInCurrentEpoch is not used, is because there is no reason to compute for jailed nodes (that were still active), as anyway for them the rating change has no effect. They still need to pay for the unjail which will set their rating to 50, no matter what it was before.

sasurobert (approved):

SebastianMarian (dismissed):

cmd/node/config/config.toml
```diff
@@ -555,7 +555,7 @@
 
 [EpochStartConfig]
     MinRoundsBetweenEpochs = 20
-    RoundsPerEpoch         = 200
+    RoundsPerEpoch         = 100
```
sasurobert, line 558: maybe change this back before merge to dev.

SebastianMarian (dismissed):

cmd/node/config/config.toml
AdoAdoAdo: will do after testing

AdoAdoAdo: done

sasurobert (approved):

SebastianMarian (approved):

stefanandy (approved):
System Test Passed

PR #2715, opened 2021-01-18
Fixed LogIfError usage

iulianpascalau:
- fixed all log.LogIfError calls to contain an odd number of arguments so the messages will be displayed properly

### DIFF
--- a/process/smartContract/process.go
+++ b/process/smartContract/process.go
@@ -575,22 +575,28 @@ func (sc *scProcessor) computeTotalConsumedFeeAndDevRwd(
 				sentGas, _ = core.SafeAddUint64(sentGas, outTransfer.GasLimit)
 				sentGas, _ = core.SafeAddUint64(sentGas, outTransfer.GasLocked)
 			}
+			displayConsumedGas := consumedGas
 			consumedGas, err = core.SafeSubUint64(consumedGas, sentGas)
 			log.LogIfError(err,
 				"function", "computeTotalConsumedFeeAndDevRwd",
-				"consumedGas", consumedGas,
+				"resulted consumedGas", consumedGas,
+				"consumedGas", displayConsumedGas,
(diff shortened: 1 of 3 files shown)


process/smartContract/process.go
```diff
@@ -576,25 +576,49 @@ func (sc *scProcessor) computeTotalConsumedFeeAndDevRwd(
...
+			"moveBalanceGasLimit", moveBalanceGasLimit,
+		)
 	}
 
 	consumedGasWithoutBuiltin, err := core.SafeSubUint64(consumedGas, accumulatedGasUsedForOtherShard)
-	log.LogIfError(err, "computeTotalConsumedFeeAndDevRwd", "accumulatedGasUsedForOtherShard")
+	log.LogIfError(err,
+		"function", "computeTotalConsumedFeeAndDevRwd",
+		"consumedGas", consumedGas,
+		"accumulatedGasUsedForOtherShard", accumulatedGasUsedForOtherShard,
```
AdoAdoAdo, line 613: can you also print the result?

SebastianMarian (approved):

sasurobert (dismissed):

process/smartContract/process.go
iulianpascalau: done

SebastianMarian (approved):

process/smartContract/process.go
sasurobert: The result is always 0.

sasurobert (approved):

process/smartContract/process.go
iulianpascalau: Correct, but in case we might change the default value returned in SafeSubUint64 and SafeAddUint64 in case of an error, this might prove useful

AdoAdoAdo (approved):

PR #2730, opened 2021-01-19
Unit test max block size reached

SebastianMarian:
* Added unit test for max block size reached edge case

### DIFF
--- a/process/block/preprocess/transactions_test.go
+++ b/process/block/preprocess/transactions_test.go
@@ -51,15 +51,6 @@ func shardedDataCacherNotifier() dataRetriever.ShardedDataCacherNotifier {
 					if reflect.DeepEqual(key, []byte("tx2_hash")) {
 						return &transaction.Transaction{Nonce: 10}, true
 					}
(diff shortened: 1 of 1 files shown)


sasurobert (dismissed):

process/block/preprocess/transactions_test.go
```diff
@@ -51,6 +51,15 @@ func shardedDataCacherNotifier() dataRetriever.ShardedDataCacherNotifier {
 					if reflect.DeepEqual(key, []byte("tx2_hash")) {
 						return &transaction.Transaction{Nonce: 10}, true
 					}
+					if reflect.DeepEqual(key, []byte("tx_hash1")) {
```
iulianpascalau, line 54: This could have worked better with:
```
if bytes.Equal(key, []byte("tx_hash1")) {
```

can remain as it is

iulianpascalau (dismissed):

iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):

PR #2731, opened 2021-01-20
Fix linter issues

mateigraura:


sasurobert (approved):

sharding/nodesSetup_test.go
```diff
@@ -651,7 +651,7 @@ func TestNodesSetup_IfNodesWithinMaxShardLimitEquivalentDistribution(t *testing.
 
 	ns = createAndAssignNodes(*ns, 2169)
 
-	ns2 := &(*ns)
+	ns2 := ns
```
AdoAdoAdo, line 654: please revert it as now the test is working properly.

Previously the test was doing the following:
1. creating creating a node setup object, storing it's reference in ns
2. filling this node setup with dummy created nodes and assigning them to shards, according with it's configured max num shards 100
3. creating a clone of this nodes setup object and assigning its reference to n2
4. changing the n2 genesis max num shards to 3 and creating again dummy nodes (using the same model as before) and assigning them to shards with respect to the new maximum number of shards
5. verifying that in the second case (ns2) we reach the exact same configuration of nodes as in ns because although even if ns had a different configuration for max num shards then ns2, both configs only created enough nodes for 3 shards which was within the limit of both. This test was verifying the backwards compatibility of adding a maximum number of shards constraint in the nodes setup.

now with this change, ns2 directly points to the same object as ns, so the test will no longer properly check the backwards compatibility of this change.

integrationTests/vm/txsFee/multiShard/scCalls_test.go
```diff
@@ -86,7 +86,7 @@ func TestScCallExecuteOnSourceAndDstShardShouldWork(t *testing.T) {
 	// execute sc result with gas refund
 	txs := testContextDst.GetIntermediateTransactions(t)
 
-	intermediateTxs = testContextDst.GetIntermediateTransactions(t)
+	_ = testContextDst.GetIntermediateTransactions(t)
```
AdoAdoAdo, line 89: please check also in this file if the second GetIntermediateTransactions call is needed.

integrationTests/vm/txsFee/multiShard/relayedTxScCalls_test.go
```diff
@@ -104,7 +104,7 @@ func TestRelayedTxScCallMultiShardShouldWork(t *testing.T) {
 
 	txs := testContextInnerSource.GetIntermediateTransactions(t)
 
-	intermediateTxs = testContextInnerSource.GetIntermediateTransactions(t)
+	_ = testContextInnerSource.GetIntermediateTransactions(t)
```
AdoAdoAdo, line 107: check here as well if the second call is needed

AdoAdoAdo (changes requested):

epochStart/metachain/baseRewards_test.go
```diff
@@ -250,12 +249,12 @@ func TestBaseRewardsCreator_addProtocolRewardToMiniblocks(t *testing.T) {
 
 	protRwShard := args.ShardCoordinator.ComputeId(protRwAddr)
 	mbSlice := createDefaultMiniBlocksSlice()
-	err = rwd.addProtocolRewardToMiniBlocks(protRwTx, mbSlice, protRwShard)
+	_ = rwd.addProtocolRewardToMiniBlocks(protRwTx, mbSlice, protRwShard)
```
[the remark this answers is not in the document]
AdoAdoAdo: only for this test, for the others it is OK to be ignored, but here we are checking specifically addProtocolRewardToMiniBlocks

integrationTests/vm/txsFee/multiShard/relayedScDeploy_test.go
```diff
@@ -80,7 +80,7 @@ func TestRelayedSCDeployShouldWork(t *testing.T) {
 
 	txs := testContextInner.GetIntermediateTransactions(t)
 
-	intermediateTxs = testContextInner.GetIntermediateTransactions(t)
+	_ = testContextInner.GetIntermediateTransactions(t)
```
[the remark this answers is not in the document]
sasurobert: These tests were done by @mihaiiuga. I think this is a simple copy paste error. One read is enough.

integrationTests/vm/txsFee/multiShard/scCalls_test.go
sasurobert: one get is enough

integrationTests/vm/txsFee/multiShard/relayedTxScCalls_test.go
sasurobert: one get is enough

integrationTests/vm/txsFee/multiShard/scCalls_test.go
```diff
@@ -180,7 +179,7 @@ func TestScCallExecuteOnSourceAndDstShardInvalidOnDst(t *testing.T) {
...
```
AdoAdoAdo, line 182: this read is unnecessary, please remove it

integrationTests/vm/txsFee/multiShard/relayedTxScCalls_test.go
```diff
@@ -225,7 +224,7 @@ func TestRelayedTxScCallMultiShardFailOnInnerTxDst(t *testing.T) {
...
```
AdoAdoAdo, line 227: this read is unnecessary, please remove

AdoAdoAdo (changes requested):

AdoAdoAdo (approved):

sasurobert (approved):

valentincraciun99 (approved):
system testing passed

PR #2735, opened 2021-01-21
new arwen with fixed memory leakage

sasurobert:
new arwen with fixed memory leakage

### DIFF
--- a/process/smartContract/testUtils.go
+++ b/process/smartContract/testUtils.go
@@ -105,3 +105,13 @@ func GetAllSCRs(scProcessorAsInterface interface{}) []data.TransactionHandler {
 
 	return scrProvider.GetIntermediateTransactions()
 }
+
+// CleanGasRefunded cleans the gas computation handler
+func CleanGasRefunded(scProcessorAsInterface interface{}) {
+	scProc, ok := scProcessorAsInterface.(*scProcessor)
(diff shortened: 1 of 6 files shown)


process/smartContract/testUtils.go
```diff
@@ -105,3 +105,13 @@ func GetAllSCRs(scProcessorAsInterface interface{}) []data.TransactionHandler {
 
 	return scrProvider.GetIntermediateTransactions()
 }
+
+// CleanGasRefunded cleans the gas computation handler
+func CleanGasRefunded(scProcessorAsInterface interface{}) {
```
iulianpascalau, line 110: Totally disagree to use this type of code even in tests. It breaks all rules regarding clean code. Let's find some other way to do the same job. I know this file contains all sort of functions but they are terrible horrible hacks and should be removed.

iulianpascalau (changes requested):

SebastianMarian (approved):

iulianpascalau (dismissed):
Created task EN-8629

SebastianMarian (approved):

iulianpascalau (approved):

LucianMincu (approved):
My ram looks bigger with this one. 👍 💻

PR #2736, opened 2021-01-21
epochStart/metachain: fix edge case in ForceEpochStart

zhiqiangxu:
According to this code at line 199:

	isNormalEpochStart := t.currentRound > t.currEpochStartRound+t.roundsPerEpoch

a round should be `> t.currEpochStartRound+t.roundsPerEpoch` to be considered in the next epoch.


epochStart/metachain/trigger.go
```diff
@@ -169,7 +169,7 @@ func (t *trigger) ForceEpochStart(round uint64) {
 	defer t.mutTrigger.Unlock()
 
 	t.nextEpochStartRound = round
-	if t.nextEpochStartRound >= t.currEpochStartRound+t.roundsPerEpoch {
+	if t.nextEpochStartRound > t.currEpochStartRound+t.roundsPerEpoch {
```
iulianpascalau, line 172: I remember writing this code. The idea was the following: if the nextEpochStartRound was higher or equal to the computed next-end of epoch, do not initiate force epoch start as the epoch start will happen in a natural way. I would keep the condition as it is.

zhiqiangxu: The natural way will happen at round `t.currEpochStartRound+t.roundsPerEpoch+1`, not `t.currEpochStartRound+t.roundsPerEpoch` .

iulianpascalau: That is true, indeed :)

iulianpascalau (dismissed):

iulianpascalau (approved):

SebastianMarian (approved):

PR #2737, opened 2021-01-21
Web server integration test + api small refactor

iulianpascalau:
- added web server test processor node
- API small refactor to apply Dependency Inversion Principle
- add integration test that prove backwards compatibility on gas cost route

### DIFF
--- a/integrationTests/interface.go
+++ b/integrationTests/interface.go
@@ -1,8 +1,19 @@
 package integrationTests
 
 import (
+	"math/big"
+
+	"github.com/ElrondNetwork/elrond-go/api"
 	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/statistics"
+	dataApi "github.com/ElrondNetwork/elrond-go/data/api"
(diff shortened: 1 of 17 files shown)


SebastianMarian (dismissed):

integrationTests/interface.go
```diff
@@ -39,3 +50,42 @@ type NetworkShardingUpdater interface {
 	UpdatePeerIdShardId(pid core.PeerID, shardId uint32)
 	IsInterfaceNil() bool
 }
+type (
```
sasurobert, line 53: add empty line before

sasurobert: why is type "(" needed ? couldn't be type Facade interface { etc }

iulianpascalau: Didn't knew this was even possible. Changing to align with the rest of the code.

iulianpascalau: Done

SebastianMarian (approved):

sasurobert (approved):

valentincraciun99 (approved):

PR #2747, opened 2021-01-26
EN-8693: economics persistent metrics

bogdan-rosianu:
Made the metrics available in `/network/economics` endpoint persistent. 

Testing procedure:
At any time (preferably once in epoch 0 and once in a higher epoch) for any node check the output of the `/network/economics` endpoint (Rest GET request). Save the output somewhere. Then restart the node and check again the endpoint. The output should be the same.


SebastianMarian (approved):

sasurobert (approved):

valentincraciun99 (approved):
system testing passed

PR #2763, opened 2021-02-01
EN-8728: gas price modifier metric update

bogdan-rosianu:
Made the `gas price modifier` metric to update when the flag is set.
Passed the `app status handler` to `economics data` using `null object design pattern` in order to avoid constructor/arguments changes. This way, the `app status handler` is initialized as a `nil implementation` on the constructor and can be changed using a `setter`.

Testing procedure:
-after starting a node check the endpoint `node-url:port/network/config` and look after the `erd_gas_price_modifier` metric (`1` for current configuration)
-after a number of epochs equal to `GasPriceModifierEnableEpoch` from `config.toml` check the endpoint again. The value should be updated to the one from `economics.toml` (`0.01` for current configuration)

### DIFF
--- a/process/economics/economicsData.go
+++ b/process/economics/economicsData.go
@@ -1,6 +1,7 @@
 package economics
 
 import (
+	"fmt"
 	"math/big"
 	"strconv"
 	"sync"
@@ -12,6 +13,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/core/check"
 	"github.com/ElrondNetwork/elrond-go/data/smartContractResult"
 	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/statusHandler"
 )
 
 var _ process.EconomicsDataHandler = (*economicsData)(nil)
@@ -43,6 +45,7 @@ type economicsData struct {
 	gasPriceModifierEnableEpoch      uint32
 	topUpGradientPoint               *big.Int
 	topUpFactor                      float64
+	statusHandler                    core.AppStatusHandler
 }
 
 // ArgsNewEconomicsData defines the arguments needed for new economics economicsData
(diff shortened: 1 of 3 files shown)


process/economics/economicsData.go
```diff
@@ -431,6 +446,17 @@ func (ed *economicsData) EpochConfirmed(epoch uint32) {
 
 	ed.flagGasPriceModifier.Toggle(epoch >= ed.gasPriceModifierEnableEpoch)
 	log.Debug("economics: gas price modifier", "enabled", ed.flagGasPriceModifier.IsSet())
+	ed.setGasPriceModifierMetric()
+}
+
+func (ed *economicsData) setGasPriceModifierMetric() {
+	if check.IfNil(ed.statusHandler) {
+		log.Debug("cannot set GasPriceModifier metric - nil status handler")
```
iulianpascalau, line 454: How is this possible? That was one of the reason we used nil status handler. To get rid of these checks.

process/economics/economicsData.go
```diff
@@ -431,6 +446,17 @@ func (ed *economicsData) EpochConfirmed(epoch uint32) {
...
+	}
+
+	if ed.flagGasPriceModifier.IsSet() {
+		ed.statusHandler.SetStringValue(core.MetricGasPriceModifier, fmt.Sprintf("%g", ed.gasPriceModifier))
```
iulianpascalau, line 458: This will not work as expected in case of rollbacks. Suggestion to get rid of the if branch

iulianpascalau (changes requested):

process/economics/economicsData.go
bogdan-rosianu: Technically it is not possible if economicsData is used right. Ok, will remove the check

process/economics/economicsData.go
bogdan-rosianu: No, I can't remove the `if` because then it will be executed at every epoch, not just when the new gas price modifier comes in.

iulianpascalau (approved):

SebastianMarian (approved):

stefanandy (approved):
System Test Passed

PR #2772, opened 2021-02-02
EN-8070: api request logging middleware

bogdan-rosianu:
Implemented and integrated a new middleware that will log different data about a request (path, request body, status code, response body, and duration) if it takes longer than a pre-defined threshold **or**it is unsuccessful (not 200 status code).
Introduced a wrapper over the gin's Writer in order to fetch the response.

Testing procedure:
-switch `LoggingEnabled` to `true` in `api.toml`
-make calls to any API endpoint
-if the execution time exceeds the threshold (at the time of opening the PR, the threshold was 100 microseconds so we can test and measure) or it is unsuccessful, you should see a `log.Debug` containing `[...] api request` in node's logs

Example:
-make a call to the balance endpoint. `nodeUrl:port/address/erd1.../balance` . alter the bech32 address so the request is failed with code 400. you should see this logged


api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return b.String()
+}
+
+type bodyWriter struct {
+	gin.ResponseWriter
+	body *bytes.Buffer
+}
+
+func (w bodyWriter) Write(b []byte) (int, error) {
```
miiu96, line 101: use `pointer receiver` instead of `value receiver ` because you change something in the object

api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
...
+			log.Error(err.Error())
+			return
+		}
+
+		if len(reqBodyBytes) > 0 {
+			request = removeWhitespacesFromString(string(reqBodyBytes))
+		}
+	}
+
+	logPrefix := prefixDurationTooLong
```
miiu96, line 69: create a method compute log prefix

api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
...
+		t := time.Now()
+
+		bw := &bodyWriter{body: bytes.NewBufferString(""), ResponseWriter: c.Writer}
+		c.Writer = bw
+
+		c.Next()
+
+		latency := time.Since(t)
+		status := c.Writer.Status()
+		if latency > rl.thresholdDurationForLoggingRequest || c.Writer.Status() != http.StatusOK {
```
miiu96, line 41: make a variable with these conditions `shouldLog` := `latency > rl.thresholdDurationForLoggingRequest || c.Writer.Status() != http.StatusOK`

api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
+package middleware
```
miiu96, line 1: tests are missing for this new functionality

api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// IsInterfaceNil returns true if there is no value under the interface
+func (rl *responseLoggerMiddleware) IsInterfaceNil() bool {
+	return rl == nil
+}
+
+func logRequestAndResponse(c *gin.Context, duration time.Duration, status int, response string) {
+	request := "n/a"
```
iulianpascalau, line 54: As discussed, we can have a flag for this (defined in api.toml) to enable the log only when desired.

api/middleware/responseLogger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	logPrefix := prefixDurationTooLong
+	if status == http.StatusBadRequest {
+		logPrefix = prefixBadRequest
+	} else if status == http.StatusInternalServerError {
+		logPrefix = prefixInternalError
+	}
+
+	log.Warn(logPrefix+" api request",
```
iulianpascalau, line 76: I think log.Debug will suffice.

api/middleware/responseLogger.go
bogdan-rosianu: I will leave it as it is

api/middleware/responseLogger.go
bogdan-rosianu: created a method for the entire title

api/middleware/responseLogger.go
bogdan-rosianu: done

api/middleware/responseLogger.go
bogdan-rosianu: added

api/middleware/responseLogger.go
bogdan-rosianu: done. added a flag for the `enabled` switch and also moved the threshold to the config file

api/middleware/responseLogger.go
bogdan-rosianu: changed

api/middleware/responseLogger.go
```diff
@@ -66,15 +74,27 @@ func logRequestAndResponse(c *gin.Context, duration time.Duration, status int, r
...
 	}
 
+	title := rlm.computeLogTitle(status)
+
+	rlm.printRequestFunc(title, c.Request.RequestURI, duration, status, request, response)
+}
+
+func (rlm *responseLoggerMiddleware) computeLogTitle(status int) string {
 	logPrefix := prefixDurationTooLong
 	if status == http.StatusBadRequest {
```
iulianpascalau, line 84: I still hate those else-if's :)

api/middleware/responseLogger_test.go
```diff
@@ -0,0 +1,203 @@
...
+		rlf.title = title
+		rlf.path = path
+		rlf.duration = duration
+		rlf.status = status
+		rlf.response = response
+		rlf.request = request
+	}
+
+	rlm := NewResponseLoggerMiddleware(thresholdDuration)
+	rlm.printRequestFunc = printHandler
```
iulianpascalau, line 76: can remain as it is. If concurrency errors happen because of this, might consider to create a new constructor in `expiort_test.go` on which you can inject the print handler.

facade/nodeFacade.go
```diff
@@ -220,8 +216,10 @@ func (nf *nodeFacade) startRest() {
 			"SameSourceResetIntervalInSec", nf.wsAntifloodConfig.SameSourceResetIntervalInSec,
 		)
 
-		responseLoggerMiddleware := middleware.NewResponseLoggerMiddleware(thresholdDurationForLoggingRequests)
-		limiters = append(limiters, responseLoggerMiddleware)
+		if nf.apiRoutesConfig.Logging.LoggingEnabled {
+			responseLoggerMiddleware := middleware.NewResponseLoggerMiddleware(time.Duration(nf.apiRoutesConfig.Logging.ThresholdInNanoSeconds) * time.Nanosecond)
```
iulianpascalau, line 220: :+1: I think this is the most efficient way to enable/disable the logging.

cmd/node/config/api.toml
```diff
@@ -1,3 +1,13 @@
+# Logging holds settings related to api requests logging
+[Logging]
+    # LoggingEnabled - if this flag is set to true, then if a requests exceeds a threshold or it is unsuccessful, then
+    # a log will be printed
+    LoggingEnabled = false
+
+    # ThresholdNanoSeconds represents the maximum duration to consider a request as normal. Above this, if the LoggingEnabled
+    # flag is set to true, then a log will be printed
+    ThresholdInNanoSeconds = 1
```
iulianpascalau, line 9: isn't this too low? Suggestion to change this to microseconds as defining time duration in nanoseconds will yield too large values. And maybe change this to 100 microseconds?

api/middleware/responseLogger_test.go
bogdan-rosianu: for now, tests pass even if I set `-race` mode when running tests

facade/nodeFacade.go
bogdan-rosianu: Yes. Also, if one forgets to update the `api.toml` then the default value of `false` will apply so no logging

cmd/node/config/api.toml
bogdan-rosianu: done

api/middleware/responseLogger.go
bogdan-rosianu: I couldn't find a better approach in order to treat the 'duration too long' vs various http codes (from which 2 are treated separately)

iulianpascalau (approved):

iulianpascalau (dismissed):

iulianpascalau (dismissed):

iulianpascalau (approved):

iulianpascalau (approved):

miiu96 (approved):

gabi-vuls (approved):
System test passed.

PR #2783, opened 2021-02-08
Unit tests for block execution invariant

SebastianMarian:
* Added unit tests for block execution invariant


iulianpascalau (approved):

stefanandy (approved):

process/coordinator/process_test.go
```diff
@@ -3017,5 +3018,1104 @@ func TestTransactionCoordinator_GetNumOfCrossShardSpecialTxsShouldWork(t *testin
...
+}
+
+func TestTransactionCoordinator_GetMaxAccumulatedAndDeveloperFeesShouldWork(t *testing.T) {
+	t.Parallel()
+
+	tx1GasLimit := uint64(100)
+	tx2GasLimit := uint64(200)
+	tx3GasLimit := uint64(300)
+
+	txHash := []byte("tx_hash1")
```
bogdan-rosianu, line 4072: `tx_hash1` is used 40 times in this file. maybe extract it to a const?

process/coordinator/process_test.go
```diff
@@ -3017,5 +3018,1104 @@ func TestTransactionCoordinator_GetNumOfCrossShardSpecialTxsShouldWork(t *testin
...
+	tx1 := &transaction.Transaction{Nonce: 1, GasLimit: tx1GasLimit, GasPrice: 1}
+	tx2 := &transaction.Transaction{Nonce: 2, GasLimit: tx2GasLimit, GasPrice: 1}
+	tx3 := &transaction.Transaction{Nonce: 3, GasLimit: tx3GasLimit, GasPrice: 1}
+
+	txHash1 := "hash1"
+	txHash2 := "hash2"
+	txHash3 := "hash3"
+
+	mapAllTxs := make(map[string]data.TransactionHandler)
+	mapAllTxs[txHash1] = tx1
```
bogdan-rosianu, line 4107: could move transactions' declarations here (no need for `tx1`, `tx2`, `tx3` variables as you only use them here). 
this comment is optional, but if you want to implement it, please check all the added tests

bogdan-rosianu (changes requested):

process/coordinator/process_test.go
SebastianMarian: Will do this in EN-8791 · Refactor TransactionCoordinator with arguments, at which I'm working it right now :)

process/coordinator/process_test.go
SebastianMarian: Will do this in EN-8791 · Refactor TransactionCoordinator with arguments, at which I'm working it right now :)

bogdan-rosianu (approved):

PR #2811, opened 2021-02-11
delegation-improvements

sasurobert:
Delegation total rewards for users
reDelegateRewards on top of delegation cap
set and get metadata
set on and off reDelegateRewards on top of delegation cap
Deleted list of delegators, added number of users instead of it.
Delete delegators when no more funds


vm/systemSmartContracts/delegation.go
```diff
@@ -988,7 +1017,9 @@ func (d *delegation) reDelegateRewards(args *vmcommon.ContractCallInput) vmcommo
 
 	delegateValue := big.NewInt(0).Set(delegator.UnClaimedRewards)
 	delegator.UnClaimedRewards.SetUint64(0)
-	return d.finishDelegateUser(globalFund, delegator, dConfig, dStatus, args.CallerAddr, args.RecipientAddr, delegateValue, false)
+
+	return d.finishDelegateUser(globalFund, delegator, dConfig, dStatus, args.CallerAddr,
```
iulianpascalau, line 1021: I would rename the `CheckCapOnReDelegateRewards` because when I first read it I thought is a function pointer or something. Maybe `FlagCheckCapOnReDelegateRewards` ?

vm/systemSmartContracts/delegation.go
```diff
@@ -1000,11 +1031,12 @@ func (d *delegation) finishDelegateUser(
 	scAddress []byte,
 	callValue *big.Int,
 	isNew bool,
+	checkDelegationCap bool,
```
iulianpascalau, line 1034: there is a naming problem here, for example, the changes on L1038 reads like this: 
if the delegation cap is enabled and **check Deglegtion cap is true** and we passed the delegation cap value, return error
So, I would rename this flag to `disableReDelegation` so the above mention condition reads ok

vm/systemSmartContracts/delegation.go
```diff
@@ -1087,12 +1121,6 @@ func (d *delegation) delegate(args *vmcommon.ContractCallInput) vmcommon.ReturnC
 	return d.delegateUser(args.CallValue, args.CallerAddr, args.RecipientAddr, dStatus)
 }
 
-func (d *delegation) addNewDelegatorToList(dStatus *DelegationContractStatus, address []byte) error {
```
iulianpascalau, line 1090: Those changes are incompatible with what we have on the testnet right now. Will provide a new, fresh testnet in order to test these in the wild.

vm/systemSmartContracts/delegation.go
```diff
@@ -1827,6 +1898,11 @@ func (d *delegation) getContractConfig(args *vmcommon.ContractCallInput) vmcommo
 		changeableServiceFee = "true"
 	}
 
+	checkCapOnReDelegate := "false"
```
iulianpascalau, line 1901: we are not being consistent when using flags. On L496 we use yes/no, here true/false. Let's standardize and use constants.

vm/systemSmartContracts/delegation.go
```diff
@@ -478,6 +486,31 @@ func (d *delegation) changeServiceFee(args *vmcommon.ContractCallInput) vmcommon
 	return vmcommon.Ok
 }
 
+func (d *delegation) setCheckCapOnReDelegateRewards(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
```
iulianpascalau, line 489: You can improve the code coverage in this function in order to get a green result from the code coverage tool

vm/systemSmartContracts/delegation.go
```diff
@@ -1621,9 +1650,51 @@ func (d *delegation) withdraw(args *vmcommon.ContractCallInput) vmcommon.ReturnC
...
+		d.eei.AddReturnMessage(err.Error())
+		return vmcommon.UserError
+	}
+
 	return vmcommon.Ok
 }
 
+func (d *delegation) deleteDelegatorIfNeeded(address []byte, delegator *DelegatorData) error {
+	if d.isOwner(address) {
+		return nil
```
iulianpascalau, line 1664: this can be easily tested to improve the code coverage

vm/systemSmartContracts/delegation.go
```diff
@@ -1621,9 +1650,51 @@ func (d *delegation) withdraw(args *vmcommon.ContractCallInput) vmcommon.ReturnC
...
+	if delegator.UnClaimedRewards.Cmp(zero) == 0 {
+		d.eei.SetStorage(address, nil)
+
+		dStatus, errGet := d.getDelegationStatus()
+		if errGet != nil {
+			return errGet
+		}
+
+		if dStatus.NumUsers > 0 {
+			dStatus.NumUsers--
```
iulianpascalau, line 1686: This is imported but not tested. Please add an unit test for this (or change an existing one)

vm/systemSmartContracts/delegation.go
```diff
@@ -2055,22 +2132,74 @@ func (d *delegation) getDelegatorFundsData(args *vmcommon.ContractCallInput) vmc
...
+}
+
+func (d *delegation) setMetaData(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	returnCode := d.checkOwnerCallValueGasAndDuplicates(args)
+	if returnCode != vmcommon.Ok {
+		return returnCode
+	}
+
+	if len(args.Arguments) != 3 {
+		d.eei.AddReturnMessage("needed 3 arguments")
```
iulianpascalau, line 2160: this can be easily tested

vm/systemSmartContracts/delegation.go
```diff
@@ -2055,22 +2132,74 @@ func (d *delegation) getDelegatorFundsData(args *vmcommon.ContractCallInput) vmc
...
+
+func (d *delegation) getMetaData(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
+	returnCode := d.checkArgumentsForGeneralViewFunc(args)
+	if returnCode != vmcommon.Ok {
+		return returnCode
+	}
+
+	marshaledData := d.eei.GetStorage([]byte(delegationMetaData))
+	if len(marshaledData) == 0 {
+		d.eei.AddReturnMessage("delegation meta data is not set")
```
iulianpascalau, line 2188: This can be easily tested

iulianpascalau (changes requested):
Mostly tested related issues

vm/systemSmartContracts/delegation.go
sasurobert: this is not a flag. it is a propriety.

vm/systemSmartContracts/delegation.go
sasurobert: reDelegation is always all enabled. It depends whether we check the total delegation cap or not in case this flag is set or not

vm/systemSmartContracts/delegation.go
sasurobert: yep. this list would have limited us to 22K users per delegation contract.

vm/systemSmartContracts/delegation.go
sasurobert: It is tested. I added tests for this.

vm/systemSmartContracts/delegation.go
sasurobert: It is tested. I added tests for this.

vm/systemSmartContracts/delegation.go
sasurobert: common. checking number of arguments. it is not a prio. tests are needed in order to check some computation/logic not to improve coverage.

vm/systemSmartContracts/delegation.go
sasurobert: would not test every if, as these are clear

sasurobert: set and get meta data was already tested.

vm/systemSmartContracts/delegation.go
sasurobert: this was already tested.

vm/systemSmartContracts/delegation.go
sasurobert: this is a view function only and it is treated in the delegation dashboard. I can change all to true / false

vm/systemSmartContracts/delegation.go
sasurobert: already improved test coverage by 12%

vm/systemSmartContracts/delegation.go
iulianpascalau: Sounds like a function

vm/systemSmartContracts/delegation.go
iulianpascalau: I know what it does but that is not clear from the naming. That is why I suggest to rename

SebastianMarian (dismissed):

vm/systemSmartContracts/delegation.go
iulianpascalau: having true / false everywhere will be great :+1:

vm/systemSmartContracts/delegation.go
iulianpascalau: Not all branches inside are tested. Some if branches can be easily tested in unit tests a few lines long.

vm/systemSmartContracts/delegation.go
```diff
@@ -493,9 +493,9 @@ func (d *delegation) setCheckCapOnReDelegateRewards(args *vmcommon.ContractCallI
 	}
 
 	switch string(args.Arguments[0]) {
-	case "yes":
+	case "true":
```
iulianpascalau, line 496: could have used consts

vm/systemSmartContracts/delegation.go
iulianpascalau: not tested

vm/systemSmartContracts/delegation.go
iulianpascalau: It is high prio when doing merges. Remember! We do not write unit tests just for fun. I know this if should work well and there should be no bugs there but remember that others might maintain the code and do some merges in which they can make mistakes.

vm/systemSmartContracts/delegation.go
iulianpascalau: It does not matter if the code is clear or not. tests should be there. Especially when we can easily write them. No excuses.

iulianpascalau (changes requested):

vm/systemSmartContracts/delegation.go
sasurobert: tested this as well.

iulianpascalau (dismissed):

SebastianMarian (approved):

iulianpascalau (approved):

vm/systemSmartContracts/delegation.go
```diff
@@ -1426,7 +1421,7 @@ func (d *delegation) computeAndUpdateRewards(callerAddress []byte, delegator *De
 		}
 
 		var rewardsForOwner *big.Int
-		percentage := float64(rewardData.ServiceFee) / float64(percentageDenominator)
+		percentage := float64(rewardData.ServiceFee) / float64(d.maxServiceFee)
```
SebastianMarian, line 1424: I think here we should add a check if percentage will be greater than 1. In this case rewardsForOwner will be greater than RewardToDistribute and than rewardForDelegator will be negative and so on. Something will definitely crash afterward.

sasurobert: ok. will add check on constructor.

iulianpascalau: Checked that through the code. Before saving the rewardData.ServiceFee there are strict checks. Only if we change the code in a buggy way we might have erroneous data

vm/systemSmartContracts/delegation.go
```diff
@@ -96,6 +96,9 @@ func NewDelegationSystemSC(args ArgsNewDelegation) (*delegation, error) {
 	if args.DelegationSCConfig.MinServiceFee > args.DelegationSCConfig.MaxServiceFee {
 		return nil, fmt.Errorf("%w minServiceFee bigger than maxServiceFee", vm.ErrInvalidDelegationSCConfig)
 	}
+	if args.DelegationSCConfig.MaxServiceFee < 1 {
```
iulianpascalau, line 99: maybe add similar condition to args.DelegationSCConfig.MinServiceFee ?

sasurobert: min service fee can be 0

iulianpascalau: ok, we can call this with -1 for min fee and 1 for max fee?

vm/systemSmartContracts/delegation.go
```diff
@@ -96,6 +96,9 @@ func NewDelegationSystemSC(args ArgsNewDelegation) (*delegation, error) {
...
+		return nil, fmt.Errorf("%w maxServiceFee must be more than 0", vm.ErrInvalidDelegationSCConfig)
```
SebastianMarian, line 100: must be higher

SebastianMarian (approved):

vm/systemSmartContracts/delegation.go
sasurobert: no. this is an UINT64

iulianpascalau (changes requested):
Code coverage decrease

iulianpascalau (approved):
Code coverage decrease on autogenerated file

stefanandy (approved):
System Test Passed

PR #2815, opened 2021-02-12
EN-8854: get all key-value pairs for an account API endpoint

bogdan-rosianu:
Implemented the `address/{address}/keys` endpoint that will return a map containing all keys and their values for an account. The value is represented in hexadecimal encoding.

Testing procedure:
-start a testnet using the corresponding proxy's branch
-find an address that must have saved something inside it's key-value storage (for example, an address that interacted with an ESDT token) and make an API call to `proxy-url/address/erd1.../keys`. it should return a map containing all its keys

### DIFF
--- a/api/address/routes.go
+++ b/api/address/routes.go
@@ -17,6 +17,7 @@ const (
 	getAccountPath  = "/:address"
 	getBalancePath  = "/:address/balance"
 	getUsernamePath = "/:address/username"
+	getKeysPath     = "/:address/keys"
 	getKeyPath      = "/:address/key/:key"
 	getESDTTokens   = "/:address/esdt"
 	getESDTBalance  = "/:address/esdt/:tokenIdentifier"
@@ -31,6 +32,7 @@ type FacadeHandler interface {
 	GetCode(account state.UserAccountHandler) []byte
 	GetESDTBalance(address string, key string) (string, string, error)
 	GetAllESDTTokens(address string) ([]string, error)
+	GetKeyValuePairs(address string) (map[string]string, error)
 	IsInterfaceNil() bool
 }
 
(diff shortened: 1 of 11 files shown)


miiu96 (changes requested):
Please add unit tests to increase coverage

SebastianMarian (approved):

SebastianMarian (dismissed):

node/node.go
```diff
@@ -440,6 +440,41 @@ func (n *Node) GetUsername(address string) (string, error) {
...
+
+	ctx := context.Background()
+	chLeaves, err := userAccount.DataTrie().GetAllLeavesOnChannel(rootHash, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	mapToReturn := make(map[string]string)
+	for leaf := range chLeaves {
+		mapToReturn[string(leaf.Key())] = hex.EncodeToString(leaf.Value())
```
sasurobert: hex.encode the leaf.Key

sasurobert (changes requested):

api/address/routes.go
```diff
@@ -31,6 +32,7 @@ type FacadeHandler interface {
 	GetCode(account state.UserAccountHandler) []byte
 	GetESDTBalance(address string, key string) (string, string, error)
 	GetAllESDTTokens(address string) ([]string, error)
+	GetKeyValuePairs(address string) (map[string]string, error)
```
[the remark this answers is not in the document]
bogdan-rosianu: I will leave it like this so we'll have all of the keys and values hex encoded

node/node.go
bogdan-rosianu: you reviewed an outdated code. it is hex encoded now

miiu96 (approved):

SebastianMarian (approved):

mateigraura (approved):
system test passed

sasurobert (approved):

PR #2816, opened 2021-02-15
[soft restart] fixed issue when checking core subcomponents

bogdan-rosianu:
fixed the deadlock situation when a method that was mutex-protected was called inside a function that was already protected by the same mutex

### DIFF
--- a/factory/coreComponentsHandler.go
+++ b/factory/coreComponentsHandler.go
@@ -123,7 +123,7 @@ func (mcc *managedCoreComponents) CheckSubcomponents() error {
 	if check.IfNil(mcc.syncTimer) {
 		return errors.ErrNilSyncTimer
 	}
-	if check.IfNil(mcc.RoundHandler()) {
+	if check.IfNil(mcc.coreComponents.RoundHandler) {
 		return errors.ErrNilRoundHandler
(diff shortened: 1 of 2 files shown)


miiu96 (approved):

factory/coreComponentsHandler.go
```diff
@@ -123,7 +123,7 @@ func (mcc *managedCoreComponents) CheckSubcomponents() error {
 	if check.IfNil(mcc.syncTimer) {
 		return errors.ErrNilSyncTimer
 	}
-	if check.IfNil(mcc.RoundHandler()) {
+	if check.IfNil(mcc.coreComponents.RoundHandler) {
```
AdoAdoAdo, line 126: please rename field in coreComponents struct from RoundHandler to roundHandler, and use here directly mcc.roundHandler

AdoAdoAdo (changes requested):

factory/coreComponentsHandler.go
bogdan-rosianu: done

AdoAdoAdo (approved):

SebastianMarian (approved):

PR #2817, opened 2021-02-15
Implement redundancy for elrond nodes

SebastianMarian:
* Implemented redundancy mechanism for nodes

**How to be tested:**

1. Start a normal system test and choose one node and with its validatorKey.pem start another two instances, 
one with flag **RedundancyLevel=1** and one with flag **RedundancyLevel=2** set in prefs.toml
3. After a while if we check the logs of those two nodes we should find in the node started with redundancy level set to 1 
the following info:
**"main or lower level redundancy machines are active   node redundancy level = 1"**
and in the node started with redundancy level set to 2 the following info:
**"main or lower level redundancy machines are active   node redundancy level = 2"**
4. After a while close the original instance and wait for a few minutes ~2. Now if we check the logs again we should find 
the following info:
**"main or lower level redundancy machines are inactive   node redundancy level = 1"** for the first node
**"main or lower level redundancy machines are active   node redundancy level = 2"** for the second node
5. After a while close the first redundancy node and after a few minutes ~2 check the logs of the second node and we should find the following info:
**"main or lower level redundancy machines are inactive   node redundancy level = 2"**
6. If after a while the original instance is restarted then after a few minutes ~2 we should notice the following info in the second node logs:
**"main or lower level redundancy machines are active   node redundancy level = 2"**
If we restarted both, the original instance and the first redundancy node, also in the logs of the first redundancy node we should find the following info:
**"main or lower level redundancy machines are active   node redundancy level = 1"**


cmd/node/main.go
```diff
@@ -372,6 +373,13 @@ VERSION:
 		Name:  "import-db-no-sig-check",
 		Usage: "This flag, if set, will cause the signature checks on headers to be skipped. Can be used only if the import-db was previously set",
 	}
+
+	// redundancyLevel defines a flag that specifies the level of redundancy for the current instance (0 = master)
```
iulianpascalau, line 377: To be "politically correct", let's replace `master` with `main` just about everywhere.

cmd/node/main.go
```diff
@@ -372,6 +373,13 @@ VERSION:
...
+	redundancyLevel = cli.Uint64Flag{
+		Name:  "redundancy-level",
+		Usage: "This flag specifies the level of redundancy used by the current instance for the node (0 = master, 1 = first slave, 2 = second slave, etc.)",
```
iulianpascalau, line 380: let's replace 'slave' with `backup` for the above mentioned reason.

consensus/spos/worker.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/redundancy"
```
iulianpascalau, line 8: go imports. Maybe set your IDE to show these as warnings?

node/node.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/redundancy"
```
iulianpascalau, line 9: go imports

node/options.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/redundancy"
```
iulianpascalau, line 7: go imports

redundancy/interface.go
```diff
@@ -0,0 +1,20 @@
+package redundancy
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+)
+
+// NodeRedundancyHandler provides functionality to handle the redundancy mechanism for a node
+type NodeRedundancyHandler interface {
```
iulianpascalau, line 8: Why do we have the interface here? Should have been wherever we consume the interface. You might redefine-it in all packages used (consensus, node....)

redundancy/interface.go
```diff
@@ -0,0 +1,20 @@
...
+	IsRedundancyNode() bool
+	IsMasterMachineActive() bool
+	AdjustInactivityIfNeeded(selfPubKey string, consensusPubKeys []string, roundIndex int64)
+	ResetInactivityIfNeeded(selfPubKey string, consensusMsgPubKey string, consensusMsgPeerID core.PeerID)
+	IsInterfaceNil() bool
+}
+
+// P2PMessenger defines a subset of the p2p.Messenger interface
+type P2PMessenger interface {
```
iulianpascalau, line 17: This is ok to be defined here as this package consumes the p2p messenger here.

redundancy/redundancy.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"sync"
+)
+
+var log = logger.GetOrCreate("redundancy")
+
+// maxRoundsOfInactivityAccepted defines the maximum rounds of inactivity accepted, after which the master or lower
+// level redundancy machines will be considered inactive
+const maxRoundsOfInactivityAccepted = 5
```
iulianpascalau, line 14: why is this so lower? In the discussed specs we wanted to disable the automatic block signing by defining a very high value for this.

cmd/node/main.go
```diff
@@ -372,6 +373,13 @@ VERSION:
...
+		Value: 0,
+	}
```
iulianpascalau, line 382: This PR does not solve the shuffle out mechanism for the redundancy nodes. Without this shuffle out this mechanism won't work.

iulianpascalau (changes requested):

cmd/node/main.go
SebastianMarian: It is solved as the node will run with the same ,pem so it will do the shuffle out mechanism

redundancy/redundancy.go
SebastianMarian: This constant is multiplied with the level given by the flag, so if we want to turn off this mechanism we should run the node with the level set to some of 100000 for example. This constant actually will define the difference from two backup machines by multiply the given level by 5

redundancy/interface.go
SebastianMarian: Yep, will move there

node/options.go
SebastianMarian: done

node/node.go
SebastianMarian: done

consensus/spos/worker.go
SebastianMarian: done

cmd/node/main.go
SebastianMarian: done

cmd/node/main.go
SebastianMarian: done

redundancy/redundancy.go
iulianpascalau: Right, didn't saw that. So, in order to disable automatic signing, we just have to put this to a large value. :+1:

cmd/node/main.go
iulianpascalau: Right, I thought the backup would have be ran as pure observers. By reusing the same key on multiple machines we simplified the process and impacted the code as minimum as we could get. :+1:

cmd/node/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"github.com/ElrondNetwork/elrond-go/redundancy"
```
iulianpascalau, line 6: forgotten go import

redundancy/redundancy.go
```diff
@@ -0,0 +1,100 @@
+package redundancy
+
+import (
+	logger "github.com/ElrondNetwork/elrond-go-logger"
```
iulianpascalau, line 4: forgotten go import

redundancy/redundnacy_test.go
```diff
@@ -0,0 +1,167 @@
+package redundancy_test
+
+import (
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/redundancy"
+	"github.com/ElrondNetwork/elrond-go/redundancy/mock"
+	"github.com/stretchr/testify/assert"
+	"testing"
```
iulianpascalau, line 8: forgotten go import

SebastianMarian: done

redundancy/redundancy.go
SebastianMarian: done

cmd/node/main.go
SebastianMarian: done

sasurobert (dismissed):

cmd/node/main.go
```diff
@@ -707,6 +707,10 @@ func startNode(ctx *cli.Context, log logger.Logger, version string) error {
 		preferencesConfig.Preferences.Identity = ctx.GlobalString(identityFlagName.Name)
 	}
 
+	if ctx.IsSet(redundancyLevel.Name) {
+		preferencesConfig.Preferences.RedundancyLevel = ctx.GlobalInt64(redundancyLevel.Name)
```
iulianpascalau, line 711: :+1:

iulianpascalau (approved):

sasurobert (approved):

stefanandy (approved):
System Test Passed

PR #2827, opened 2021-02-18
En 9027 transaction simulate query param

miiu96:
Extend API route `/transaction/simulate` with an extra query parameter `checkSignature` in order to can simulate a transaction without the correct signature.

By default, if the user uses this API route without this parameter the transaction signature **will be verified**. 

In order to ignore the transaction signature user have to specify to ignore it (`?checkSignature=false`).


node/node.go
```diff
@@ -998,14 +999,19 @@ func (n *Node) commonTransactionValidation(tx *transaction.Transaction) (process
 	currentEpoch := n.epochStartTrigger.Epoch()
 	enableSignWithTxHash := currentEpoch >= n.enableSignTxWithHashEpoch
 
+	txSingleSigner := n.txSingleSigner
+	if bypassSignature {
+		txSingleSigner = &disabledSig.DisabledSingleSig{}
```
iulianpascalau, line 1004: The power of polymorphism :+1:

api/transaction/routes.go
```diff
@@ -33,7 +33,7 @@ type FacadeHandler interface {
 	CreateTransaction(nonce uint64, value string, receiver string, receiverUsername []byte, sender string, senderUsername []byte, gasPrice uint64,
 		gasLimit uint64, data []byte, signatureHex string, chainID string, version uint32, options uint32) (*transaction.Transaction, []byte, error)
 	ValidateTransaction(tx *transaction.Transaction) error
-	ValidateTransactionForSimulation(tx *transaction.Transaction) error
+	ValidateTransactionForSimulation(tx *transaction.Transaction, bypassSignature bool) error
```
iulianpascalau, line 36: If I would to blame something in this PR I would blame the fact that now we have "bypass signature = true" which means "do not check the signature". I would have made the flag, query param and all variable names sound more like "checkSignature" (with default to true if not specified). In this way when we pass true I would automatically think that the signature will be checked. "bypass signature = false" will require that everyone that will read the code will have to do a mental double negation operation which will make the code a little more hard to understand.

miiu96: Query parameter `bypassSignature` was renamed, now parameter is `checkSignature` and is by default true.

api/transaction/routes.go
```diff
@@ -538,18 +541,18 @@ func ComputeTransactionGasLimit(c *gin.Context) {
 }
 
 func getQueryParamWithResults(c *gin.Context) (bool, error) {
-	withResultsStr := c.Request.URL.Query().Get("withResults")
+	withResultsStr := c.Request.URL.Query().Get(queryParamWithResults)
```
iulianpascalau, line 544: :+1:

node/node.go
```diff
@@ -999,7 +999,7 @@ func (n *Node) commonTransactionValidation(tx *transaction.Transaction, bypassSi
 	enableSignWithTxHash := currentEpoch >= n.enableSignTxWithHashEpoch
 
 	txSingleSigner := n.txSingleSigner
-	if bypassSignature {
+	if checkSignature {
```
iulianpascalau, line 1002: `if !checkSignature`

api/transaction/routes.go
```diff
@@ -525,10 +541,19 @@ func ComputeTransactionGasLimit(c *gin.Context) {
...
 		return false, nil
 	}
 
 	return strconv.ParseBool(withResultsStr)
 }
+
+func getQueryParameterCheckSignature(c *gin.Context) (bool, error) {
+	bypassSignatureStr := c.Request.URL.Query().Get(queryParamCheckSignature)
+	if bypassSignatureStr == "" {
+		return true, nil
```
bogdan-rosianu, line 555: is this intended behavior?
I think that now if the query param is missing, then it would be `true` as default

node/node.go
miiu96: good catch.

api/transaction/routes.go
bogdan-rosianu: please refactor: use everywhere in this PR `bypassSignatureCheck` instead of `checkSignature`. So adjust the logic. if `bypassSignatureCheck` then use a disabled sig verifier, and so on. It is misleading right now. Checking signature is the default option, so bypassing it should be activated by a bool flag..

facade/nodeFacade.go
```diff
@@ -311,8 +311,8 @@ func (nf *nodeFacade) ValidateTransaction(tx *transaction.Transaction) error {
 }
 
 // ValidateTransactionForSimulation will validate a transaction for the simulation process
-func (nf *nodeFacade) ValidateTransactionForSimulation(tx *transaction.Transaction) error {
-	return nf.node.ValidateTransactionForSimulation(tx)
+func (nf *nodeFacade) ValidateTransactionForSimulation(tx *transaction.Transaction, checkSignature bool) error {
+	return nf.node.ValidateTransactionForSimulation(tx, checkSignature)
```
bogdan-rosianu, line 315: `bypassSignature` instead of `checkSignature`

node/node.go
```diff
@@ -960,8 +961,8 @@ func (n *Node) ValidateTransaction(tx *transaction.Transaction) error {
 }
 
 // ValidateTransactionForSimulation will validate a transaction for use in transaction simulation process
-func (n *Node) ValidateTransactionForSimulation(tx *transaction.Transaction) error {
-	txValidator, intTx, err := n.commonTransactionValidation(tx)
+func (n *Node) ValidateTransactionForSimulation(tx *transaction.Transaction, checkSignature bool) error {
```
bogdan-rosianu, line 964: `bypassSignature` instead of `checkSignature`

node/node.go
```diff
@@ -975,7 +976,7 @@ func (n *Node) ValidateTransactionForSimulation(tx *transaction.Transaction) err
 	return err
 }
 
-func (n *Node) commonTransactionValidation(tx *transaction.Transaction) (process.TxValidator, process.TxValidatorHandler, error) {
+func (n *Node) commonTransactionValidation(tx *transaction.Transaction, checkSignature bool) (process.TxValidator, process.TxValidatorHandler, error) {
```
bogdan-rosianu, line 979: `bypassSignature` instead of `checkSignature`

node/node.go
```diff
@@ -997,14 +998,19 @@ func (n *Node) commonTransactionValidation(tx *transaction.Transaction) (process
 	currentEpoch := n.epochStartTrigger.Epoch()
 	enableSignWithTxHash := currentEpoch >= n.enableSignTxWithHashEpoch
 
+	txSingleSigner := n.txSingleSigner
+	if !checkSignature {
```
bogdan-rosianu, line 1002: `bypassSignature` instead of `checkSignature`
so:
```
if bypassSignature {
   // use disabled signer
}
```

bogdan-rosianu (changes requested):

node/node.go
bogdan-rosianu: That's why I said it is misleading. Should definitely be renamed to `bypassSignatureCheck`

bogdan-rosianu (approved):

iulianpascalau (dismissed):

iulianpascalau (approved):

bogdan-rosianu (approved):

stefanandy (approved):
System Test Passed

PR #2832, opened 2021-02-21
Add claimRewards value to TotalCumulatedRewards

mihaieremia:
- Fixed the situation when the TotalCumulatedRewards rewards did not contain the `claimRewards` value


sasurobert (approved):

iulianpascalau (approved):

iulianpascalau:
I do not think this is backwards compatible with the current testnet (no worries, we will just have new one). cc @sasurobert

PR #2837, opened 2021-02-22
Fix unStake from waiting at end of epoch

sasurobert:
fixed the scenario of unstaking a node at end of epoch from the waiting list

### DIFF
--- a/epochStart/metachain/systemSCs.go
+++ b/epochStart/metachain/systemSCs.go
@@ -302,7 +302,7 @@ func (s *systemSCProcessor) unStakeNodesWithNotEnoughFunds(
 
 		validatorInfo := getValidatorInfoWithBLSKey(validatorInfos, blsKey)
 		if validatorInfo == nil {
-			log.Debug("unStaked nodes which was in waiting only", "blsKey", blsKey)
+			log.Debug("unStaked node which was in additional queue", "blsKey", blsKey)
 			continue
 		}
 
--- a/epochStart/metachain/systemSCs.go
+++ b/epochStart/metachain/systemSCs.go
@@ -342,8 +342,8 @@ func (s *systemSCProcessor) unStakeOneNode(blsKey []byte, epoch uint32) error {
 		return err
 	}
 
-	account, errLoad := s.peerAccountsDB.GetExistingAccount(blsKey)
-	if errLoad != nil {
+	account, errExists := s.peerAccountsDB.GetExistingAccount(blsKey)
+	if errExists != nil {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


epochStart/metachain/systemSCs.go
```diff
@@ -340,9 +342,14 @@ func (s *systemSCProcessor) unStakeOneNode(blsKey []byte, epoch uint32) error {
 		return err
 	}
 
-	peerAccount, errGet := s.getPeerAccount(blsKey)
-	if errGet != nil {
-		return errGet
+	account, errLoad := s.peerAccountsDB.GetExistingAccount(blsKey)
+	if errLoad != nil {
+		return nil
```
iulianpascalau, line 347: log.Debug the error?

sasurobert: this is not an error. but if account does not exist there is no need to unStake

sasurobert: renamed to errExist it is used in multiple places

epochStart/metachain/systemSCs.go
```diff
@@ -294,14 +294,16 @@ func (s *systemSCProcessor) unStakeNodesWithNotEnoughFunds(
...
+		log.Debug("unStake at end of epoch for node", "blsKey", blsKey)
 		err = s.unStakeOneNode(blsKey, epoch)
 		if err != nil {
 			return 0, err
 		}
 
 		validatorInfo := getValidatorInfoWithBLSKey(validatorInfos, blsKey)
 		if validatorInfo == nil {
-			return 0, epochStart.ErrNilValidatorInfo
+			log.Debug("unStaked nodes which was in waiting only", "blsKey", blsKey)
```
SebastianMarian, line 305: "unStake node which was" or "unStake nodes which were"

vm/systemSmartContracts/validator.go
```diff
@@ -1730,7 +1730,6 @@ func (v *validatorSC) getTotalStakedTopUpStakedBlsKeys(args *vmcommon.ContractCa
 	topUp := big.NewInt(0).Set(registrationData.TotalStakeValue)
 	topUp.Sub(topUp, stakeForNodes)
 	if topUp.Cmp(zero) < 0 {
-		log.Warn("topup value is less than 0")
```
SebastianMarian, line 1733: Maybe log.Debug instead deletion

sasurobert: no need for this log. as we can see the results of this in several other places.

epochStart/metachain/systemSCs.go
sasurobert: pushed

iulianpascalau (approved):

SebastianMarian (approved):

valentincraciun99 (approved):
system testing passed

PR #2872, opened 2021-03-02
Gas schedule error when loading fails

iulianpascalau:
- added new gas schedule + version
- added prints when the gas schedule fails to load
- fixed local scripts


vm/factory/systemSCFactory.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
+var log = logger.GetOrCreate("core/forking")
```
sasurobert, line 17: Is this the correct package?

vm/factory/systemSCFactory.go
```diff
@@ -148,6 +152,8 @@ func (scf *systemSCFactory) GasScheduleChange(gasSchedule map[string]map[string]
```
sasurobert, line 152: Log error on line 150 as well

sasurobert (changes requested):

vm/factory/systemSCFactory.go
iulianpascalau: :man_facepalming:

iulianpascalau: done

vm/factory/systemSCFactory.go
iulianpascalau: done

sasurobert (approved):

raduchis (approved):

valentincraciun99 (approved):
system testing passed

PR #2874, opened 2021-03-02
Fix variables.sh

catalinnnn:
Other scripts expect this variable to be defined, best is to define and initialize it with the default value of 0.2
Without this, ./config.sh won't work without manually setting that variable


iulianpascalau:
already fixed in #2872. Thanks for pointing this out.

PR #2908, opened 2021-03-16
SaveAccount without LoadAccount first

BeniaminDrasovean:
Add the dataTrie to tries cacher if SaveAccount is called, but the dataTrie is not present in the hasher.


SebastianMarian (dismissed):

data/state/accountsDB.go
```diff
@@ -442,6 +442,10 @@ func (adb *AccountsDB) saveDataTrie(accountHandler baseAccountHandler) error {
 		"new root hash", accountHandler.GetRootHash(),
 	)
 
+	if adb.dataTries.Get(accountHandler.AddressBytes()) == nil {
```
SebastianMarian, line 445: check.IfNil ?

SebastianMarian: if Get returns an interface

SebastianMarian, line 445: check.IfNil if Get returns an interface

SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System Test passed

PR #2923, opened 2021-03-23
Split miniblocks based on type txs scs

SebastianMarian:
* Implemented splitting mechanism for mini blocks "from me" based on the type of transactions they contain inside (non smart contract calls txs or smart contract calls txs)
* Added ScheduledMiniBlocksEbableEpoch flag
* Sketched the implementation for adding scheduled mini blocks in the proposed block

**How to be tested**

Start a normal system test with txs and sc calls
In epoch 5 scheduled mini blocks mechanism will be activated.
After epoch 5 will passed we should check the log files (some messages like: "createScheduledMiniBlocks has been started", "createScheduledMiniBlocks has been finished" should be found)


process/block/preprocess/transactionsV2.go
```diff
@@ -0,0 +1,491 @@
...
+	numTxsBad                          int
+	numTxsSkipped                      int
+	numTxsFailed                       int
+	numTxsWithInitialBalanceConsumed   int
+	numCrossShardScCallsOrSpecialTxs   int
+	totalTimeUsedForProcess            time.Duration
+	totalTimeUsedForComputeGasConsumed time.Duration
+}
+
+func (txs *transactions) createAndProcessMiniBlocksFromMeV2(
```
sasurobert, line 35: this function is too long. can we split again ?

sasurobert (changes requested):

process/block/preprocess/transactionsV2.go
SebastianMarian: Will try to split in the latest PR, as if it will be done in this one it will affect/conflict with all the others which started with this base

sasurobert (approved):

PR #2934, opened 2021-03-29
NTP and p2p fixes

iulianpascalau:
- made the ntp syncer ignore out of bound clock offset values
- made the p2p messenger less stringent about discarding too new messages


ntp/syncTime.go
```diff
@@ -195,17 +197,19 @@ func (s *syncTime) sync() {
 
 	clockOffsetsWithoutEdges := s.getClockOffsetsWithoutEdges(clockOffsets)
 	clockOffsetHarmonicMean := s.getHarmonicMean(clockOffsetsWithoutEdges)
+	if clockOffsetHarmonicMean < -outOfBoundsDuration || clockOffsetHarmonicMean > outOfBoundsDuration {
```
sasurobert, line 200: why don't you use abs ?

p2p/libp2p/netMessenger.go
```diff
@@ -52,7 +52,7 @@ const durationCheckConnections = time.Second
 const refreshPeersOnTopic = time.Second * 3
 const ttlPeersOnTopic = time.Second * 10
 const pubsubTimeCacheDuration = 10 * time.Minute
-const acceptMessagesInAdvanceDuration = 5 * time.Second //we are accepting the messages with timestamp in the future only fot this delta
+const acceptMessagesInAdvanceDuration = 20 * time.Second //we are accepting the messages with timestamp in the future only fot this delta
```
sasurobert, line 55: only *for this delta - not fot

sasurobert (changes requested):

SebastianMarian (approved):

ntp/syncTime.go
iulianpascalau: done, however, I have created a function for Abs because math.Abs operates on floats.

sasurobert (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed. 
Mention: Some of the cloud machines got stuck because of the low processing capabilities.

PR #2946, opened 2021-03-31
Fix trie sync OOM problems

iulianpascalau:
- fix trie sync OOM problems

### DIFF
--- a/data/syncer/userAccountsSyncer.go
+++ b/data/syncer/userAccountsSyncer.go
@@ -141,12 +141,12 @@ func (u *userAccountsSyncer) syncAccountDataTries(rootHashes [][]byte, ssh data.
 
 func (u *userAccountsSyncer) syncDataTrie(rootHash []byte, ssh data.SyncStatisticsHandler, ctx context.Context) error {
 	u.throttler.StartProcessing()
+	defer u.throttler.EndProcessing()
 
 	u.syncerMutex.Lock()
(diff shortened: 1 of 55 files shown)


update/sync/syncAccountsDBs.go
```diff
@@ -155,7 +155,8 @@ func (st *syncAccountsDBs) syncAccountsOfType(accountType genesis.Type, trieID s
 		return err
 	}
 
-	st.setTries(shardId, accAdapterIdentifier, rootHash, accountsDBSyncer.GetSyncedTries())
+	//TODO(iulian) fix here in this PR
```
sasurobert, line 158: this whole thing works without this :D maybe from accountsDBSyncer you can take those.

iulianpascalau: will test :+1:

sasurobert (dismissed):

sasurobert (dismissed):

data/syncer/userAccountsSyncer.go
```diff
@@ -157,7 +156,9 @@ func (u *userAccountsSyncer) syncDataTrie(rootHash []byte, ssh data.SyncStatisti
 		return err
```
SebastianMarian, line 156: Not need to call u.throttler.EndProcessing() on return err cases in this method ?

sasurobert (approved):

SebastianMarian (approved):

PR #2949, opened 2021-03-31
Formatted NFT identifier in api response

bogdan-rosianu:
Formatted the NFT identifier in `/address/:address/esdt` endpoint. 
Now, if a token is regular esdt, the key for it would follow the syntax `TICKER-3e3e3e` and if it the token is a NFT, then it would follow the syntax `TICKER-3e3e3e-02` where `02` is the hex encoded nonce.

Before this PR, the token identifier for NFTs would have resulted in non readable characters

### DIFF
--- a/cmd/node/config/api.toml
+++ b/cmd/node/config/api.toml
@@ -49,10 +49,7 @@
         { Name = "/:address/esdt/:tokenIdentifier", Open = true },
 
         # /address/:address/nft/:tokenIdentifier/nonce/:nonce will return data of an nft esdt token for a given account, tokenID and nonce
-        { Name = "/:address/nft/:tokenIdentifier/nonce/:nonce", Open = true },
-
-        # /address/:address/completeEsdts will return the list of esdt tokens and all the underlying data for a given account
-        { Name = "/:address/completeEsdts", Open = true }
+        { Name = "/:address/nft/:tokenIdentifier/nonce/:nonce", Open = true }
 	]
 
 [APIPackages.hardfork]
--- a/node/node.go
+++ b/node/node.go
@@ -60,6 +60,8 @@ import (
 // SendTransactionsPipe is the pipe used for sending new transactions
 const SendTransactionsPipe = "send transactions pipe"
 
+const esdtTickerNumChars = 6
+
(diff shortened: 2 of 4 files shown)


sasurobert (approved):
Maybe a unit test? :)

sasurobert (approved):

cmd/node/config/api.toml
```diff
@@ -49,10 +49,7 @@
         { Name = "/:address/esdt/:tokenIdentifier", Open = true },
 
         # /address/:address/nft/:tokenIdentifier/nonce/:nonce will return data of an nft esdt token for a given account, tokenID and nonce
```
miiu96, line 51: `data of a nft` instead `data of an nft`

node/node.go
```diff
@@ -60,6 +60,8 @@ import (
 // SendTransactionsPipe is the pipe used for sending new transactions
 const SendTransactionsPipe = "send transactions pipe"
 
+const esdtTickerNumChars = 6
```
miiu96, line 63: group these constants 
const(
 ..
 ..
)

node/node_test.go
```diff
@@ -390,6 +390,73 @@ func TestNode_GetAllESDTTokens(t *testing.T) {
 	assert.Equal(t, esdtData, value[esdtToken])
 }
 
+func TestNode_GetAllESDTTokensShouldReturnEsdtAndFormattedNft(t *testing.T) {
+	acc, _ := state.NewUserAccount([]byte("newaddress"))
```
miiu96, line 394: t.Parallel ?

cmd/node/config/api.toml
bogdan-rosianu: done

node/node.go
bogdan-rosianu: done

node/node_test.go
bogdan-rosianu: no test in this file contains `t.Parallel()` so I would leave it as it is

sasurobert (dismissed):

sasurobert (approved):

miiu96 (approved):

PR #2954, opened 2021-04-01
Fix stuck situation created by too many scrs generated in one single mini block

SebastianMarian:
* Fixed out of gas situation created by scrs generated in destination shard and added in one single mini block and sent back to the source shard

**How to be tested**

Repeat scenario in which there are unstaked many nodes, forcing meta to create a mini block with too many scrs, which would lead to a stuck situation in the initial source shard

After the scenario will be finished, search after message "basePostProcessor.splitMiniBlocksIfNeeded: gas limit exceeded". If this message will be found that means that a stuck situation was avoided, by splitting the resulted scrs mini block in more mini blocks each of them containing a limited number of txs, whose gas used is not exceed the max gas limit accepted per one mini block.

### DIFF
--- a/process/block/postprocess/basePostProcess.go
+++ b/process/block/postprocess/basePostProcess.go
@@ -1,7 +1,6 @@
 package postprocess
 
 import (
-	"bytes"
 	"sync"
 
 	"github.com/ElrondNetwork/elrond-go-logger"
@@ -41,6 +40,7 @@ type basePostProcessor struct {
 	interResultsForBlock    map[string]*txInfo
 	mapTxToResult           map[string][]string
 	intraShardMiniBlock     *block.MiniBlock
+	economicsFee            process.FeeHandler
 }
 
 // SaveCurrentIntermediateTxToStorage saves all current intermediate results to the provided storage unit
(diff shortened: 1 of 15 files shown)


process/block/postprocess/oneMBPostProcessor.go
```diff
@@ -113,9 +118,14 @@ func (opp *oneMBPostProcessor) CreateAllInterMiniBlocks() []*block.MiniBlock {
 // VerifyInterMiniBlocks verifies if the receipts/bad transactions added to the block are valid
 func (opp *oneMBPostProcessor) VerifyInterMiniBlocks(body *block.Body) error {
 	scrMbs := opp.CreateAllInterMiniBlocks()
-	createdMapMbs := make(map[uint32]*block.MiniBlock)
+	createdMapMbs := make(map[uint32][]*block.MiniBlock)
 	for _, mb := range scrMbs {
-		createdMapMbs[mb.ReceiverShardID] = mb
+		_, ok := createdMapMbs[mb.ReceiverShardID]
```
sasurobert, line 123: this is duplicated code with the one in intermediateResults.go - you could move this map creation to a base func

process/block/postprocess/basePostProcess.go
```diff
@@ -171,3 +179,54 @@ func (bpp *basePostProcessor) RemoveProcessedResultsFor(txHashes [][]byte) {
...
+
+	for _, miniBlock := range miniBlocks {
+		receiverShardID := miniBlock.ReceiverShardID
+		currentMiniBlock := createEmptyMiniBlock(miniBlock)
+		mapGasLimitInReceiverShard[receiverShardID] = 0
+
+		for _, txHash := range miniBlock.TxHashes {
+			interResult, ok := bpp.interResultsForBlock[string(txHash)]
+			if !ok {
+				log.Warn("basePostProcessor.splitMiniBlocksIfNeeded: missing tx", "hash", txHash)
```
sasurobert, line 195: this should be a critical bug, not ?

sasurobert (changes requested):

process/block/postprocess/basePostProcess.go
```diff
@@ -171,3 +179,54 @@ func (bpp *basePostProcessor) RemoveProcessedResultsFor(txHashes [][]byte) {
...
+func (bpp *basePostProcessor) splitMiniBlocksIfNeeded(miniBlocks []*block.MiniBlock) []*block.MiniBlock {
+	splitMiniBlocks := make([]*block.MiniBlock, 0)
+	mapGasLimitInReceiverShard := make(map[uint32]uint64)
+
+	for _, miniBlock := range miniBlocks {
+		receiverShardID := miniBlock.ReceiverShardID
+		currentMiniBlock := createEmptyMiniBlock(miniBlock)
+		mapGasLimitInReceiverShard[receiverShardID] = 0
+
+		for _, txHash := range miniBlock.TxHashes {
```
AdoAdoAdo, line 192: can we extract this inner loop into a separate function?

process/block/postprocess/oneMBPostProcessor.go
SebastianMarian: done

process/block/postprocess/basePostProcess.go
SebastianMarian: Yes, this will be caught up in VerifyInterMiniBlocks when the hash of the each created mbs will not matched with the hash of each received mini block. Because CreateAllInterMiniBlocks does not return error, and this method is called from there, it would have been useless to return error from here just to log it there

sasurobert: ok. can remain

process/block/postprocess/basePostProcess.go
SebastianMarian: done

AdoAdoAdo (approved):

sasurobert (approved):

gabi-vuls (approved):
System test passed.

AdoAdoAdo (approved):

stefanandy (approved):
System Test Passed

PR #2960, opened 2021-04-06
Different default shard as observer

bogdan-rosianu:
Before this fix, if no destination shard as observer was set (so `DestinationShardAsObserver = "disabled"` in `prefs.toml`) then the "chosen" shard would have been `0`. In order to avoid this, a random (or at least one that isn't always `0`) shard should be generated.

Because the default `0` value was set in multiple instances, a deterministic algorithm for the same entry was needed. So, the last byte from the public key is divided by the number of shards, assigning the reminder to the shard ID.

### DIFF
--- a/core/converters.go
+++ b/core/converters.go
@@ -201,8 +201,8 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
 	return uint32(val), err
 }
 
-// AssignShardForPubKeyWhenNotSpecified will return the same shard ID when the same seed source is set to the randomizer
-// This function sums all the bytes from the public key and based on a modulo operation it will return a shard ID
+// AssignShardForPubKeyWhenNotSpecified will return the same shard ID when it is called with the same parameters
+// This function fetched the last byte of the public key and based on a modulo operation it will return a shard ID
 func AssignShardForPubKeyWhenNotSpecified(pubKey []byte, numShards uint32) uint32 {
 	if len(pubKey) == 0 {
 		return 0 // should never happen
--- a/core/converters.go
+++ b/core/converters.go
@@ -204,14 +204,14 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
 // AssignShardForPubKeyWhenNotSpecified will return the same shard ID when the same seed source is set to the randomizer
 // This function sums all the bytes from the public key and based on a modulo operation it will return a shard ID
 func AssignShardForPubKeyWhenNotSpecified(pubKey []byte, numShards uint32) uint32 {
-	sum := 0
-	for _, b := range pubKey {
-		sum += int(b)
+	if len(pubKey) == 0 {
+		return 0 // should never happen
 	}
 
+	lastByte := pubKey[len(pubKey)-1]
 	numShardsIncludingMeta := numShards + 1
 
-	randomShardID := uint32(sum) % numShardsIncludingMeta
+	randomShardID := uint32(lastByte) % numShardsIncludingMeta
 	if randomShardID == numShards {
 		randomShardID = MetachainShardId
 	}
--- a/core/converters.go
+++ b/core/converters.go
@@ -200,3 +200,22 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
 
 	return uint32(val), err
 }
+
+// AssignShardForPubKeyWhenNotSpecified will return the same shard ID when the same seed source is set to the randomizer
+// This function sums all the bytes from the public key and based on a modulo operation it will return a shard ID
+func AssignShardForPubKeyWhenNotSpecified(pubKey []byte, numShards uint32) uint32 {
+	sum := 0
+	for _, b := range pubKey {
+		sum += int(b)
+	}
+
+	numShardsIncludingMeta := numShards + 1
+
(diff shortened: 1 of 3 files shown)


core/converters.go
```diff
@@ -200,3 +200,22 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
...
+func AssignShardForPubKeyWhenNotSpecified(pubKey []byte, numShards uint32) uint32 {
+	sum := 0
+	for _, b := range pubKey {
+		sum += int(b)
+	}
+
+	numShardsIncludingMeta := numShards + 1
+
+	randomShardID := uint32(sum) % numShardsIncludingMeta
+
```
iulianpascalau, line 215: remove empty line?

factory/shardingFactory.go
```diff
@@ -38,7 +38,13 @@ func CreateShardCoordinator(
 			return nil, "", err
 		}
 		if selfShardId == core.DisabledShardIDAsObserver {
-			selfShardId = uint32(0)
+			pubKeyBytes, err := pubKey.ToByteArray()
+			if err != nil {
+				log.Warn("cannot get public key bytes", "error", err)
```
iulianpascalau, line 43: I think the error escalates anyway. Might wrap the error for more context.

factory/shardingFactory.go
```diff
@@ -100,7 +106,13 @@ func CreateNodesCoordinator(
 		return nil, err
 	}
 	if shardIDAsObserver == core.DisabledShardIDAsObserver {
-		shardIDAsObserver = uint32(0)
+		pubKeyBytes, err := pubKey.ToByteArray()
+		if err != nil {
+			log.Warn("cannot get public key bytes", "error", err)
```
iulianpascalau, line 111: I think the error escalates anyway. Might wrap the error for more context.

core/converters.go
bogdan-rosianu: removed

factory/shardingFactory.go
bogdan-rosianu: done

factory/shardingFactory.go
bogdan-rosianu: done

iulianpascalau (approved):

factory/shardingFactory.go
```diff
@@ -38,7 +38,12 @@ func CreateShardCoordinator(
 			return nil, "", err
 		}
 		if selfShardId == core.DisabledShardIDAsObserver {
-			selfShardId = uint32(0)
+			pubKeyBytes, err := pubKey.ToByteArray()
```
raduchis, line 41: is line 32 correct? if err == sharding.ErrPublicKeyNotFoundInGenesis -> just genesis nodes?

raduchis (approved):

core/converters.go
```diff
@@ -200,3 +200,21 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
...
+// This function sums all the bytes from the public key and based on a modulo operation it will return a shard ID
+func AssignShardForPubKeyWhenNotSpecified(pubKey []byte, numShards uint32) uint32 {
+	sum := 0
+	for _, b := range pubKey {
+		sum += int(b)
+	}
+
+	numShardsIncludingMeta := numShards + 1
+
+	randomShardID := uint32(sum) % numShardsIncludingMeta
```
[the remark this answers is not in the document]
bogdan-rosianu: you're right. done

factory/shardingFactory.go
bogdan-rosianu: If you refer to the fact that a node can belong to a shard even if it is not in genesis (for example a staked node added after genesis), it is ok, because this is the "genesis shard coordinator". After this, the real shard (coordinator) will be calculated inside the `epoch bootstrapper`

iulianpascalau (dismissed):

factory/shardingFactory.go
raduchis: ok!

core/converters.go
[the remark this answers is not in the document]
raduchis: ok!

core/converters.go
```diff
@@ -200,3 +200,21 @@ func ProcessDestinationShardAsObserver(destinationShardIdAsObserver string) (uin
 
 	return uint32(val), err
 }
+
+// AssignShardForPubKeyWhenNotSpecified will return the same shard ID when the same seed source is set to the randomizer
+// This function sums all the bytes from the public key and based on a modulo operation it will return a shard ID
```
raduchis, line 205: not correct any more

bogdan-rosianu: updated

raduchis (dismissed):

core/converters_test.go
```diff
@@ -273,3 +275,36 @@ func TestAssignShardForPubKeyWhenNotSpecifiedShouldReturnSameShardForSameKey(t *
 
 	require.Equal(t, result0, result1)
 }
+
+func TestShardAssignment(t *testing.T) {
```
iulianpascalau, line 279: Not quite a test :D can remain as it is

iulianpascalau (approved):

raduchis (approved):

miiu96 (approved):

PR #2966, opened 2021-04-07
Feat/trie sync fixes

iulianpascalau:
Sync trie fixes so the node will not raise an Out Of Memory exception during bootstrap
Added a new variant of trie syncer which should perform better


sasurobert (approved):

gabi-vuls (approved):
System test passed.

raduchis (approved):

PR #2968, opened 2021-04-07
Test exec on dest with async call

sasurobert:
Test and new arwen


process/smartContract/process.go
```diff
@@ -2294,7 +2294,7 @@ func (sc *scProcessor) EpochConfirmed(epoch uint32) {
 	sc.flagReturnDataToLastTransfer.Toggle(epoch > sc.returnDataToLastTransferEnableEpoch)
 	log.Debug("scProcessor: return data to last transfer", "enabled", sc.flagReturnDataToLastTransfer.IsSet())
 
-	sc.flagSenderInOutTransfer.Toggle(epoch > sc.senderInOutTransferEnableEpoch)
+	sc.flagSenderInOutTransfer.Toggle(epoch >= sc.senderInOutTransferEnableEpoch)
```
bogdan-rosianu, line 2297: should L2294 and L2291 also be changed also?

bogdan-rosianu (approved):

SebastianMarian (approved):

process/smartContract/process.go
sasurobert: no - that was already launched on mainnet.

PR #2980, opened 2021-04-09
Delete nil esdt data

sasurobert:
Delete nil esdt data

### DIFF
--- a/process/smartContract/builtInFunctions/esdtTransfer.go
+++ b/process/smartContract/builtInFunctions/esdtTransfer.go
@@ -265,12 +265,26 @@ func checkFrozeAndPause(
 	return nil
 }
 
+func arePropertiesEmpty(properties []byte) bool {
+	for _, property := range properties {
+		if property != 0 {
+			return false
+		}
+	}
+	return true
+}
+
 func saveESDTData(
 	userAcnt state.UserAccountHandler,
 	esdtData *esdt.ESDigitalToken,
 	key []byte,
 	marshalizer marshal.Marshalizer,
 ) error {
+	isValueZero := esdtData.Value.Cmp(zero) == 0
+	if isValueZero && arePropertiesEmpty(esdtData.Properties) {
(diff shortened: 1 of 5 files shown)


process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -265,12 +265,26 @@ func checkFrozeAndPause(
...
+
 func saveESDTData(
 	userAcnt state.UserAccountHandler,
 	esdtData *esdt.ESDigitalToken,
 	key []byte,
 	marshalizer marshal.Marshalizer,
 ) error {
+	isValueZero := esdtData.Value.Cmp(zero) == 0
+	if isValueZero && arePropertiesEmpty(esdtData.Properties) {
+		return userAcnt.DataTrieTracker().SaveKeyValue(key, nil)
```
bogdan-rosianu, line 285: will this remove the key also? or the key remains but with `null` value?

process/smartContract/builtInFunctions/esdtTransfer.go
```diff
@@ -265,12 +265,26 @@ func checkFrozeAndPause(
 	return nil
 }
 
+func arePropertiesEmpty(properties []byte) bool {
+	for _, property := range properties {
```
bogdan-rosianu, line 269: could use an early `if len(properties) > 0 { return false }`

bogdan-rosianu (approved):

iulianpascalau (approved):

process/smartContract/builtInFunctions/esdtTransfer.go
sasurobert: this removes the key.

process/smartContract/builtInFunctions/esdtTransfer.go
sasurobert: no. as if you do len(make([]byte, 10)) it returns 10

PR #2988, opened 2021-04-11
Fixed esdt owner

iulianpascalau:
- Fixed ESDT owner comparison in functions

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -571,7 +571,7 @@
 
 [EpochStartConfig]
     MinRoundsBetweenEpochs = 20
-    RoundsPerEpoch         = 200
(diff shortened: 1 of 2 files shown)


bogdan-rosianu (approved):
It is ok this way, but I see 2 ways of refactoring:
1. Why don't rename the `getEsdtConfig` to `updateEsdtConfig` so it will update the struct members' values and use struct members everywhere? the rest of the members will remain outdate (base issuing cost, min token name length, max token name length)
2. I see that these members are only used in contructor. If so, and we always read the new configuration from storage, why don't we remove the `baseIssueingCost` and `ownerAddress` from the struct and use the constructor parameters only in constructor

sasurobert (approved):

PR #2992, opened 2021-04-12
balanced destination shard as observer

bogdan-rosianu:
different destination shard as an observer when no shard is specified. same as #2960 but targeted to `master` branch

Testing procedure:
- start an observer by using this branch
- if the key is not in consensus and the node will start as observer, it should be routed to a "random" shard. The same key in multiple tries should be routed to the same shard.
- try with multiple keys in order to see the shard balancing


iulianpascalau (approved):

miiu96 (approved):

miiu96 (approved):

gabi-vuls (approved):
Ssytem test passed

PR #2994, opened 2021-04-12
Improve consensus in subround EndRound to avoid forks

SebastianMarian:
* Improve consensus in subround EndRound to avoid forks

**How to be tested**

**_CASE 1_**

Start a system test #allin with induced forks. After a while we should search and find this message in log files. 

_"doEndRoundJob: time is out -> cancel broadcasting final info and header"_

Also the hit rate should be much higher compared with other system tests ran with induced forks.

**_CASE 2_**

Start a system test #allin **WITHOUT** induced forks. After a while we should search in log files after this message: 

_"doEndRoundJob: time is out -> cancel broadcasting final info and header"_ . **THIS MESSAGE SHOULD NOT BE FOUND!**


raduchis (approved):

consensus/spos/bls/subroundEndRound.go
```diff
@@ -204,6 +204,14 @@ func (sr *subroundEndRound) doEndRoundJobByLeader() bool {
 
 	// broadcast section
 
+	//TODO: Move this if before L205 before committing it, just to keep the induced hard forks system test functionality
+	if sr.Rounder().Index() > sr.RoundIndex {
+		log.Debug("doEndRoundJob: time is out -> cancel broadcasting final info and header",
```
iulianpascalau, line 209: I do not think this will work as expected because the rounder did not manage to increment its index.

SebastianMarian: Good catch, I will fix it

iulianpascalau (dismissed):

sasurobert (approved):

consensus/spos/bls/subroundEndRound.go
AdoAdoAdo: can we add also some tolerance?
we should test with some heavy load and make sure this does not stop the producing of blocks.

raduchis (approved):

sasurobert (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #2995, opened 2021-04-12
Keep max num nodes limit

sasurobert:
Clean waiting list nodes when no nodes are qualified
Keep the maximum number of nodes property always


iulianpascalau (approved):

vm/systemSmartContracts/staking.go
```diff
@@ -1592,41 +1649,81 @@ func (s *stakingSC) stakeNodesFromQueue(args *vmcommon.ContractCallInput) vmcomm
...
 
-	mapCheckedOwners[string(owner)] = true
+	numActiveNodes := int64(0)
+	for _, blsKey := range validatorData.BlsPubKeys {
+		stakedData, errGet := s.getOrCreateRegisteredData(blsKey)
+		if errGet != nil {
+			return nil, errGet
+		}
+
+		if stakedData.Staked || stakedData.Waiting || stakedData.Jailed {
```
AdoAdoAdo, line 1710: why do we consider stakedData.Waiting as an active node?
is this referring to staking queue or shard waiting list?
if it is referring to staking queue then I think this is incorrect.

vm/systemSmartContracts/staking.go
```diff
@@ -1592,41 +1649,81 @@ func (s *stakingSC) stakeNodesFromQueue(args *vmcommon.ContractCallInput) vmcomm
...
+		return nil, err
 	}
 
 	numRegisteredKeys := int64(len(validatorData.BlsPubKeys))
 	numQualified := big.NewInt(0).Div(validatorData.TotalStakeValue, s.minNodePrice).Int64()
 
-	if numQualified < numRegisteredKeys {
-		mapCheckedOwners[string(owner)] = false
-		return false, nil
+	if numQualified >= numRegisteredKeys {
```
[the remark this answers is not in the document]
AdoAdoAdo: can you please add a test to validate that we do not stake an aditional node from staking queue if we already have exactly numQualified nodes active and an aditional node in the staking queue on the first position?

AdoAdoAdo (changes requested):

epochStart/metachain/systemSCs.go
```diff
@@ -1169,7 +1169,7 @@ func (s *systemSCProcessor) stakeNodesFromQueue(
 	nodesToStake uint32,
 	nonce uint64,
 ) error {
-	if nodesToStake == 0 {
+	if nodesToStake == 0 && !s.flagCorrectNumNodesToStake.IsSet() {
```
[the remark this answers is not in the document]
sasurobert: exactly. otherwise the same waiting list operations would happen twice.

vm/systemSmartContracts/staking.go
```diff
@@ -232,6 +232,11 @@ func (s *stakingSC) canStake() bool {
 	return stakeConfig.StakedNodes < stakeConfig.MaxNumNodes
 }
 
+func (s *stakingSC) canStakeIfOneRemoved() bool {
+	stakeConfig := s.getConfig()
+	return stakeConfig.StakedNodes <= stakeConfig.MaxNumNodes
```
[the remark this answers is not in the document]
sasurobert: not exactly - as in case of unStake - before the actual unStake this would return 0. and it would be the same problem. I added this function to have as little backward compatibility as possible.

vm/systemSmartContracts/staking.go
```diff
@@ -1528,6 +1541,50 @@ func (s *stakingSC) resetLastUnJailedFromQueue(args *vmcommon.ContractCallInput)
...
+func (s *stakingSC) cleanAdditionalQueueNotEnoughFunds(
+	waitingListData *waitingListReturnData,
+	mapCheckedOwners map[string]*validatorFundInfo,
+) error {
+	if !s.flagCorrectLastUnjailed.IsSet() {
+		return nil
+	}
+
+	for i := len(waitingListData.blsKeys) - 1; i >= 0; i-- {
+		stakedData := waitingListData.stakedDataList[i]
```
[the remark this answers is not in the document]
sasurobert: it is enforced

vm/systemSmartContracts/staking.go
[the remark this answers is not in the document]
sasurobert: The test is already added. All the nodes from the additional queue need those 2500eGLD - thus if numQualified == numRegistered - it means the validator has enough funds to have all of its nodes active.

vm/systemSmartContracts/staking.go
sasurobert: Waiting is considered a node for which 2500eGLD was paid. It is referring to the staking queue and this function is correct. The same is used when a validator stakes a node - it is verified if there is enough funds and all of its nodes are qualified and whether you can add one more or not.

vm/systemSmartContracts/staking.go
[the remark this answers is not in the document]
AdoAdoAdo: I would still add a check, just in case, before the loop.

if len(waitingListData.blsKeys) != len(waitingListData.stakedDataList) return an error

vm/systemSmartContracts/staking.go
AdoAdoAdo: so an active node here is a key with at least 2500 staked, even though that key is not currently part of the 3200 allowed active nodes?

Then the naming is confusing, can we rename this to numRegisteredNodes instead of numActiveNodes?

sasurobert: done

epochStart/metachain/systemSCs.go
```diff
@@ -1183,12 +1191,57 @@ func (s *systemSCProcessor) updateSystemSCContractsCode(contractMetadata []byte)
 	return nil
 }
 
+func (s *systemSCProcessor) cleanAdditionalQueue() error {
+	vmInput := &vmcommon.ContractCallInput{
```
iulianpascalau, line 1195: please add this:
```GO
        sw := core.NewStopWatch()
	sw.Start("systemSCProcessor")
	defer func() {
		sw.Stop("systemSCProcessor")
		log.Info("systemSCProcessor.cleanAdditionalQueue time measurements", sw.GetMeasurements()...)
	}()
```

vm/systemSmartContracts/staking.go
```diff
@@ -1543,32 +1545,31 @@ func (s *stakingSC) resetLastUnJailedFromQueue(args *vmcommon.ContractCallInput)
 
 func (s *stakingSC) cleanAdditionalQueueNotEnoughFunds(
```
iulianpascalau, line 1546: cleanAdditionalQueueNotEnoughFundsReturningUnstaked is a better name now

epochStart/metachain/systemSCs.go
```diff
@@ -232,6 +232,13 @@ func (s *systemSCProcessor) ProcessSystemSmartContract(
 		}
 	}
 
+	if s.flagCorrectNumNodesToStake.IsSet() {
+		err := s.cleanAdditionalQueue()
+		if err != nil {
+			return err
```
iulianpascalau, line 238: should we consider this a critical error? Shouldn't we just log.Error the error?

epochStart/metachain/systemSCs.go
sasurobert: done.

vm/systemSmartContracts/staking.go
sasurobert: that is too long.

epochStart/metachain/systemSCs.go
sasurobert: it here we have a critical error - the contracts will not work as expected.

iulianpascalau: ok, makes sense. Returning error here will avoid of having an inconsistent state. The metachain will halt and we have to do the fixes so it will resume the operation.

iulianpascalau (approved):

AdoAdoAdo (approved):

LucianMincu (approved):

PR #2997, opened 2021-04-12
fix nft transfer and execute from user to smart contract

sasurobert:
fix nft transfer and execute from user to smart contract

### DIFF
--- a/process/smartContract/process.go
+++ b/process/smartContract/process.go
@@ -346,7 +346,7 @@ func (sc *scProcessor) executeSmartContractCall(
 	userErrorVmOutput := &vmcommon.VMOutput{
 		ReturnCode: vmcommon.UserError,
 	}
-	vmExec, err := findVMByTransaction(sc.vmContainer, tx)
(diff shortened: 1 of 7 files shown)


camilbancioiu (approved):

process/smartContract/process.go
```diff
@@ -936,13 +935,17 @@ func (sc *scProcessor) treatExecutionAfterBuiltInFunc(
 		return true, userErrorVmOutput, newVMInput, sc.ProcessIfError(acntSnd, vmInput.CurrentTxHash, tx, err.Error(), []byte(""), snapshot, vmInput.GasLocked)
 	}
 
-	err = sc.checkUpgradePermission(acntDst, newVMInput)
+	newDestSC, err := sc.getAccountFromAddress(vmInput.RecipientAddr)
```
iulianpascalau, line 938: Is this backwards compatible?

sasurobert: yes. it is. there is no execution after built in only after esdt is enabled.

iulianpascalau (approved):

PR #3032, opened 2021-04-22
ensure the token randomness length in all cases

sasurobert:
ensure the token randomness length in all cases

### DIFF
--- a/vm/systemSmartContracts/esdt.go
+++ b/vm/systemSmartContracts/esdt.go
@@ -22,7 +22,6 @@ import (
 const numOfRetriesForIdentifier = 50
 const tickerSeparator = "-"
 const tickerRandomSequenceLength = 3
-const encodedRandomSequenceLength = tickerRandomSequenceLength * 2
 const minLengthForTickerName = 3
 const maxLengthForTickerName = 10
 const minLengthForInitTokenName = 10
@@ -417,11 +416,7 @@ func (e *esdt) createNewTokenIdentifier(caller []byte, ticker []byte) ([]byte, e
 
 	one := big.NewInt(1)
 	for i := 0; i < numOfRetriesForIdentifier; i++ {
-		encoded := []byte(hex.EncodeToString(newRandomAsBigInt.Bytes()))
-		lenEncoded := len(encoded)
-		if lenEncoded < encodedRandomSequenceLength {
-			encoded = append(make([]byte, encodedRandomSequenceLength-lenEncoded), encoded...)
-		}
+		encoded := fmt.Sprintf("%06x", newRandomAsBigInt)
 		newIdentifier := append(tickerPrefix, encoded...)
 		data := e.eei.GetStorage(newIdentifier)
 		if len(data) == 0 {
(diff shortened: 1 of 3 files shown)


dorin-iancu (dismissed):

vm/systemSmartContracts/esdt.go
```diff
@@ -409,16 +409,20 @@ func isTokenNameHumanReadable(tokenName []byte) bool {
...
 	newRandomAsBigInt := big.NewInt(0).SetBytes(newRandomForTicker)
 
 	one := big.NewInt(1)
 	for i := 0; i < numOfRetriesForIdentifier; i++ {
-		encoded := hex.EncodeToString(newRandomAsBigInt.Bytes())
-		newIdentifier := append(tickerPrefix, []byte(encoded)...)
+		encoded := []byte(hex.EncodeToString(newRandomAsBigInt.Bytes()))
+		lenEncoded := len(encoded)
+		if lenEncoded < encodedRandomSequenceLength {
+			encoded = append(make([]byte, encodedRandomSequenceLength-lenEncoded), encoded...)
```
andrei-marinica, line 423: wouldn't it make more sense to prepend the character '0' instead of \x00 bytes?

raduchis: :+1:

raduchis (approved):

vm/systemSmartContracts/esdt.go
andrei-marinica: `fmt.Sprintf("%06x", big.NewInt(123))` does it in one line, the only problem is I don't think you can provide the number of digits as a variable. But I think encodedRandomSequenceLength is a constant anyway, up to you to decide

vm/systemSmartContracts/esdt.go
```diff
@@ -409,16 +409,20 @@ func isTokenNameHumanReadable(tokenName []byte) bool {
 
 func (e *esdt) createNewTokenIdentifier(caller []byte, ticker []byte) ([]byte, error) {
 	newRandomBase := append(caller, e.eei.BlockChainHook().CurrentRandomSeed()...)
-	newRandom := sha256.Sha256{}.Compute(string(newRandomBase))
+	newRandom := e.hasher.Compute(string(newRandomBase))
```
iulianpascalau, line 412: is this backwards compatible?

vm/systemSmartContracts/esdt.go
iulianpascalau: ...or you could use bytes.Repeat function

vm/systemSmartContracts/esdt.go
sasurobert: with mainnet it is.

vm/systemSmartContracts/esdt.go
sasurobert: done as @andrei-marinica mentioned.

raduchis (approved):

andrei-marinica (approved):

iulianpascalau (approved):

bogdan-rosianu (approved):
System tests passed

gabi-vuls (approved):
System test passed

PR #3034, opened 2021-04-22
consensus: monitor also shard-to-meta miniblocks topics for rebroadcast

AdoAdoAdo:
In case of miniblocks with transactions to be processed by metachain, validators in consensus groups were not monitoring the right topic in order to find when the agreed miniblocks were broadcast. 
This caused all consensus group validators to consider every shard to meta miniblocks not broadcast by leader and the other validators in consensus, and each of them trying to re-broadcast it on their slot.

TODO: move back the changed log to trace

### DIFF
--- a/consensus/broadcast/delayedBroadcast.go
+++ b/consensus/broadcast/delayedBroadcast.go
@@ -589,7 +589,8 @@ func (dbb *delayedBlockBroadcaster) registerInterceptorsCallbackForShard(
 	rootTopic string,
 	cb func(topic string, hash []byte, data interface{}),
 ) error {
-	for idx := uint32(0); idx < dbb.shardCoordinator.NumberOfShards(); idx++ {
+	shardIDs := dbb.shardIdentifiers()
+	for idx := range shardIDs {
 		// interested only in cross shard data
 		if idx == dbb.shardCoordinator.SelfId() {
 			continue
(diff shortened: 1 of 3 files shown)


consensus/broadcast/delayedBroadcast.go
```diff
@@ -606,6 +607,16 @@ func (dbb *delayedBlockBroadcaster) registerInterceptorsCallbackForShard(
 	return nil
 }
 
+func (dbb *delayedBlockBroadcaster) shardIdentifiers() map[uint32]struct{} {
```
iulianpascalau, line 610: This should have been part of the shardcoordinator implementation. Maybe we will do it one day.

process/interceptors/baseDataInterceptor.go
```diff
@@ -84,7 +84,7 @@ func (bdi *baseDataInterceptor) processInterceptedData(data process.InterceptedD
 		return
 	}
 
-	log.Trace("intercepted data is processed",
+	log.Debug("intercepted data is processed",
```
iulianpascalau, line 87: add a TODO to revert this?

iulianpascalau (approved):

process/interceptors/baseDataInterceptor.go
AdoAdoAdo: it is in the PR description

gabi-vuls (dismissed):
System test passed

iulianpascalau (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed

PR #3047, opened 2021-04-27
New unBondTokens implementation

iulianpascalau:
- new implementation for unBondTokens as to circumvent found issue

Testing scenario that should happen in epochs 1-4 and should be carried on the `test-fix-unbond-tokens` branch:
1. stake one node in the queue with 2619 eGLD.
2. unstake 119 eGLD top up and check that you have to wait 3 epoch changes.
3. unstake 2500 eGLD from the node (direct from wallet, use the 3 dot button beside the BLS key)
4. finalize the withdraw of 2500 eGLD part, should get "no tokens that can be unbond at this time" error

Wait for epoch 4 that activates the new unBondTokens variant, repeat step 4, should unbond the 2500 value (after a total of 5 epochs between the first unstake, another unBond call should return also the last 119 eGLD).
repeat steps 1-3 with another key. Step 4 should not error, should return the 2500 eGLD and after another 5 epoch changes, a new unBond call should return the last 119 eGLD.

### DIFF
--- a/integrationTests/vm/testInitializer.go
+++ b/integrationTests/vm/testInitializer.go
@@ -71,6 +71,7 @@ type ArgEnableEpoch struct {
 	DeployEnableEpoch              uint32
 	MetaProtectionEnableEpoch      uint32
 	RelayedTxEnableEpoch           uint32
+	UnbondTokensV2                 uint32
 }
 
 // VMTestContext -
@@ -443,6 +444,7 @@ func CreateVMAndBlockchainHookMeta(
 	accnts state.AccountsAdapter,
 	gasSchedule map[string]map[string]uint64,
 	shardCoordinator sharding.Coordinator,
+	arg ArgEnableEpoch,
 ) (process.VirtualMachinesContainer, *hooks.BlockChainHookImpl) {
 	actualGasSchedule := gasSchedule
 	if gasSchedule == nil {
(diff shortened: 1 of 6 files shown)


sasurobert (approved):

vm/systemSmartContracts/validator.go
```diff
@@ -53,8 +53,10 @@ type validatorSC struct {
 	mutExecution             sync.RWMutex
 	endOfEpochAddress        []byte
 	enableDelegationMgrEpoch uint32
+	enableUnbondTokensV2     uint32
```
SebastianMarian, line 56: enableUnbondTokensV2Epoch

integrationTests/vm/testInitializer.go
```diff
@@ -71,6 +71,7 @@ type ArgEnableEpoch struct {
 	DeployEnableEpoch              uint32
 	MetaProtectionEnableEpoch      uint32
 	RelayedTxEnableEpoch           uint32
+	UnbondTokensV2                 uint32
```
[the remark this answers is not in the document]
iulianpascalau: done

vm/systemSmartContracts/validator.go
iulianpascalau: done

sasurobert (approved):

SebastianMarian (approved):

vm/systemSmartContracts/validator.go
```diff
@@ -1695,6 +1698,18 @@ func (v *validatorSC) unBondTokens(args *vmcommon.ContractCallInput) vmcommon.Re
 func (v *validatorSC) unBondTokensFromRegistrationData(
 	registrationData *ValidatorDataV2,
 	valueToUnBond *big.Int,
+) (*big.Int, vmcommon.ReturnCode) {
+	isV1Active := !v.flagUnbondTokensV2.IsSet()
```
bogdan-rosianu, line 1702: why using double negation?
`if v2 -> return for v2` instead of `if !v2 -> return for v1`

iulianpascalau: because this pattern will be extended to be used on v2, v3 and so on

gabi-vuls (approved):
System test passed.

PR #3060, opened 2021-04-29
EN 9565 peer heartbeat interceptor

iulianpascalau:
- new peer heartbeat interceptor implementation

### DIFF
--- a/process/interceptors/peerHeartbeatInterceptor.go
+++ b/process/interceptors/peerHeartbeatInterceptor.go
@@ -1,0 +1,127 @@
+package interceptors
+
+import (
+	"errors"
+
+	"github.com/ElrondNetwork/elrond-go/core"
+	"github.com/ElrondNetwork/elrond-go/core/check"
+	"github.com/ElrondNetwork/elrond-go/debug/resolver"
+	"github.com/ElrondNetwork/elrond-go/marshal"
+	"github.com/ElrondNetwork/elrond-go/p2p"
+	"github.com/ElrondNetwork/elrond-go/process"
+)
+
+// ArgPeerHeartbeatInterceptor is the argument for the peer heartbeat interceptor
+type ArgPeerHeartbeatInterceptor struct {
+	ArgSingleDataInterceptor
+	Marshalizer         marshal.Marshalizer
+	ValidatorChecker    process.ValidatorChecker
+	ProcessingThrottler process.InterceptorThrottler
+	HeartbeatProcessor  process.PeerHeartbeatProcessor
+}
+
+type peerHeartbeatInterceptor struct {
+	*baseDataInterceptor
+	validatorChecker                 process.ValidatorChecker
+	peerHeartbeatProcessingThrottler process.InterceptorThrottler
+	peerHeartbeatProcessor           process.PeerHeartbeatProcessor
+}
+
+// NewPeerHeartbeatInterceptor hooks a new interceptor for packed multi data containing peer heartbeat instances
+func NewPeerHeartbeatInterceptor(arg ArgPeerHeartbeatInterceptor) (*peerHeartbeatInterceptor, error) {
+	err := checkArguments(arg.ArgSingleDataInterceptor)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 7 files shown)


process/interceptors/peerHeartbeatInterceptor.go
```diff
@@ -0,0 +1,127 @@
...
+		interceptedData, err = phi.interceptedData(dataBuff, message.Peer(), fromConnectedPeer)
+		if err != nil {
+			phi.throttler.EndProcessing()
+			return err
+		}
+
+		peerHb, ok := interceptedData.(process.InterceptedPeerHeartbeat)
+		if !ok {
+			//intercepted data is not of type interceptedPeerInfo
+			cause := "intercepted data is not of type process.InterceptedPeerInfo"
```
raduchis, line 88: type process.InterceptedPeerHeartbeat?

process/interceptors/peerHeartbeatInterceptor.go
```diff
@@ -0,0 +1,127 @@
...
+		_, shardID, err = phi.validatorChecker.GetValidatorWithPublicKey(peerHb.PublicKey())
+		isSkippableObserver := err != nil && !phi.peerHeartbeatProcessingThrottler.CanProcess()
+		if isSkippableObserver {
+			continue
+		}
+
+		//validators will always get the chance to get their message processed despite the number of observers messages
+		//being broadcast
+		phi.peerHeartbeatProcessingThrottler.StartProcessing()
+		peerHb.SetShardID(shardID)
```
raduchis, line 105: for an Observer will the shardID be set always to 0 if the CanProcess is true? The GetValidatorWithPublicKey returns nil, 0, error.

process/interceptors/peerHeartbeatInterceptor.go
```diff
@@ -0,0 +1,127 @@
...
+		peerHbs = append(peerHbs, peerHb)
+	}
+
+	go func(originator, fromConnectedPeer core.PeerID) {
+		for _, pi := range peerHbs {
+			errProcess := phi.peerHeartbeatProcessor.Process(pi)
```
raduchis, line 111: peerHeartbeatProcessor is implemented just by a stub. Is it ok? The new implementation still comes?

iulianpascalau: yes, it will follow in the next PR

process/interceptors/peerHeartbeatInterceptor.go
iulianpascalau: Exactly. We should set the shard ID only for the validator nodes. Will change

process/interceptors/peerHeartbeatInterceptor.go
raduchis: ok

process/interceptors/peerHeartbeatInterceptor.go
raduchis: ok

process/interceptors/peerHeartbeatInterceptor.go
raduchis: ok

sasurobert (dismissed):

process/interceptors/peerHeartbeatInterceptor.go
sasurobert: I think it would be important to set for obsevers as well as they route a lot of messages when they are connected. Could we get that information from the heartbeat ?

iulianpascalau: Yes, actually the peer heartbeat structure will contain 2 shardID values (exactly in the same way the current implementation have 2 shardID fields): one for computed shardID and one for the advertised shardID (aka the shardID send from that peer through the peer heartbeat message). Will rename as to avoid confusions.

iulianpascalau: done

sasurobert (approved):

raduchis (approved):

PR #3064, opened 2021-05-04
added command "make lint"

miiu96:
Added in the Makefile command "make lint" that will run` golangci linter.`
This command will check if the `golangci linter` is installed ( in case of not installed, will install the latest tag of the `golangci linter`) and after that will run the linter.

### DIFF
--- a/crypto/signing/multisig/pubkeys.go
+++ b/crypto/signing/multisig/pubkeys.go
@@ -6,8 +6,8 @@ import (
 
 func convertStringsToPubKeys(pubKeys []string, kg crypto.KeyGenerator) ([]crypto.PublicKey, error) {
 	// convert pubKeys
-	pk := make([]crypto.PublicKey, len(pubKeys))
-	for idx, pubKeyStr := range pubKeys {
+	pk := make([]crypto.PublicKey, 0, len(pubKeys))
+	for _, pubKeyStr := range pubKeys {
 		if pubKeyStr == "" {
 			return nil, crypto.ErrEmptyPubKeyString
 		}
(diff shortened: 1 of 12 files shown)


iulianpascalau (dismissed):

bogdan-rosianu (changes requested):
works only for the first issue right now

bogdan-rosianu (approved):

crypto/signing/multisig/pubkeys.go
```diff
@@ -5,10 +5,9 @@ import (
 )
 
 func convertStringsToPubKeys(pubKeys []string, kg crypto.KeyGenerator) ([]crypto.PublicKey, error) {
-	var pk []crypto.PublicKey
-
-	//convert pubKeys
-	for _, pubKeyStr := range pubKeys {
+	// convert pubKeys
+	pk := make([]crypto.PublicKey, len(pubKeys))
```
iulianpascalau, line 9: please use:
```GO
pk := make([]crypto.PublicKey, 0, len(pubKeys))
for _, pubKeyStr := range pubKeys {
		if pubKeyStr == "" {
			return nil, crypto.ErrEmptyPubKeyString
		}
	        //................
		pk = append(pk, pubKey)
}
```

iulianpascalau (changes requested):

crypto/signing/multisig/pubkeys.go
miiu96: done.

iulianpascalau (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.

PR #3070, opened 2021-05-06
En 9798 start in epoch fix

raduchis:
fixed panic if start-in-epoch = false on a validator node.


AdoAdoAdo (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3083, opened 2021-05-11
En 9787 intercepted peer heartbeat implementation + factory

iulianpascalau:
- added intercepted peer heartbeat implementation + associated data factory
- changed the location of the heartbeat proto description as to align with the rest of the project


data/heartbeat/proto/heartbeat.proto
```diff
@@ -0,0 +1,26 @@
...
+
+option go_package = "heartbeat";
+
+// Heartbeat represents the heartbeat message that is sent between peers from the same shard containing
+// current node status
+message Heartbeat {
+    bytes   Payload         = 1;
+    uint32  ShardID         = 2;
+    string  VersionNumber   = 3;
+    string  NodeDisplayName = 4;
```
sasurobert, line 13: do we have validation for all these fields for maximum length ?

sasurobert (dismissed):

data/heartbeat/proto/heartbeat.proto
iulianpascalau: yes, basically the same implementation (copied because the old heartbeat package will simply disappear) in process/heartbeat/interceptedPeerHeartbeat.go CheckValidity function.

AdoAdoAdo (changes requested):

process/heartbeat/interceptedPeerHeartbeat.go
```diff
@@ -0,0 +1,168 @@
...
+
+// PublicKey returns the public key as byte slice
+func (iph *interceptedPeerHeartbeat) PublicKey() []byte {
+	return iph.peerHeartbeat.Pubkey
+}
+
+// SetComputedShardID sets the computed shard ID. Should normally be called by the peer heartbeat interceptor
+//only once, before the processing start
+func (iph *interceptedPeerHeartbeat) SetComputedShardID(shardId uint32) {
+	iph.computedShardID = shardId
```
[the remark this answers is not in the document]
iulianpascalau: done

AdoAdoAdo (approved):

PR #3089, opened 2021-05-12
Refactor trie insert and commit

BeniaminDrasovean:
Refactor trie insert and commit in preparation for the trie checkpoint optimizations.

### DIFF
--- a/data/trie/branchNode.go
+++ b/data/trie/branchNode.go
@@ -388,61 +388,68 @@ func (bn *branchNode) getNext(key []byte, db data.DBWriteCacher) (node, []byte, 
 	return bn.children[childPos], key, nil
 }
 
-func (bn *branchNode) insert(n *leafNode, db data.DBWriteCacher) (bool, node, [][]byte, error) {
+func (bn *branchNode) insert(n *leafNode, db data.DBWriteCacher) (node, [][]byte, error) {
 	emptyHashes := make([][]byte, 0)
 	err := bn.isEmptyOrNil()
 	if err != nil {
-		return false, nil, emptyHashes, fmt.Errorf("insert error %w", err)
+		return nil, emptyHashes, fmt.Errorf("insert error %w", err)
 	}
-	if len(n.Key) == 0 {
-		return false, nil, emptyHashes, ErrValueTooShort
+
+	insertedKey := n.Key
+	if len(insertedKey) == 0 {
+		return nil, emptyHashes, ErrValueTooShort
 	}
-	childPos := n.Key[firstByte]
+	childPos := insertedKey[firstByte]
 	if childPosOutOfRange(childPos) {
-		return false, nil, emptyHashes, ErrChildPosOutOfRange
+		return nil, emptyHashes, ErrChildPosOutOfRange
 	}
-	n.Key = n.Key[1:]
+
+	n.Key = insertedKey[1:]
 	err = resolveIfCollapsed(bn, childPos, db)
 	if err != nil {
-		return false, nil, emptyHashes, err
+		return nil, emptyHashes, err
 	}
 
-	if bn.children[childPos] != nil {
-		var dirty bool
-		var newNode node
-		var oldHashes [][]byte
-
-		dirty, newNode, oldHashes, err = bn.children[childPos].insert(n, db)
-		if !dirty || err != nil {
-			return false, bn, emptyHashes, err
-		}
-
-		if !bn.dirty {
-			oldHashes = append(oldHashes, bn.hash)
-		}
-
-		bn.children[childPos] = newNode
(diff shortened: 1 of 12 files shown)


data/trie/leafNode.go
```diff
@@ -111,20 +111,39 @@ func (ln *leafNode) hashNode() ([]byte, error) {
...
 	ln.dirty = false
 	return encodeNodeAndCommitToDB(ln, targetDb)
 }
+func (ln *leafNode) commitCheckpoint(_ data.DBWriteCacher, targetDb data.DBWriteCacher) error {
+	err := ln.isEmptyOrNil()
+	if err != nil {
+		return fmt.Errorf("commit checkpoint error %w", err)
+	}
+
+	//TODO add early return if should not commit to checkpoint db and set commited flag to true
```
iulianpascalau, line 133: implement both TODOs?

data/trie/trieStorageManager.go
```diff
@@ -408,14 +411,19 @@ func (tsm *trieStorageManager) takeSnapshot(snapshot *snapshotsQueueEntry, msh m
 		return
 	}
 
-	maxTrieLevelInMemory := uint(5)
-	err = newRoot.commit(true, 0, maxTrieLevelInMemory, tsm.db, db)
-	if err != nil {
-		log.Error("trie storage manager: commit", "error", err.Error())
+	if snapshot.newDb {
```
iulianpascalau, line 414: maybe use a StopWatch to measure the time taken by this method?

data/trie/extensionNode.go
```diff
@@ -285,63 +325,68 @@ func (en *extensionNode) getNext(key []byte, db data.DBWriteCacher) (node, []byt
...
 
-		newEn, err = newExtensionNode(en.Key, newNode, en.marsh, en.hasher)
-		if err != nil {
-			return false, nil, emptyHashes, err
-		}
+func (en *extensionNode) insertInSameEn(n *leafNode, keyMatchLen int, db data.DBWriteCacher) (node, [][]byte, error) {
+	n.Key = n.Key[keyMatchLen:]
+	newNode, oldHashes, err := en.child.insert(n, db)
+	if check.IfNil(newNode) || err != nil {
+		return nil, [][]byte{}, err
```
AdoAdoAdo, line 355: Could you have nil newNode and nil err? 
then you end up with nil err

if it cannot happen then you don't need the check.IfNil right?

data/trie/branchNode.go
```diff
@@ -388,61 +446,68 @@ func (bn *branchNode) getNext(key []byte, db data.DBWriteCacher) (node, []byte,
...
-	oldHash := make([][]byte, 0)
+	modifiedHashes := make([][]byte, 0)
+	modifiedHashes = bn.modifyNodeAfterInsert(modifiedHashes, childPos, newLn)
+
+	return bn, modifiedHashes, nil
+}
+
+func (bn *branchNode) insertOnExistingChild(n *leafNode, childPos byte, db data.DBWriteCacher) (node, [][]byte, error) {
+	newNode, modifiedHashes, err := bn.children[childPos].insert(n, db)
+	if check.IfNil(newNode) || err != nil {
```
AdoAdoAdo, line 492: Could you have nil newNode and nil err?
then you end up with nil err

if it cannot happen then you don't need the check.IfNil right?

data/trie/leafNode.go
BeniaminDrasovean: Will be implemented in the following PR, because it requires a lot of code changes.

data/trie/trieStorageManager.go
BeniaminDrasovean: Used a stopWatch in the accountsDB to measure the time it takes to snapshot/checkpoint the whole state.

data/trie/extensionNode.go
BeniaminDrasovean: There is a case where you can have a nil newNode and a nil err: if you try to insert the same value at the same key (`leafNode.go` L233). It is not an error, but there is no newNode to return. The `insert()` return is treated in `particiaMerkleTrie.go` on L132 - L138.

data/trie/branchNode.go
BeniaminDrasovean: There is a case where you can have a nil newNode and a nil err: if you try to insert the same value at the same key (leafNode.go L233). It is not an error, but there is no newNode to return. The insert() return is treated in particiaMerkleTrie.go on L132 - L138.

AdoAdoAdo (approved):

data/trie/leafNode.go
iulianpascalau: ok

iulianpascalau (approved):

PR #3094, opened 2021-05-14
Re-delegate dust check

iulianpascalau:
- added redelegation dust check on the delegation system SC
- fixed some issues regarding ValidatorToDelegationEnableEpoch flags
- refactored economics.toml file

Testing scenario:
start with the `test-redelegate-dust-check` that will activate the feature in epoch 6.
In epoch 2 launch a delegation SC. In epoch 3, merge an existing validator into the delegation SC. 
Take 2 addresses (let's call them A and B), delegate a few egld on that delegation SC. 
In epochs 3-5 (whenever is possible), check rewards for address A. It should be in the interval (0, 1) eGLD. 
Undelegate everything. After that, do a redelegate. Now A just delegated some dust, which is not ok.
After epoch 6 activates the extra check, try the same with address B. Remember, claimable rewards should be up until 1eGLD. 
When B does a redelegate, that transaction should fail, stating something like `can not re-delegate as the remaining value will be below the minimum required`. Rewards should be still claimable (please check, claim should work)
Also, please check after the fix is available (epoch 6 and higher) that a call to redelegate of the owner of the delegation SC works.

### DIFF
--- a/vm/errors.go
+++ b/vm/errors.go
@@ -229,3 +229,6 @@ var ErrNotEnoughInitialOwnerFunds = errors.New("not enough initial owner funds")
 
 // ErrNFTCreateRoleAlreadyExists signals that NFT create role already exists
 var ErrNFTCreateRoleAlreadyExists = errors.New("NFT create role already exists")
+
+// ErrRedelegateValueWillGenerateDust signals that the re-delegate value will generate dust in the delegation SC
+var ErrRedelegateValueWillGenerateDust = errors.New("re-delegate value will generate dust in the delegation SC")
--- a/vm/errors.go
+++ b/vm/errors.go
@@ -230,5 +230,5 @@ var ErrNotEnoughInitialOwnerFunds = errors.New("not enough initial owner funds")
 // ErrNFTCreateRoleAlreadyExists signals that NFT create role already exists
 var ErrNFTCreateRoleAlreadyExists = errors.New("NFT create role already exists")
 
-// ErrRedelegateValueWillGenerateDust signals that the re-delegate value will generate dust in the delegation SC
-var ErrRedelegateValueWillGenerateDust = errors.New("re-delegate value will generate dust in the delegation SC")
+// ErrRedelegateValueBelowMinimum signals that the re-delegate added to the remaining value will be below the minimum required
+var ErrRedelegateValueBelowMinimum = errors.New("can not re-delegate as the remaining value will be below the minimum required")
--- a/vm/systemSmartContracts/delegation.go
+++ b/vm/systemSmartContracts/delegation.go
@@ -60,6 +60,8 @@ type delegation struct {
 	stakingV2Enabled                 atomic.Flag
 	flagValidatorToDelegation        atomic.Flag
 	validatorToDelegationEnableEpoch uint32
+	flagReDelegateDustCheck          atomic.Flag
+	reDelegateDustCheckEnableEpoch   uint32
 }
 
 // ArgsNewDelegation defines the arguments to create the delegation smart contract
@@ -76,6 +78,7 @@ type ArgsNewDelegation struct {
 	Marshalizer                      marshal.Marshalizer
 	EpochNotifier                    vm.EpochNotifier
 	ValidatorToDelegationEnableEpoch uint32
+	ReDelegateDustCheckEnableEpoch   uint32
 }
 
 // NewDelegationSystemSC creates a new delegation system SC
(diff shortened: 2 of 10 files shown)


vm/systemSmartContracts/delegation.go
```diff
@@ -1322,6 +1332,31 @@ func (d *delegation) reDelegateRewards(args *vmcommon.ContractCallInput) vmcommo
...
+	if !d.flagReDelegateDustCheck.IsSet() {
+		return nil
+	}
+
+	delegationManagement, err := getDelegationManagement(d.eei, d.marshalizer, d.delegationMgrSCAddress)
+	if err != nil {
+		return err
+	}
+
+	fund, err := d.getFund(delegator.ActiveFund)
```
sasurobert, line 1345: here you may have the scenario that the user does not have any active fund - thus this get will return error - the user might still have more than 1 eGLD as rewards.

sasurobert (changes requested):

vm/systemSmartContracts/delegation.go
iulianpascalau: Does the active fund gets erased even in the case the delegator undelegated everything but did not claimed its rewards?

vm/errors.go
```diff
@@ -229,3 +229,6 @@ var ErrNotEnoughInitialOwnerFunds = errors.New("not enough initial owner funds")
 
 // ErrNFTCreateRoleAlreadyExists signals that NFT create role already exists
 var ErrNFTCreateRoleAlreadyExists = errors.New("NFT create role already exists")
+
+// ErrRedelegateValueWillGenerateDust signals that the re-delegate value will generate dust in the delegation SC
+var ErrRedelegateValueWillGenerateDust = errors.New("re-delegate value will generate dust in the delegation SC")
```
raduchis, line 234: I wouldn't call it dust... just cause it can be smaller than 1 EGLD, which is not negligible. 
call it somehting like: redelegateBelowThreshold or redelegateBelowMinDelegationAmount

vm/systemSmartContracts/delegation.go
```diff
@@ -1322,6 +1332,31 @@ func (d *delegation) reDelegateRewards(args *vmcommon.ContractCallInput) vmcommo
...
+	if err != nil {
+		return err
+	}
+
+	newActiveFundValue := big.NewInt(0).Set(fund.Value)
+	newActiveFundValue.Add(newActiveFundValue, delegator.UnClaimedRewards)
+	willRemainDust := newActiveFundValue.Cmp(delegationManagement.MinDelegationAmount) < 0
```
raduchis, line 1352: willRemainDust -> belowMinDelegationAmount

vm/systemSmartContracts/delegation.go
iulianpascalau: I think this will return the error on L1289 because the computeAndUpdateRewards try to fetch the active found object. So it is ok to have the code as it is

vm/errors.go
iulianpascalau: refactored

vm/systemSmartContracts/delegation.go
iulianpascalau: refactored

vm/systemSmartContracts/delegation.go
iulianpascalau: fixing

vm/errors.go
raduchis: ok!

vm/systemSmartContracts/delegation.go
sasurobert: yes. active fund is deleted.

sasurobert (dismissed):

vm/systemSmartContracts/delegation.go
```diff
@@ -2772,6 +2805,9 @@ func (d *delegation) EpochConfirmed(epoch uint32) {
 
 	d.flagValidatorToDelegation.Toggle(epoch >= d.validatorToDelegationEnableEpoch)
 	log.Debug("validator to delegation", "enabled", d.flagValidatorToDelegation.IsSet())
+
+	d.flagReDelegateBelowMinCheck.Toggle(epoch >= d.reDelegateBelowMinCheckEnableEpoch)
+	log.Debug("re-delegate dust check", "enabled", d.flagReDelegateBelowMinCheck.IsSet())
```
raduchis, line 2810: "dust" remained here

raduchis (approved):

vm/systemSmartContracts/delegation.go
raduchis: ok

sasurobert (dismissed):

raduchis (dismissed):

raduchis (approved):

sasurobert (approved):

mihaieremia (approved):

LucianMincu (approved):

mihaieremia:
Well done guys 🙏

PR #3109, opened 2021-05-21
EN 10033 - PeerAuthentication and interceptor optimization

iulianpascalau:
- rename the peerHeartbeat to peerAuthentication & all related implementations
- optimized the peerAuthenticationInterceptor.go as to not process nor re-broadcast messages that contain messages from observer peers


process/heartbeat/interceptedPeerAuthentication.go
```diff
@@ -0,0 +1,177 @@
...
+
+// SetComputedShardID sets the computed shard ID. Concurrency safe.
+func (ipa *interceptedPeerAuthentication) SetComputedShardID(shardId uint32) {
+	ipa.mutComputedShardID.Lock()
+	ipa.computedShardID = shardId
+	ipa.mutComputedShardID.Unlock()
+}
+
+// CheckValidity will check the validity of the received peer heartbeat
+// However, this will not perform the signature validation since that can be an attack vector
```
bogdan-rosianu, line 111: no need for such details. simply mention that it won't do sig checks

process/interceptors/peerAuthenticationInterceptor.go
```diff
@@ -0,0 +1,116 @@
...
+			//intercepted data is not of type interceptedPeerInfo
+			cause := "intercepted data is not of type process.InterceptedPeerInfo"
+			pai.blackListPeers(cause, nil, message.Peer(), fromConnectedPeer)
+			pai.throttler.EndProcessing()
+
+			return errors.New(cause)
+		}
+
+		var shardID uint32
+		_, shardID, err = pai.validatorChecker.GetValidatorWithPublicKey(peerAuth.PublicKey())
```
bogdan-rosianu, line 90: are we sure that an error can be returned only if it is an observer? Maybe make the condition more specific, like `err == ErrPubKeyIsNotAValidator`

bogdan-rosianu (changes requested):

process/heartbeat/interceptedPeerAuthentication.go
iulianpascalau: done

process/interceptors/peerAuthenticationInterceptor.go
iulianpascalau: I would keep this as it is for one single reason: p2p package will depend on the sharding package for that type of error check. Could have also redefined the error but that would have caused some unexpected behavior (hard to catch in integration tests) when the indexedHashedCoordinator would have changed that particular error.

bogdan-rosianu: ok

bogdan-rosianu (approved):

process/interceptors/peerAuthenticationInterceptor.go
```diff
@@ -97,7 +97,7 @@ func (pai *peerAuthenticationInterceptor) ProcessReceivedMessage(message p2p.Mes
 		_, shardID, err = pai.validatorChecker.GetValidatorWithPublicKey(peerAuth.PublicKey())
 		peerAuth.SetComputedShardID(shardID)
 
-		isObserver := err == nil
+		isObserver := err != nil
```
AdoAdoAdo, line 100: also don't propagate messages if len(multiDataBuff)>1

process/errors.go
```diff
@@ -1027,3 +1021,9 @@ var ErrPropertyTooShort = errors.New("property too short")
 
 // ErrNilPeerSignatureHandler signals that a nil peerSignatureHandler object has been provided
 var ErrNilPeerSignatureHandler = errors.New("trying to set nil peerSignatureHandler")
+
+// ErrPeerAuthenticationForObservers signals (to the network messenger) that the current received message containing bulk
+// peer authentication messages that should be dropped as an observer public key wasn't checked by the processors
+// Otherwise, there is a slight change of an attack on a peer that does sync in a previous epoch and not able to check all
```
AdoAdoAdo, line 1027: slight change -> slight chance, also too long.

process/interceptors/peerAuthenticationInterceptor.go
iulianpascalau: done

process/errors.go
iulianpascalau: comment re-written

AdoAdoAdo (approved):

bogdan-rosianu (approved):

PR #3114, opened 2021-05-25
Refactor mocks

BeniaminDrasovean:
Create a stub for AccountsAdapter and for Trie in testscommon package, and use it everywhere.


node/node_test.go
```diff
@@ -528,9 +537,12 @@ func TestNode_GetAllIssuedESDTs(t *testing.T) {
 
 				return ch, nil
 			},
+			RootCalled: func() ([]byte, error) {
```
bogdan-rosianu, line 540: if `TrieStub` would have a nil check for the handler functions, then you wouldn't need this nil func implementation

bogdan-rosianu (approved):

PR #3125, opened 2021-05-31
Added get/restore snapshot methods to blockchain hook

matei-p:


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/ElrondNetwork/elrond-go
 go 1.13
 
 require (
-	github.com/ElrondNetwork/arwen-wasm-vm v1.2.14
+	github.com/ElrondNetwork/arwen-wasm-vm v1.2.15-0.20210531084552-d81a17a3049d
```
sasurobert, line 6: need a proper release. It will be funny that we will always need to change the implementation for v1.2 - even when it is about some interface changes.

sasurobert (approved):

bogdan-rosianu (approved):

PR #3126, opened 2021-05-31
Rename DoesFileExist function to FileExists

gabi-vuls:
Rename `DoesFileExist` to `FileExists`


iulianpascalau (commented):
The code is ok, you have to make another PR and use signed commits

PR #3127, opened 2021-05-31
En 10032 rename does file exist with signed commits

gabi-vuls:
Rename DoesFileExist with FileExists.


iulianpascalau (approved):

iulianpascalau (changes requested):
signed commits are not working

PR #3136, opened 2021-05-31
Consensus scheduled block

AdoAdoAdo:
This pull request starts the pipeline processing for the proposer and validators during consensus for the v2 blocks with scheduled miniblocks and validates the end result.

The proposer does not send the block final info if either there was an error with the block processing (coding errors - should be caught during testing phase if any) or timeout during execution - this is tricky and could happen in case of bad gas estimation, or starved processor.

As the consensus round is elastic, with early finalization in case all signatures are received ahead of time, the checking for the scheduled miniblocks execution result is delayed until either a final status is available or the scheduled mb execution has timedout.


consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return sp.processingTime - elapsedTime
+}
+
+// IsProcessedOK returns true if the scheduled processing was finalized without error
+// Function is blocking until the allotted time for processing is finished
+func (sp *scheduledProcessor) IsProcessedOK() bool {
+	status := sp.getStatus()
+
+loop:
```
iulianpascalau, line 101: the break loop stuff could have been avoided by extracting the for loop in a new function and used return statement.

data/block/block.go
```diff
@@ -437,3 +437,8 @@ func (h *Header) SetAdditionalData(_ headerVersionData.HeaderAdditionalData) err
 	// first header version does not have any additional data
 	return nil
 }
+
+// HasScheduledSupport returns true if the header version supports scheduled data
```
iulianpascalau, line 441: HasScheduledSupport returns false as the first block version does not support scheduled data

data/block/blockV2.go
```diff
@@ -468,6 +468,11 @@ func (hv2 *HeaderV2) SetEpochStartMetaHash(hash []byte) error {
 	return hv2.Header.SetEpochStartMetaHash(hash)
 }
 
+// HasScheduledSupport returns true if the header version supports scheduled data
```
iulianpascalau, line 471: HasScheduledSupport returns true as the second block version does support scheduled data

data/block/metaBlock.go
```diff
@@ -392,6 +392,11 @@ func (m *MetaBlock) GetMiniBlockHeaderHandlers() []data.MiniBlockHeaderHandler {
 	return mbHeaderHandlers
 }
 
+// HasScheduledSupport sets the miniBlock headers from the given miniBlock header handlers
```
iulianpascalau, line 395: wrong comment, might use the one defined in the shard block

integrationTests/consensus/consensus_test.go
```diff
@@ -223,6 +227,7 @@ func TestConsensusBLSFullTest(t *testing.T) {
 		t.Skip("this is not a short test")
 	}
 
+	_ = logger.SetLogLevel("*:DEBUG")
```
iulianpascalau, line 230: might remove this

consensus/spos/scheduledProcessor.go
AdoAdoAdo: :+1:

consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,180 @@
...
+	for status == inProgress {
+		remainingExecutionTime := sp.computeRemainingProcessingTime()
```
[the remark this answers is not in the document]
AdoAdoAdo: negative remaining execution will have no impact on time.After, it will immediately return

consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,180 @@
...
+	processingNotStarted processingStatus = 0
+	processingError      processingStatus = 1
+	inProgress           processingStatus = 2
+	processingOK         processingStatus = 3
+
+	processingNotStartedString = "processing not started"
+	processingErrorString      = "processing error"
+	inProgressString           = "processing in progress"
+	processingOKString         = "processing OK"
+	unexpectedString           = "processing OK"
```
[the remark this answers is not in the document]
AdoAdoAdo: done

consensus/spos/bls/subroundBlock.go
```diff
@@ -551,6 +553,8 @@ func (sr *subroundBlock) processReceivedBlock(cnsDta *consensus.Message) bool {
 		return false
 	}
 
+	sr.ConsensusCoreHandler.ScheduledProcessor().StartScheduledProcessing(sr.Header, sr.Body)
```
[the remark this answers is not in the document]
AdoAdoAdo: The block subround doesn't really need to wait on scheduled execution completion. 
The result will be verified on end round subround.

consensus/spos/bls/subroundEndRound.go
```diff
@@ -188,6 +188,11 @@ func (sr *subroundEndRound) doEndRoundJobByLeader() bool {
 		return false
 	}
 
+	ok := sr.ConsensusCoreHandler.ScheduledProcessor().IsProcessedOK()
+	if !ok {
+		return false
```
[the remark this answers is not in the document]
AdoAdoAdo: logged inside IsProcessedOK

consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// ScheduledProcessorArgs holds the arguments required to instantiate the pipelineExecution
+type ScheduledProcessorArgs struct {
+	SyncTimer                  ntp.SyncTimer
+	Processor                  process.ScheduledBlockProcessor
+	ProcessingTimeMilliSeconds uint32
+}
+
+type scheduledProcessor struct {
```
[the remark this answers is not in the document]
AdoAdoAdo: renamed

consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,180 @@
...
+
+// NewScheduledProcessor creates a new processor for scheduled transactions
+func NewScheduledProcessor(args ScheduledProcessorArgs) (*scheduledProcessor, error) {
+	if check.IfNil(args.SyncTimer) {
+		return nil, process.ErrNilSyncTimer
+	}
+	if check.IfNil(args.Processor) {
+		return nil, process.ErrNilBlockProcessor
+	}
+
```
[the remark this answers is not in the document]
AdoAdoAdo: done

data/block/block.go
AdoAdoAdo: done

data/block/blockV2.go
AdoAdoAdo: done

data/block/metaBlock.go
AdoAdoAdo: done

integrationTests/consensus/consensus_test.go
AdoAdoAdo: done

consensus/spos/bls/subroundBlock.go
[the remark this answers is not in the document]
iulianpascalau: ok

consensus/spos/bls/subroundEndRound.go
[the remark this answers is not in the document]
iulianpascalau: ok

consensus/spos/scheduledProcessor.go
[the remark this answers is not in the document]
iulianpascalau: ok

iulianpascalau (approved):

consensus/spos/scheduledProcessor.go
```diff
@@ -0,0 +1,188 @@
...
+		sp.setStatus(processingOK)
+		return
+	}
+
+	sp.oneInstance.Lock()
+
+	sp.setStatus(inProgress)
+	log.Debug("scheduledProcessorWrapper.StartScheduledProcessing - scheduled processing has been started")
+
+	go func() {
```
sasurobert, line 138: a lot of my previous questions about running in go routine were answered here. Although I thing it would be safer if this function gets a context and the go routine can be closed from the outside.

consensus/spos/bls/subroundBlock.go
```diff
@@ -160,8 +162,8 @@ func (sr *subroundBlock) createBlock(header data.HeaderHandler) (data.HeaderHand
 
 // sendBlockBodyAndHeader method sends the proposed block body and header in the subround Block
 func (sr *subroundBlock) sendBlockBodyAndHeader(
```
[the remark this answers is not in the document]
AdoAdoAdo: renamed, the parameter order was changed for consistency with the other methods.

consensus/spos/bls/subroundBlock.go
```diff
@@ -92,11 +92,13 @@ func (sr *subroundBlock) doBlockJob() bool {
 		return false
 	}
 
-	sentWithSuccess := sr.sendBlock(body, header)
+	sentWithSuccess := sr.sendBlock(header, body)
 	if !sentWithSuccess {
 		return false
 	}
 
+	sr.ConsensusCoreHandler.ScheduledProcessor().StartScheduledProcessing(header, body)
```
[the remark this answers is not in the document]
AdoAdoAdo: 1. Yes it starts the processing and it has an allotted time (and gas) for finishing (same flow we have for regular txs processing which have a bounded time for execution)
2. The scheduled miniblocks can also be cross shard with the current shard as destination (probably in the follow up PR these need to be treated as well as I believe there were some TODOS), as long as the miniblock contains only smart contract calls (currently this is one limitation but we might decide afterwards if we want to keep this or allow also regular txs).
3. In the case where something fails (e.g implementation bug, not enough time due to wrong gas estimation) then the proposer will not send the final data, so consensus will fail. In case it fails on the validator side but a malicious proposer sends the final info, then the decision will be on the next proposer, which is the scheduled root hash it starts with and which is the previous block it links to.
4. The routine does the actual execution of the already proposed scheduled transactions, so if validators sent their signatures, all of them agreed that they will process. It is the same with cross shard smart contract miniblocks that need to be executed on a different shard, on the source shard it was agreed that the miniblock contains the set of transactions that MUST be executed in that specific order on destination shard. Here as well it is a commitment for execution (but in the same round and on the same shard).

consensus/spos/scheduledProcessor.go
AdoAdoAdo: added a forceStop for scheduled execution

consensus/spos/scheduledProcessor.go
```diff
@@ -79,24 +84,29 @@ func NewScheduledProcessorWrapper(args ScheduledProcessorWrapperArgs) (*schedule
...
+// ForceStopScheduledExecutionBlocking forces the execution to stop
+func (sp *scheduledProcessorWrapper) ForceStopScheduledExecutionBlocking() {
+	status := sp.getStatus()
+	if status != inProgress {
+		return
 	}
 
-	return sp.processingTime - elapsedTime
+	sp.stopExecution <- struct{}{}
+	for status == inProgress {
```
iulianpascalau, line 99: this can easily become:
```GO
for status == inProgress {
    time.Sleep(processingCheckStep)
    status = sp.getStatus()
}
```

consensus/spos/scheduledProcessor.go
```diff
@@ -149,6 +159,44 @@ func (sp *scheduledProcessorWrapper) StartScheduledProcessing(header data.Header
 	}()
 }
 
+// IsInterfaceNil returns true if there is no value under the interface
+func (sp *scheduledProcessorWrapper) IsInterfaceNil() bool {
+	return sp == nil
+}
+
+func (sp *scheduledProcessorWrapper) emptyStopChannel() {
```
iulianpascalau, line 167: this resembled with the core.EmptyChannel function. could have used that by changing stopExecution to be a buffered chan of type bool.

consensus/spos/scheduledProcessor.go
AdoAdoAdo: done

iulianpascalau (approved):

sasurobert (approved):
Schedule Block Processing has to be called in sync as well.

PR #3167, opened 2021-06-10
Always-not-synced exception in block syncer during import-db operation

iulianpascalau:
- added an exception to the block syncer as to always consider itself not synchronized in the import-db process

Testing scenario: in the normal operation, the network should behave normally, the WARN message `using always-not-synced status because the node is running in import-db` should not appear on any node. During the import-db operation, the above WARN message should appear and the node should not wait more that 10 rounds between epochs on the metachain.
When testing the import-db operation, the branch `test-always-not-synced-in-import-db` should be used.

### DIFF
--- a/process/sync/baseSync.go
+++ b/process/sync/baseSync.go
@@ -536,9 +536,7 @@ func (boot *baseBootstrap) incrementSyncedWithErrorsForNonce(nonce uint64) uint3
 // in the blockchain, and all this mechanism will be reiterated for the next block.
 func (boot *baseBootstrap) syncBlock() error {
 	boot.computeNodeState()
-	nodeState := boot.GetNodeState()
-	shouldSync := nodeState == core.NsNotSynchronized || boot.isInImportMode
-	if !shouldSync {
+	if !boot.shouldSync() {
 		return nil
 	}
 
@@ -632,6 +630,15 @@ func (boot *baseBootstrap) syncBlock() error {
 	boot.cleanNoncesSyncedWithErrorsBehindFinal()
 
 	return nil
+}
+
+func (boot *baseBootstrap) shouldSync() bool {
+	if boot.isInImportMode {
+		return true
+	}
+
+	nodeState := boot.GetNodeState()
+	return nodeState == core.NsNotSynchronized
 }
 
 func (boot *baseBootstrap) cleanNoncesSyncedWithErrorsBehindFinal() {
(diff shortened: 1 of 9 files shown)


process/sync/metablock.go
```diff
@@ -60,6 +60,11 @@ func NewMetaBootstrap(arguments ArgMetaBootstrapper) (*MetaBootstrap, error) {
 		uint64Converter:     arguments.Uint64Converter,
 		poolsHolder:         arguments.PoolsHolder,
 		indexer:             arguments.Indexer,
+		isInImportMode:      arguments.IsInImportMode,
+	}
+
+	if arguments.IsInImportMode {
```
SebastianMarian, line 66: Maybe to be sure that isImportMode was correctly or was not missed to be set with the value of argument.IsImportMode you can check directly with isInImportMode in this condition

process/sync/shardblock.go
```diff
@@ -57,6 +57,11 @@ func NewShardBootstrap(arguments ArgShardBootstrapper) (*ShardBootstrap, error)
 		uint64Converter:     arguments.Uint64Converter,
 		poolsHolder:         arguments.PoolsHolder,
 		indexer:             arguments.Indexer,
+		isInImportMode:      arguments.IsInImportMode,
+	}
+
+	if arguments.IsInImportMode {
```
SebastianMarian, line 63: The same here

process/sync/metablock.go
iulianpascalau: done

process/sync/shardblock.go
iulianpascalau: done

SebastianMarian (approved):

process/sync/baseSync.go
```diff
@@ -632,6 +632,15 @@ func (boot *baseBootstrap) syncBlock() error {
 	return nil
 }
 
+func (boot *baseBootstrap) shouldSync() bool {
+	if boot.isInImportMode {
+		return true
+	}
+
+	nodeState := boot.GetNodeState()
```
sasurobert, line 640: nice!

sasurobert (dismissed):

sasurobert (approved):

SebastianMarian (approved):

LucianMincu (approved):

PR #3179, opened 2021-06-15
Trie nodes chunks processor integration

iulianpascalau:
- integrated the trie nodes chunks processor in interceptors

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -350,6 +350,12 @@
     Type = "SizeLRU"
     SizeInBytes = 104857600 #100MB
 
+[TrieNodesChunksDataPool]
+    Name = "TrieNodesDataPool"
+    Capacity = 400
+    Type = "SizeLRU"
+    SizeInBytes = 104857600 #100MB
+
 [SmartContractDataPool]
     Name = "SmartContractDataPool"
     Capacity = 900000
--- a/dataRetriever/dataPool/dataPool.go
+++ b/dataRetriever/dataPool/dataPool.go
@@ -16,6 +16,7 @@ type dataPool struct {
 	miniBlocks           storage.Cacher
 	peerChangesBlocks    storage.Cacher
 	trieNodes            storage.Cacher
+	trieNodesChunks      storage.Cacher
 	currBlockTxs         dataRetriever.TransactionCacher
 	smartContracts       storage.Cacher
 }
@@ -29,6 +30,7 @@ func NewDataPool(
 	miniBlocks storage.Cacher,
 	peerChangesBlocks storage.Cacher,
 	trieNodes storage.Cacher,
+	trieNodesChunks storage.Cacher,
 	currBlockTxs dataRetriever.TransactionCacher,
 	smartContracts storage.Cacher,
 ) (*dataPool, error) {
@@ -57,6 +59,9 @@ func NewDataPool(
 	if check.IfNil(trieNodes) {
 		return nil, dataRetriever.ErrNilTrieNodesPool
 	}
+	if check.IfNil(trieNodesChunks) {
+		return nil, dataRetriever.ErrNilTrieNodesChunksPool
+	}
 	if check.IfNil(smartContracts) {
 		return nil, dataRetriever.ErrNilSmartContractsPool
 	}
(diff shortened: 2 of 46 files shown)


cmd/node/config/config.toml
```diff
@@ -350,6 +350,12 @@
     Type = "SizeLRU"
     SizeInBytes = 104857600 #100MB
 
+[TrieNodesChunksDataPool]
+    Name = "TrieNodesDataPool"
```
BeniaminDrasovean, line 354: There is another data pool named "TrieNodesDataPool". Is this intentional?

process/interceptors/processor/chunk/chunk.go
```diff
@@ -37,8 +37,7 @@ func (c *chunk) Put(chunkIndex uint32, buff []byte) {
 func (c *chunk) TryAssembleAllChunks() []byte {
 	gotAllParts := c.maxChunks > 0 && len(c.data) == int(c.maxChunks)
 	if !gotAllParts {
-		//TODO(iulian) remove this log
-		log.Warn("not all parts got", "max chunks", c.maxChunks, "len chunk", len(c.data))
+		log.Trace("not all parts got", "max chunks", c.maxChunks, "len chunk", len(c.data))
```
BeniaminDrasovean, line 40: "not all parts received" instead of "not all parts got"?

process/factory/containers/interceptorsContainer.go
```diff
@@ -123,6 +125,22 @@ func (ic *interceptorsContainer) Iterate(handler func(key string, interceptor pr
 	}
 }
 
+// Close will call the close method on all contained interceptors
+func (ic *interceptorsContainer) Close() error {
+	var errFound error
+	ic.Iterate(func(key string, interceptor process.Interceptor) bool {
```
bogdan-rosianu, line 131: using the generic mutex map seems to be 5 times slower for the iterating purpose because of the needed casts and getters.

``` GOLANG
func BenchmarkMutexMap(b *testing.B) {
	mm := NewMutexMap()

	for i := 0; i < 100; i++ {
		mm.Insert(fmt.Sprintf("key-%d", i), "value")
	}

	for i := 0; i < b.N; i++ {
		keys := mm.Keys()
		for _, keyVal := range keys {
			key, ok := keyVal.(string)
			if !ok {
				continue
			}

			_, ok = mm.Get(key)
			if !ok {
				continue
			}
		}
	}
}

func BenchmarkFixedTypeMap(b *testing.B) {
	mm := make(map[string]string)
	mt := sync.RWMutex{}

	for i := 0; i < 100; i++ {
		mt.Lock()
		mm[fmt.Sprintf("key-%d", i)] = "value"
		mt.Unlock()
	}

	for i := 0; i < b.N; i++ {
		mt.RLock()
		for range mm {

		}
		mt.RUnlock()
	}
}
```
```
// BenchmarkMutexMap-12    	  220370	      5368 ns/op
// BenchmarkFixedTypeMap-12    	 1000000	      1016 ns/op
```

It can remain the same because there aren't many interceptors, but for other use cases I'd say we should use a specific map with a specific mutex

bogdan-rosianu (changes requested):

process/factory/containers/interceptorsContainer.go
iulianpascalau: interesting comment :+1: . Will add a TODO to change the implementations. We have have several implementations like this. Added task EN-10283

cmd/node/config/config.toml
iulianpascalau: yes, the first one will keep the complete trie nodes, this one will keep the partial trie nodes along with the chunk information

process/interceptors/processor/chunk/chunk.go
iulianpascalau: done

dataRetriever/dataPool/dataPool.go
```diff
@@ -29,6 +30,7 @@ func NewDataPool(
 	miniBlocks storage.Cacher,
 	peerChangesBlocks storage.Cacher,
 	trieNodes storage.Cacher,
+	trieNodesChunks storage.Cacher,
```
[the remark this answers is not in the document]
iulianpascalau: extracted now

process/factory/containers/interceptorsContainer.go
iulianpascalau: Will remain as it is because of the duplicated code reduction reason. Would be fixed when we will migrate to the golang variant that will support generics.

BeniaminDrasovean (approved):

bogdan-rosianu (approved):

PR #3181, opened 2021-06-16
Docker image improvement

huttarichard:
Improvements for docker image:

- Got rid of `grep arwen-wasm-vm` replaced with `go list -f '{{.Dir}}`. This is better control over what version is used.
- Added `-w -s`, -s  disable symbol table, -w disable DWARF generation. Usually this is not needed. But might be handy if you want to use for example gdb or pprof. Which I assume nobody will in docker.
- Removed some not-needed instructions. 
- We might use alpine linux to make image even smaller. 

```
docker build -f docker/elrond/Dockerfile -t elrdnode .
docker run -it elrdnode
```

Please verify the changes.

```
# ldd /usr/bin/node
	linux-vdso.so.1 (0x00007ffc43dec000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f5b083cd000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f5b0802f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5b07e10000)
	libwasmer_linux_amd64.so => /lib/libwasmer_linux_amd64.so (0x00007f5b0799c000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5b07784000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5b07393000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5b08756000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5b0718f000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f5b06f87000)
```

There is `libwasmer_linux_amd64` properly linked, so I assume it should be fine.
```
libwasmer_linux_amd64.so => /lib/libwasmer_linux_amd64.so (0x00007f5b0799c000)
```


docker/elrond/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+# Elrond node
 WORKDIR /go/elrond-go/cmd/node
-RUN go build -i -v -ldflags="-X main.appVersion=$(git describe --tags --long --dirty)"
-RUN cp /go/pkg/mod/github.com/!elrond!network/arwen-wasm-vm@$(cat /go/elrond-go/go.mod | grep arwen-wasm-vm | sed 's/.* //' | tail -n 1)/wasmer/libwasmer_linux_amd64.so /lib/libwasmer_linux_amd64.so
+RUN go build -i -v -ldflags="-w -s -X main.appVersion=$(git describe --tags --long --dirty)"
+
+# Build elrond vm
 WORKDIR /go/elrond-go
-RUN go get github.com/ElrondNetwork/arwen-wasm-vm/cmd/arwen@$(cat /go/elrond-go/go.mod | grep arwen-wasm-vm | sed 's/.* //' | head -n 1)
+RUN cp $(go list -f '{{.Dir}}' -m github.com/ElrondNetwork/arwen-wasm-vm/v1_3)/wasmer/libwasmer_linux_amd64.so /lib/libwasmer_linux_amd64.so
```
miiu96, line 14: Can you let this line as it is, because it is required to use the latest version of `libwasmer_linux_amd64.so` ( you hardcoded `arwen-wasm-vm/v1_3` and after a while, we will have v1_4 and so on, and this will have to be changed manually).

miiu96 (changes requested):

docker/elrond/Dockerfile
huttarichard: Please can this be explained to me? Im having trouble to get why is this rather complicated.

You have one module imported twice. Why?

- `github.com/ElrondNetwork/arwen-wasm-vm`
- `github.com/ElrondNetwork/arwen-wasm-vm/v1_3`

I might think of better solution. The reason why I propose these changes was to get rid of grep. As this could resolve into literally anything having `arwen-wasm-vm` in a name. 

You want to be able to control this bahaviour, otherwise wrong libraris might be linked and you are not going to spot it at docker build time.

huttarichard: why not to use just `github.com/ElrondNetwork/arwen-wasm-vm` alone?

huttarichard: it normal in wild to use `/v1` `/v2` in path as it will indicate braking changes, but is it necessary to use it with `v1_3`, `v1_4`

miiu96: @camilbancioiu can you provide more explanations here, please.

camilbancioiu: > Please can this be explained to me? Im having trouble to understand why is this rather complicated.

Both versions need to be imported, because the versions will be switched at runtime, at the discretion of the node. Moreover, more Arwen versions will be imported simultaneously as development goes on.

> As this could resolve into literally anything having arwen-wasm-vm in a name.

The essential part is the final `head -n 1`, which specifically takes the first match of `arwen-wasm-vm` and no other. It is unambiguous.

> it normal in wild to use /v1 /v2 in path as it will indicate braking changes, but is it necessary to use it with v1_3, v1_4, minor component?

The difference between `arwen-wasm-vm` (v1.2) and `arwen-wasm-vm/v1_3` is a breaking change. Creating v1.4 would also imply a breaking change. However, smart contracts will see no difference between v1.2, v1.3 and the future v1.4. This is why v1 remains.

Either way, the versioning scheme for Arwen (and package naming) is subject to change in the near-future, but could not be simplified just yet. The new, simplified scheme is being developed as we speak. 

> You want to be able to control this bahaviour, otherwise wrong libraris might be linked and you are not going to spot it at docker build time.

The command to compile Arwen is only relevant to the very first version mentioned in `go.mod` and no other. That is why there's `head -n 1` after the `grep`. No other version of Arwen needs to be compiled into an executable.

If you really want to get rid of `grep` for some reason, please make sure the correct `libwasmer` will be used. And make sure that no other compilation is attempted except for `arwen-wasm-vm`. Not for `arwen-wasm-vm/v1_3` or any future one.

Also, if you somehow hard-code `arwen-wasm-vm` in your Docker file, there may be issues when we fix the versioning scheme. We are aiming to write `arwen-wasm-vm/v1_2` and `arwen-wasm-vm/v1_3` in `go.mod`, but the Docker file must know how to select only `arwen-wasm-vm/v1_2` for compilation to binary, which means it will need to be updated. Or keep `grep | sed | head -n 1`, which automatically selects the first `arwen-wasm-vm`, regardless of the versioning scheme.

Hope this helps.

huttarichard: @camilbancioiu 

> The essential part is the final head -n 1, which specifically takes the first match of arwen-wasm-vm and no other. It is unambiguous.

Putting replace directive (go.mod) at beginning with `arwen-wasm-vm` in name will create the issue.

> Either way, the versioning scheme for Arwen (and package naming) is subject to change in the near-future, but could not be simplified just yet. The new, simplified scheme is being developed as we speak.

I'll wait then, I'll be happy to contribute even more.

huttarichard: @camilbancioiu thanks.

PR #3201, opened 2021-06-22
added ec benchmark

andrei-diaconescu:
for benchmark I needed a lot of different input data which was obtained by using certain function.

For example, before using AddEc/IsOnCurveEc() I used the Double() function before each Add/IsOnCurveEc() call to generate diferent parameters each time.

For Unmarshal or UnmarshalCompressed() I used the Marshal() first to get the data byte array first then put it as input in Unmarshal.


sasurobert (dismissed):

integrationTests/vm/arwen/testdata/c-api-tests/ecBenchmark/ecBenchmark.c
```diff
@@ -0,0 +1,461 @@
...
+    int p521Handle = p521Ec();
+    int p521XBasePointHandle = bigIntNew(0);
+    int p521YBasePointHandle = bigIntNew(0);
+    ellipticCurveGetValues(p521Handle,fieldOrderHandle,basePointOrderHandle,eqConstantHandle,p521XBasePointHandle,p521YBasePointHandle);
+
+    int i;
+    for (i = 0; i < LIMIT/4; i++)
+    {
+        addEC(p224XBasePointHandle,p224YBasePointHandle,p224Handle,p224XBasePointHandle,p224YBasePointHandle,p224XBasePointHandle,p224YBasePointHandle);
+    }
```
alyn509, line 76: you can eventually extract this for in a function. not the most pleasant thing seeing it repeated 4 times

integrationTests/vm/arwen/testdata/c-api-tests/ecBenchmark/ecBenchmark.c
```diff
@@ -0,0 +1,461 @@
...
+        doubleEC(p256XBasePointHandle,p256YBasePointHandle,p256Handle,p256XBasePointHandle,p256YBasePointHandle);
+    }
+    for ( i = 0; i < LIMIT/4; i++)
+    {
+        doubleEC(p384XBasePointHandle,p384YBasePointHandle,p384Handle,p384XBasePointHandle,p384YBasePointHandle);
+    }
+    for ( i = 0; i < LIMIT/4; i++)
+    {
+        doubleEC(p521XBasePointHandle,p521YBasePointHandle,p521Handle,p521XBasePointHandle,p521YBasePointHandle);
+    }
```
alyn509, line 129: same as above

integrationTests/vm/arwen/testdata/c-api-tests/ecBenchmark/ecBenchmark.c
alyn509: you can try something like this with a lambda: https://stackoverflow.com/questions/61329542/how-to-pass-a-lambda-function-as-a-parameter

PR #3204, opened 2021-06-22
Refactor state snapshot

BeniaminDrasovean:
On state checkpoint and snapshot, when traversing the main trie to verify what nodes need to be commited to the snapshot DB, return all leaves found on a channel, so that subsequent data tries can be commited directly, without another main trie traversing.

### DIFF
--- a/data/state/accountsDB.go
+++ b/data/state/accountsDB.go
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	createNewDb   = true
-	useExistingDb = false
+	createNewDb       = true
+	useExistingDb     = false
+	leavesChannelSize = 100
 )
 
 var numCheckpointsKey = []byte("state checkpoint")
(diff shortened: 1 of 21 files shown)


data/state/accountsDB.go
```diff
@@ -962,8 +962,9 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
 	go func() {
 		stopWatch := core.NewStopWatch()
 		stopWatch.Start("snapshotState")
-		trieStorageManager.TakeSnapshot(rootHash, createNewDb)
-		adb.snapshotUserAccountDataTrie(rootHash, true)
+		leavesChannel := make(chan core.KeyValueHolder, 100)
```
bogdan-rosianu, line 965: why 100? can you extract a constant at least?

data/trie/leafNode.go
```diff
@@ -139,19 +144,49 @@ func (ln *leafNode) commitCheckpoint(_ data.DBWriteCacher, targetDb data.DBWrite
...
 
+	err = sendNodeOnChannel(ln, leavesChan)
+	if err != nil {
+		return err
+	}
+
 	return encodeNodeAndCommitToDB(ln, targetDb)
 }
 
+func sendNodeOnChannel(ln *leafNode, leavesChan chan core.KeyValueHolder) error {
```
bogdan-rosianu, line 174: rename to `writeNodeOnChannel`?

bogdan-rosianu (changes requested):

data/state/accountsDB.go
BeniaminDrasovean: Copied it from GetAllLeavesOnChannel :D  Extracted it in a const.

data/trie/leafNode.go
BeniaminDrasovean: Done.

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #3218, opened 2021-06-25
events driver integration

mateigraura:
This pr handles the integration of the [event notifier driver](https://github.com/ElrondNetwork/notifier-go)


sasurobert (approved):

cmd/node/config/external.toml
```diff
@@ -13,3 +13,22 @@
     # EnabledIndexes represents a slice of indexes that will be enabled for indexing. Full list is:
     # ["tps", "rating", "transactions", "blocks", "validators", "miniblocks", "rounds", "accounts", "accountshistory"]
     EnabledIndexes    = ["tps", "rating", "transactions", "blocks", "validators", "miniblocks", "rounds", "accounts", "accountshistory"]
+
+# EventNotifierConnector defines settings needed to configure and launch the event notifier component
+[EventNotifierConnector]
+    # Enabled signals whether the driver is enabled or not
```
iulianpascalau, line 19: this is a config. I think "signals" can easily confuse people. Might rename to 
 `# Enabled will turn on or off the event notifier connector`
valid for all comments.
Also, this is enabled. What happens if the connector can not connect to the ProxyUrl? Does it errors? The validators should need to have this enabled or not?

go.sum
```diff
@@ -69,6 +71,12 @@ github.com/ElrondNetwork/elrond-go v1.1.38-0.20210322140841-6a27c533e8a5/go.mod
 github.com/ElrondNetwork/elrond-go v1.1.42/go.mod h1:mxhdRRV7PY+IHzznJ3MnX5/2YSMNxl2znD6f6Y8nuCM=
 github.com/ElrondNetwork/elrond-go v1.1.51/go.mod h1:0gT8uLRXSqmKDrrn55aXTSca0XGpg8aszx+Q5OPMNpY=
 github.com/ElrondNetwork/elrond-go v1.1.59-0.20210526130950-2a93e2e11c39/go.mod h1:0aZfahOLX4vO+1D6Ejx8XwOLTzd7pfcx94xzAOpt2bs=
+github.com/ElrondNetwork/elrond-go v1.1.64-0.20210617123526-cf2f884b9cec/go.mod h1:bGnt6NgYeLm5tkbUhqjsUvH13hZpxXEdyl2Nd5XMiVg=
```
iulianpascalau, line 74: go mod tidy?

outport/factory/outportFactory.go
```diff
@@ -3,10 +3,16 @@ package factory
 import (
 	indexerFactory "github.com/ElrondNetwork/elastic-indexer-go/factory"
 	"github.com/ElrondNetwork/elrond-go/outport"
+	notifierFactory "github.com/ElrondNetwork/notifier-go/factory"
 )
 
+type OutportFactoryArgs struct {
```
iulianpascalau, line 9: missing comments on exported stuff

iulianpascalau (commented):
minor stuff

go.sum
mateigraura: done

outport/factory/outportFactory.go
mateigraura: done

cmd/node/config/external.toml
mateigraura: Changed the comment. Set the `Enabled` flag to `false`, it shouldn't be active by default, good catch 👍🏻 
Validators don't need this enabled, the events driver is meant to be used with an observer. In case it can't connect to the ProxyUrl, it simply logs the error but doesn't break the node, as it simply returns.

iulianpascalau: Great. :+1: Good to know!

iulianpascalau (approved):

sasurobert (approved):

PR #3230, opened 2021-06-30
EN-10317: wrong config message for full archive + validator

bogdan-rosianu:
Implemented a mechanism that detects if a node is in the same time a validator  + has full archive flag set to true. This is an invalid setup so after all components are closed, an infinite loop printing the configuration issue will be triggered.

Testing procedure:
- start a testnet
- at some point, go to a validator node, stop it, set `FullArchive = true` inside prefs.toml and start it again. It should stop processing and responding to Rest API and multiple logs containg `validator cannot be a full archive node` should be observed. If reset the `FullArchive` to `false`, the node should advance
- the rest of the network should work as usual


AdoAdoAdo (approved):

raduchis (dismissed):

raduchis (approved):

gabi-vuls (approved):
System test passed.

AdoAdoAdo (approved):

PR #3234, opened 2021-07-01
added timestamp in get tx by hash endpoint

bogdan-rosianu:
added an additional field `timestamp` in the endpoint that returns the transaction based on hash

Testing procedure:
- use proxy from branch `logs-transaction-and-update` 
- start a testnet with `DbLookupExtensions.Enabled` set to `true` inside `config.toml`

Acceptance criteria:
- get the hash of some transactions and use `<proxy>/transaction/<hash>` . The timestamp field should be filled and match the block's timestamp


miiu96 (approved):

raduchis (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3241, opened 2021-07-07
integrate return call after error

sasurobert:
Treat scr calls differently when there are return calls with error.


iulianpascalau (commented):
Please add a description

SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3247, opened 2021-07-08
BlockChainHook return same pointer on LoadAccounts

sasurobert:
return same pointer if bytes are equal on load account


bogdan-rosianu (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #3275, opened 2021-07-16
Update Arwen with hotfix

camilbancioiu:
This PR updates dependency on Arwen v1.3.25.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/ElrondNetwork/arwen-wasm-vm/v1_2 v1.2.26
-	github.com/ElrondNetwork/arwen-wasm-vm/v1_3 v1.3.24
+	github.com/ElrondNetwork/arwen-wasm-vm/v1_3 v1.3.25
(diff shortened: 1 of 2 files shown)


sasurobert (commented):
go mod tidy - need to see changes in go.sum as well.

sasurobert (approved):

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/ElrondNetwork/arwen-wasm-vm/v1_2 v1.2.26
-	github.com/ElrondNetwork/arwen-wasm-vm/v1_3 v1.3.24
+	github.com/ElrondNetwork/arwen-wasm-vm/v1_3 v1.3.25
```
iulianpascalau, line 7: does the fix in arwen is backwards compatible?

iulianpascalau: According to the latest commit, it looks like it should not break the backwards compatibility.

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3291, opened 2021-07-23
additional logs for gin server starting and closing

bogdan-rosianu:
added additional logs when closing gin webserver


api/gin/webServer.go
```diff
@@ -281,6 +281,10 @@ func (ws *webServer) Close() error {
 	err := ws.httpServer.Close()
 	ws.Unlock()
 
+	if err != nil {
+		log.Warn("cannot close http server", "error", err)
```
iulianpascalau, line 285: or, we can wrap the error here (since it is already logged)
like
```GO
if err != nil {
    err = fmt.Errorf("%w while closing the http server in gin/webServer", err)
}
```

SebastianMarian (approved):

api/gin/webServer.go
bogdan-rosianu: done

iulianpascalau (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3294, opened 2021-07-23
Make install proto

miiu96:
Added an extra command in the Makefile in order to install protobuff
This new command `make install-proto`  will download an prepare protobuff dependency in order to can run `go generate ./...`


install-proto.sh
```diff
@@ -0,0 +1,41 @@
+mkdir "${GOPATH}"/src/github.com/gogo
+
```
iulianpascalau, line 2: so many empty lines

install-proto.sh
```diff
@@ -0,0 +1,41 @@
...
+cd protobuf
+
+git fetch origin pull/659/head:casttypewith
+
+git checkout casttypewith
+
+PROTO_COMPILER=protoc-3.17.3-linux-x86_64.zip
+TEMP_FOLDER_NAME=temp-001
+
+mkdir ~/${TEMP_FOLDER_NAME}
```
iulianpascalau, line 16: `mkdir /tmp/${TEMP_FOLDER_NAME}` instead. Also on L18 & L29. Will this make the script fail if the temp folder exists?

install-proto.sh
```diff
@@ -0,0 +1,41 @@
...
+git clone https://github.com/gogo/protobuf.git
+
+cd protobuf
+
+git fetch origin pull/659/head:casttypewith
+
+git checkout casttypewith
+
+PROTO_COMPILER=protoc-3.17.3-linux-x86_64.zip
+TEMP_FOLDER_NAME=temp-001
```
iulianpascalau, line 14: `temp-proto-001`

SebastianMarian (approved):

install-proto.sh
miiu96: removed.

install-proto.sh
miiu96: refactored.

install-proto.sh
miiu96: done.

iulianpascalau (approved):

iulianpascalau (commented):
Maybe update the readme.md (workflow + makefile run) ?

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #3306, opened 2021-07-27
trigger registry versioned header support

AdoAdoAdo:
Fixed trigger registry marshal/unmarshal to take into account the header version

TODO:
the hardfork export/import needs also to implement the versioned header instantiation.


epochStart/shardchain/triggerRegistry.go
```diff
@@ -55,6 +52,26 @@ func (t *trigger) LoadState(key []byte) error {
 	return nil
 }
 
+// UnmarshalTrigger unmarshalls the trigger
+func (t *trigger) UnmarshalTrigger(data []byte) (*TriggerRegistry, error) {
+	state := &TriggerRegistry{
+		EpochStartShardHeader: &block.Header{},
```
SebastianMarian, line 58: Maybe first we can try with &block.HeaderV2{} here and in line 66 with &block.Header{}, exactly as we did in CreateShardHeader, as after the point of the activation epoch, it should work with only one unmarshal

AdoAdoAdo: done

SebastianMarian (approved):

raduchis (approved):

raduchis (approved):

epochStart/shardchain/triggerRegistry.go
```diff
@@ -103,15 +103,15 @@ func (t *trigger) saveState(key []byte) error {
...
-	_ = registry.SetMetaEpoch(t.metaEpoch)
-	_ = registry.SetEpoch(t.epoch)
-	_ = registry.SetCurrentRoundIndex(t.currentRoundIndex)
-	_ = registry.SetEpochStartRound(t.epochStartRound)
-	_ = registry.SetEpochMetaBlockHash(t.epochMetaBlockHash)
-	_ = registry.SetIsEpochStart(t.isEpochStart)
-	_ = registry.SetNewEpochHeaderReceived(t.newEpochHdrReceived)
-	_ = registry.SetEpochFinalityAttestingRound(t.epochFinalityAttestingRound)
-	_ = registry.SetEpochStartHeaderHandler(t.epochStartShardHeader)
+	err := registry.SetMetaEpoch(t.metaEpoch)
```
SebastianMarian: You can use _ instead err here as it is not used below anyway

SebastianMarian (approved):

PR #3321, opened 2021-08-01
New crypto repo reference & multisigner mocks cleanup

iulianpascalau:
- new crypto version
- multisig stubs/mocks cleanup


miiu96 (dismissed):

testscommon/cryptoMocks/multisignerMock.go
```diff
@@ -0,0 +1,237 @@
...
+	aggSig      []byte
+	aggCom      []byte
+	commSecret  []byte
+	commHash    []byte
+	commitments [][]byte
+	sigs        [][]byte
+	pubkeys     []string
+	selfId      uint16
+
+	VerifyCalled                           func(msg []byte, bitmap []byte) error
```
bogdan-rosianu, line 20: this is a combination of a mock and a stub. we have a separate stub, so please remove the handler functions and leave it as a mock only.

bogdan-rosianu (changes requested):

testscommon/cryptoMocks/multisignerMock.go
iulianpascalau: I know, the thing is that the multisigner is a complex implementation underneath. The tests are written in such way that a complete refactoring will take a while. Will remain as it is for now.

miiu96 (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.
image

PR #3339, opened 2021-08-09
Merge dev in feat

SebastianMarian:
* Merged dev in feat/scheduled-sc-execution


iulianpascalau (changes requested):
You need to rework the merge process because you have only one commit

PR #3341, opened 2021-08-09
[MERGE] dev into feat-scheduled-sc-execution

SebastianMarian:
* Merged dev in feat scheduled sc execution


process/common.go
```diff
@@ -749,3 +749,39 @@ func GetScheduledSCRsFromStorage(
 
 	return mapScheduledSCRs, nil
 }
+
+// CreateShardHeader creates a shard header from the given byte array
+func CreateShardHeader(marshalizer marshal.Marshalizer, hdrBuff []byte) (data.ShardHeaderHandler, error) {
+	hdr, err := CreateHeaderV2(marshalizer, hdrBuff)
```
iulianpascalau, line 755: An idea to be implemented in the feat branch: what if we specify an epoch (defined in the epoch config file) that states: after this epoch, the node accepts and will decode blocks only with this version. In this way the sub-optimal (try block v2, if it fails, try block v1 and so on) will be temporary.

iulianpascalau (approved):

AdoAdoAdo (approved):

PR #3354, opened 2021-08-17
Fix OOM exceptions

iulianpascalau:
- fixed oom by closing the snapshot db's used in snapshot process
- minor code changes around some log prints
- refactored pruningStorerFactory.go so it does not use slice pointers
- bump gin version to v1.7.4

### DIFF
--- a/storage/factory/pruningStorerFactory.go
+++ b/storage/factory/pruningStorerFactory.go
@@ -250,12 +250,14 @@ func (psf *StorageServiceFactory) CreateForShard() (dataRetriever.StorageService
 	store.AddStorer(dataRetriever.ReceiptsUnit, receiptsUnit)
 	store.AddStorer(dataRetriever.TrieEpochRootHashUnit, trieEpochRootHashStorageUnit)
 
-	err = psf.setupDbLookupExtensions(store, &successfullyCreatedStorers)
+	createdStorers, err := psf.setupDbLookupExtensions(store)
+	successfullyCreatedStorers = append(successfullyCreatedStorers, createdStorers...)
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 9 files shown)


gabi-vuls (approved):
System test passed. 
image

sasurobert (approved):

storage/factory/pruningStorerFactory.go
```diff
@@ -250,12 +250,14 @@ func (psf *StorageServiceFactory) CreateForShard() (dataRetriever.StorageService
 	store.AddStorer(dataRetriever.ReceiptsUnit, receiptsUnit)
 	store.AddStorer(dataRetriever.TrieEpochRootHashUnit, trieEpochRootHashStorageUnit)
 
-	err = psf.setupDbLookupExtensions(store, &successfullyCreatedStorers)
+	createdStorers, err := psf.setupDbLookupExtensions(store)
+	successfullyCreatedStorers = append(successfullyCreatedStorers, createdStorers...)
```
bogdan-rosianu, line 254: maybe move the `append` sections after the error check? it's useless to append to a slice if you return an error so the slice you have just appended to is never used

bogdan-rosianu (changes requested):

storage/factory/pruningStorerFactory.go
iulianpascalau: Won't do. Please check the whole function. The `successfullyCreatedStorers` is used in the defer function as to cleanup the already-opened storers. It should stay as it is, otherwise the defer function will not close the storers that could have been opened by one of those 2 functions.

bogdan-rosianu (approved):

miiu96 (approved):

PR #3358, opened 2021-08-18
factory, genesis, process, update: processing type for miniblock header

AdoAdoAdo:


process/block/interceptedBlocks/common.go
```diff
@@ -111,8 +113,9 @@ func checkShardData(sd data.ShardDataHandler, coordinator sharding.Coordinator)
 			return process.ErrInvalidShardId
 		}
 
-		if len(smbh.GetReserved()) > 0 {
-			return process.ErrReservedFieldNotSupportedYet
+		fmt.Println("len =", len(smbh.GetReserved()))
```
SebastianMarian, line 116: Is this print needed ?

process/block/preprocess/transactions.go
```diff
@@ -273,14 +273,40 @@ func (txs *transactions) RestoreBlockDataIntoPools(
 	return txsRestored, nil
 }
 
+func (txs *transactions) isScheduledMode(header data.HeaderHandler, body *block.Body) (bool, error) {
+	if header.GetEpoch() < txs.scheduledMiniBlocksEnableEpoch {
```
SebastianMarian, line 277: We can use here directly !txs.flagScheduledMiniBlocks.IsSet()

PR #3364, opened 2021-08-20
Esdt total supply[WIP]

miiu96:


dblookupext/esdtSupply/esdtSuppliesProcessor_test.go
```diff
@@ -0,0 +1 @@
+package esdtSupply
```
iulianpascalau, line 1: missing tests. This is targeted towards development branch so it should contain tests.

dblookupext/esdtSupply/logsGetter.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return logsDB, nil
+}
+
+func (lg *logsGetter) getLogsBasedOnMB(mb *block.MiniBlock) map[string]data.LogHandler {
+	dbLogs := make(map[string]data.LogHandler)
+	for _, txHash := range mb.TxHashes {
+		txLog, ok := lg.getTxLog(txHash)
+		if !ok {
+			continue
```
iulianpascalau, line 54: by using this mechanism you will just print some warns but fail to correctly process data. This will result in presenting erroneous information. This should be considered processing error and should prevent the advancement of the node

dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -0,0 +1,132 @@
...
+			core.BuiltInFunctionESDTLocalBurn:      {},
+			core.BuiltInFunctionESDTLocalMint:      {},
+			core.BuiltInFunctionESDTNFTCreate:      {},
+			core.BuiltInFunctionESDTNFTAddQuantity: {},
+			core.BuiltInFunctionESDTNFTBurn:        {},
+		},
+	}
+}
+
+func (lp *logsProcessor) processLogs(logs map[string]data.LogHandler, isRevert bool) error {
```
iulianpascalau, line 38: bool param. This function clearly does 2 things. Should split in 2.

dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -0,0 +1,132 @@
...
+				supplies[tokenIDStr] = tokenSupply
+				continue
+			}
+
+			supplies[tokenIDStr].Supply.Add(supplies[tokenIDStr].Supply, tokenSupply.Supply)
+		}
+
+	}
+
+	return lp.saveSupplies(supplies)
```
iulianpascalau, line 76: :+1:

dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (lp *logsProcessor) processLogs(logs map[string]data.LogHandler, isRevert bool) error {
+	supplies := make(map[string]*SupplyESDT)
+	for _, logHandler := range logs {
+		if check.IfNil(logHandler) {
+			continue
+		}
+
+		for _, entryHandler := range logHandler.GetLogEvents() {
```
iulianpascalau, line 45: could have been another function

dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -0,0 +1,132 @@
...
+		}, tokenIdentifier, nil
+	}
+
+	supplyFromStorage := &SupplyESDT{}
+	err = lp.marshalizer.Unmarshal(supplyFromStorage, supplyFromStorageBytes)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newSupply := big.NewInt(0).Add(supplyFromStorage.Supply, bigValue)
```
iulianpascalau, line 122: does this work as expected? What happens if you have the same ESDT token involved on 2 different transactions in the same block? Should use the get from storage only if you haven't already worked with the same ESDT token in the current block

dblookupext/historyRepository.go
```diff
@@ -432,6 +444,10 @@ func (hr *historyRepository) IsEnabled() bool {
 	return true
 }
 
+func (hr *historyRepository) RevertBlock(blockHeader data.HeaderHandler, blockBody data.BodyHandler) error {
```
iulianpascalau, line 447: missing comment

dblookupext/interface.go
```diff
@@ -35,3 +37,9 @@ type BlockTracker interface {
 	RegisterFinalMetachainHeadersHandler(func(shardID uint32, headers []data.HeaderHandler, headersHashes [][]byte))
 	IsInterfaceNil() bool
 }
+
+type SuppliesHandler interface {
```
iulianpascalau, line 41: missing comment

dblookupext/nilHistoryRepository.go
```diff
@@ -36,6 +36,11 @@ func (nhr *nilHistoryRepository) IsEnabled() bool {
 	return false
 }
 
+// RevertBlock -
+func (nhr *nilHistoryRepository) RevertBlock(_ data.HeaderHandler, _ data.BodyHandler) error {
+	return nil
```
iulianpascalau, line 41: either add tests for this or move the implementation in a `disabled` package

process/sync/baseSync.go
```diff
@@ -726,6 +731,13 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
 			)
 		}
 
+		err = boot.historyRepo.RevertBlock(currHeader, currBody)
+		if err != nil {
+			log.Debug("boot.historyRepo.RevertBlock",
```
iulianpascalau, line 736: why isn't this a critical error?

dblookupext/esdtSupply/logsGetter.go
miiu96: I know but in this case, I don't know if a transaction has logs or not.

dblookupext/esdtSupply/logsProcessor.go
miiu96: If I split it into two functions will have a lot of duplicated code.
This function does one thing ---> process a log ---> IF is revert in case of Burn ESDT or NFT will add to supply instead of decrease.

dblookupext/esdtSupply/logsProcessor.go
miiu96: done.

dblookupext/esdtSupply/logsProcessor.go
miiu96: No. Goot catch. Refactored this part and updated unit tests to confirm that is working as expected now

dblookupext/historyRepository.go
miiu96: added comment.

dblookupext/interface.go
miiu96: done.

dblookupext/nilHistoryRepository.go
miiu96: moved.

process/sync/baseSync.go
miiu96: I followed the pattern on the previous operation. Should I make this a critical error?

dblookupext/esdtSupply/esdtSuppliesProcessor_test.go
miiu96: added

dblookupext/esdtSupply/logsGetter.go
iulianpascalau: getTxLog function with L72. That is my main concern. Should refactor so the function will return a struct containing the log handler, error and if it contained logs or not so it should be skipped or not on L53

dblookupext/esdtSupply/logsProcessor.go
iulianpascalau: could have used a "multiplier" factor (1 or -1) instead of bool

process/sync/baseSync.go
iulianpascalau: I think it will be safer to make it critical error. As this might hide processing/storing data errors from the history db lookup extension mechanism.

dblookupext/esdtSupply/logsGetter.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	logsDB := make(map[string]data.LogHandler)
+	for _, mb := range body.MiniBlocks {
+		shouldIgnore := mb.Type != block.TxBlock && mb.Type != block.SmartContractResultBlock
+		if shouldIgnore {
+			continue
+		}
+
+		dbLogsMb := lg.getLogsBasedOnMB(mb)
```
sasurobert, line 41: why do we need this ? in case of indexer - the node already prepares the complete logs on commitBlock. can't we use the input as full logs directly ?

dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (lp *logsProcessor) processEvent(txLog *transaction.Event, supplies map[string]*SupplyESDT, isRevert bool) error {
+	tokenIdentifier := txLog.Topics[0]
```
sasurobert, line 95: add check for length of txLog.Topics - this can panic

sasurobert (changes requested):

dblookupext/interface.go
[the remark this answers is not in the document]
miiu96: done.

dblookupext/esdtSupply/logsGetter.go
miiu96: This is needed in case of rollback. When method RevertBlock is called I have `only header and body` based on the transactions hashes from miniblocks I can read the logs from storage and revert the changes for the current header.

dblookupext/esdtSupply/logsProcessor.go
miiu96: added extra check.

process/sync/baseSync.go
miiu96: done.

iulianpascalau (approved):

sasurobert (approved):

PR #3395, opened 2021-08-31
Fix edge case negative fee 

miiu96:
- Bug-fix: fixed the edge case when computed transaction fee was negative


iulianpascalau (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.
The new error is coming from shared cpu machines and is harmless. 
image

PR #3403, opened 2021-09-02
Optimized consensus signature share verification

iulianpascalau:
- optimized consensus signature share verification


consensus/spos/bls/subroundSignature.go
```diff
@@ -158,6 +158,11 @@ func (sr *subroundSignature) receivedSignature(cnsDta *consensus.Message) bool {
 	}
 
 	currentMultiSigner := sr.MultiSigner()
+	err = currentMultiSigner.VerifySignatureShare(uint16(index), cnsDta.SignatureShare, sr.GetData(), nil)
+	if err != nil {
+		return false
```
SebastianMarian, line 163: You can log the error,  just to know which is the error when this method returns false in a critical situation and to keep the pattern used in this method:

		log.Debug("receivedSignature.VerifySignatureShare",
			"index", index,
			"error", err.Error())

iulianpascalau: done

BeniaminDrasovean (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed. 
image

raduchis (approved):
Tested in production and endround time was decreased from ~600 ms to ~100 ms

PR #3410, opened 2021-09-06
Backward compatibility fix on Transfer And Exec

sasurobert:
fix backward compatibility issue


iulianpascalau (approved):
tested with the exact encountered edge-case and found that is fully backwards compatibile. :+1:

SebastianMarian (approved):

gabi-vuls (approved):
System test passed 
image

PR #3411, opened 2021-09-06
Fix trie sync timeout when the node is building the list of data tries to be synced

iulianpascalau:
- fixed trie sync timeout false positive trigger by resetting the watchdog when the node builds up its list of data tries to be synced


sasurobert (approved):

BeniaminDrasovean (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3413, opened 2021-09-06
Feat/web server refactor

bogdan-rosianu:
Refactored the gin web server by using endpoints' handlers as functions of a structure instead of simple exported functions. This brings a small speed improvement, and, on top of that, the gin server restarting needed for "node is starting..." message is no longer needed, as the `facade` interface will be updated directly.


iulianpascalau (approved):

gabi-vuls (approved):
System test passed. 
image

sasurobert (approved):

PR #3430, opened 2021-09-09
EN-10574 Get block by round

mariusmihaic:


dblookupext/factory/historyRepositoryFactory_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/ElrondNetwork/elrond-go/config"
 	"github.com/ElrondNetwork/elrond-go/dataRetriever"
 	"github.com/ElrondNetwork/elrond-go/dblookupext/factory"
+	"github.com/ElrondNetwork/elrond-go/process"
+	mock2 "github.com/ElrondNetwork/elrond-go/process/mock"
```
miiu96, line 13: mock2 ?

node/blockAPI/metaBlock.go
```diff
@@ -65,6 +65,21 @@ func (mbp *metaAPIBlockProcessor) GetBlockByHash(hash []byte, withTxs bool) (*ap
 	return mbp.computeStatusAndPutInBlock(blockAPI, dataRetriever.MetaHdrNonceHashDataUnit)
 }
 
+// GetBlockByRound will return a shard APIBlock by round
```
miiu96, line 68: // GetBlockByRound will return a `meta` APIBlock by round

api/block/routes.go
```diff
@@ -130,6 +172,15 @@ func getQueryParamNonce(c *gin.Context) (uint64, error) {
 	return strconv.ParseUint(nonceStr, 10, 64)
 }
 
+func getQueryParamRound(c *gin.Context) (uint64, error) {
+	roundStr := c.Param("round")
+	if roundStr == "" {
```
iulianpascalau, line 177: lines L177 - L179 can be safely removed

dblookupext/factory/historyRepositoryFactory_test.go
iulianpascalau: Consider replacing `mock2` with `processorMock` alias

dblookupext/historyRepository.go
```diff
@@ -156,9 +167,21 @@ func (hr *historyRepository) RecordBlock(
...
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
+func (hr *historyRepository) putRoundByNonce(header data.HeaderHandler) error {
+	roundToByteSlice := hr.uint64ByteSliceConverter.ToByteSlice(header.GetRound())
+	nonceToByteSlice := hr.uint64ByteSliceConverter.ToByteSlice(header.GetNonce())
```
iulianpascalau, line 180: If we record only the round-nonce pairs, then we will 3 DB accesses to fetch a block (one for retrieving the nonce, one for retrieving the hash from the nonce and the later for retrieving the block from the hash). 
Wondering if we were to record directly the round-hash we would needed only 2 DB accesses.

dblookupext/historyRepository_test.go
```diff
@@ -8,8 +8,11 @@ import (
 	"github.com/ElrondNetwork/elrond-go-core/core"
 	"github.com/ElrondNetwork/elrond-go-core/data"
 	"github.com/ElrondNetwork/elrond-go-core/data/block"
+	mock3 "github.com/ElrondNetwork/elrond-go-core/data/mock"
```
iulianpascalau, line 11: please replace `mock2` and `mock3` with something else

node/blockAPI/shardBlock.go
```diff
@@ -67,6 +67,21 @@ func (sbp *shardAPIBlockProcessor) GetBlockByHash(hash []byte, withTxs bool) (*a
 	return sbp.computeStatusAndPutInBlock(blockAPI, storerUnit)
 }
 
+// GetBlockByRound will return a shard APIBlock by round
+func (sbp *shardAPIBlockProcessor) GetBlockByRound(round uint64, withTxs bool) (*api.Block, error) {
```
iulianpascalau, line 71: duplicated code with metaBlock.go. We usually fix these by creating a base struct with the common functionality and extend it in all implementations.

node/nodeBlocks.go
```diff
@@ -35,6 +35,15 @@ func (n *Node) GetBlockByNonce(nonce uint64, withTxs bool) (*api.Block, error) {
 	return apiBlockProcessor.GetBlockByNonce(nonce, withTxs)
 }
 
+func (n *Node) GetBlockByRound(round uint64, withTxs bool) (*api.Block, error) {
+	apiBlockProcessor, err := n.createAPIBlockProcessor()
```
iulianpascalau, line 39: @miiu96 please check this. Why do we need to create an apiBlockProcessor for **each request**? We should definitely change this.

iulianpascalau (changes requested):

dblookupext/factory/historyRepositoryFactory_test.go
mariusmihaic: Renamed it to `processMock`

node/blockAPI/metaBlock.go
mariusmihaic: Yes, fixed it here and for `GetBlockByHash` as well

api/block/routes.go
mariusmihaic: Removed it from here as well as from `getQueryParamNonce`

dblookupext/historyRepository_test.go
mariusmihaic: Renamed them

node/blockAPI/shardBlock.go
mariusmihaic: Yes. I was aware about it, but couldn't use the same implementation for both meta and shard blocks, because the returned value is given by: `GetBlockByNonce(nonce, withTxs)` -> which has different implementations for meta/shard block.
However, I've extracted some common code in a separate function, which is now part of `baseBlocks`:

```go
// GetBlockByRound will return a shard APIBlock by round
func (bap *baseAPIBlockProcessor) getNonceParamByRound(round uint64, withTxs bool) (uint64, error) {
	roundToByteSlice := bap.uint64ByteSliceConverter.ToByteSlice(round)
	nonceBytes, err := bap.store.Get(dataRetriever.RoundNonceUnit, roundToByteSlice)
	if err != nil {
		return 0, err
	}

	nonce, err := bap.uint64ByteSliceConverter.ToUint64(nonceBytes)
	if err != nil {
		return 0 , err
	}
	return nonce, nil
}
```

dblookupext/historyRepository.go
mariusmihaic: Applied your suggestion and now we save block's hash by round, therefore needing only 2 DB accesses, instead of 3.

node/blockAPI/shardBlock.go
mariusmihaic: The above comment is now obsolete, since the block's hash is now saved by round instead of saving block's round by nonce

node/nodeBlocks.go
mariusmihaic: @miiu96 ?

dataRetriever/interface.go
```diff
@@ -91,8 +91,10 @@ const (
 	TrieEpochRootHashUnit UnitType = 17
 	// ESDTSuppliesUnit is the ESDT supplies storage unit identifier
 	ESDTSuppliesUnit UnitType = 18
-	// RoundNonceUnit is the round nonce storage unit identifier
-	RoundNonceUnit UnitType = 19
+	// MetaHdrRoundHashDataUnit is the meta header round-hash storage data pair unit identifier
+	MetaHdrRoundHashDataUnit UnitType = 49
+	// ShardHdrRoundHashDataUnit is the shard round-hash storage data unit identifier
+	ShardHdrRoundHashDataUnit UnitType = 50
```
iulianpascalau, line 97: let's put this on the 200 value (or even better ~400 as to allow the creation of 256 shards) and place it after the ShardHdrNonceHashDataUnit definition. MetaHdrRoundHashDataUnit can be 19. Adjust the comment below to include the ShardHdrNonceHashDataUnit.

dblookupext/historyRepository_test.go
```diff
@@ -94,12 +94,12 @@ func TestNewHistoryRepository(t *testing.T) {
 	require.NotNil(t, repo)
 }
 
-func TestHistoryRepository_RecordBlock_InvalidBlockRoundByNonceStorer_ExpectError(t *testing.T) {
+func TestHistoryRepository_RecordBlock_InvalidBlockRoundByHashStorer_ExpectError(t *testing.T) {
```
iulianpascalau, line 97: too many _ in name. To align with the rest of the project, please rename this to: 
```
TestHistoryRepository_RecordBlockInvalidBlockRoundByHashStorerExpectError
```

storage/factory/pruningStorerFactory.go
```diff
@@ -519,19 +519,23 @@ func (psf *StorageServiceFactory) setupDbLookupExtensions(chainStorer *dataRetri
...
+	blockHashByRoundBloomFilter := GetBloomFromConfig(blockHashByRoundConfig.Bloom)
+	blockHashByRoundUnit, err := storageUnit.NewStorageUnitFromConf(blockHashByRoundCacherConfig, blockHashByRoundDBConfig, blockHashByRoundBloomFilter)
 	if err != nil {
 		return createdStorers, err
 	}
 
-	createdStorers = append(createdStorers, blockRoundByNonceUnit)
-	chainStorer.AddStorer(dataRetriever.RoundNonceUnit, blockRoundByNonceUnit)
+	createdStorers = append(createdStorers, blockHashByRoundUnit)
+	chainStorer.AddStorer(hashByRoundUnitData, blockHashByRoundUnit)
```
iulianpascalau, line 538: Ok, one question here: why do we need a different value for blockHashByRoundUnit (according to the shard in which the node is operating) but we **always** open the same storer from the same path? This can be simplified by stating a hardcoded value in dataretrieve/interface.go for this unit type (not added to the shard ID value), something like 19 or 20 or something.

iulianpascalau (changes requested):

dblookupext/historyRepository_test.go
mariusmihaic: Renamed it to be consistent with the rest of the tests

iulianpascalau (approved):

miiu96 (approved):

PR #3449, opened 2021-09-17
fix missing pprof endpoint

bogdan-rosianu:
After merging `feat/web-server-refactor` - the pprof route was missing. This PR adds it back.

Testing procedure:
- start a testnet with the `--profile-mode` added to nodes
- `go tool pprof <node-ip>:<node-port>/debug/pprof/goroutine` - should work and not return 404 code

### DIFF
--- a/node/nodeRunner.go
+++ b/node/nodeRunner.go
@@ -29,7 +29,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/dataRetriever"
 	dbLookupFactory "github.com/ElrondNetwork/elrond-go/dblookupext/factory"
 	"github.com/ElrondNetwork/elrond-go/facade"
-	"github.com/ElrondNetwork/elrond-go/facade/disabled"
+	"github.com/ElrondNetwork/elrond-go/facade/initial"
 	mainFactory "github.com/ElrondNetwork/elrond-go/factory"
 	"github.com/ElrondNetwork/elrond-go/genesis/parsing"
 	"github.com/ElrondNetwork/elrond-go/health"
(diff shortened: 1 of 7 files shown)


sasurobert (dismissed):

SebastianMarian (approved):

iulianpascalau (approved):

node/nodeRunner.go
```diff
@@ -491,7 +491,7 @@ func (nr *nodeRunner) createApiFacade(
 
 func (nr *nodeRunner) createHttpServer() (shared.UpgradeableHttpServerHandler, error) {
 	httpServerArgs := gin.ArgsNewWebServer{
-		Facade:          disabled.NewDisabledNodeFacade(nr.configs.FlagsConfig.RestApiInterface),
+		Facade:          initial.NewInitialNodeFacade(nr.configs.FlagsConfig.RestApiInterface, nr.configs.FlagsConfig.EnablePprof),
```
iulianpascalau, line 494: could have provided the whole flags config struct.

SebastianMarian (approved):

node/nodeRunner.go
bogdan-rosianu: no need for this now as it only uses 2 fields of the huge list of flags. maybe if we need another or a few more flags, we can parse the whole struct

sasurobert (dismissed):

node/nodeRunner.go
iulianpascalau: ok

iulianpascalau (approved):

sasurobert (approved):

SebastianMarian (approved):

iulianpascalau (approved):

PR #3452, opened 2021-09-19
Optimize gas used in cross dest me mbs

SebastianMarian:
* Optimized gas consumption in cross shard mini blocks with destination in self shard

**Testing scenario:**

1. Start a normal system test
2. In epoch 2, OptimizeGasUsedInCrossMiniBlocks flag will be activated
3. Start txgen from this branch and with this param:
 _**txgen branch: temp-increase-gas-erc20-esdt
 txgen starting command: ./start.sh start 100 base**_
4. After 1-2 epochs we should check the logs for "gas consumed and refunded info" messages and we should have in some of
them "total gas consumed" higher than 1.5 billion, but "total gas consumed" - "total gas refunded" always lower than 1.5 billion


process/block/preprocess/basePreProcess.go
```diff
@@ -415,3 +415,21 @@ func (bpp *basePreProcess) getTxMaxTotalCost(txHandler data.TransactionHandler)
...
+	if bpp.gasHandler.TotalGasConsumed() > bpp.gasHandler.TotalGasRefunded() {
+		totalGasConsumed = bpp.gasHandler.TotalGasConsumed() - bpp.gasHandler.TotalGasRefunded()
+	}
+
+	return totalGasConsumed
+}
+
+func (bpp *basePreProcess) updateGasConsumedWithGasRefunded(
+	txHash []byte,
+	gasConsumedByMiniBlockInReceiverShard *uint64,
```
iulianpascalau, line 430: let's not use these pointer input arguments as outputs. Maybe create a struct that will wrap these 3 params and pass it as pointer and change the values inside? The function states that an update will occur but using uint pointers is a little bit ugly.

process/block/preprocess/basePreProcess.go
```diff
@@ -415,3 +415,21 @@ func (bpp *basePreProcess) getTxMaxTotalCost(txHandler data.TransactionHandler)
...
 	return cost
 }
+
+func (bpp *basePreProcess) getTotalGasConsumed() uint64 {
+	totalGasConsumed := uint64(0)
+	if bpp.gasHandler.TotalGasConsumed() > bpp.gasHandler.TotalGasRefunded() {
+		totalGasConsumed = bpp.gasHandler.TotalGasConsumed() - bpp.gasHandler.TotalGasRefunded()
+	}
+
+	return totalGasConsumed
```
iulianpascalau, line 425: it is ok to return 0 here if the TotalGasConsumed <= TotalGasRefunded  ?

SebastianMarian: This situation, to refund more than you received to be used, should never occur. But if this will happen because of an introduced bug, at least it is protected to against overflow situation, as both of them are uint64. So you consider that nothing has been processed yet.

process/block/preprocess/basePreProcess.go
SebastianMarian: This is done in a big refactor which has been occurred over this component in scheduled txs feature.

process/block/preprocess/basePreProcess.go
```diff
@@ -415,3 +415,21 @@ func (bpp *basePreProcess) getTxMaxTotalCost(txHandler data.TransactionHandler)
...
+}
+
+func (bpp *basePreProcess) updateGasConsumedWithGasRefunded(
+	txHash []byte,
+	gasConsumedByMiniBlockInReceiverShard *uint64,
+	totalGasConsumedInSelfShard *uint64) {
```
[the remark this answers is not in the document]
SebastianMarian: Ok

sasurobert (changes requested):
We need an activation flag.

iulianpascalau:
> We need an activation flag.

As discussed we need an activation flag because, if the blocks will be full from the first day when we upgrade the network, there will be a lot of missed blocks as some un-upgraded nodes will refuse the proposed blocks by an upgraded proposer. Also, the nodes that take longer to upgrade (exchanges for example) will get un-synced.

sasurobert (approved):

iulianpascalau (approved):

iulianpascalau (approved):

sasurobert (approved):

config/tomlConfig_test.go
```diff
@@ -668,17 +668,22 @@ func TestEnableEpochConfig(t *testing.T) {
 			ESDTTransferRoleEnableEpoch:                 34,
 			BuiltInFunctionOnMetaEnableEpoch:            35,
 			ComputeRewardCheckpointEnableEpoch:          36,
-			ESDTNFTCreateOnMultiShardEnableEpoch:        37,
-			OptimizeGasUsedInCrossMiniBlocksEnableEpoch: 38,
+			SCRSizeInvariantCheckEnableEpoch:            37,
```
iulianpascalau: we really need to refactor this test :D

iulianpascalau (dismissed):

genesis/process/shardGenesisBlockCreator.go
```diff
@@ -472,6 +473,7 @@ func createProcessorsForShardGenesisBlock(arg ArgsGenesisBlockCreator, enableEpo
 	disabledBlockTracker := &disabled.BlockTracker{}
 	disabledBlockSizeComputationHandler := &disabled.BlockSizeComputationHandler{}
 	disabledBalanceComputationHandler := &disabled.BalanceComputationHandler{}
+	disabledEpochNotifier := &disabled.EpochNotifier{}
```
sasurobert, line 476: you should have normal epoch notifier here. as this is called in hardfork as well.

SebastianMarian: done

PR #3482, opened 2021-09-29
P2P bootstrap fix + enable epochs fix in genesis processors

iulianpascalau:
- fixed the kad-dht discoverer to recover faster after a first connection fail to a bunch of seeders, added unit test
- fixed some enable epochs for the genesis processors 

Local tested the import-db with the mainnet DBs for a few hundred blocks, all good :heavy_check_mark: .

Testing scenarios:
- start a node using the `master` branch. do not connect it to a seeder (use seeder address from this branch in p2p.toml). Let it for about 6-7 minutes. Grep the logs for `error connecting to seeder`. You will find multiple entries. The first one is apart form the second for ~5 minutes, the rest are apart of each other  for ~10 seconds.
- repeat the above mentioned procedure, building the node from this branch. Grep from the same string. You will find all entries apart from each other for ~10 seconds.


sasurobert (approved):

sasurobert (approved):

raduchis (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3488, opened 2021-10-01
Added memory ballast flag

bogdan-rosianu:
added a flag for memory ballast optimization

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -23,7 +23,8 @@ const (
 )
 
 var (
-	nodeHelpTemplate = `NAME:
+	memoryBallastObject []byte
+	nodeHelpTemplate    = `NAME:
    {{.Name}} - {{.Usage}}
 USAGE:
    {{.HelpName}} {{if .VisibleFlags}}[global options]{{end}}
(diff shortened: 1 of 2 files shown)


cmd/node/main.go
```diff
@@ -104,6 +105,14 @@ func startNodeRunner(c *cli.Context, log logger.Logger, version string) error {
 		return err
 	}
 
+	memBallastValue := c.GlobalUint64(memBallast.Name)
+	if memBallastValue > 0 {
+		// memory ballast is an optimization for golang's garbage collector. If set to a high value, it can decrease
+		// the number of times when GC performs STW processes, that results is a better performance over high load
+		memoryBallastObject = make([]byte, memBallastValue*core.MegabyteSize)
+		log.Debug("initialized memory ballast object", "size", core.ConvertBytes(uint64(len(memoryBallastObject))))
```
iulianpascalau, line 113: I think the compiler might optimize the code and remove the memBallastValue after this call. Might add a new print on L132 that uses memBallastValue. The node might call `runtime.GC` after shuffling out a shard. This call should theoretically remove the `memoryBallastObject` object.

sasurobert (approved):

iulianpascalau (approved):

cmd/node/main.go
bogdan-rosianu: added some prints and tested locally. The object won't be removed/freed

PR #3490, opened 2021-10-04
Trie sync oom fix

BeniaminDrasovean:
Fixed a possible oom situation: if there are a lot of dataTries, the array that contains all the rootHashes of the dataTries can lead to an oom issue. To fix this, the array has been removed, and each dataTrie is synced as soon as we get access to it's rootHash.

### DIFF
--- a/state/syncer/userAccountsSyncer.go
+++ b/state/syncer/userAccountsSyncer.go
@@ -105,69 +105,7 @@ func (u *userAccountsSyncer) SyncAccounts(rootHash []byte) error {
 
 	log.Debug("main trie synced, starting to sync data tries", "num data tries", len(u.dataTries))
 
-	rootHashes, err := u.findAllAccountRootHashes(mainTrie)
-	if err != nil {
-		return err
-	}
-
-	err = u.syncAccountDataTries(rootHashes, tss, ctx)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
-
-func (u *userAccountsSyncer) syncAccountDataTries(rootHashes [][]byte, ssh data.SyncStatisticsHandler, ctx context.Context) error {
-	var errFound error
-	errMutex := sync.Mutex{}
-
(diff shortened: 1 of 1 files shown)


state/syncer/userAccountsSyncer.go
```diff
@@ -229,13 +174,65 @@ func (u *userAccountsSyncer) findAllAccountRootHashes(mainTrie common.Trie) ([][
...
+		if len(account.RootHash) == 0 {
+			continue
+		}
+
+		err = u.checkGoRoutinesThrottler(ctx)
+		if err != nil {
+			return err
+		}
+
+		//throttler.StartProcessing is required to be here as it prevent the following edge-case:
```
iulianpascalau, line 186: comments no longer required

state/syncer/userAccountsSyncer.go
```diff
@@ -229,13 +174,65 @@ func (u *userAccountsSyncer) findAllAccountRootHashes(mainTrie common.Trie) ([][
...
+				errMutex.Lock()
+				errFound = newErr
+				errMutex.Unlock()
+			}
+			atomic.AddInt32(&u.numTriesSynced, 1)
+			log.Trace("finished sync data trie", "roothash", trieRootHash)
+			wg.Done()
+		}(account.RootHash)
+
+		atomic.AddInt32(&u.numMaxTries, 1)
```
iulianpascalau, line 210: please move this on L194 as it makes more sense

state/syncer/userAccountsSyncer.go
BeniaminDrasovean: Removed.

state/syncer/userAccountsSyncer.go
BeniaminDrasovean: Done.

iulianpascalau (approved):

PR #3535, opened 2021-10-25
Round Activation handler

mariusmihaic:
- Added `enableRound.toml` which contains all round activations (for now -> dummy)
- Added `RoundActivationHandler` interface which can be queried with a specific `event name` + `round`

- Added `roundActivation` struct which implements `RoundActivationHandler`
- Included `roundActivation` in `bootstrapComponents`


AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

process/roundActivation/roundActivation.go
```diff
@@ -3,50 +3,31 @@ package roundActivation
...
-
-func (ra *roundActivation) isCurrentRound(round uint64) bool {
-	return round == uint64(ra.roundHandler.Index())
-}
-
-// IsEnabledInCurrentRound checks if the queried round flag name is enabled in current round
-func (ra *roundActivation) IsEnabledInCurrentRound(name string) bool {
-	if round, exists := ra.roundByNameMap[name]; exists {
-		return ra.isCurrentRound(round)
+	r, exists := ra.roundByNameMap[name]
```
iulianpascalau, line 28: no mutex protection? The method is exposed and this struct is prone to be used in concurrent manner.

iulianpascalau: I know it does only reads but the code is fragile in this way. 
Reason: adding a write **with mutex protection** in a new function but forgetting to add the mutex in all other methods could easily panic the node. And we might encounter that in production environment.

iulianpascalau (approved):

process/roundActivation/roundActivation.go
mariusmihaic: Added mutex

AdoAdoAdo (approved):

sasurobert (approved):

PR #3542, opened 2021-10-26
Disabled checkpointing in import-db

iulianpascalau:
- completely disabled checkpointing in import-db process


valentin-lup (approved):
Import-DB completed without errors

PR #3561, opened 2021-11-04
Further pruning storer fixes - part 3

iulianpascalau:
- more improvements for pruning storer's concurrent operations\
- removed a poorly usage of a second mutex when creating a new instance of a DB
- fixes around the extension of a pruning storer DB: prevented a deadlock


iulianpascalau:
faulty. Opened #3562 instead

PR #3563, opened 2021-11-05
Reject transactions with gas limit higher than max accepted

SebastianMarian:
* Reject transactions with gas limit higher than max accepted

**How to be tested**

Start an #allin system test. In epoch 2 we should send some transactions with gas limit higher than 600k. They should not pass the interceptor and they should not be added in transactions pool.


raduchis (approved):

gabi-vuls (approved):
Scenario passed. 
System test passed. 
image

sasurobert (approved):

iulianpascalau (approved):

PR #3566, opened 2021-11-06
Extend esdt supply api route

miiu96:
Extend the api route `/network/esdt/supply/:token`
Now this API route will return also the `minted` value and `burned` value


dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -125,47 +125,55 @@ func (lp *logsProcessor) processEvent(txLog *transaction.Event, supplies map[str
...
-	supplies[tokenIDStr] = &SupplyESDT{
-		Supply: newSupply,
-	}
+	supplies[tokenIDStr] = supply
+	lp.updateTokenSupply(supplies[tokenIDStr], valueFromEvent, string(txLog.Identifier), isRevert)
 
 	return nil
 }
 
+func (lp *logsProcessor) updateTokenSupply(tokenSupply *SupplyESDT, valueFromEvent *big.Int, eventIdentifier string, isRevert bool) {
```
iulianpascalau, line 148: :+1:

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/ElrondNetwork/concurrent-map v0.1.3
 	github.com/ElrondNetwork/covalent-indexer-go v1.0.3
 	github.com/ElrondNetwork/elastic-indexer-go v1.0.11
-	github.com/ElrondNetwork/elrond-go-core v1.1.2
+	github.com/ElrondNetwork/elrond-go-core v1.1.4-0.20211108091911-651af340e6ef
```
iulianpascalau, line 12: proper release

node/node.go
```diff
@@ -482,8 +482,17 @@ func (n *Node) GetESDTsRoles(address string) (map[string][]string, error) {
...
-func (n *Node) GetTokenSupply(token string) (string, error) {
-	return n.processComponents.HistoryRepository().GetESDTSupply(token)
+func (n *Node) GetTokenSupply(token string) (*api.ESDTSupply, error) {
+	esdtSupply, err := n.processComponents.HistoryRepository().GetESDTSupply(token)
+	if err != nil {
+		return nil, err
+	}
+
+	return &api.ESDTSupply{
+		Supply: esdtSupply.Supply.String(),
```
iulianpascalau, line 492: this works because how it is implemented right now. Since we are talking about 2 different components, I would put some nil checks for all fields (someone might change the behavior of the `HistoryRepository().GetESDTSupply` implementation in the future and the node will simply panic)

go.mod
miiu96: done.

bogdan-rosianu (dismissed):

node/node.go
miiu96: `esdtSupply.Supply.String() `
if esdtSupply.Supply is nil will not panic.

iulianpascalau: Right and we will index `<nil>` which, of course is a correctly formatted number :man_facepalming: 
Please add nil checks,

miiu96: added and unit tests

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed. 
image

iulianpascalau (approved):

PR #3573, opened 2021-11-09
New OutportDriver interface

iulianpascalau:
- new outport driver interface


outport/outport.go
```diff
@@ -25,73 +25,108 @@ func NewOutport() *outport {
...
 
 // SaveBlock will save block for every driver
-func (o *outport) SaveBlock(args *indexer.ArgsSaveBlockData) {
+func (o *outport) SaveBlock(args *indexer.ArgsSaveBlockData) error {
 	o.mutex.RLock()
 	defer o.mutex.RUnlock()
 
 	for _, driver := range o.drivers {
-		driver.SaveBlock(args)
+		err := driver.SaveBlock(args)
```
AdoAdoAdo, line 33: so in case we have multiple drivers and one of them has an issue we no longer call SaveBlock for the rest?

maybe call for all SaveBlock, and if at least one of them returned an error log all of them and report the last error?

could be done for the rest of the methods in outport

process/sync/baseSync.go
```diff
@@ -740,7 +740,14 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
 			return err
 		}
 
-		boot.outportHandler.RevertIndexedBlock(currHeader, currBody)
+		err = boot.outportHandler.RevertIndexedBlock(currHeader, currBody)
+		if err != nil {
+			log.Debug("boot.outportHandler.RevertIndexedBlock",
```
AdoAdoAdo, line 745: here we return the error and log it as well, is there a reason?

iulianpascalau: copied from the rest of the function. Removing the log.Debug

outport/outport.go
iulianpascalau: sure, we can do that

mateigraura (approved):

AdoAdoAdo (dismissed):

gabi-vuls (approved):
System test passed. 
image

AdoAdoAdo (approved):

liviuancas-elrond (dismissed):
test passed :+1: 
steps: 

- node is in sync
-  external service goes down
-  node falls behind the chain
- external service is back up and responds correctly to pushed events\
- node gets back in sync

image

process/block/metablock.go
```diff
@@ -576,38 +576,28 @@ func (mp *metaProcessor) indexBlock(
...
+			"epoch: %d, hash: %s", err, metaBlock.GetEpoch(), hex.EncodeToString(headerHash))
 	}
 
 	if shardCoordinatorShardID != nodesCoordinatorShardID {
-		log.Debug("indexBlock",
-			"epoch", epoch,
-			"shardCoordinator.ShardID", shardCoordinatorShardID,
-			"nodesCoordinator.ShardID", nodesCoordinatorShardID)
-		return
+		return fmt.Errorf("%w in metaProcessor.indexBlock, shard ID mismach, "+
```
SebastianMarian, line 592: Typo: mismach

process/block/metablock.go
```diff
@@ -1223,7 +1224,11 @@ func (mp *metaProcessor) CommitBlock(
 		mp.blockTracker.CleanupInvalidCrossHeaders(header.Epoch, header.Round)
 	}
 
-	mp.indexBlock(header, headerHash, body, lastMetaBlock, notarizedHeadersHashes, rewardsTxs)
+	err = mp.indexBlock(header, headerHash, body, lastMetaBlock, notarizedHeadersHashes, rewardsTxs)
+	if err != nil {
+		return err
```
SebastianMarian, line 1229: Is this critical? We should return here? If yes, this return will block the node as the commit has been already done above. Also the set of current block header and current block hash in blockchain component is already done. I would move this line as upper it is possible, although this will not solve the blocking situation until we can move this call above L1161.

process/block/shardblock.go
```diff
@@ -561,39 +562,26 @@ func (sp *shardProcessor) indexBlockIfNeeded(
...
+			"epoch: %d, hash: %s", err, header.GetEpoch(), hex.EncodeToString(headerHash))
 	}
 
 	if shardId != nodesCoordinatorShardID {
-		log.Debug("indexBlockIfNeeded: shardId != nodesCoordinatorShardID",
-			"epoch", epoch,
-			"shardCoordinator.ShardID", shardId,
-			"nodesCoordinator.ShardID", nodesCoordinatorShardID)
-		return
+		return fmt.Errorf("%w in shardProcessor.indexBlockIfNeeded, shard ID mismach, "+
```
SebastianMarian, line 576: Typo: mismach

process/block/shardblock.go
```diff
@@ -895,7 +892,11 @@ func (sp *shardProcessor) CommitBlock(
 	}
 
 	sp.blockChain.SetCurrentBlockHeaderHash(headerHash)
-	sp.indexBlockIfNeeded(bodyHandler, headerHash, headerHandler, lastBlockHeader)
+	err = sp.indexBlockIfNeeded(bodyHandler, headerHash, headerHandler, lastBlockHeader)
```
SebastianMarian, line 895: The same comment here like in metablock.go. By moving this call above L803, if it is possible, could save the node from a blocking situation.

process/sync/baseSync.go
```diff
@@ -733,14 +733,13 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
...
-				"error", err.Error(),
-			)
-
 			return err
 		}
 
-		boot.outportHandler.RevertIndexedBlock(currHeader, currBody)
+		err = boot.outportHandler.RevertIndexedBlock(currHeader, currBody)
+		if err != nil {
+			return err
```
SebastianMarian, line 741: I think that this return and also the one from line 736 should be avoided. Maybe only log.Debug with error, as the rollback has been already done in L713, and after this line the code should not return err, just for the consistency.

iulianpascalau:
Not a good solution. Will close this one and will continue working on #3581

PR #3581, opened 2021-11-11
New outport driver interface - take 2

iulianpascalau:
- new outport driver interface
- made the outport handler wait for all drivers to return nil error on each function

### DIFF
--- a/outport/outport.go
+++ b/outport/outport.go
@@ -1,7 +1,9 @@
 package outport
 
 import (
+	"fmt"
 	"sync"
+	"time"
 
 	"github.com/ElrondNetwork/elrond-go-core/core/check"
 	"github.com/ElrondNetwork/elrond-go-core/data"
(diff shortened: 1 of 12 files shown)


liviuancas-elrond (approved):
test passed :+1: 
steps:

node is in sync
external service goes down
node falls behind the chain
external service is back up and responds correctly to pushed events
node gets back in sync

SebastianMarian (approved):

outport/outport.go
```diff
@@ -1,27 +1,39 @@
...
-var log = logger.GetOrCreate("outport")
-
 // NewOutport will create a new instance of proxy
-func NewOutport() *outport {
-	return &outport{
-		drivers: make([]Driver, 0),
-		mutex:   sync.RWMutex{},
+func NewOutport(retrialInterval time.Duration) (*outport, error) {
+	if retrialInterval < minimumRetrialInterval {
+		return nil, fmt.Errorf("%w, provided: %d, minimum: %d", ErrInvalidRetrialInterval, retrialInterval, minimumRetrialInterval)
```
AdoAdoAdo, line 28: :+1:  
we can avoid high(100%) cpu load if flag is missing

AdoAdoAdo (approved):

outport/outport.go
iulianpascalau: exactly. We should always have minimums when doing endless fors.

gabi-vuls (approved):
System test passed
image

PR #3587, opened 2021-11-12
Gas schedule adjustment for v5

iulianpascalau:
- gas schedule adjustment for v5 in order to comply with the rest of the new settings


camilbancioiu (approved):

cmd/node/config/gasSchedules/gasScheduleV5.toml
```diff
@@ -255,7 +255,7 @@
     I64Store16 = 5
     I64Store32 = 5
     MemorySize = 5
-    MemoryGrow = 5
+    MemoryGrow = 50000000
```
sasurobert, line 258: why do we want this ? I would change the maxLimit to 10 and should be enough.

PR #3589, opened 2021-11-15
Fix tx interceptor

iulianpascalau:
- fixed fee handler when checking an intercepted transaction


AdoAdoAdo (approved):

sasurobert (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3591, opened 2021-11-19
epochStart: fix pending miniblocks computation at start in epoch

AdoAdoAdo:
fixes the computation of pending miniblocks and processed miniblocks at start in epoch

Some details:
At start in epoch, the pending/processed miniblocks are computed out of the start of epoch metablock and used for bootstrapping and setting the tracker with the correct information.
In cases where there are partially processed Metablocks referenced in the start of epoch metablock for a shard, the miniblocks That were already processed are not set as processed.

One other fix, also for the pending miniblocks at bootstrap is that the shardID for the pending miniblocks was wrongly set to the sender shard instead of the receiver shard, although this does not have an effect, as the shard ID is not used afterwards.

### DIFF
--- a/epochStart/bootstrap/shardStorageHandler_test.go
+++ b/epochStart/bootstrap/shardStorageHandler_test.go
@@ -4,16 +4,23 @@ import (
 	"os"
 	"testing"
 
+	"github.com/ElrondNetwork/elrond-go-core/core"
 	"github.com/ElrondNetwork/elrond-go-core/core/check"
 	"github.com/ElrondNetwork/elrond-go-core/data"
 	"github.com/ElrondNetwork/elrond-go-core/data/block"
+	"github.com/ElrondNetwork/elrond-go-core/data/typeConverters"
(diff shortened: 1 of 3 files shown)


epochStart/bootstrap/shardStorageHandler_test.go
```diff
@@ -165,3 +172,263 @@ func TestGetAllMiniBlocksWithDst(t *testing.T) {
 	assert.Equal(t, shardMbHeaders[string(hash1)], shardMiniBlockHeader)
 	assert.NotNil(t, shardMbHeaders[string(hash2)])
 }
+
+func TestShardStorageHandler_getProcessedAndPendingMiniBlocksErrorGettingEpochStartShardData(t *testing.T) {
```
iulianpascalau, line 176: :+1: for all new unit tests

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.
image

SebastianMarian (approved):

PR #3594, opened 2021-11-22
mined wallet key

AdoAdoAdo:


cmd/keygenerator/main.go
```diff
@@ -206,6 +236,51 @@ func generateKey(keyGen crypto.KeyGenerator, list []key) ([]key, error) {
 	return list, nil
 }
 
+func generateMinedWalletKeys(keyGen crypto.KeyGenerator, list []key, startingHexPattern string, shardID int) ([]key, error) {
+	pattern := nopattern != startingHexPattern
```
iulianpascalau, line 240: isPatternProvided ?

cmd/keygenerator/main.go
```diff
@@ -206,6 +236,51 @@ func generateKey(keyGen crypto.KeyGenerator, list []key) ([]key, error) {
...
+		return keys, nil
+	}
+}
+
+func keyHasPattern(key []byte, pattern []byte) bool {
+	return bytes.HasPrefix(key, pattern)
+}
+
+func keyInShard(keyBytes []byte, shardID byte) bool {
+	lastByte := keyBytes[len(keyBytes)-1]
```
iulianpascalau, line 280: I think this doesn't work well: for example, if we provide the shardID as 1 only those addresses ending with the value 1 are selected. There are plenty more (like 3 or 5 or ...)
Now, this functionality is not mandatory to be changed as, for example, we want an address that ends with a specific byte in order to be sure that no matter how many shard splits we will do, we end up in the desired shard. This will be most likely used whenever we want to generate an address in the same shard with another one. 
Since shard ID now means a value out of [0-2], we can rename the variable + flag to something like ending byte

AdoAdoAdo: with 3 or 5 will not work when we have more shards, here we ensure that the address will always be in shard1 (except for when there is only 1 shard - shard 0)

iulianpascalau (dismissed):

iulianpascalau (dismissed):

cmd/keygenerator/main.go
```diff
@@ -70,10 +76,11 @@ VERSION:
 	keyType = cli.StringFlag{
 		Name: "key-type",
 		Usage: fmt.Sprintf(
-			"What king of keys should generate. Available options: %s, %s, %s",
+			"What king of keys should generate. Available options: %s, %s, %s, %s",
```
SebastianMarian, line 79: Steal the crown?

cmd/keygenerator/CLI.md
```diff
@@ -17,9 +17,8 @@ AUTHOR:
 GLOBAL OPTIONS:
    --num-keys value  How many keys should generate. Example: 1 (default: 1)
    --key-type value  What king of keys should generate. Available options: validator, wallet, both (default: "validator")
-   --help, -h        show help
-   --version, -v     print the version
-   
-
+   --help, -h        Show help
+   --version, -v     Print the version
+   --hex-key-prefix  Value special pattern for the hex key prefix: Example 00f00 (default nopattern)
```
raduchis, line 22: I think that exactly this pattern will not work, because it should have an even number of hexadecimal digits, right?

SebastianMarian (approved):

SebastianMarian (approved):

iulianpascalau (approved):

raduchis (approved):

LucianMincu (approved):

PR #3596, opened 2021-11-22
Proof verify benchmark

mariusmihaic:
`elrond-go-core` integration + refactor
-
- Deleted `constants.go` which contained all proof ID's, since they are now moved in `elrond-go-core`. Referenced PR: [Proof construction benchmark](https://github.com/ElrondNetwork/elrond-go-core/pull/34)
- Deleted all `proof.GetType()` funcs
- Refactored all tests to use proof stubs instead of proof constructors from `elrond-go-core`
- Extracted all common constructor arguments for multiple header detectors to `MultipleHeaderDetectorArgs` struct

Benchmarks
-
- Added `common_test.go` which contains necessary funcs to create a real testing environment for a multiple header detector(e.g.: nodes coordinator, bls multi signers)
- Added `slashResGenerator_test.go` which creates different slash restults (see `generateMultipleHeaderProposalSlashResult` & `generateMultipleHeaderSigningSlashResult`) based on some slashable headers (see `generateSlashableHeaders`)
- Added detector(multiple header signing/proposal)  benchmarks for `VerifyData` and `ValidateProof`. All benchmarks yield very good performances, except for `ValidateProof` for a multiple header singing proof. A future PR with improvements will follow on this topic.

For reference, here are some results, using the following setup:
- `metaConsensusGroupSize  = 400`
-  `maliciousValidators = 1/4 * metaConsensusGroupSize`
- `slashableHeaders = 3`
- `hasher = Blake2bWithSize(16)`
- `singers = BLSMultiSig`
- `cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz`

-> MultipleHeaderSigningDetector:
```
BenchmarkMultipleHeaderSigningDetector_VerifyData  	1000000000	         0.001656 ns/op
BenchmarkMultipleHeaderSigningDetector_ValidateProof   	       1	29592135712 ns/op
```

-> MultipleHeaderProposalDetector
```
BenchmarkMultipleHeaderProposalDetector_VerifyData 	1000000000	         0.0008890 ns/op
BenchmarkMultipleHeaderProposalDetector_ValidateProof 	     198	   6341860 ns/op
```


process/slash/detector/common_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/ElrondNetwork/elrond-go-crypto/signing/mcl/singlesig"
+	"github.com/ElrondNetwork/elrond-go-crypto/signing/multisig"
+	mockEpochStart "github.com/ElrondNetwork/elrond-go/epochStart/mock"
+	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/process/headerCheck"
+	"github.com/ElrondNetwork/elrond-go/process/mock"
+	"github.com/ElrondNetwork/elrond-go/process/slash"
+	"github.com/ElrondNetwork/elrond-go/process/slash/detector"
+	"github.com/ElrondNetwork/elrond-go/sharding"
+	shardingMock "github.com/ElrondNetwork/elrond-go/sharding/mock"
```
AdoAdoAdo, line 23: can you import mocks from testscommon instead?

AdoAdoAdo (changes requested):

process/slash/detector/benchmarkMultipleHeaderProposal_test.go
```diff
@@ -0,0 +1,87 @@
...
+	crypto "github.com/ElrondNetwork/elrond-go-crypto"
+	"github.com/ElrondNetwork/elrond-go-crypto/signing"
+	"github.com/ElrondNetwork/elrond-go-crypto/signing/mcl"
+	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/process/slash/detector"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkMultipleHeaderProposalDetector_VerifyData(b *testing.B) {
+	if testing.Short() {
```
raduchis, line 18: I don't think that the testing.Short is necessary. The Benchmarks should be run only with the -bench flag

process/slash/detector/slashResGenerator_test.go
```diff
@@ -0,0 +1,254 @@
...
+		bitmapSize++
+	}
+	return uint32(bitmapSize)
+}
+
+func selectMaliciousSigners(noOfMaliciousSigners uint32, allMultiSigData map[string]multiSignerData) map[string]struct{} {
+	maliciousSigners := make(map[string]struct{}, noOfMaliciousSigners)
+
+	ct := uint32(0)
+	for pubKey := range allMultiSigData {
```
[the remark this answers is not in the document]
mariusmihaic: 1. Regarding first part, unfortunately it is not possible since `allMultiSigData` is a `[string]multiSignerData`, therefore in this case:
- `ct` would be `key->str` and
- `pubKey` would be `multiSignerData`
2.
>also take care that noOfMaliciousSigners is below len(allMultiSigData)

There is no need for this check. If `noOfMaliciousSigners` > ` len(allMultiSigData)`, code will not panic, it would simply select  `noOfMaliciousSigners` = ` len(allMultiSigData)`

process/slash/detector/slashResGenerator_test.go
```diff
@@ -0,0 +1,254 @@
...
+	return minRequiredSigners, bitmap, sigSharesData
+}
+
+func signHeader(
+	b *testing.B,
+	header data.HeaderHandler,
+	signers []crypto.MultiSigner,
+	bitmap []byte,
+	sigSharesData []sigShareData,
+	leaderPrivateKey crypto.PrivateKey) {
```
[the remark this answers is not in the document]
mariusmihaic: Yes, fixed it

process/slash/detector/benchmarkMultipleHeaderProposal_test.go
```diff
@@ -0,0 +1,87 @@
...
+		})
+	}
+}
+
+func benchmarkVerifyDataMultipleHeaderProposal(
+	b *testing.B,
+	hasher hashing.Hasher,
+	keyGenerator crypto.KeyGenerator,
+	blsSigners map[string]multiSignerData,
+	interceptedHeaders []process.InterceptedHeader) {
```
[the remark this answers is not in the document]
mariusmihaic: Yes, fixed it

process/slash/detector/common_test.go
mariusmihaic: Replaced only `&shardingMock.NodesCoordinatorCacheMock{}`. Other mocks were not in testscommon and I think it's better not to extend this PR with any further refactoring

process/slash/detector/benchmarkMultipleHeaderProposal_test.go
mariusmihaic: Right, removed them

process/slash/detector/slashResGenerator_test.go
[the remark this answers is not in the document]
AdoAdoAdo: 1. Right, for some reason I was thinking it was an array.

process/slash/detector/common_test.go
AdoAdoAdo: All new code should import from testscommon, otherwise we add more refactor work for later.

mariusmihaic: Refactored them

raduchis (approved):

AdoAdoAdo (approved):

PR #3603, opened 2021-11-24
process, cmd, factory: add front running protection

AdoAdoAdo:


process/block/preprocess/transactions.go
```diff
@@ -1186,8 +1210,10 @@ func (txs *transactions) computeSortedTxs(
 	log.Debug("computeSortedTxs.GetSortedTransactions")
 	sortedTxs := sortedTransactionsProvider.GetSortedTransactions()
 
-	SortTransactionsBySenderAndNonce(sortedTxs)
-	return sortedTxs, nil
+	selectedTxs := txs.preFilterTransactions(sortedTxs, gasBandwidth)
```
sasurobert, line 1213: the preFilter function should go inside the sortedTransactionsProvider - You should push the gasBandwidth there

process/block/preprocess/transactions.go
```diff
@@ -1317,8 +1343,41 @@ func (txs *transactions) IsInterfaceNil() bool {
...
+
+	txs.sortTransactionsBySenderAndNonceWithFrontRunningProtection(transactions, randomness)
+}
+
+func (txs *transactions) sortTransactionsBySenderAndNonceWithFrontRunningProtection(transactions []*txcache.WrappedTransaction, randomness []byte) {
+	// make sure randomness is 32bytes and uniform
+	randSeed := txs.hasher.Compute(string(randomness))
+	xoredAddresses := make(map[string][]byte)
+
+	for _, tx := range transactions {
```
sasurobert, line 1361: can we add the sorting to be done By GasPrice buckets ? so sort those with gasPrice X - after that with GasPrice with X-1 and so on ? or buckets in GasPrice of intervals of 100 ?

process/factory/metachain/preProcessorsContainerFactory.go
```diff
@@ -155,25 +158,28 @@ func (ppcm *preProcessorsContainerFactory) Create() (process.PreProcessorsContai
...
+		Accounts:             ppcm.accounts,
+		OnRequestTransaction: ppcm.requestHandler.RequestTransaction,
+		EconomicsFee:         ppcm.economicsFee,
+		GasHandler:           ppcm.gasHandler,
+		BlockTracker:         ppcm.blockTracker,
+		BlockType:            block.TxBlock,
+		PubkeyConverter:      ppcm.pubkeyConverter,
+		BlockSizeComputation: ppcm.blockSizeComputation,
+		BalanceComputation:   ppcm.balanceComputation,
+		EpochNotifier:        ppcm.epochNotifier,
```
iulianpascalau, line 177: align?

process/factory/shard/preProcessorsContainerFactory.go
```diff
@@ -187,25 +190,28 @@ func (ppcm *preProcessorsContainerFactory) Create() (process.PreProcessorsContai
...
+		OnRequestTransaction: ppcm.requestHandler.RequestTransaction,
+		EconomicsFee:         ppcm.economicsFee,
+		GasHandler:           ppcm.gasHandler,
+		BlockTracker:         ppcm.blockTracker,
+		BlockType:            block.TxBlock,
+		PubkeyConverter:      ppcm.pubkeyConverter,
+		BlockSizeComputation: ppcm.blockSizeComputation,
+		BalanceComputation:   ppcm.balanceComputation,
+		EpochNotifier:        ppcm.epochNotifier,
+		OptimizeGasUsedInCrossMiniBlocksEnableEpoch: ppcm.optimizeGasUsedInCrossMiniBlocksEnableEpoch,
```
iulianpascalau, line 210: align?

process/block/preprocess/transactions.go
iulianpascalau: I think this will defeat the purpose of front running protection. The higher the gas price on a transaction, the higher chances to be selected on every block to executed. Indeed, it might not get to be executed in the first block after entering the pool but will eventually be executed as the randomness will yield at some point a favorable value.

AdoAdoAdo: currently anyway we prefilter the "first batch" of transactions, which we got from the smart tx cache, fitting the allowed "gas bandwidth". The smart txcache gives you the higher scored senders first, so in a way there is a priority to get into the batch for the higher price transactions, but the order inside the batch is not guaranteed, as it depends on the randomness.
This is I think the effect we wanted.

process/factory/shard/preProcessorsContainerFactory.go
AdoAdoAdo: this is the way gofmt formats it for me, can't find the option to change it.

process/block/preprocess/transactions.go
```diff
@@ -731,11 +740,24 @@ func (txs *transactions) getAllTxsFromMiniBlock(
...
+	if gasConsumed < maxGasPerBlock {
+		gasBandwidth = maxGasPerBlock - gasConsumed
+	}
+	return gasBandwidth
+}
+
 // CreateAndProcessMiniBlocks creates miniblocks from storage and processes the transactions added into the miniblocks
 // as long as it has time
-func (txs *transactions) CreateAndProcessMiniBlocks(haveTime func() bool) (block.MiniBlockSlice, error) {
+func (txs *transactions) CreateAndProcessMiniBlocks(haveTime func() bool, randomness []byte) (block.MiniBlockSlice, error) {
```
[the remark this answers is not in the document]
AdoAdoAdo: On the scheduled SC calls feature, we had to push the header being processed to the transaction preprocessor, and the randomness will be available from the header as well, this is why I chose to temporarily push the randomness.
When merging the feature to dev, this will be replaced with header, and the randomness will be taken from it.

process/block/preprocess/transactions.go
AdoAdoAdo: prefiltering is only required when one is creating the block, so for the consensus group leader, not for observers or the other validators. 
All validators/observers are calling the sortTransactions, this is why it was not moved inside.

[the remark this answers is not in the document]
AdoAdoAdo: as discussed, the prefiltering is only done on the leader, not on the validators or observers, so this change is backwards compatible.

process/block/preprocess/transactions.go
```diff
@@ -240,8 +248,8 @@ func (txs *transactions) RestoreBlockDataIntoPools(
 			txs.txPool.AddData([]byte(txHash), &tx, tx.Size(), strCache)
 		}
 
-		//TODO: Should be analyzed if restoring into pool only cross-shard miniblocks with destination in self shard,
-		//would create problems or not
+		//TODO: Should be analyzed if restoring into pool only cross-shard miniBlocks with destination in self shard
```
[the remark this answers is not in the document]
AdoAdoAdo: right, removed the TODO

process/block/preprocess/transactions.go
```diff
@@ -731,11 +740,24 @@ func (txs *transactions) getAllTxsFromMiniBlock(
...
+}
+
 // CreateAndProcessMiniBlocks creates miniblocks from storage and processes the transactions added into the miniblocks
 // as long as it has time
-func (txs *transactions) CreateAndProcessMiniBlocks(haveTime func() bool) (block.MiniBlockSlice, error) {
+func (txs *transactions) CreateAndProcessMiniBlocks(haveTime func() bool, randomness []byte) (block.MiniBlockSlice, error) {
 	startTime := time.Now()
-	sortedTxs, err := txs.computeSortedTxs(txs.shardCoordinator.SelfId(), txs.shardCoordinator.SelfId())
+
+	gasBandwidth := uint64(float64(txs.getRemainingGasPerBlock()) * (1 + selectionGasBandwidthIncrease))
```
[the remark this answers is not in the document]
AdoAdoAdo: changed to allow for double the remaining gas as bandwidth, now that the front running protection is available

process/block/preprocess/transactions.go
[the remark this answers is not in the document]
sasurobert: even on scheduled. this txsPreprocessor could have simply access to chainHandler - and take everything from these.

process/block/preprocess/transactions.go
sasurobert: No. GetSortedTransactions is called only by the leader on the CreateBlock flow. It is not called by validators, obsservers.

iulianpascalau (dismissed):

process/block/preprocess/transactions.go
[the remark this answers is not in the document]
AdoAdoAdo: Chain handler does not help with the block currently being processed.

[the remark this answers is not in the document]
sasurobert: blockchainHook has access to current Header.

process/block/preprocess/transactions.go
AdoAdoAdo: I somehow thought of the sorting that is done afterwards. 
Yes this could theoretically be moved there but in this case the logic needs to be changed more. 
I will leave this as a todo.

process/block/preprocess/transactions.go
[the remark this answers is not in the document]
AdoAdoAdo: will be refactored after merging scheduled sc calls feature.

sasurobert (dismissed):

iulianpascalau (approved):

iulianpascalau (approved):

LucianMincu (approved):
System tests passed.

PR #3623, opened 2021-12-06
No snapshot after start in epoch

BeniaminDrasovean:
In this PR: Add a specific key-value to each trie storer. When snapshoting, check if that value is present in the last storage, and only then start the snapshot. This way, after start in epoch no snapshot is triggered.


trie/trieStorageManager.go
```diff
@@ -613,10 +635,14 @@ func (tsm *trieStorageManager) SetEpochForPutOperation(epoch uint32) {
 	storer.SetEpochForPutOperation(epoch)
 }
 
+// EpochConfirmed is called whenever a new epoch is confirmed
 func (tsm *trieStorageManager) EpochConfirmed(epoch uint32, _ uint64) {
 	tsm.flagDisableOldStorage.Toggle(epoch >= tsm.disableOldStorageEpoch)
 	log.Debug("old trie storage", "disabled", tsm.flagDisableOldStorage.IsSet())
 
+	err := tsm.mainStorer.Put([]byte(activeDBKey), []byte(activeDBVal))
```
bogdan-rosianu, line 643: it looks like this is the only place when you set the value for active DB. I guess it is meant to be set only on the first confirmed epoch, right? So if starting from genesis, epoch 0 -> epoch 1 will trigger the value to be set; or when starting in epoch, only the next epoch?

bogdan-rosianu (changes requested):

trie/trieStorageManager.go
```diff
@@ -493,6 +502,19 @@ func (tsm *trieStorageManager) takeCheckpoint(checkpointEntry *snapshotsQueueEnt
 	}
 }
 
+func isActiveDb(stsm *snapshotTrieStorageManager) bool {
+	val, err := stsm.Get([]byte(activeDBKey))
```
SebastianMarian, line 506: Check the err for nil?

trie/trieStorageManager.go
```diff
@@ -493,6 +502,19 @@ func (tsm *trieStorageManager) takeCheckpoint(checkpointEntry *snapshotsQueueEnt
...
+	if bytes.Equal(val, []byte(activeDBVal)) {
+		return true
+	}
+
+	log.Debug("snapshotTrieStorageManager get",
+		"err", err.Error(),
+		"value", val,
```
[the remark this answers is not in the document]
BeniaminDrasovean: Refactored

trie/trieStorageManager.go
```diff
@@ -613,10 +635,14 @@ func (tsm *trieStorageManager) SetEpochForPutOperation(epoch uint32) {
 	storer.SetEpochForPutOperation(epoch)
 }
 
+// EpochConfirmed is called whenever a new epoch is confirmed
 func (tsm *trieStorageManager) EpochConfirmed(epoch uint32, _ uint64) {
 	tsm.flagDisableOldStorage.Toggle(epoch >= tsm.disableOldStorageEpoch)
 	log.Debug("old trie storage", "disabled", tsm.flagDisableOldStorage.IsSet())
 
+	err := tsm.mainStorer.Put([]byte(activeDBKey), []byte(activeDBVal))
+	log.LogIfError(err, "error", "set db as activeDB error")
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

trie/trieStorageManager.go
BeniaminDrasovean: This func is called every time when an epoch is changed, or when the trieStorageManager is initialized, so `activeDBKey` will be saved in each storer (each epoch a new storer is created)

trie/trieStorageManager.go
BeniaminDrasovean: Done.

bogdan-rosianu (dismissed):

SebastianMarian (approved):

bogdan-rosianu (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed. 
image

PR #3628, opened 2021-12-08
Feat/scheduled sc execution

AdoAdoAdo:
This feature branch introduces Scheduled Smart Contract Execution.

The scheduling of smart contract calls is done for the current round by the block producer and also executed during the current round when the scheduling is done, but the results are only available in the next block. This is done in order to not affect the consensus time, but to better use the processing capacity of the validators during consensus.

Activating this feature allows the validators in the consensus group (including the block producers) to use up to 3 seconds of the 6 seconds block time, instead of only 1.5 seconds, to execute transactions (and especially in the second half smart contract calls), which leads to a doubling of the processing capacity of each shard, in the scenarios where the majority of gas is used for smart contract calls processing (which is also the current case of the network).

The implementation follows the internal EIP, which should be checked before reviewing.


LucianMincu (approved):
System tests passed.

raduchis (approved):

sasurobert (approved):

iulianpascalau (commented):
Proper releases can be done in a separate PR directly on the `development` branch

PR #3637, opened 2021-12-09
Improve txs execution and cross mbs size

SebastianMarian:
* Resized the max gas in one mini block for better balancing of cross-shard and in-shard txs processing
* Increased the max number of pending mini blocks per shard allowed, after which a trigger for shard is stuck is activated
* Balanced percentage of gas used for txs with dest me vs txs from me to 50%-50%
* Fixed isShardStuck management for self to self txs
* Fixed an edge case situation in meta-chain when reverting of successful mini-blocks processed was not done completely


cmd/node/config/economics.toml
```diff
@@ -38,9 +38,9 @@
 
 [FeeSettings]
     GasLimitSettings = [
-        {EnableEpoch = 0, MaxGasLimitPerBlock = "1500000000", MaxGasLimitPerMiniBlock = "1500000000", MaxGasLimitPerMetaBlock = "15000000000", MaxGasLimitPerMetaMiniBlock = "15000000000", MinGasLimit = "50000"},
-        {EnableEpoch = 2, MaxGasLimitPerBlock = "1500000000", MaxGasLimitPerMiniBlock = "600000000", MaxGasLimitPerMetaBlock = "15000000000", MaxGasLimitPerMetaMiniBlock = "600000000", MinGasLimit = "50000"},
-        {EnableEpoch = 3, MaxGasLimitPerBlock = "1500000000", MaxGasLimitPerMiniBlock = "500000000", MaxGasLimitPerMetaBlock = "15000000000", MaxGasLimitPerMetaMiniBlock = "500000000", MinGasLimit = "50000"},
+        {EnableEpoch = 0, MaxGasLimitPerBlock = "1500000000", MaxGasLimitPerMiniBlock = "1500000000", MaxGasLimitPerMetaBlock = "15000000000", MaxGasLimitPerMetaMiniBlock = "15000000000", MaxGasLimitPerTx = "1500000000", MaxGasLimitPerTx = "1500000000", MinGasLimit = "50000"},
```
iulianpascalau, line 41: duplicated `MaxGasLimitPerTx` value set. On all 3 settings. This can have unpredictable results.

process/block/preprocess/smartContractResults.go
```diff
@@ -518,6 +519,16 @@ func (scr *smartContractResults) ProcessMiniBlock(miniBlock *block.MiniBlock, ha
 
 		processedTxHashes = append(processedTxHashes, miniBlockTxHashes[index])
 
+		if scr.flagOptimizeGasUsedInCrossMiniBlocks.IsSet() {
+			//TODO: Remove the commented code if the meta-chain revert problem is fixed
```
iulianpascalau, line 523: TODO still relevant?

process/block/preprocess/transactions.go
```diff
@@ -1276,6 +1277,16 @@ func (txs *transactions) ProcessMiniBlock(
 
 		processedTxHashes = append(processedTxHashes, miniBlockTxHashes[index])
 
+		if txs.flagOptimizeGasUsedInCrossMiniBlocks.IsSet() {
+			//TODO: Remove the commented code if the meta-chain revert problem is fixed
```
iulianpascalau, line 1281: TODO still relevant?

process/constants.go
```diff
@@ -84,7 +84,7 @@ const MinForkRound = uint64(0)
 
 // MaxNumPendingMiniBlocksPerShard defines the maximum number of pending miniblocks per shard, after which a shard
 // could be considered stuck if this value multiplied by the number of shards is reached
-const MaxNumPendingMiniBlocksPerShard = 30
+const MaxNumPendingMiniBlocksPerShard = 60
```
iulianpascalau, line 87: since this is strictly related to the maximum number of miniblocks resulted by the split, I suggest moving this also in the economics.toml file or somehow automatically computed by splitting the max gas limit per block with the current miniblock gas limit multiplied by a const like 10.

cmd/node/config/economics.toml
SebastianMarian: Good find. Removed duplicated values

process/block/preprocess/smartContractResults.go
SebastianMarian: If the system test will pass the TODO and the commented code will be removed

process/block/preprocess/transactions.go
SebastianMarian: If the system test will pass the TODO and the commented code will be removed

process/constants.go
SebastianMarian: I think the second suggestion is better

process/coordinator/process.go
```diff
@@ -667,6 +677,19 @@ func (tc *transactionCoordinator) CreateMbsAndProcessCrossShardTransactionsDstMe
 	return miniBlocks, nrTxAdded, allMBsProcessed, nil
 }
 
+func (tc *transactionCoordinator) revertIfNeeded(txsToBeReverted [][]byte) {
+	shouldRevert := tc.shardCoordinator.SelfId() == core.MetachainShardId && len(txsToBeReverted) > 0
```
sasurobert, line 681: so no revert at all for metachain ? for miniblocks processing ?

process/coordinator/process.go
```diff
@@ -555,13 +555,18 @@ func (tc *transactionCoordinator) CreateMbsAndProcessCrossShardTransactionsDstMe
...
+	processedTxHashes := make([][]byte, 0)
 
 	if check.IfNil(hdr) {
 		return miniBlocks, nrTxAdded, false, nil
 	}
 
 	shouldSkipShard := make(map[uint32]bool)
 
+	if tc.shardCoordinator.SelfId() == core.MetachainShardId {
+		tc.initProcessedTxsResults()
```
sasurobert, line 567: why this ?

SebastianMarian: Metachain should init post processor map which hold all scrs created after last init, right here, before starting to process miniblocks from the given shard hdr, just to keep track of all scrs created also from successful mbs processed. Then, if for example, the last mb processing will fail it could revert all the scrs coresponding to all mbs processed, no matter if they were successful or not.

process/coordinator/process.go
SebastianMarian: They are reverted, as this method is called by metachain with all txs already processed from successful mbs execution when complete processing of the whole shard hdr fails.

process/coordinator/process.go
```diff
@@ -950,7 +973,9 @@ func (tc *transactionCoordinator) processCompleteMiniBlock(
 ) error {
 
 	snapshot := tc.accounts.JournalLen()
-	tc.initProcessedTxsResults()
+	if tc.shardCoordinator.SelfId() != core.MetachainShardId {
```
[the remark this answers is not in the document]
SebastianMarian: Metachain will do this initialization upper, just before to start processing the given shard hdr, because it needs to fully process the whole shard hdr and it needs to keep track for all the scrs created after each mb processed, to be able to revert all in case of processing of one mb will fail later. In the case of shard node this initialization should be done here, just before to start processing the given mb, because it needs to fully process the whole mb instead of the whole metablock.

process/block/preprocess/basePreProcess.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/ElrondNetwork/elrond-vm-common/atomic"
 )
 
+const maxGasLimitPercentUsedForDestMeTxs = 50
```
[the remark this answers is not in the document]
SebastianMarian: I would keep this here for now, just to be sure that this value would not be easily modified back to 100% by validators, therefore canceling this optimization.

sasurobert (approved):

process/track/baseBlockTrack.go
```diff
@@ -630,7 +632,8 @@ func (bbt *baseBlockTrack) IsShardStuck(shardID uint32) bool {
 		}
 	}
 
-	maxNumPendingMiniBlocks := process.MaxNumPendingMiniBlocksPerShard * bbt.shardCoordinator.NumberOfShards()
+	maxNumMiniBlocksForSameReceiverInOneBlock := bbt.feeHandler.MaxGasLimitPerBlockForSafeCrossShard() / bbt.feeHandler.MaxGasLimitPerMiniBlockForSafeCrossShard()
+	maxNumPendingMiniBlocks := maxNumMiniBlocksForSameReceiverInOneBlock * process.MaxMetaNoncesBehind * uint64(bbt.shardCoordinator.NumberOfShards())
```
iulianpascalau, line 636: :+1:

iulianpascalau (dismissed):

sasurobert (approved):

gabi-vuls (dismissed):
System test passed.
Errors are safe. 
image

sasurobert (approved):

iulianpascalau (approved):

gabi-vuls (approved):

PR #3644, opened 2021-12-14
Arwen with safer upgradeFromSource

sasurobert:
just new arwen
Test: intrashard upgrade from source with not enough gas should error - before this branch it will not.


andrei-marinica (approved):

mariusmihaic (approved):

gabi-vuls (approved):
System test passed.
Errors are safe.
image

PR #3646, opened 2021-12-15
Fixes after review - genesis indexing

ssd04:
- fixes after review
- more unit tests
- set TxCount for genesis block header

### DIFF
--- a/examples/address_test.go
+++ b/examples/address_test.go
@@ -25,18 +25,6 @@ func TestHexAddressToBech32Address(t *testing.T) {
 	require.Equal(t, "erd14uqxan5rgucsf6537ll4vpwyc96z7us5586xhc5euv8w96rsw95sfl6a49", bech32Address)
 }
 
-func TestHexAddressToBech32GenesisMintingAddress(t *testing.T) {
-	t.Parallel()
-
-	hexEncodedAddress := "f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0"
-
-	hexEncodedAddressBytes, err := hex.DecodeString(hexEncodedAddress)
-	require.NoError(t, err)
-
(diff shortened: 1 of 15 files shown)


bogdan-rosianu (changes requested):

examples/address_test.go
```diff
@@ -25,6 +25,18 @@ func TestHexAddressToBech32Address(t *testing.T) {
 	require.Equal(t, "erd14uqxan5rgucsf6537ll4vpwyc96z7us5586xhc5euv8w96rsw95sfl6a49", bech32Address)
 }
 
+func TestHexAddressToBech32GenesisMintingAddress(t *testing.T) {
```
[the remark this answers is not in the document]
ssd04: :+1:

genesis/parsing/accountsParser_test.go
```diff
@@ -74,15 +74,17 @@ func createDelegatedInitialAccount(address string, delegatedBytes []byte, delega
...
-	ap, err := parsing.NewAccountsParser(
-		"./testdata/genesis_ok.json",
-		nil,
-		"",
-		createMockHexPubkeyConverter(),
-		&mock.KeyGeneratorStub{},
-		&mock.HasherMock{},
-		&mock.MarshalizerMock{},
-	)
+	args := genesis.AccountsParserArgs{
```
iulianpascalau, line 77: usually we create an unexported function in the test file, something like `createMockAccountsParserArgs` and return a populated argument struct that will work as it is in the xxx_ShouldWork test.
On the other tests, like in the TestNewAccountsParser_NilEntireBalanceShouldErr we could have something like:
```GO
args := createMockAccountsParserArgs()
args.EntireSupply = nil
////.......
```
In this way the duplicate code can be removed :)

ssd04: added additional function :+1:

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #3649, opened 2021-12-17
No snapshot after start in epoch

BeniaminDrasovean:
When starting in epoch, do not snapshot the first time. Also, do not try to get from a db if it is closed.


sasurobert (dismissed):

sasurobert (approved):

ssd04 (approved):

gabi-vuls (approved):
System test passed.

PR #3651, opened 2021-12-20
Create commit proof tx from multiple hot wallets

mariusmihaic:
- Added `KeyPair` struct, which defines a crypto.PrivateKey <-> crypto.PublicKey pair. This is used to sign/send txs
- When constructing a `SlashingNotifier`, one must defined its keys in a `map<ShardID, KeyPair>`
- Whenever detecting a slashing event, `SlashingNotifier` will select first `KeyPair` that is not in the same shard where the slashing event has been detected to create a tx commitment proof


process/slash/notifier/slashingNotifier.go
```diff
@@ -161,13 +173,35 @@ func (sn *slashingNotifier) computeTxData(proof coreSlash.SlashingProofHandler)
...
 
-func (sn *slashingNotifier) signTx(tx *transaction.Transaction) error {
+// Selects first KeyPair that is not in the same shard
+// where the slashing event has been detected
+func (sn *slashingNotifier) selectKeyPair() KeyPair {
+	var shardID uint32
+	selfID := sn.shardCoordinator.SelfId()
+
+	first := true
+	for currShardID := range sn.keyPairs {
```
AdoAdoAdo, line 183: this selection looks a bit weird
I think you can remove the "first" variable and just initialize shardID with selfID

AdoAdoAdo (changes requested):

process/slash/notifier/slashingNotifier.go
mariusmihaic: >this selection looks a bit weird

Yes, I'm aware it looks weird, but only because I didn't want to iterate through the map anymore if I've already found a the first suitable keyPair. 
However, considering that the map containing the `keyPairs` will most likely be very small in size (most probably 3 wallets, one in each shard) I can change the code to make more readable.

>I think you can remove the "first" variable and just initialize shardID with currShardID

Right, applied your suggestion

consensus/mock/keyGenMock.go
```diff
@@ -7,16 +7,16 @@ import (
 // PrivateKeyMock mocks a private key implementation
 type PrivateKeyMock struct {
```
ssd04, line 8: Not necessarily in this PR, but I think we can move this mock into testscommon, and remove the duplicates from the repo

process/slash/notifier/slashingNotifier.go
```diff
@@ -179,12 +179,8 @@ func (sn *slashingNotifier) selectKeyPair() KeyPair {
...
-	first := true
 	for currShardID := range sn.keyPairs {
-		if first == true {
-			shardID = currShardID
-			first = false
-		}
+		shardID = currShardID
 
 		if currShardID != selfID {
 			shardID = currShardID
```
AdoAdoAdo, line 186: this is no longer needed if initialized on top with currShardID

consensus/mock/keyGenMock.go
mariusmihaic: Right, this thing needs a refactor in a future PR

ssd04 (approved):

process/slash/notifier/slashingNotifier.go
mariusmihaic: Good catch, removed the extra line

PR #3660, opened 2021-12-22
Fix some trie prints on dev

BeniaminDrasovean:
Do not snapshot after start in epoch on peer trie. (this was done for the user trie on a previous PR)

### DIFF
--- a/trie/trieStorageManager.go
+++ b/trie/trieStorageManager.go
@@ -301,7 +301,10 @@ func (tsm *trieStorageManager) Get(key []byte) ([]byte, error) {
 		return nil, ErrContextClosing
 	}
 
-	val, _ := tsm.mainStorer.Get(key)
+	val, err := tsm.mainStorer.Get(key)
+	if checkIfClosingError(err) {
+		return nil, err
+	}
 	if len(val) != 0 {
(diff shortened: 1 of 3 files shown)


sasurobert (approved):

trie/trieStorageManager.go
```diff
@@ -301,7 +301,10 @@ func (tsm *trieStorageManager) Get(key []byte) ([]byte, error) {
 		return nil, ErrContextClosing
 	}
 
-	val, _ := tsm.mainStorer.Get(key)
+	val, err := tsm.mainStorer.Get(key)
+	if checkIfClosingError(err) {
```
iulianpascalau, line 305: is this ok? Shouldn't have been `!checkIfClosingError(err)` ? Also, I would call this simply `isClosingError` 

Here and on L317, L329

BeniaminDrasovean: If there is a closing err, it should not continue to search in other dbs, but return the err. Changed the func name.

iulianpascalau (approved):

sasurobert (approved):

LucianMincu (approved):
System tests passed.

PR #3668, opened 2022-01-05
Index altered accounts

bogdan-rosianu:
index altered accounts as well when saving a block into outport drivers - helpful in the future when outport drivers libraries won't be directly injected into the node


process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+
+	alteredAccounts[encodedAddress] = &indexer.AlteredAccount{
+		Address: encodedAddress,
+		Balance: userAccount.GetBalance().String(),
+		Nonce:   userAccount.GetNonce(),
+	}
+
+	if len(markedAccountTokens) > 0 {
+		for _, tokenData := range markedAccountTokens {
+			err = aap.fetchTokensDataForMarkedAccount(encodedAddress, tokenData, alteredAccounts)
```
mariusmihaic, line 144: Why not pass `userAccount` instead of `encodedAddress`?
This way, in `fetchTokensDataForMarkedAccount` you would have saved **3 duplicated code checks**:
1. Decode address
2. Load account from DB
3. Convert account to user account

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+		return err
+	}
+
+	alteredAccount := alteredAccounts[encodedAddress]
+
+	if alteredAccount.Tokens == nil {
+		alteredAccount.Tokens = make([]*indexer.AccountTokenData, 0)
+	}
+
+	alteredAccount.Tokens = append(alteredAccount.Tokens, &indexer.AccountTokenData{
```
mariusmihaic, line 204: Does this mean that if an account has, let's say 3 operations with `token1`, each of them with nonce = 1,2,3 => all of them will be indexed, right? Do we want that? Or do we only want to store the latest nonce operation?

process/block/alteredaccounts/errors.go
```diff
@@ -0,0 +1,15 @@
+package alteredaccounts
+
+import "errors"
+
+// ErrNilShardCoordinator signals that a nil shard coordinator has been provided
+var ErrNilShardCoordinator = errors.New("nil shard coordinator")
```
mariusmihaic, line 6: Please do not duplicate these errors for each created package. This whole file can be deleted.
Use existing ones.

process/block/baseProcess_test.go
```diff
@@ -364,6 +364,7 @@ func CreateMockArguments(
 			EpochNotifier:      &mock.EpochNotifierStub{},
 			RoundNotifier:      &mock.RoundNotifierStub{},
 			GasHandler:         &mock.GasHandlerMock{},
+			AlteredAccountsProvider: &testscommon.AlteredAccountsProviderStub{},
```
mariusmihaic, line 367: Please reformat the code here.

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -0,0 +1,430 @@
...
+				RcvAddr: []byte("shard2 addr - tx3"),
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, 5, len(res))
+
+	for key := range res {
+		decodedKey, _ := args.AddressConverter.Decode(key)
+		require.True(t, strings.HasPrefix(string(decodedKey), "shard0"))
```
mariusmihaic, line 225: I highly recommend here and in the above tests that you check for the full address instead of prefix only.
E.g.: Here you have 6 addresses which have `shard0` prefix, and you expect 5 of them as output. Which one of them do you expect? How can you be sure that those 5 are the ones you actually expect? The prefix is not enough.

It would be also useful to check if for the corresponding address, you have loaded the correct balance/nonce

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -0,0 +1,430 @@
...
+						},
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	encodedAddr := hex.EncodeToString([]byte("addr"))
+	require.Equal(t, expectedToken.Value.String(), res[encodedAddr].Tokens[0].Balance)
```
mariusmihaic, line 314: Perhaps create an expected `AccountTokenData` object and compare it with `res[encodedAddr].Tokens[0]` so you can be sure that all fields are the same (e.g.: esdt name = `token0`)

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -0,0 +1,430 @@
...
```
mariusmihaic, line 361: Same remark as above

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -0,0 +1,430 @@
...
+						},
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	encodedAddr := hex.EncodeToString([]byte("addr"))
+	require.Equal(t, 2, len(res[encodedAddr].Tokens))
```
mariusmihaic, line 420: Please check for actual token values instead of only `len`.
Also, perhaps:
```suggestion
	require.Len(t, res[encodedAddr].Tokens, 2)
```

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+		args := getMockArgs()
+		aap, err := NewAlteredAccountsProvider(args)
+
+		require.NotNil(t, aap)
+		require.NoError(t, err)
+	})
+}
+
+func TestAlteredAccountsProvider_ExtractAlteredAccountsFromPool(t *testing.T) {
```
mariusmihaic, line 78: Maybe add an extra test for an account which has multiple ESDT/NFT transfers with the same ID, but different nonces?

process/block/alteredaccounts/alteredAccountsProvider.go
miiu96: :+1:

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+
+	var esdtToken esdt.ESDigitalToken
+	err = aap.marshalizer.Unmarshal(&esdtToken, esdtTokenBytes)
+	if err != nil {
+		return err
+	}
+
+	alteredAccount := alteredAccounts[encodedAddress]
+
+	if alteredAccount.Tokens == nil {
```
miiu96, line 200: you can remove this.

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	tokenID := topics[0]
+
+	encodedAddress := aap.addressConverter.Encode(address)
+	_, exists := markedAlteredAccounts[encodedAddress]
+	if !exists {
+		markedAlteredAccounts[encodedAddress] = &markedAlteredAccount{}
+	}
+
+	markedAccount := markedAlteredAccounts[encodedAddress]
+	if len(markedAccount.tokens) == 0 {
```
miiu96, line 319: you can remove this

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	selfShardID := aap.shardCoordinator.SelfId()
+	for _, regularTx := range txPool.Txs {
+		senderShardID := aap.shardCoordinator.ComputeId(regularTx.GetSndAddr())
+		receiverShardID := aap.shardCoordinator.ComputeId(regularTx.GetRcvAddr())
+		if senderShardID != selfShardID && receiverShardID != selfShardID {
+			continue
+		}
+
+		if senderShardID == selfShardID {
+			aap.addMoveBalanceAddressInMap(regularTx.GetSndAddr(), markedAlteredAccounts)
```
iulianpascalau, line 229: we call this twice for all ESDT/NFTs operations. Can we make this be called once for these types of transactions?

iulianpascalau (changes requested):

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/ElrondNetwork/elrond-go-core/data/esdt"
+	"github.com/ElrondNetwork/elrond-go-core/data/indexer"
+	"github.com/ElrondNetwork/elrond-go-core/marshal"
+	logger "github.com/ElrondNetwork/elrond-go-logger"
+	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/ElrondNetwork/elrond-go/state"
+)
+
+var (
+	log        = logger.GetOrCreate("process/block/alteredaccounts")
```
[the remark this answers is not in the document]
bogdan-rosianu: yes, it's preferable that packages' names are all lowercase

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+type markedAlteredAccountToken struct {
+	identifier string
+	nonce      uint64
+}
+
+type markedAlteredAccount struct {
+	tokens  []*markedAlteredAccountToken
+}
+
+// ArgsAlteredAccountsProvider
```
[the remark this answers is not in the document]
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	if check.IfNil(args.Marshalizer) {
+		return nil, ErrNilMarshalizer
+	}
+
+	return &alteredAccountsProvider{
+		shardCoordinator: args.ShardCoordinator,
+		addressConverter: args.AddressConverter,
+		accountsDB:       args.AccountsDB,
+		marshalizer:      args.Marshalizer,
+		fungibleTokensIdentifier: map[string]struct{}{
```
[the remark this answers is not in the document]
bogdan-rosianu: This isn't right. fungible tokens and nfts are only identified by the those built in functions. I cannot use the whole list becase the list is specific for each of them

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+			core.BuiltInFunctionESDTNFTBurn:          {},
+			core.BuiltInFunctionESDTNFTAddQuantity:   {},
+			core.BuiltInFunctionESDTNFTCreate:        {},
+			core.BuiltInFunctionMultiESDTNFTTransfer: {},
+		},
+	}, nil
+}
+
+// ExtractAlteredAccountsFromPool will extract and return altered accounts from the pool
+func (aap *alteredAccountsProvider) ExtractAlteredAccountsFromPool(txPool *indexer.Pool) (map[string]*indexer.AlteredAccount, error) {
```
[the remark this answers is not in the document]
bogdan-rosianu: added

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+
+	markedAlteredAccounts[encodedAddress] = &markedAlteredAccount{}
+}
+
+func (aap *alteredAccountsProvider) extractESDTAccounts(
+	txPool *indexer.Pool,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) error {
+	var err error
+	for _, txLog := range txPool.Logs {
```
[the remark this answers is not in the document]
bogdan-rosianu: for esdts/nft it's the best way to look into the logs only

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+			aap.addMoveBalanceAddressInMap(regularTx.GetRcvAddr(), markedAlteredAccounts)
+		}
+	}
+}
+
+func (aap *alteredAccountsProvider) addMoveBalanceAddressInMap(
+	address []byte,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) {
+	encodedAddress := aap.addressConverter.Encode(address)
```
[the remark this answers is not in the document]
bogdan-rosianu: you're right. saved in map as string(address bytes) and only encoded in the end

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	return nil
+}
+
+func (aap *alteredAccountsProvider) processEvent(
+	event data.EventHandler,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) error {
+	_, isEsdtOperation := aap.fungibleTokensIdentifier[string(event.GetIdentifier())]
+	if isEsdtOperation {
+		err := aap.extractEsdtData(event, zeroBigInt, markedAlteredAccounts)
```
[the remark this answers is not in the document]
bogdan-rosianu: that's the convention. esdts have nonce = 0

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	nonce *big.Int,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) error {
+	address := event.GetAddress()
+	topics := event.GetTopics()
+	if len(topics) == 0 {
+		return nil
+	}
+
+	tokenID := topics[0]
```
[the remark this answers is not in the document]
bogdan-rosianu: added

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+			return err
+		}
+	}
+
+	_, isNftOperation := aap.nonFungibleTokensIdentifier[string(event.GetIdentifier())]
+	if isNftOperation {
+		topics := event.GetTopics()
+		if len(topics) == 0 {
+			return nil
+		}
```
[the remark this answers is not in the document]
bogdan-rosianu: no, it's much better with the nonce param

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+		return err
+	}
+
+	alteredAccounts[encodedAddress] = &indexer.AlteredAccount{
+		Address: encodedAddress,
+		Balance: userAccount.GetBalance().String(),
+		Nonce:   userAccount.GetNonce(),
+	}
+
+	if len(markedAccountTokens) > 0 {
```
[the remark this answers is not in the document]
bogdan-rosianu: right. removed

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: true. sent user account as well. kept the encodedAddress because it tells me directly the map key

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: all tokens an address interacted with during the block will be indexed

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: no need for full address check because the output is a map and keys are unique. balance/nonce checks are done on other unit tests

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: added

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+
+	return nil
+}
+
+func (aap *alteredAccountsProvider) extractMoveBalanceAddresses(
+	txPool *indexer.Pool,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) {
+	selfShardID := aap.shardCoordinator.SelfId()
+	for _, regularTx := range txPool.Txs {
```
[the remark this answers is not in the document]
bogdan-rosianu: good catch. done

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: removed

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: removed

factory/blockProcessorCreator.go
```diff
@@ -779,6 +792,15 @@ func (pcf *processComponentsFactory) newMetaBlockProcessor(
 	return metaProcessor, vmFactoryTxSimulator, nil
 }
 
+func (pcf *processComponentsFactory) createAlteredAccountsProvider() (process.AlteredAccountsProviderHandler, error) {
+	return alteredaccounts.NewAlteredAccountsProvider(alteredaccounts.ArgsAlteredAccountsProvider{
```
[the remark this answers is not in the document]
bogdan-rosianu: no need for an additional variable, the reason is pretty obvious, so I'd leave it like this

integrationTests/testSyncNode.go
```diff
@@ -218,6 +218,7 @@ func (tpn *TestProcessorNode) initBlockProcessorWithSync() {
 		EpochNotifier:      tpn.EpochNotifier,
 		RoundNotifier:      coreComponents.RoundNotifier(),
 		GasHandler:         tpn.GasHandler,
+		AlteredAccountsProvider: &testscommon.AlteredAccountsProviderStub{},
```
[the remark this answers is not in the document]
bogdan-rosianu: fixed

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return err
+	}
+
+	userAccount, ok := account.(state.UserAccountHandler)
+	if !ok {
+		return nil
+	}
+
+	esdtTokenBytes, err := userAccount.RetrieveValueFromDataTrieTracker(storageKey)
```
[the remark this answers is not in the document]
bogdan-rosianu: it can in case of invalid transactions. should I only log errors, then ?

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	event data.EventHandler,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) error {
+	_, isEsdtOperation := aap.fungibleTokensIdentifier[string(event.GetIdentifier())]
+	if isEsdtOperation {
+		err := aap.extractEsdtData(event, zeroBigInt, markedAlteredAccounts)
+		if err != nil {
+			log.Debug("cannot extract esdt data", "error", err)
+			return err
+		}
```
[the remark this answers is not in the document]
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+			return nil
+		}
+
+		nonce := topics[1]
+		nonceBigInt := big.NewInt(0).SetBytes(nonce)
+		err := aap.extractEsdtData(event, nonceBigInt, markedAlteredAccounts)
+		if err != nil {
+			log.Debug("cannot extract nft data", "error", err)
+			return nil
+		}
```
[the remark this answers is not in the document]
bogdan-rosianu: done

process/block/metablock.go
```diff
@@ -117,6 +117,7 @@ func NewMetaProcessor(arguments ArgMetaProcessor) (*metaProcessor, error) {
 		processDataTriesOnCommitEpoch: arguments.Config.Debug.EpochStart.ProcessDataTrieOnCommitEpoch,
 		gasConsumedProvider:           arguments.GasHandler,
 		economicsData:                 arguments.CoreComponents.EconomicsData(),
+		alteredAccountsProvider:       arguments.AlteredAccountsProvider,
```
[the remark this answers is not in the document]
bogdan-rosianu: it is checked on `checkProcessorNilParameters`

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+
+	return alteredAccounts, nil
+}
+
+func (aap *alteredAccountsProvider) processMarkedAccountData(
+	encodedAddress string,
+	markedAccountTokens []*markedAlteredAccountToken,
+	alteredAccounts map[string]*indexer.AlteredAccount,
+) error {
+	addressBytes, err := aap.addressConverter.Decode(encodedAddress)
```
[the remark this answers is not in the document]
bogdan-rosianu: refactored the addresses management to avoid continuous encodes and decodes

process/block/shardblock.go
```diff
@@ -87,6 +87,7 @@ func NewShardProcessor(arguments ArgShardProcessor) (*shardProcessor, error) {
 		processDataTriesOnCommitEpoch: arguments.Config.Debug.EpochStart.ProcessDataTrieOnCommitEpoch,
 		gasConsumedProvider:           arguments.GasHandler,
 		economicsData:                 arguments.CoreComponents.EconomicsData(),
+		alteredAccountsProvider:       arguments.AlteredAccountsProvider,
```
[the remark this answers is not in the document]
bogdan-rosianu: it is checked on checkProcessorNilParameters

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+	markedAccount.tokens = append(markedAccount.tokens, &markedAlteredAccountToken{
+		identifier: string(tokenID),
+		nonce:      nonce.Uint64(),
+	})
+
+	return nil
+}
+
+func (aap *alteredAccountsProvider) isTokenAlreadyMarked(markedAccount *markedAlteredAccount, identifier string, nonce uint64) bool {
+	for _, tokenData := range markedAccount.tokens {
```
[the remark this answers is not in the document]
bogdan-rosianu: yes, refactored by using a map

process/block/alteredaccounts/alteredAccountsProvider.go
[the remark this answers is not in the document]
iulianpascalau: Then we need a mechanism as to change the list in case we might add a new function (maybe in vmcommon). 
How do you propose to accomplish this?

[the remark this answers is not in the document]
bogdan-rosianu: I don't think there is an easy way of achieving this. When logs are generated, the identifier is set programmatically (hardcoded if you like this naming better). So the only way right now is to know and use the identifiers the logs are generated for and their purpose.

process/block/alteredaccounts/tokensProcessor.go
```diff
@@ -0,0 +1,164 @@
...
+	nonce *big.Int,
+	tokenID string,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) error {
+	if !tp.isSameShard(address) {
+		return nil
+	}
+
+	addressStr := string(address)
+	_, exists := markedAlteredAccounts[addressStr]
```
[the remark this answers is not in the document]
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider.go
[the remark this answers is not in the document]
mariusmihaic: Possible solution in the future would be to add a wrapper over esdt/nft operations in `elrond-vm-common`.
Until then, we can leave it like this

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -213,133 +175,46 @@ func (aap *alteredAccountsProvider) fetchTokensDataForMarkedAccount(
...
+}
+
+func (aap *alteredAccountsProvider) extractAddressesFromTxsHandlers(
+	selfShardID uint32,
+	txsHandlers map[string]data.TransactionHandler,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) {
+	for _, txHandler := range txsHandlers {
+		senderShardID := aap.shardCoordinator.ComputeId(txHandler.GetSndAddr())
+		receiverShardID := aap.shardCoordinator.ComputeId(txHandler.GetRcvAddr())
```
mariusmihaic, line 197: [optional]
Could extract sender & receiver address in separate variables, e.g.:
```go
senderAddr := txHandler.GetSndAddr()
receiverAddr :=txHandler.GetRcvAddr()
```

process/block/alteredaccounts/tokensProcessor.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// in case of esdt, nft or multi esdt transfers, the 3rd index of the topics contains the destination address
+	tokenID := topics[idxTokenIDInTopics]
+	err := tp.processEsdtDataForAddress(address, nonce, string(tokenID), markedAlteredAccounts)
+	if err != nil {
+		return err
+	}
+
+	if len(topics) > idxReceiverAddressInTopics {
+		destinationAddress := topics[idxReceiverAddressInTopics]
```
mariusmihaic, line 117: **destination**Address := topics[idx**Receiver**AddressInTopics]

Is that correct?

process/block/alteredaccounts/tokensProcessor.go
```diff
@@ -0,0 +1,164 @@
...
+	addressStr := string(address)
+	_, exists := markedAlteredAccounts[addressStr]
+	if !exists {
+		markedAlteredAccounts[addressStr] = &markedAlteredAccount{}
+	}
+
+	markedAccount := markedAlteredAccounts[addressStr]
+	if markedAccount.tokens == nil {
+		markedAccount.tokens = make(map[string]*markedAlteredAccountToken)
+	}
```
mariusmihaic, line 146: Would this work to avoid an extra check?

```suggestion
	if !exists {
		markedAlteredAccounts[addressStr] = &markedAlteredAccount{
			tokens: make(map[string]*markedAlteredAccountToken),
		}
	}

	markedAccount := markedAlteredAccounts[addressStr]
```

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -37,57 +38,46 @@ type ArgsAlteredAccountsProvider struct {
...
-			core.BuiltInFunctionMultiESDTNFTTransfer: {},
-		},
-		nonFungibleTokensIdentifier: map[string]struct{}{
-			core.BuiltInFunctionESDTNFTTransfer:      {},
-			core.BuiltInFunctionESDTNFTBurn:          {},
-			core.BuiltInFunctionESDTNFTAddQuantity:   {},
-			core.BuiltInFunctionESDTNFTCreate:        {},
-			core.BuiltInFunctionMultiESDTNFTTransfer: {},
-		},
+		tokensProc:       newTokensProcessor(args.ShardCoordinator),
```
mariusmihaic, line 69: To be discussed if it is the best idea to internally initialize a processor inside your altered accounts provider.
The major disadvantage  is that you cannot mock it for tests

process/block/alteredaccounts/tokensProcessor.go
```diff
@@ -0,0 +1,164 @@
+package alteredaccounts
```
mariusmihaic, line 1: :+1:

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -37,57 +38,46 @@ type ArgsAlteredAccountsProvider struct {
...
-	shardCoordinator            sharding.Coordinator
-	addressConverter            core.PubkeyConverter
-	accountsDB                  state.AccountsAdapter
-	marshalizer                 marshal.Marshalizer
-	fungibleTokensIdentifier    map[string]struct{}
-	nonFungibleTokensIdentifier map[string]struct{}
+	shardCoordinator   sharding.Coordinator
+	addressConverter   core.PubkeyConverter
+	accountsDB         state.AccountsAdapter
+	marshalizer        marshal.Marshalizer
```
mariusmihaic, line 44: Perhaps `marshaller` ?

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -139,50 +126,29 @@ func (aap *alteredAccountsProvider) processMarkedAccountData(
...
 
-	if len(markedAccountTokens) > 0 {
-		for _, tokenData := range markedAccountTokens {
-			err = aap.fetchTokensDataForMarkedAccount(encodedAddress, tokenData, alteredAccounts)
-			if err != nil {
-				log.Warn("cannot fetch token data for marked account", "error", err)
-				return err
-			}
+	for tokenKey, tokenData := range markedAccountTokens {
+		err = aap.fetchTokensDataForMarkedAccount([]byte(tokenKey), encodedAddress, userAccount, tokenData, alteredAccounts)
```
mariusmihaic, line 130: This function should **only** do what the name says: fetch tokens data for an account.
It should return `&indexer.AccountTokenData`, then here you can add it to your altered accounts.

process/block/alteredaccounts/errors.go
mariusmihaic: We should definitely(work for another PR) move all high-frequently used errors in a separate package.

process/block/alteredaccounts/alteredAccountsProvider_test.go
mariusmihaic: The only reason why this particular test works without checking for full address is because at `hash1` you have a `self` transaction, address being: `shard0 addr 3 - tx1`. But again, looking at your function as a `black-box`,  you cannot know how addresses are stored in the map. You could even think that the addresses are altered during execution. Even though it holds unique keys, you still don't know exactly which keys are stored (again, looking at it as a black-box).

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -311,7 +370,12 @@ func testExtractAlteredAccountsFromPoolShouldIncludeESDT(t *testing.T) {
 	require.NoError(t, err)
 
 	encodedAddr := hex.EncodeToString([]byte("addr"))
```
mariusmihaic, line 372: It looks quite confusing why the encoded address is the result of: `hex.EncodeToString`. It requires that you are aware that the pubKeyConverter uses hex package as encoder. If the Mock changes code inside, then this test will fail. Please use the following:
```suggestion
	encodedAddr := args.AddressConverter.Encode([]byte("addr"))
```

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -358,7 +422,64 @@ func testExtractAlteredAccountsFromPoolShouldIncludeNFT(t *testing.T) {
...
```
mariusmihaic, line 424: Same remark regarding `encodedAddr`

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -358,7 +422,64 @@ func testExtractAlteredAccountsFromPoolShouldIncludeNFT(t *testing.T) {
...
+					Address: []byte("addr"),
+					Events: []*transaction.Event{
+						{
+							Address:    []byte("addr"),
+							Identifier: []byte(core.BuiltInFunctionESDTNFTTransfer),
+							Topics: [][]byte{
+								[]byte("token0"),
+								big.NewInt(38).Bytes(),
+								nil,
+								receiverOnDestination,
```
mariusmihaic, line 469: This test and the one above are identical, except that here you also add `receiver` address. You could: 
- Safely delete the one above, since it won't decrease any code coverage
- Also, below, instead of checking `len(res)`, please check the stored account token data for each address

process/block/alteredaccounts/tokensProcessor.go
```diff
@@ -0,0 +1,164 @@
...
+		markedAccount = &markedAlteredAccount{}
+		markedAlteredAccounts[addressStr] = markedAccount
+	}
+
+	if markedAccount.tokens == nil {
+		markedAccount.tokens = make(map[string]*markedAlteredAccountToken)
+	}
+
+	tokenKey := tokenID + string(nonce.Bytes())
+	_, alreadyExists := markedAccount.tokens[tokenKey]
```
mariusmihaic, line 149: [just for my clarification]
If you have, let's say 3 logs with same address = address1, same esdt token = esdt1 => each of them will be called here with nonce =0 => we will only store it once (becase `alreadyExists==true`). Doesn't that mean that we somehow "loose" some intermediary possible balance changes, since you only get the balance at the end of the execution?

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -417,7 +538,124 @@ func testExtractAlteredAccountsFromPoolAddressHasBalanceChangeEsdtAndfNft(t *tes
...
+	for _, token := range res[encodedAddr].Tokens {
+		if token.Identifier == "esdttoken" {
+			require.Equal(t, &indexer.AccountTokenData{
+				Identifier: "esdttoken",
+				Balance:    expectedToken0.Value.String(),
+				Nonce:      0,
+				MetaData:   nil,
+			}, token)
+		}
+		if token.Identifier == "nft-o" {
```
mariusmihaic, line 642: This test is truly _"interesting"_  .
You have no token identifier == `nft-o` -> note there is an `o`, not a `0`, **yet** your test passes => **WRONG**.
That is because your code never reaches this `if`, therefore never calls `require.Equal`.
Also, this `for loop` is also not required(useless), since len(ret)==3 AND you have 3 `ifs`.
Perhaps use 3 require.Contains conditions for each token and drop the for loop

process/block/alteredaccounts/alteredAccountsProvider_test.go
mariusmihaic: Which one is it, exactly?

process/block/alteredaccounts/tokensProcessor.go
mariusmihaic: If it is desired, then perhaps add some unit tests where:
- You have the same esdt token(let's say 2-3 times) for an account => you **only store it once with nonce = 0**
- You have the same nft token(for let's say 2-3 times) for an account => you store **the same nft multiple times with different nonces**

mariusmihaic (changes requested):
Please check the new comments as well as the first ones which have not been marked as `resolver`

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: ok, check that the resulting map contains all the addresses I am interested in

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: `testExtractAlteredAccountsFromPoolAddressHasMultipleNfts`

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: done

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: I just split the functionality into multiple components so the code is easier to track. In this situation, I am not interested in mocking the tokens processor, but on the contrary, I want to test the component like in production (so the real tokens processor)

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: The interface is named `Marshalizer` so I just kept the naming for consistency. Should be renamed everywhere

process/block/alteredaccounts/alteredAccountsProvider.go
bogdan-rosianu: renamed to `addTokensDataForMarkedAccount` 
I wouldn't return the AccountTokenData as the account the function is adding to already exists in the map, so I think it is better to update the map inside the function, instead of returning it and adding to the map where I've just read it from (inside the function).

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: you're right. changed

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: fixed

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: I was not really interested in the code coverage, but rather in testing as many scenarios as possible. This way, the 2 tests (even if they are >80% similar) will trigger different branches/flows that I wanted to check.

process/block/alteredaccounts/tokensProcessor.go
bogdan-rosianu: As discussed, I am only interested in fetching the data at the end of the block. This way I am sure the data is accurate, regardless of the operations and their order.

Regarding tests:
1. added an additional check on `testExtractAlteredAccountsFromPoolShouldIncludeESDT` to ensure 2 entries won't be added
2. we already have `testExtractAlteredAccountsFromPoolAddressHasMultipleNfts`

process/block/alteredaccounts/alteredAccountsProvider.go
mariusmihaic: Agree, but this makes it easy to only test "happy-paths". If you want to "inject" an error to check how different code paths are executed, it would be harder, but still doable. Can leave it as it is, though.

process/block/alteredaccounts/alteredAccountsProvider.go
mariusmihaic: Agree

process/block/alteredaccounts/alteredAccountsProvider_test.go
mariusmihaic: Agree on the first point, but could you please follow-up on the second point?

process/block/alteredaccounts/alteredAccountsProvider_test.go
mariusmihaic: I still think this `for loop` could be replaced with a simple `require.Contains`. This way, if you want to add another entry in logs for this test, you don't have to also change the counter, the `require.Contains` will take care of it

mariusmihaic (commented):
No new findings, please check the remaining open comments/topics.

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: if you run the test `testExtractAlteredAccountsFromPoolShouldIncludeNFT` with coverage, you will notice that it won't enter within this `if`: `if len(topics) > idxReceiverAddressInTopics {` 
if you run the test `testExtractAlteredAccountsFromPoolShouldIncludeDestinationFromTokensLogsTopics` with coverage, you will notice that it will enter the same `if`

process/block/alteredaccounts/tokensProcessor.go
bogdan-rosianu: I understand now. Yes, good catch. done

process/block/alteredaccounts/alteredAccountsProvider_test.go
bogdan-rosianu: Done. However, I need that for loop to iterate over the slice and extract the tokens, but then I check all the extracted tokens, so if any of them is missing, the test will fail

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -199,10 +199,10 @@ func (aap *alteredAccountsProvider) extractAddressesFromTxsHandlers(
 		senderShardID := aap.shardCoordinator.ComputeId(senderAddress)
 		receiverShardID := aap.shardCoordinator.ComputeId(receiverAddress)
 
-		if senderShardID == selfShardID {
+		if senderShardID == selfShardID && len(senderAddress) > 0 {
```
mariusmihaic, line 202: :+1:

process/block/alteredaccounts/alteredAccountsProvider_test.go
```diff
@@ -642,40 +654,30 @@ func testExtractAlteredAccountsFromPoolAddressHasMultipleNfts(t *testing.T) {
...
-			require.Equal(t, &indexer.AccountTokenData{
-				Identifier: "nft-1",
-				Balance:    expectedToken2.Value.String(),
-				Nonce:      expectedToken2.TokenMetaData.Nonce,
-				MetaData:   expectedToken2.TokenMetaData,
-			}, token)
-			numChecks++
-		}
-	}
+	require.Contains(t, res[encodedAddr].Tokens, &indexer.AccountTokenData{
```
mariusmihaic, line 660: :+1:

mariusmihaic (approved):

iulianpascalau (dismissed):

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,226 @@
...
+	aap.extractAddressesFromTxsHandlers(selfShardID, txPool.Invalid, markedAlteredAccounts)
+}
+
+func (aap *alteredAccountsProvider) extractAddressesFromTxsHandlers(
+	selfShardID uint32,
+	txsHandlers map[string]data.TransactionHandler,
+	markedAlteredAccounts map[string]*markedAlteredAccount,
+) {
+	for _, txHandler := range txsHandlers {
+		senderAddress := txHandler.GetSndAddr()
```
miiu96, line 196: I think you should ignore the sender of the rewards transactions. 
Also, you should ignore the receiver of the invalid transactions.

bogdan-rosianu: i've added a check for invalid addresses. if the address is not valid, don't mark it as altered account

process/block/alteredaccounts/alteredAccountsProvider.go
```diff
@@ -0,0 +1,348 @@
...
+		Balance: userAccount.GetBalance().String(),
+		Nonce:   userAccount.GetNonce(),
+	}
+
+	if len(markedAccountTokens) > 0 {
+		for _, tokenData := range markedAccountTokens {
+			err = aap.fetchTokensDataForMarkedAccount(encodedAddress, tokenData, alteredAccounts)
+			if err != nil {
+				log.Warn("cannot fetch token data for marked account", "error", err)
+				return err
```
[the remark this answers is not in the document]
bogdan-rosianu: the comment was left on outdated code, can you look again? Also, I think it is ok to return error because otherwise we won't have token data available in outport driver and if there is a problem, we want to see it

miiu96 (approved):

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/ElrondNetwork/arwen-wasm-vm/v1_4 v1.4.34
 	github.com/ElrondNetwork/concurrent-map v0.1.3
 	github.com/ElrondNetwork/covalent-indexer-go v1.0.6
-	github.com/ElrondNetwork/elastic-indexer-go v1.1.32
-	github.com/ElrondNetwork/elrond-go-core v1.1.8-0.20220105140215-4643166f499b
+	github.com/ElrondNetwork/elastic-indexer-go v1.1.33
+	github.com/ElrondNetwork/elrond-go-core v1.1.10-0.20220201122512-86d7af4818f4
```
iulianpascalau, line 12: do not forget about proper release

iulianpascalau (approved):

miiu96 (approved):

PR #3669, opened 2022-01-06
Freeze account setGuardian

mariusmihaic:


vm/systemSmartContracts/freezeAccount.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// todo; check if guardian is already stored?
+
+// 1. User does NOT have any guardian => set guardian
+// 2. User has ONE guardian pending => does not set, wait until first one is set
+// 3. User has ONE guardian enabled => add it
+// 4. User has TWO guardians. FIRST is enabled, SECOND is pending => change pending with current one / does nothing until it is set
+// 5. User has TWO guardians. FIRST is enabled, SECOND is enabled => replace oldest one
+func (fa *freezeAccount) setGuardian(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
```
sasurobert, line 119: This is a built in function - not a system smart contract. System SCs are only on metachain.

sasurobert (changes requested):

vm/systemSmartContracts/freezeAccount.go
```diff
@@ -0,0 +1,284 @@
...
+		return err
+	}
+
+	account, err := fa.systemEI.BlockChainHook().GetUserAccount(address)
+	if err != nil {
+		return err
+	}
+
+	key := calcProtectedPrefixedKey(GuardiansKey)
+	return account.AccountDataHandler().SaveKeyValue(key, marshalledData)
```
sasurobert, line 218: This is wrong. You never write from a SC directly to a UserAccount - you always write into the cache simply. you should use fa.systemEI.SetStorage. But this has to be moved to built in functions, those are the actual SC for every account.

sasurobert (changes requested):

vm/systemSmartContracts/freezeAccount.go
mariusmihaic: Thanks for pointing out. Will move this code to `elrond-vm-common`.

vm/systemSmartContracts/freezeAccount.go
mariusmihaic: Thanks for clarification, will change it :pray:

PR #3680, opened 2022-01-10
get from cache when taking snapshot

BeniaminDrasovean:
When a snapshot is in progress, try and get the needed values from cache first.


iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3694, opened 2022-01-12
status metrics - concurrent operations

bogdan-rosianu:
introduced rwmutexes for all operations done on status metrics component
also, changed from sync.Map to 3 regular maps as sync.Map's concurrency is not needed anymore as now we have a mutex for each operation

### DIFF
--- a/statusHandler/statusMetricsProvider.go
+++ b/statusHandler/statusMetricsProvider.go
@@ -123,40 +123,37 @@ func (sm *statusMetrics) StatusP2pMetricsMap() map[string]interface{} {
 }
 
 func (sm *statusMetrics) getMetricsWithKeyFilterMutexProtected(filterFunc func(input string) bool) map[string]interface{} {
+	statusMetricsMap := make(map[string]interface{})
+
 	sm.mutUint64Operations.RLock()
-	uint64Metrics := sm.uint64Metrics
+	for key, value := range sm.uint64Metrics {
+		if !filterFunc(key) {
+			continue
+		}
+
+		statusMetricsMap[key] = value
+	}
 	sm.mutUint64Operations.RUnlock()
 
 	sm.mutStringOperations.RLock()
-	stringMetrics := sm.stringMetrics
+	for key, value := range sm.stringMetrics {
+		if !filterFunc(key) {
+			continue
+		}
+
+		statusMetricsMap[key] = value
+	}
 	sm.mutStringOperations.RUnlock()
 
 	sm.mutInt64Operations.RLock()
-	int64Metrics := sm.int64Metrics
+	for key, value := range sm.int64Metrics {
+		if !filterFunc(key) {
+			continue
+		}
+
+		statusMetricsMap[key] = value
+	}
 	sm.mutInt64Operations.RUnlock()
-
-	statusMetricsMap := make(map[string]interface{})
-	for key, value := range uint64Metrics {
-		if !filterFunc(key) {
-			continue
-		}
-
-		statusMetricsMap[key] = value
-	}
-	for key, value := range stringMetrics {
-		if !filterFunc(key) {
-			continue
-		}
-
-		statusMetricsMap[key] = value
-	}
-	for key, value := range int64Metrics {
-		if !filterFunc(key) {
-			continue
-		}
-
-		statusMetricsMap[key] = value
-	}
 
 	return statusMetricsMap
 }
@@ -168,7 +165,6 @@ func (sm *statusMetrics) StatusMetricsWithoutP2PPrometheusString() string {
 	sm.mutUint64Operations.RLock()
 	shardID := sm.uint64Metrics[common.MetricShardId]
 	sm.mutUint64Operations.RUnlock()
-
 
 	stringBuilder := strings.Builder{}
 	for key, value := range metrics {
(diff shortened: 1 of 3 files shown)


statusHandler/statusMetricsProvider.go
```diff
@@ -97,40 +116,46 @@ func (sm *statusMetrics) Close() {
...
+	sm.mutOperations.RLock()
+	defer sm.mutOperations.RUnlock()
+
 	statusMetricsMap := make(map[string]interface{})
-	sm.nodeMetrics.Range(func(key, value interface{}) bool {
-		keyString := key.(string)
-		if !strings.Contains(keyString, "_p2p_") {
-			return true
+	for key, value := range sm.nodeMetrics {
+		if strings.Contains(key, "_p2p_") {
```
iulianpascalau, line 141: bug, forgotten `!` : `if !strings.Contains(key, "_p2p_") {`

statusHandler/statusMetricsProvider.go
```diff
@@ -10,13 +10,14 @@ import (
 
 // statusMetrics will handle displaying at /node/details all metrics already collected for other status handlers
 type statusMetrics struct {
-	nodeMetrics *sync.Map
+	nodeMetrics   map[string]interface{}
```
iulianpascalau, line 13: since we only have 3 types of values: strings, uint64 and int64 we could have had 3 different maps. We would have avoid all those casts (which consume CPU) and the memory footprint should have been almost the same

statusHandler/statusMetricsProvider.go
```diff
@@ -97,40 +116,46 @@ func (sm *statusMetrics) Close() {
...
-		return true
-	})
+		statusMetricsMap[key] = value
+	}
 
 	return statusMetricsMap
 }
 
 // StatusP2pMetricsMap will return the p2p metrics in a map
 func (sm *statusMetrics) StatusP2pMetricsMap() map[string]interface{} {
```
iulianpascalau, line 135: we could have made this function (along with its counterpart more generic). Something like:
`StatusMetricsMapWhereKeyContains(key string) map[string]interface{}`

statusHandler/statusMetricsProvider.go
```diff
@@ -97,40 +116,46 @@ func (sm *statusMetrics) Close() {
...
-		return true
-	})
+		statusMetricsMap[key] = value
+	}
 
 	return statusMetricsMap
 }
 
 // StatusMetricsWithoutP2PPrometheusString returns the metrics in a string format which respects prometheus style
 func (sm *statusMetrics) StatusMetricsWithoutP2PPrometheusString() string {
```
iulianpascalau, line 152: do we still need this?

statusHandler/statusMetricsProvider.go
bogdan-rosianu: thanks fixed

statusHandler/statusMetricsProvider.go
bogdan-rosianu: done. also, each map has its own mutex

statusHandler/statusMetricsProvider.go
bogdan-rosianu: I didn't change at the interface level, but applied the advice for internal usage

statusHandler/statusMetricsProvider.go
bogdan-rosianu: yes, we cannot tell is anyone uses this endpoint. since it's used for prometheus exported metrics, there are chances that a few people monitor their node by using this

statusHandler/statusMetricsProvider.go
```diff
@@ -284,28 +301,20 @@ func (sm *statusMetrics) NetworkMetrics() map[string]interface{} {
 
 // must be called under mutex protection
 func (sm *statusMetrics) loadUint64Metric(metric string) uint64 {
```
iulianpascalau, line 303: this function became a little bit useless. we can just call `xxxxvalue := sm.uint64Metrics[metric]` as the uint64/int64 defaults are 0 and the string default is ""

Valid also for `loadStringMetric`

bogdan-rosianu: done, removed

statusHandler/statusMetricsProvider.go
```diff
@@ -28,86 +36,72 @@ func (sm *statusMetrics) IsInterfaceNil() bool {
 
 // Increment method increment a metric
 func (sm *statusMetrics) Increment(key string) {
-	sm.mutOperations.Lock()
-	defer sm.mutOperations.Unlock()
+	sm.mutUint64Operations.Lock()
+	defer sm.mutUint64Operations.Unlock()
 
-	keyValueI, ok := sm.nodeMetrics[key]
+	value, ok := sm.uint64Metrics[key]
```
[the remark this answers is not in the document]
bogdan-rosianu: it would be a pretty big change as this pattern is used in a lot of places. I would leave it as it is for now

iulianpascalau (approved):

statusHandler/statusMetricsProvider.go
```diff
@@ -97,40 +110,62 @@ func (sm *statusMetrics) Close() {
...
+	}
+	sm.mutUint64Operations.RUnlock()
+
+	sm.mutStringOperations.RLock()
+	for key, value := range sm.stringMetrics {
+		if !filterFunc(key) {
+			continue
+		}
+
+		statusMetricsMap[key] = value
```
raduchis, line 144: if the same key is present in multiple maps, the previous key/value will be overwritten. There are no cases with same keys in different maps, right?

raduchis (approved):

dragos-rebegea (approved):

gabi-vuls (commented):
System test passed. 
image

PR #3698, opened 2022-01-13
No process error on shuffle out

iulianpascalau:
- suppressed original processing errors when the node is shuffling between shards

### DIFF
--- a/common/context.go
+++ b/common/context.go
@@ -2,8 +2,12 @@ package common
 
 import "context"
 
-// WasContextClosed will return true if the provided context has been closed
-func WasContextClosed(ctx context.Context) bool {
+// IsContextDone will return true if the provided context signals it is done
+func IsContextDone(ctx context.Context) bool {
+	if ctx == nil {
+		return true
+	}
+
 	select {
 	case <-ctx.Done():
 		return true
--- a/common/context.go
+++ b/common/context.go
@@ -1,0 +1,13 @@
+package common
+
+import "context"
+
+// WasContextClosed will return true if the provided context has been closed
+func WasContextClosed(ctx context.Context) bool {
+	select {
+	case <-ctx.Done():
+		return true
+	default:
+		return false
+	}
+}
--- a/consensus/spos/bls/subroundBlock.go
+++ b/consensus/spos/bls/subroundBlock.go
@@ -1,6 +1,7 @@
 package bls
 
 import (
+	"context"
 	"time"
 
 	"github.com/ElrondNetwork/elrond-go-core/core"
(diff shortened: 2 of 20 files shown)


common/context.go
```diff
@@ -0,0 +1,13 @@
+package common
+
+import "context"
+
+// WasContextClosed will return true if the provided context has been closed
+func WasContextClosed(ctx context.Context) bool {
```
bogdan-rosianu, line 6: I would rename it to `IsContextDone` because:
1. the context can also be canceled in case of timeout or cancel(), not just closed
2. Was -> Is because the context can be closed/canceled when within the function so putting the action into the present is more meaningful.

bogdan-rosianu: also, please guard the `ctx` for nil values. the code would panic. I'd say that you should return true if it is nil

consensus/spos/bls/subroundBlock.go
```diff
@@ -107,6 +108,15 @@ func (sr *subroundBlock) doBlockJob() bool {
 	return true
 }
 
+func printLogMessage(ctx context.Context, baseMessage string, err error) {
+	if common.WasContextClosed(ctx) {
+		log.Debug(baseMessage + "context is closing")
```
bogdan-rosianu, line 113: add a space. right now, for this if, it would result in:
DEBUG doBlockJob.createHeadercontext is closing

bogdan-rosianu (changes requested):

consensus/spos/bls/subroundBlock.go
iulianpascalau: done, thanks

common/context.go
iulianpascalau: done, thanks

bogdan-rosianu (dismissed):

bogdan-rosianu (approved):

process/sync/baseSync.go
```diff
@@ -502,6 +502,11 @@ func (boot *baseBootstrap) syncBlocks(ctx context.Context) {
 
 		err := boot.syncStarter.SyncBlock()
```
SebastianMarian, line 503: There is another err log in this method which could be avoided if we also send ctx there. I don't have a strong opinion if it is mandatory to be done also there, as we log here that the go routine is stopping.

SebastianMarian (approved):

process/sync/baseSync.go
iulianpascalau: Right, the correct process there is to replace the error

SebastianMarian (dismissed):

SebastianMarian (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed 
image

PR #3710, opened 2022-01-19
Check frozen account on intercepted tx

mariusmihaic:


AdoAdoAdo:
changes will be taken on #3893

PR #3712, opened 2022-01-20
Critical section in txcache

iulianpascalau:
- added critical section when adding/removing a transaction in the tx cache. Also on clear operation.


storage/txcache/txCache.go
```diff
@@ -70,15 +71,17 @@ func (cache *TxCache) AddTx(tx *WrappedTransaction) (ok bool, added bool) {
 		cache.doEviction()
 	}
 
+	cache.mutAddTx.Lock()
```
AdoAdoAdo, line 74: can you add protection also for remove & clar

iulianpascalau: done

sasurobert (approved):

AdoAdoAdo (approved):

PR #3720, opened 2022-01-24
Fixed error on genesis block save

iulianpascalau:
- fixed error on genesis block save
- switched some prints to log level TRACE

### DIFF
--- a/dataRetriever/chainStorer.go
+++ b/dataRetriever/chainStorer.go
@@ -1,6 +1,7 @@
 package dataRetriever
 
 import (
+	"fmt"
 	"sync"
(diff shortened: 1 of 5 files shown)


dragos-rebegea (dismissed):

andreibancioiu (approved):
:+1: 

I confirm that the method extraction (which resulted in the new `saveMetaBlock()` and `saveShardBlock()`) is a safe refactoring (checked diff).

dragos-rebegea (approved):

LucianMincu (approved):
System tests passed.

PR #3722, opened 2022-01-26
Move shuffler nodescoordinator lite

ssd04:
Refactor sharding:
- move shuffler into nodes coordinator package


ssd04:
The requirements for this task has changed, so the approach has also changed. This issue is addressed in another PR -> https://github.com/ElrondNetwork/elrond-go/pull/3728

PR #3733, opened 2022-01-31
Import-db fix for new pruning storers

iulianpascalau:
- fixed import-db process when opening storers in advance: open the pruning storers in order with no gaps


sasurobert (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #3739, opened 2022-02-01
Sign with libp2p key

iulianpascalau:
- added the possibility to sign a payload with the lip2p's private key
- added the possibility to verify a (payload, peer ID, signature) on the netMessenger


p2p/libp2p/p2pSigner.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type p2pSigner struct {
+	privateKey *libp2pCrypto.Secp256k1PrivateKey
+}
+
+// Sign will sign a payload with the internal private key
+func (signer *p2pSigner) Sign(payload []byte) ([]byte, error) {
+	return signer.privateKey.Sign(payload)
```
bogdan-rosianu, line 18: since this struct doesn't have a constructor, maybe check if `signer.privateKey` is nil?

p2p/libp2p/netMessenger_test.go
```diff
@@ -1799,3 +1799,29 @@ func TestNetworkMessenger_Bootstrap(t *testing.T) {
...
+	err := mes1.ConnectToPeer(getConnectableAddress(mes2))
+	assert.Nil(t, err)
+
+	defer func() {
+		_ = mes1.Close()
+		_ = mes2.Close()
+	}()
+
+	payload := []byte("payload")
+	sig, err := mes1.Sign(payload)
```
bogdan-rosianu, line 1819: can you rename `mes1` and `mes2` to `messenger1` and `messenger2`? `mes` could also be the short form for `message`

p2p/libp2p/p2pSigner.go
sstanculeanu, line 18: perhaps a check for privateKey not nil should be added here

p2p/libp2p/p2pSigner_test.go
```diff
@@ -0,0 +1,128 @@
...
+			case 0:
+				_, errSign := signer.Sign(payload2)
+				assert.Nil(t, errSign)
+			case 1:
+				errVerify := signer.Verify(payload1, pid, sig1)
+				assert.Nil(t, errVerify)
+			}
+
+			wg.Done()
+		}(i)
```
sstanculeanu, line 124: shouldn't this be like?
```suggestion
		}(i%2)
```

p2p/libp2p/p2pSigner.go
sstanculeanu: already mentioned by Bogdan

iulianpascalau: this is component was specifically made without a constructor because I wanted to be used as a sub-component. 
There is no reason for adding a nil check as that will be a setup error. The unit test should prevent this to happen.

p2p/libp2p/netMessenger_test.go
iulianpascalau: done in the whole file

p2p/libp2p/p2pSigner.go
iulianpascalau: no, we generally do not check for the internal components over and over again for nil checks as this will degrade the performance.

iulianpascalau: second, if we add a nil check there but we forgot to properly instantiate the netMessenger.p2pSigner the code will still panic as the signer.xxxxx will trigger a nil dereference exception. We would just clutter the code with useless nil checks that will never be triggered in a correctly setup component. The unit test(s)/future integration tests ensure that the code will not get damaged accidentally.

sstanculeanu (approved):

bogdan-rosianu (approved):

PR #3741, opened 2022-02-01
Messages wrappers

sstanculeanu:
Added wrappers for both HeartbeatV2 and PeerAuthentication messages


process/heartbeat/interceptedPeerAuthentication.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	err = verifyPropertyLen(payloadProperty, ipa.peerAuthentication.Payload)
+	if err != nil {
+		return err
+	}
+	err = verifyPropertyLen(payloadSignatureProperty, ipa.peerAuthentication.PayloadSignature)
+	if err != nil {
+		return err
+	}
+
```
iulianpascalau, line 84: we should also check the signatures here & if the signatures are worth of checking:
1. we should check if the BLS public key is a validator or not. This can be achieved by using an interface that defines the function `GetValidatorWithPublicKey(publicKey []byte) (validator Validator, shardId uint32, err error)` (it is implemented by the nodesCoordinator)
2. For the payload sig, we already have the implementation done in the netMessenger the possibility to check if the sig was done correctly, so one more interface should be passed on the argument, something that defines the 
`Verify(payload []byte, pid core.PeerID, signature []byte) error` method
Then, we need to check if the payload is a valid timestamp and that timestamp is still valid as it belongs to an interval [current time - expiry_duration, current time + small_threshold (might be a const of around 10 seconds)]
3. For the BLS sig, we should use the interface PeerSignatureHandler (it is already implemented by a component that is also used in the consensus mechanism). This also caches the bls sig, bls pk and the pid tuple. 
the BLS sig verification is deliberately left as the last check in order to check the sig only if the message is considered valid

process/heartbeat/interceptedHeartbeat.go
```diff
@@ -68,6 +67,11 @@ func createHeartbeat(marshalizer marshal.Marshalizer, buff []byte) (*heartbeat.H
 	if err != nil {
 		return nil, err
 	}
+	payload := &heartbeat.Payload{}
```
iulianpascalau, line 70: maybe save this payload instance in the interceptedHeartbeat pointer? We almost certain will need it

process/heartbeat/interceptedPeerAuthentication.go
```diff
@@ -136,6 +195,25 @@ func (ipa *interceptedPeerAuthentication) String() string {
...
 }
 
+func (ipa *interceptedPeerAuthentication) verifyPayload() error {
+	payload := &heartbeat.Payload{}
+	err := ipa.marshalizer.Unmarshal(payload, ipa.peerAuthentication.Payload)
+	if err != nil {
+		return err
+	}
+
+	currentTimeStamp := uint64(time.Now().Unix())
```
iulianpascalau, line 205: suggestion to simplify:
```GO
    currentTimeStamp := time.Now().Unix()
    messageTimeStamp := payload.Timestamp
    minTimestampAllowed := currentTimeStamp - ipa.expiryTimespanInSec
    maxTimestampAllowed := currentTimeStamp + payloadExpiryThresholdInSec
    if messageTimeStamp < minTimestampAllowed || messageTimeStamp > maxTimestampAllowed {
        return process.ErrMessageExpired
    }
```
and make the Payload.Timestamp from the proto file an int64

process/heartbeat/interceptedPeerAuthentication.go
```diff
@@ -36,31 +46,56 @@ func NewInterceptedPeerAuthentication(arg ArgInterceptedPeerAuthentication) (*in
...
 
+func checkArg(arg ArgInterceptedPeerAuthentication) error {
+	err := checkBaseArg(arg.ArgBaseInterceptedHeartbeat)
+	if err != nil {
+		return err
+	}
+	if check.IfNil(arg.NodesCoordinator) {
+		return process.ErrNilNodesCoordinator
+	}
+	if arg.SignaturesHandler == nil {
```
iulianpascalau, line 69: check.IfNil

iulianpascalau (approved):

process/heartbeat/interceptedHeartbeat.go
```diff
@@ -0,0 +1,139 @@
...
+	hb, payload, err := createHeartbeat(arg.Marshalizer, arg.DataBuff)
+	if err != nil {
+		return nil, err
+	}
+
+	intercepted := &interceptedHeartbeat{
+		heartbeat: *hb,
+		peerId:    arg.PeerId,
+		payload:   *payload,
+	}
```
dragos-rebegea, line 51:
```suggestion
	intercepted := &interceptedHeartbeat{
		heartbeat: *hb,
		payload:   *payload,
		peerId:       arg.PeerId,
	}
```
please use the same order in creation as in declaration

process/heartbeat/interceptedHeartbeat.go
```diff
@@ -0,0 +1,139 @@
...
+	if ihb.heartbeat.PeerSubType != uint32(core.RegularPeer) && ihb.heartbeat.PeerSubType != uint32(core.FullHistoryObserver) {
+		return process.ErrInvalidPeerSubType
+	}
+	return nil
+}
+
+// IsForCurrentShard always returns true
+func (ihb *interceptedHeartbeat) IsForCurrentShard() bool {
+	return true
```
dragos-rebegea, line 106: why is this always returning true ?

process/heartbeat/interceptedHeartbeat_test.go
```diff
@@ -0,0 +1,190 @@
...
+		ihb, err := NewInterceptedHeartbeat(arg)
+		assert.False(t, ihb.IsInterfaceNil())
+		assert.Nil(t, err)
+	})
+}
+
+func Test_interceptedHeartbeat_CheckValidity(t *testing.T) {
+	t.Parallel()
+	t.Run("payloadProperty too short", testInterceptedHeartbeatPropertyLen(payloadProperty, false))
+	t.Run("payloadProperty too short", testInterceptedHeartbeatPropertyLen(payloadProperty, true))
```
dragos-rebegea, line 117:
```suggestion
	t.Run("payloadProperty too long", testInterceptedHeartbeatPropertyLen(payloadProperty, true))
```
same for all tests description

dragos-rebegea (approved):

dragos-rebegea (changes requested):
check the comments above

dragos-rebegea (approved):

iulianpascalau (approved):

PR #3749, opened 2022-02-03
PeerAuthenticationResolver

sstanculeanu:


dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
```diff
@@ -149,13 +149,15 @@ func (brcf *baseResolversContainerFactory) createTxResolver(
 	}
 
 	arg := resolvers.ArgTxResolver{
-		SenderResolver:    resolverSender,
+		ArgBaseResolver: resolvers.ArgBaseResolver{
```
iulianpascalau, line 152: :+1:

dataRetriever/resolvers/peerAuthenticationResolver.go
```diff
@@ -0,0 +1,218 @@
...
+		&dataRetriever.RequestData{
+			Type:  dataRetriever.HashArrayType,
+			Value: buffHashes,
+		},
+		hashes,
+	)
+}
+
+// RequestDataFromReferenceAndChunk requests a peer authentication chunk by specifying the reference and the chunk index
+func (res *peerAuthenticationResolver) RequestDataFromReferenceAndChunk(hash []byte, chunkIndex uint32) error {
```
iulianpascalau, line 102: I think we can have a method here like: RequestDataFromChunk and provide only the chunk index. Hash will not be required. Also, we can define a new dataRetriever.ChunkType value so we can do the filtering in the ProcessReceivedMessage function

dataRetriever/resolvers/peerAuthenticationResolver.go
```diff
@@ -0,0 +1,218 @@
...
+		return err
+	}
+
+	return res.Send(buffToSend, pid)
+}
+
+func (res *peerAuthenticationResolver) fetchPeerAuthenticationMessagesForHash(hash []byte) [][]byte {
+	var messages [][]byte
+
+	keys := res.peerAuthenticationPool.Keys()
```
iulianpascalau, line 191: considering the keys from the pool, might generate different views among nodes. 
Suggestion to do the following:
1. inject a nodes coordinator interface in this resolver
2. get all validators list. concatenate the list from each shard. sort validators based on public key
3. slice the resulting list based on the chunk index + maximum number of peers.
4. iterate the sub-slice and try to get (if existing) the peerAuth message from the pool

iulianpascalau (commented):
GJ for the necessary cleanup! :+1:

dataRetriever/resolvers/peerAuthenticationResolver.go
```diff
@@ -142,59 +151,171 @@ func (res *peerAuthenticationResolver) ProcessReceivedMessage(message p2p.Messag
...
+		return nil, err
+	}
+
+	validatorsPKs := make([][]byte, 0)
+	for _, shardValidators := range validatorsPKsMap {
+		validatorsPKs = append(validatorsPKs, shardValidators...)
 	}
 
+	sort.Slice(validatorsPKs, func(i, j int) bool {
+		return bytes.Compare(validatorsPKs[i], validatorsPKs[j]) < 0
```
iulianpascalau, line 207: :+1: this sort by the bls pks should make the distribution among shards quite even

dataRetriever/resolvers/peerAuthenticationResolver.go
```diff
@@ -142,59 +151,171 @@ func (res *peerAuthenticationResolver) ProcessReceivedMessage(message p2p.Messag
...
+	if err != nil {
+		return err
+	}
+
+	var lastErr error
+	errorsFound := 0
+	dataSlice := make([][]byte, 0, res.maxNumOfPeerAuthenticationInResponse)
+	for _, pk := range pksChunk {
+		peerAuth, tmpErr := res.fetchPeerAuthenticationAsByteSlice(pk)
+		if tmpErr != nil {
```
iulianpascalau, line 175: I do not think we should add the errors here. It will just clutter the node in normal circumstances when some peer authentication messages are missing. Instead, I would return an error if the data slice is empty, outside this for loop

dataRetriever/resolvers/peerAuthenticationResolver.go
```diff
@@ -142,59 +151,171 @@ func (res *peerAuthenticationResolver) ProcessReceivedMessage(message p2p.Messag
...
+// fetchPeerAuthenticationSlicesForHash fetches all peer authentications for the matching pks to hash
+func (res *peerAuthenticationResolver) fetchPeerAuthenticationSlicesForHash(hash []byte) [][]byte {
 	var messages [][]byte
 
 	keys := res.peerAuthenticationPool.Keys()
-	sort.Slice(keys, func(i, j int) bool {
-		return bytes.Compare(keys[i], keys[j]) < 0
-	})
-
 	for _, key := range keys {
```
iulianpascalau, line 317: didn't realized how this would have actually end up implemented. 
This is way too sub-optimal to be left as it is.
We should use a get operation here and send the full public key in the request until we figure out how to optimally resolve this.

iulianpascalau (commented):
:+1:

iulianpascalau (dismissed):

dataRetriever/resolvers/peerAuthenticationResolver_test.go
```diff
@@ -0,0 +1,599 @@
...
+		t.Parallel()
+
+		arg := createMockArgPeerAuthenticationResolver()
+		res, err := resolvers.NewPeerAuthenticationResolver(arg)
+		assert.Nil(t, err)
+		assert.False(t, res.IsInterfaceNil())
+	})
+}
+
+func Test_peerAuthenticationResolver_ProcessReceivedMessage(t *testing.T) {
```
dragos-rebegea, line 143:
```suggestion
func TestPeerAuthenticationResolver_ProcessReceivedMessage(t *testing.T) {
```

dataRetriever/resolvers/peerAuthenticationResolver_test.go
```diff
@@ -0,0 +1,599 @@
...
+		chunkIndex := uint32(0)
+		providedHashes, err := arg.Marshalizer.Marshal(batch.Batch{Data: providedKeys})
+		assert.Nil(t, err)
+		err = res.ProcessReceivedMessage(createRequestMsgWithChunkIndex(dataRetriever.HashArrayType, providedHashes, epoch, chunkIndex), fromConnectedPeer)
+		assert.Nil(t, err)
+		assert.Equal(t, 2, messagesSent)
+	})
+}
+
+func Test_peerAuthenticationResolver_RequestShouldError(t *testing.T) {
```
dragos-rebegea, line 538:
```suggestion
func TestPeerAuthenticationResolver_RequestShouldError(t *testing.T) {
```

dataRetriever/resolvers/peerAuthenticationResolver_test.go
```diff
@@ -0,0 +1,599 @@
...
+
+		hashes := make([][]byte, 0)
+		hashes = append(hashes, []byte("pk"))
+		err = res.RequestDataFromHashArray(hashes, 0)
+		assert.Equal(t, expectedErr, err)
+	})
+
+}
+
+func Test_peerAuthenticationResolver_RequestShouldWork(t *testing.T) {
```
dragos-rebegea, line 574:
```suggestion
func TestPeerAuthenticationResolver_RequestShouldWork(t *testing.T) {
```

dragos-rebegea (approved):

iulianpascalau (approved):

PR #3762, opened 2022-02-07
Unit tests and bugfix requestEpochStartInfo

mariusmihaic:
Tests
--
- Added tests for `requestEpochStartInfo` from `shardBlock.go` 
- Since nil pointer checks for `shardBlock.go` are done in `baseProcesss.go`( `CheckProcessorNilParameters`), I've moved all nil pointer check tests + created new ones from `shardBlock.go` -> `baseProcesss.go`

Bugfixes
--
- Fixed a bug in `NewShardProcessor` where arguments.DataComponents.Blockchain().**GetGenesisHeader()** could have been a nil pointer.
- Fixed a bug in `requestEpochStartInfo`: Before entering in the foor loop to request headers (`L343`), meta header is requested once `sp.requestHandler.RequestMetaHeader(header.GetEpochStartMetaHash())` (`L340`).
While inside the loop, if header is not found in pool (`headersPool.GetHeaderByHash`, from various reasons we did not have a response to our query) => error is returned, but we do not make any more requests for it . One line code on `L354` would fix it:

```go
	if err != nil {
			go sp.requestHandler.RequestMetaHeader(header.GetEpochStartMetaHash()) // THIS LINE
			continue
		}
```


sasurobert (dismissed):

process/block/shardblock_test.go
```diff
@@ -135,193 +136,88 @@ func CreateMockArgumentsMultiShard(
...
+		expectedErr error
+	}{
+		{
+			args: func() blproc.ArgShardProcessor {
+				return CreateMockArgumentsMultiShard(coreComponents, nil, bootstrapComponents, statusComponents)
+			},
+			expectedErr: process.ErrNilDataComponentsHolder,
+		},
+		{
+			args: func() blproc.ArgShardProcessor {
```
iulianpascalau, line 155: almost the same number of lines of code as t.Run :D
Can be left as it is

process/block/shardblock_test.go
```diff
@@ -1367,6 +1263,303 @@ func TestShardProcessor_ProcessBlockWithWrongMiniBlockHeaderShouldErr(t *testing
...
+		}
+
+		shardProc, _ := blproc.NewShardProcessor(args)
+		header := &testscommon.HeaderHandlerStub{
+			IsStartOfEpochBlockCalled: func() bool {
+				return false
+			},
+		}
+		err := shardProc.RequestEpochStartInfo(header, haveTime)
+		require.Nil(t, err)
```
iulianpascalau, line 1297: it is mandatory to add a time.Sleep here of about 1 second in order to be sure no request is made
Reason: the request is done on a go routine that will need some time to get executed.
Valid on all tests like this

process/block/shardblock_test.go
```diff
@@ -1367,6 +1263,303 @@ func TestShardProcessor_ProcessBlockWithWrongMiniBlockHeaderShouldErr(t *testing
...
+			RequestMetaHeaderCalled: func(hash []byte) {
+				require.Equal(t, headerHash, hash)
+				requestsCt.Increment()
+			},
+		}
+
+		shardProc, _ := blproc.NewShardProcessor(args)
+		err := shardProc.RequestEpochStartInfo(blockStartOfEpoch, haveTime)
+		require.Nil(t, err)
+		require.Equal(t, int64(1), requestsCt.Get())
```
iulianpascalau, line 1368: for the same reason, this test might fail sometimes. A time.sleep is mandatory before this statement

iulianpascalau (changes requested):
GJ! :+1:

process/block/shardblock_test.go
mariusmihaic: Right, missed this part. Added a `time.Sleep(time.Second)` for all tests to be sure

process/block/shardblock_test.go
mariusmihaic: Added `time.Sleep` for all tests

process/block/shardblock.go
```diff
@@ -65,8 +65,11 @@ func NewShardProcessor(arguments ArgShardProcessor) (*shardProcessor, error) {
 	if check.IfNil(arguments.DataComponents.Datapool().Transactions()) {
 		return nil, process.ErrNilTransactionPool
 	}
-
 	genesisHdr := arguments.DataComponents.Blockchain().GetGenesisHeader()
+	if check.IfNil(genesisHdr) {
```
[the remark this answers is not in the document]
mariusmihaic: This might've been the reason why this nil check was missed. Discovered it while setting up mocks for the tests

process/block/shardblock_test.go
mariusmihaic: Will leave it like this then

iulianpascalau (approved):

ssd04 (approved):

PR #3802, opened 2022-02-15
Integrated peer authentication resolver

sstanculeanu:
- added generatePeerAuthenticationResolver() method which is used to create the peer auth resolver
- added request handler for peer auth, the component that will make the requests for peer auth messages. Its integration will be done in the next PR


dataRetriever/requestHandlers/requestHandler_test.go
```diff
@@ -1159,3 +1151,261 @@ func TestResolverRequestHandler_RequestTrieNodeNotAValidResolver(t *testing.T) {
 	rrh.RequestTrieNode([]byte("hash"), "topic", 1)
 	assert.True(t, called)
 }
+
+//------- RequestPeerAuthentications
```
iulianpascalau, line 1155: we can drop these comments as now we mainly use one test with inner tests.

iulianpascalau (approved):

dragos-rebegea (approved):

PR #3809, opened 2022-02-16
fix tokensParser + more unit tests

bogdan-rosianu:
fix for "ExtractTokenIDAndNonceFromTokenStorageKey" when the nonce could have been represented as a hyphen when the key is converted to string. also, added more guards in regards to the ticker's min/max length + optimize by using `Index` instead of `Split`

### DIFF
--- a/common/tokensParser.go
+++ b/common/tokensParser.go
@@ -5,8 +5,19 @@ import (
 	"strings"
 )
 
-// esdtTickerNumRandChars represents the number of hex-encoded random characters sequence of a ticker
-const esdtTickerNumRandChars = 6
+const (
+	// esdtTickerNumRandChars represents the number of hex-encoded random characters sequence of a ticker
+	esdtTickerNumRandChars = 6
+
+	// separatorChar represents the character that separated the token ticker by the random sequence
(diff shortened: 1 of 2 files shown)


common/tokensParser.go
```diff
@@ -5,8 +5,19 @@ import (
...
-const esdtTickerNumRandChars = 6
+const (
+	// esdtTickerNumRandChars represents the number of hex-encoded random characters sequence of a ticker
+	esdtTickerNumRandChars = 6
+
+	// separatorChar represents the character that separated the token ticker by the random sequence
+	separatorChar = "-"
+
+	// minLengthForTickerName represents the minimum number of characters a token's ticker can have
+	minLengthForTickerName = 3
```
iulianpascalau, line 16: do we have these consts in the vm-common package?

bogdan-rosianu: unfortunately, no. we don't have exported constants for these

sstanculeanu (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3812, opened 2022-02-17
Feat/header verification

ssd04:
Refactor `sharding` package:
- move nodesCoordinator to a separate package inside sharding directory

Extend node api with `/internal` route in order to be able to fetch raw marshalized data for headers and miniblocks:
- separate route for shardblock and metablock
- each case has 3 different options: `by-hash`, `by-nonce`, `by-round`

Update `/network` route:
- added `/ratings` endpoint
- updated `/config` endpoint


gabi-vuls (dismissed):
System test passed.

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed

PR #3814, opened 2022-02-17
LevelDB storer Destroy improvements

bogdan-rosianu:
Improved the destroying of the level db storers

### DIFF
--- a/storage/leveldb/leveldbSerial.go
+++ b/storage/leveldb/leveldbSerial.go
@@ -260,7 +260,7 @@ func (s *SerialDB) Close() error {
 	// (just to close some go routines started as edge cases that would otherwise hang)
 	defer s.closer.Close()
 
-	return s.prepareClosingUnprotected()
+	return s.doClose()
 }
 
 // Remove removes the data associated to the given key
@@ -284,7 +284,7 @@ func (s *SerialDB) Destroy() error {
 	// (just to close some go routines started as edge cases that would otherwise hang)
 	defer s.closer.Close()
 
-	err := s.prepareClosingUnprotected()
+	err := s.doClose()
 	if err == nil {
 		return os.RemoveAll(s.path)
 	}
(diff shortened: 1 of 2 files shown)


storage/leveldb/leveldbSerial.go
```diff
@@ -260,15 +260,7 @@ func (s *SerialDB) Close() error {
...
-	_ = s.putBatch()
-	s.cancel()
-
-	db := s.makeDbPointerNilReturningLast()
-	if db != nil {
-		return db.Close()
-	}
-
-	return nil
+	return s.prepareClosingUnprotected()
```
iulianpascalau, line 263: suggestion to rename to `doClose` for 2 reasons:
1. unprotected is usually called in conjunction to having a mutex and this method should not be called outside that mutex locking/unlocking
2. it does not "prepare" the closing, it actually "execute" the closing

storage/leveldb/leveldbSerial.go
```diff
@@ -319,6 +301,20 @@ func (s *SerialDB) DestroyClosed() error {
 	return err
 }
 
+// prepareClosingUnprotected will prepare the internal component for closing the database
+// must be called under mutex protection
+func (s *SerialDB) prepareClosingUnprotected() error {
```
iulianpascalau, line 306: can be done in another PR: extract the common code in the level.go file also

storage/leveldb/leveldbSerial.go
bogdan-rosianu: done

storage/leveldb/leveldbSerial.go
bogdan-rosianu: added a todo

iulianpascalau (approved):

storage/leveldb/leveldbSerial.go
```diff
@@ -319,6 +301,21 @@ func (s *SerialDB) DestroyClosed() error {
 	return err
 }
 
+// doClose will handle the closing of the internal components
+// must be called under mutex protection
+// TODO: re-use this function in leveldb.go as well
+func (s *SerialDB) doClose() error {
+	_ = s.putBatch()
+	s.cancel()
```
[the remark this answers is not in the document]
bogdan-rosianu: no need for this. `cancel()` is set on constructor and in the entire project we rely on the fact that the constructor will return a good object, with no null values

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #3816, opened 2022-02-17
Peer auth processor

sstanculeanu:
- added peerAuthenticationRequestsProcessor which triggers requests for peer auth messages via requests handler in order to collect enough messages from other peers
- integrated this component into node via heartbeatV2Components


heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -0,0 +1,234 @@
...
+	ShardId                uint32
+	Epoch                  uint32
+	MessagesInChunk        uint32
+	MinPeersThreshold      float32
+	DelayBetweenRequests   time.Duration
+	MaxTimeout             time.Duration
+}
+
+// PeerAuthenticationRequestsProcessor defines the component that sends the requests for peer authentication messages
+type PeerAuthenticationRequestsProcessor struct {
```
iulianpascalau, line 40: unexported?

heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -0,0 +1,234 @@
...
+			missingKeys = append(missingKeys, []byte(mKey))
+		}
+	}
+
+	return missingKeys
+}
+
+// Close closes the internal components
+func (processor *PeerAuthenticationRequestsProcessor) Close() error {
+	if processor.cancel != nil {
```
iulianpascalau, line 223: no need to check for nil

heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -0,0 +1,234 @@
...
+		if processor.isThresholdReached(sortedValidatorsKeys) {
+			log.Debug("received enough messages, closing PeerAuthenticationRequestsProcessor go routine")
+			return
+		}
+
+		requestsTimer.Reset(processor.delayBetweenRequests)
+		select {
+		case <-requestsTimer.C:
+			processor.requestMissingKeys(sortedValidatorsKeys)
+		case <-timeoutTimer.C:
```
iulianpascalau, line 138: this can be safely removed and instead use another type of context on the constructor:
```GO
ctx, processor.cancel = context.WithTimeout(context.Background(), args.MaxTimeout)
```

heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf("%w for MaxTimeout, provided %d, min expected %d",
+			heartbeat.ErrInvalidTimeDuration, args.MaxTimeout, minTimeout)
+	}
+
+	return nil
+}
+
+func (processor *PeerAuthenticationRequestsProcessor) startRequestingMessages(ctx context.Context) {
+	defer func() {
+		if processor.cancel != nil {
```
iulianpascalau, line 112: no need for the nil check

heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	missingKeys := make([][]byte, 0)
+	for mKey, mVal := range validatorsMap {
+		if !mVal {
+			missingKeys = append(missingKeys, []byte(mKey))
+		}
+	}
+
+	return missingKeys
```
iulianpascalau, line 218: this will work for about 10000 missing keys.
Before returning this, I would add a new function that "randomly" creates a sub-slice out of all missing keys if they exceed a certain number of items (lets say 1000 - can be configurable)

iulianpascalau (commented):
partial review -> suggestion only on the production code

heartbeat/processor/peerAuthenticationRequestsProcessor.go
```diff
@@ -215,20 +216,41 @@ func (processor *PeerAuthenticationRequestsProcessor) getMissingKeys(sortedValid
...
+	tmpKeys := make([][]byte, lenMissingKeys)
+	copy(tmpKeys, missingKeys)
+
+	randomizer := &random.ConcurrentSafeIntRandomizer{}
+	randMissingKeys := make([][]byte, 0)
+	for len(randMissingKeys) != int(processor.maxMissingKeysInResponse) {
+		randomIndex := randomizer.Intn(lenMissingKeys)
+		randMissingKeys = append(randMissingKeys, tmpKeys[randomIndex])
+
+		tmpKeys[randomIndex] = tmpKeys[lenMissingKeys-1]
```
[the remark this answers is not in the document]
iulianpascalau: :+1:

factory/heartbeatV2Components.go
```diff
@@ -87,6 +88,19 @@ func checkHeartbeatV2FactoryArgs(args ArgHeartbeatV2ComponentsFactory) error {
 
 // Create creates the heartbeatV2 components
 func (hcf *heartbeatV2ComponentsFactory) Create() (*heartbeatV2Components, error) {
+	if !hcf.networkComponents.NetworkMessenger().HasTopic(common.PeerAuthenticationTopic) {
```
iulianpascalau, line 91: We might think of this when we will refactor the network messenger:
This pattern is widely used: if !hasTopic.... createTopic. We can have something like: createTopicIfMissing

iulianpascalau (approved):

iulianpascalau (approved):

dragos-rebegea (approved):

PR #3818, opened 2022-02-18
En 11661 staking v4 init

mariusmihaic:


epochStart/metachain/systemSCs.go
```diff
@@ -313,6 +317,13 @@ func (s *systemSCProcessor) ProcessSystemSmartContract(
 		}
 	}
 
+	if s.flagStakingV4Enabled.IsSet() {
```
sasurobert, line 320: rename to flag Init

sasurobert: 287 you need a disable flag

cmd/node/config/enableEpochs.toml
```diff
@@ -187,6 +187,10 @@
     # StakeLimitsEnableEpoch represents the epoch when stake limits on validators are enabled
     StakeLimitsEnableEpoch = 5
 
+    # StakingV4InitEnableEpoch represents the epoch when staking v4 is initialized. This is the epoch in which
+    # all nodes from staking queue are moved in the auction list
+    StakingV4InitEnableEpoch = 1000000
```
iulianpascalau, line 192: for testing purposes you can set this to 3 or 4

epochStart/metachain/systemSCs.go
```diff
@@ -313,6 +320,13 @@ func (s *systemSCProcessor) ProcessSystemSmartContract(
 		}
 	}
 
+	if s.flagInitStakingV4Enabled.IsSet() {
```
iulianpascalau, line 323: this will be called only on the epoch when we activate the staking v4. Is this correct?
Do we need this flag to execute something only on the activation epoch?

epochStart/metachain/systemSCs.go
mariusmihaic: - Renamed to `flagStakingV4InitEnabled`
- Also added a disable flag

epochStart/metachain/systemSCs.go
mariusmihaic: Yes. This is the expected behavior. We only want these actions to happen only once, hence the name StakingV4**Init**EnableEpoch. 
On this epoch, the only thing we want is to move all nodes from staking queue to auction list ("remove" waiting list).
Starting next epoch, using another flag -StakingV4EnableEpoch- we will start sorting auction nodes based on top up

iulianpascalau (approved):

PR #3826, opened 2022-02-22
FIX block API

miiu96:
- Fix the check there is set `sender` and `receiver` address for smart contract results


bogdan-rosianu (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3835, opened 2022-02-23
epoch activation ESDT

dragos-rebegea:


process/epochActivation/epochActivation.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type epochActivation struct {
+	config        config.EnableEpochs
+	epoch         uint32
+	epochNotifier vm.EpochNotifier
+	mutex         sync.RWMutex
+}
+
+func (ea *epochActivation) EpochConfirmed(epoch uint32, _ uint64) {
```
iulianpascalau, line 18: missing comments

process/epochActivation/epochActivation.go
```diff
@@ -0,0 +1,85 @@
...
+	return &epochActivation{
+		config: config,
+		mutex:  sync.RWMutex{},
+	}, nil
+}
+
+func (ea *epochActivation) IsESDTEnabled() bool {
+	ea.mutex.RLock()
+	defer ea.mutex.RUnlock()
+	return ea.config.ESDTEnableEpoch >= ea.epoch
```
iulianpascalau, line 43: I would have preferred to compute a bool var for each flag at the end of epoch and simply return it here. 
We can leave this at is is as a uint32 comparison is extremely fast.

vm/systemSmartContracts/esdt.go
```diff
@@ -55,21 +55,7 @@ type esdt struct {
...
-	flagTransferRole                       atomic.Flag
-	nftCreateONMultiShardEnableEpoch       uint32
-	flagNFTCreateONMultiShard              atomic.Flag
-	metaESDTEnableEpoch                    uint32
-	flagMetaESDT                           atomic.Flag
-	transformToMultiShardCreateEnableEpoch uint32
-	flagTransformToMultiShardCreate        atomic.Flag
-	registerAndSetAllRolesEnableEpoch      uint32
-	flagRegisterAndSetAllRoles             atomic.Flag
+	process.EpochActivationHandler
```
iulianpascalau, line 58: we need a variable here as to protect against mistakenly accesses

vm/systemSmartContracts/esdt.go
```diff
@@ -1463,7 +1429,7 @@ func getFirstAddressWithGivenRole(token *ESDTDataV2, definedRole []byte) ([]byte
 }
 
 func (e *esdt) changeToMultiShardCreate(args *vmcommon.ContractCallInput) vmcommon.ReturnCode {
-	if !e.flagTransformToMultiShardCreate.IsSet() {
+	if !e.IsESDTNFTCreateOnMultiShardEnabled() {
```
iulianpascalau, line 1432: is this ok?

PR #3837, opened 2022-02-23
[esdt api] export token's properties as hex

bogdan-rosianu:
When calling node's API for fetching tokens data, the properties are now exported as hex instead of plain string, in order to avoid non readable characters

### DIFF
--- a/api/groups/addressGroup.go
+++ b/api/groups/addressGroup.go
@@ -1,6 +1,7 @@
 package groups
 
 import (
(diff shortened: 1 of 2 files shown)


andreibancioiu (approved):

sstanculeanu (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3842, opened 2022-02-24
HeartbeatV2 api integration

sstanculeanu:
- added monitor for heartbeatV2 messages
- now GetHeartbeats from node merges the messages from v1 and v2, prioritizing v2


heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,200 @@
...
+
+	pids := monitor.cache.Keys()
+
+	heartbeatsV2 := make([]data.PubKeyHeartbeat, 0)
+	for idx := 0; idx < len(pids); idx++ {
+		pid := pids[idx]
+		peerId := core.PeerID(pid)
+		hb, ok := monitor.cache.Get(pid)
+		if !ok {
+			log.Debug("could not get data from cache for pid", "pid", peerId.Pretty())
```
iulianpascalau, line 102: I would remove this debug as some data might be evicted from the cache while iterating in this for loop
or maybe turn-it log.Trace

heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,200 @@
...
+		heartbeatData, err := monitor.parseMessage(peerId, hb)
+		if err != nil {
+			log.Debug("could not parse message for pid", "pid", peerId.Pretty(), "error", err.Error())
+			continue
+		}
+
+		heartbeatsV2 = append(heartbeatsV2, heartbeatData)
+	}
+
+	for idx := range heartbeatsV2 {
```
iulianpascalau, line 115:
```GO
for idx, hbData := range heartbeatsV2 {
	pk := hbData.PublicKey
	hbData.NumInstances = monitor.numInstances[pk]
}
```
this is slightly optimal as the slice indexing is only done once

heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,200 @@
...
+	if args.HideInactiveValidatorInterval < minDuration {
+		return fmt.Errorf("%w on HideInactiveValidatorInterval, provided %d, min expected %d",
+			heartbeat.ErrInvalidTimeDuration, args.HideInactiveValidatorInterval, minDuration)
+	}
+
+	return nil
+}
+
+// GetHeartbeats returns the heartbeat status
+func (monitor *heartbeatV2Monitor) GetHeartbeats() []data.PubKeyHeartbeat {
```
iulianpascalau, line 91: this is not a concurrent safe function. Because the usage of the internal monitor.numInstances map this struct might panic the node, if 2 or more requests are done in the same time.
Either remove this property map, or add mutex protection here

heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,200 @@
...
+	err := monitor.marshaller.Unmarshal(&payload, heartbeatV2.Payload)
+	if err != nil {
+		return pubKeyHeartbeat, err
+	}
+
+	peerInfo := monitor.peerShardMapper.GetPeerInfo(pid)
+
+	crtTime := time.Now()
+	messageAge := monitor.getMessageAge(crtTime, payload.Timestamp)
+	stringType := string(rune(peerInfo.PeerType))
```
iulianpascalau, line 145: do we really need the conversion to rune?

heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (monitor *heartbeatV2Monitor) getMessageAge(crtTime time.Time, messageTimestamp int64) time.Duration {
+	messageTime := time.Unix(messageTimestamp, 0)
+	msgAge := crtTime.Sub(messageTime)
+	return msgAge
+}
+
+func (monitor *heartbeatV2Monitor) isActive(messageAge time.Duration) bool {
+	if messageAge < 0 {
```
iulianpascalau, line 178: let's not test this with 0 as the current's node RTC might be slightly behind others. Let's use a const duration such as -10 seconds

node/node.go
```diff
@@ -827,15 +827,34 @@ func (n *Node) GetCode(codeHash []byte) []byte {
 
 // GetHeartbeats returns the heartbeat status for each public key defined in genesis.json
 func (n *Node) GetHeartbeats() []heartbeatData.PubKeyHeartbeat {
```
iulianpascalau, line 829: a test for these changes would be great

heartbeat/monitor/monitor.go
sstanculeanu: removed

heartbeat/monitor/monitor.go
sstanculeanu:
```go
for idx := range heartbeatsV2 {
	hbData := &heartbeatsV2[idx]
	pk := hbData.PublicKey
	hbData.NumInstances = numInstances[pk]
}
```
Added this solution as the proposed one won't reflect the changes outside the for loop, as `hbData` would be just a copy not a reference.

heartbeat/monitor/monitor.go
sstanculeanu: :+1:  removed the property map as it was not needed anyway

heartbeat/monitor/monitor.go
sstanculeanu: not needed, fixed

heartbeat/monitor/monitor.go
sstanculeanu: Used the same approach as in old heartbeat implementation, with a max between 0 and current value.

node/node.go
sstanculeanu: :+1: added

iulianpascalau (approved):

heartbeat/monitor/monitor.go
```diff
@@ -0,0 +1,202 @@
...
+		return pubKeyHeartbeat, err
+	}
+
+	peerInfo := monitor.peerShardMapper.GetPeerInfo(pid)
+
+	crtTime := time.Now()
+	messageAge := monitor.getMessageAge(crtTime, payload.Timestamp)
+	stringType := peerInfo.PeerType.String()
+	if monitor.shouldSkipMessage(messageAge, stringType) {
+		return pubKeyHeartbeat, fmt.Errorf("validator should be skipped")
```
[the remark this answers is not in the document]
sstanculeanu: added new error

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #3863, opened 2022-03-02
Close network components first

ssd04:
Changed closable components order on close: set network components to be closed at first, so that network traffic will not be initiated while data and state components are still active.

Added nil checks for status handler goroutine which will try to fetch network components metric:
- nil check for network messenger component


AdoAdoAdo (approved):

iulianpascalau (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #3865, opened 2022-03-02
Improve scheduled SC execution dest me

SebastianMarian:
* Improved scheduled SC execution by getting all mini blocks cross shard dest me with type TxBlock, no matter if they were not executed as scheduled in the source shard
* Avoided splitting each mini block of type TxBlock from me, in two mini blocks (one with not SC calls txs and one with only SC calls txs)


iulianpascalau (dismissed):
shallow review, looks good

AdoAdoAdo (approved):

raduchis (approved):

andreibancioiu:
For preparing scheduled transactions, perhaps we should use a smaller `batchSizePerSender` (e.g. `=1 `) when calling `SelectTransactionsWithBandwidth(numRequested int, batchSizePerSender int, bandwidthPerSender uint64)`. That is, to increase "sender diversity" among the selection transactions. 

Can you confirm that we already do that / or should?

AdoAdoAdo (dismissed):

raduchis (dismissed):

update/sync/syncTransactions.go
```diff
@@ -165,16 +165,13 @@ func (ts *transactionsSync) requestTransactionsFor(miniBlock *block.MiniBlock) i
...
 	case block.SmartContractResultBlock:
 		ts.requestHandler.RequestUnsignedTransactions(miniBlock.SenderShardID, missingTxs)
-		// TODO - uncomment this
-		// ts.requestHandler.RequestUnsignedTransactions(miniBlock.ReceiverShardID, missingTxs)
+		ts.requestHandler.RequestUnsignedTransactions(miniBlock.ReceiverShardID, missingTxs)
 	case block.RewardsBlock:
 		ts.requestHandler.RequestRewardTransactions(miniBlock.SenderShardID, missingTxs)
-		// TODO - uncomment this
-		// ts.requestHandler.RequestRewardTransactions(miniBlock.ReceiverShardID, missingTxs)
+		ts.requestHandler.RequestRewardTransactions(miniBlock.ReceiverShardID, missingTxs)
```
sasurobert, line 174: Why is this component used for bootstrap now ? In case of update - there is a different process.

process/block/preprocess/smartContractResults.go
```diff
@@ -333,13 +333,6 @@ func (scr *smartContractResults) SaveTxsToStorage(body *block.Body) error {
 		if miniBlock.Type != block.SmartContractResultBlock {
 			continue
 		}
```
sasurobert, line 335: It is still little bit useless to these miniblocks. This is mostly needed by destination shard. not sender.

update/sync/syncTransactions.go
sasurobert: In case of hard fork - this helps the requesting and getting of data.

epochStart/bootstrap/process.go
```diff
@@ -1099,6 +1099,7 @@ func (e *epochStartBootstrap) createRequestHandler() error {
 
 	storageService := disabled.NewChainStorer()
 
+	log.Debug("epochStartBootstrap.createRequestHandler", "shard", e.shardCoordinator.SelfId())
```
sasurobert, line 1102: the request handlers - as metaResolversContainers should be used for Validator sync - epochstart metablocks. After that the node should know in which shard it is and have proper request handlers. With this, the node knows it is in shard 2 for example, but it is still pretending on being in meta.

sasurobert: maybe add a TODO, a refactor here.

update/sync/syncTransactions.go
AdoAdoAdo: for bootstrap with scheduled active, we need to request txs from last blocks

sasurobert: Even in case of shuffle out without scheduled we request tes from last blocks.

epochStart/bootstrap/process.go
iulianpascalau: added TODO

SebastianMarian:
Looks good :)

sasurobert (dismissed):

sasurobert (approved):

iulianpascalau (changes requested):
backwards incompatibility detected on this PR

iulianpascalau (approved):
import-db process is now working :+1:

AdoAdoAdo (approved):

raduchis (approved):

gabi-vuls (approved):
System test passed.

PR #3869, opened 2022-03-03
[TO BE TESTED]Improve scheduled sc execution dest me test

iulianpascalau:


dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -196,7 +197,8 @@ func (rrh *resolverRequestHandler) RequestRewardTransactions(destShardID uint32,
 
 // RequestMiniBlock method asks for miniblock from the connected peers
 func (rrh *resolverRequestHandler) RequestMiniBlock(destShardID uint32, miniblockHash []byte) {
-	if !rrh.testIfRequestIsNeeded(miniblockHash) {
+	suffix := fmt.Sprintf("miniblocks_%d", destShardID)
```
sasurobert, line 200: I think smaller suffix would help.

sasurobert: Like 2 letters - and maybe define them as costs

iulianpascalau: on it

iulianpascalau: done

PR #3875, opened 2022-03-04
Handle epoch zero

ssd04:


api/groups/networkGroup.go
```diff
@@ -38,9 +40,15 @@ type networkFacadeHandler interface {
 	StatusMetrics() external.StatusMetricsHandler
 	GetAllIssuedESDTs(tokenType string) ([]string, error)
 	GetTokenSupply(token string) (*api.ESDTSupply, error)
+	GetGenesisNodesPubKeys() (map[uint32][][]byte, map[uint32][][]byte)
 	IsInterfaceNil() bool
 }
 
+type GenesisNodesConfig struct {
```
iulianpascalau, line 47: missing comment

api/groups/networkGroup.go
```diff
@@ -314,6 +327,32 @@ func (ng *networkGroup) getRatingsConfig(c *gin.Context) {
 	)
 }
 
+// getGenesisNodesConfig return genesis nodes configuration
+func (ng *networkGroup) getGenesisNodesConfig(c *gin.Context) {
+	start := time.Now()
+	eligibleNodesConfig, waitingNodesConfig := ng.getFacade().GetGenesisNodesPubKeys()
+	log.Debug(fmt.Sprintf("GetGenesisNodesConfig took %s", time.Since(start)))
+
+	if eligibleNodesConfig == nil || waitingNodesConfig == nil {
```
iulianpascalau, line 336: could have returned an error from the ng.getFacade().GetGenesisNodesPubKeys() as to signal the operation failed.
This is slightly cleaner as we do not need to always remember to test both returned values. Just an error, the if it is not nil it just failed and that's it.

node/external/blockAPI/internalBlock.go
```diff
@@ -82,13 +83,20 @@ func (ibp *internalBlockProcessor) GetInternalShardBlockByRound(format common.Ap
...
 
-func (ibp *internalBlockProcessor) convertShardBlockBytesToInternalBlock(blockBytes []byte) (*block.Header, error) {
-	blockHeader := &block.Header{}
-	err := ibp.marshalizer.Unmarshal(blockHeader, blockBytes)
+func (ibp *internalBlockProcessor) convertShardBlockBytesToInternalBlock(blockBytes []byte) (interface{}, error) {
+	hdr, err := process.CreateShardHeader(ibp.marshalizer, blockBytes)
 	if err != nil {
 		return nil, err
 	}
+	blockHeaderV2, ok := hdr.(*block.HeaderV2)
```
iulianpascalau, line 91: Please remove these casts and directly return the hdr.
Reason: If we add a new type of header (let's say v3) we need to remember that somewhere in the node's block api struct we have a duplicated logic only on some casts. 

The function will become:
```GO
func (ibp *internalBlockProcessor) convertShardBlockBytesToInternalBlock(blockBytes []byte) (interface{}, error) {
	return process.CreateShardHeader(ibp.marshalizer, blockBytes)
}
```

node/external/nodeApiResolver.go
```diff
@@ -141,7 +149,7 @@ func (nar *nodeApiResolver) GetBlockByRound(round uint64, withTxs bool) (*api.Bl
 	return nar.apiBlockHandler.GetBlockByRound(round, withTxs)
 }
 
-// GetInternalMetaBlockByHash wil return a meta block by hash
+// GetInternalMetaBlockByHash will return a meta block by hash
```
iulianpascalau, line 152: :+1:

node/external/nodeApiResolver_test.go
```diff
@@ -200,7 +204,7 @@ func TestNodeApiResolver_GetTotalStakedValue(t *testing.T) {
 	t.Parallel()
 
 	wasCalled := false
-	arg := createMockAgrs()
+	arg := createMockArgs()
```
iulianpascalau, line 207: :+1:

node/external/blockAPI/internalBlock.go
ssd04: since the header will be exposed to gin as json, it has to be concrete type, not interface, so in the end i have to cast it, or to have a separate function for this

api/groups/networkGroup.go
ssd04: done

api/groups/networkGroup.go
ssd04: done

node/external/blockAPI/internalBlock.go
iulianpascalau: json can handle happily "interfaces". Remember that during execution, the will be concrete types flowing through here. Please do a test.

ssd04: thanks :+1: just checked and it works fine

iulianpascalau (approved):

api/groups/networkGroup.go
```diff
@@ -38,9 +40,16 @@ type networkFacadeHandler interface {
...
 	GetAllIssuedESDTs(tokenType string) ([]string, error)
 	GetTokenSupply(token string) (*api.ESDTSupply, error)
+	GetGenesisNodesPubKeys() (map[uint32][][]byte, map[uint32][][]byte, error)
 	IsInterfaceNil() bool
 }
 
+// GenesisNodesConfig defines the eligible and waiting nodes configurations
+type GenesisNodesConfig struct {
+	Eligible map[uint32][][]byte `json:"eligible"`
+	Waiting  map[uint32][][]byte `json:"waiting"`
```
bogdan-rosianu, line 50: please expose them as string (hex representation, aka by using validator pub key converter)

api/groups/networkGroup_test.go
```diff
@@ -50,6 +50,16 @@ type esdtTokensResponse struct {
 	Code  string
 }
 
+type genesisNodesConfigResponse struct {
+	Data  genesisNodesConfigData `json:"data"`
+	Error string                 `json:"error"`
+	Code  string
```
bogdan-rosianu, line 56: forgotten `json:"code"`

cmd/node/config/api.toml
```diff
@@ -224,7 +227,7 @@
         { Name = "/raw/miniblock/by-hash/:hash/epoch/:epoch", Open = true },
 
         # /internal/json/miniblock/by-hash/:hash will return the miniblock in JSON format based on hash for a specified epoch
-        { Name = "/json/miniblock/by-hash/:hash/epoch/:epoch", Open = true }
+        { Name = "/json/miniblock/by-hash/:hash/epoch/:epoch", Open = true },
```
bogdan-rosianu, line 230: the last item in a toml list doesn't need an appended `,`

facade/initial/initialNodeFacade.go
```diff
@@ -341,6 +341,11 @@ func (inf *initialNodeFacade) GetTokenSupply(_ string) (*api.ESDTSupply, error)
 	return nil, errNodeStarting
 }
 
+// GetGenesisNodesPubKeys returns nothing
```
bogdan-rosianu, line 344: actually it returns something - the node is starting error

facade/nodeFacade.go
```diff
@@ -552,6 +552,16 @@ func (nf *nodeFacade) convertVmOutputToApiResponse(input *vmcommon.VMOutput) *vm
 	}
 }
 
+// GetGenesisNodesPubKeys will return genesis nodes public keys by shard
+func (nf *nodeFacade) GetGenesisNodesPubKeys() (map[uint32][][]byte, map[uint32][][]byte, error) {
+	eligible, waiting := nf.apiResolver.GetGenesisNodesPubKeys()
+	if eligible == nil || waiting == nil {
```
bogdan-rosianu, line 558: might change the condition to `eligible == nil && waiting == nil` as if we'd have for example only eligible and no waiting, then you'd return error but it would be better to see the actual eligible and a `null` for waiting

node/external/nodeApiResolver.go
```diff
@@ -197,6 +205,24 @@ func (nar *nodeApiResolver) GetInternalMiniBlock(format common.ApiOutputFormat,
 	return nar.apiInternalBlockHandler.GetInternalMiniBlock(format, decodedHash, epoch)
 }
 
+// GetGenesisNodesPubKeys will return genesis nodes public keys by shard
+func (nar *nodeApiResolver) GetGenesisNodesPubKeys() (map[uint32][][]byte, map[uint32][][]byte) {
+	eligibleNodesConfig, waitingNodesConfig := nar.genesisNodesSetupHandler.InitialNodesInfo()
```
bogdan-rosianu, line 210: as mentioned in another comment, please use validator pub key converter so in the entire PR you should have `map[uint32][]string` instead of `map[uint32][][]byte`. 
you can convert the bytes in the `getInitialNodesPubKeysBytes` function

bogdan-rosianu (changes requested):

api/groups/networkGroup.go
iulianpascalau: why not leaving them as []byte?

bogdan-rosianu: because in almost every endpoint that we have "access" to return the data in a desired format (vm-query return data excluded) we expose them in human readable format.

node/external/nodeApiResolver.go
ssd04: will instantiate a pubKeyConverter in sdk also, maybe with a const for address length

facade/nodeFacade.go
ssd04: modified the condition :+1:

facade/initial/initialNodeFacade.go
ssd04: :+1:

AdoAdoAdo (approved):

api/groups/networkGroup_test.go
ssd04: added. thanks

api/groups/networkGroup.go
ssd04: changed to expose as hex encoded string

node/external/nodeApiResolver_test.go
```diff
@@ -340,7 +342,9 @@ func TestNodeApiResolver_GetGenesisNodesPubKeys(t *testing.T) {
 	t.Parallel()
 
 	pubKey1 := []byte("pubKey1")
+	expPubKey1 := hex.EncodeToString(pubKey1)
```
bogdan-rosianu, line 345: Could have used validator key converter instead of hex

bogdan-rosianu (approved):

PR #3884, opened 2022-03-07
return node is starting for all api calls when needed

bogdan-rosianu:
refactored the way `node is starting` error is returned. before this PR, there were some endpoints that didn't return an error and also returned code 200


api/groups/nodeGroup_test.go
```diff
@@ -125,6 +126,66 @@ func TestHeartbeatstatus(t *testing.T) {
 	assert.NotEqual(t, "", statusRsp.Message)
 }
 
+func TestP2PMetrics_ShouldReturnErrorIfFacadeReturnsError(t *testing.T) {
```
sstanculeanu, line 129: perhaps a new `TestQueryDebug_ShouldReturnErrorIfFacadeReturnsError` for extra coverage

facade/initial/disabledStatusMetrics.go
```diff
@@ -13,44 +12,42 @@ func NewDisabledStatusMetricsHandler() *disabledStatusMetricsHandler {
...
+func (d *disabledStatusMetricsHandler) NetworkMetrics() (map[string]interface{}, error) {
+	return getReturnValues()
 }
 
-func getReturnMap() map[string]interface{} {
-	return map[string]interface{}{
-		responseKey: responseValue,
-	}
+func getReturnValues() (map[string]interface{}, error) {
+	return map[string]interface{}{}, errNodeStarting
```
sstanculeanu, line 50:
```suggestion
	return map[string]interface{}{
		responseKey: responseValue,
	}, errNodeStarting
```
or remove responseValue as it is not used anymore

integrationTests/mock/statusMetricsStub.go
```diff
@@ -2,54 +2,54 @@ package mock
 
 // StatusMetricsStub -
 type StatusMetricsStub struct {
```
ssd04, line 4: i think we can remove also this mock and reference the newly created one from testscommon

api/groups/nodeGroup_test.go
bogdan-rosianu: the situation is already covered in `TestQueryDebug_GetQueryErrorsShouldErr`

facade/initial/disabledStatusMetrics.go
bogdan-rosianu: removed

integrationTests/mock/statusMetricsStub.go
bogdan-rosianu: good catch. done

sstanculeanu (dismissed):

ssd04 (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #3898, opened 2022-03-14
small fix for loading data from storage in epoch start bootstrap

bogdan-rosianu:
due to bad configs/dbs, the nodes configurations may not have the latest epoch, which could result in a nil pointer dereference. this can not happen when the configs/dbs are ok


iulianpascalau (approved):

gabi-vuls (approved):
System test passed

PR #3904, opened 2022-03-15
process: add global shard stuck condition for chain recovery

AdoAdoAdo:
This PR adds another layer of protection for stuck shard extreme scenarios.
In case there are a lot of cross shard transactions for one shard, the previous condition was not sufficient to allow the shard to catch up with processing the metachain blocks. 
To ensure the recovery, another condition was added to stop all shard originating transactions while the pending miniblocks or delta metachain block nonces are above a new defined threshold.


iulianpascalau (approved):

iulianpascalau (dismissed):

SebastianMarian (dismissed):

bogdan-rosianu (approved):

iulianpascalau (approved):

bogdan-rosianu (dismissed):

iulianpascalau (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.

PR #3907, opened 2022-03-16
Shard validator info handler map interface

mariusmihaic:
- Added `ShardValidatorsInfoMapHandler` interface which shall be used to manage operations inside a `<shardID,[]ValidatorInfoHandler>` map in a concurrent-safe way. This is used to replace operations with pointers on a `<shardID, []*ValidatorInfo>` map. Integration PRs with this interface will follow. For now, basic functionalities have been provided, this interface being subject to change in future PRs
- Added `ValidatorInfoHandler` (similar with `HeaderHandler` - getters + setters) interface to be used instead of `*ValidatorInfo`
- Added full branch coverage + concurrency unit tests


state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+)
+
+type shardValidatorsInfoMap struct {
+	mutex      sync.RWMutex
+	valInfoMap map[uint32][]ValidatorInfoHandler
+}
+
+// NewShardValidatorsInfoMap creates an instance of shardValidatorsInfoMap which manages a
+// <shardID, validatorsInfo> map internally
+func NewShardValidatorsInfoMap(numOfShards uint32) *shardValidatorsInfoMap {
```
iulianpascalau, line 19: you can simply this as not providing the numOfShards
we do not need a size for a map variable

state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+		valInfoMap: make(map[uint32][]ValidatorInfoHandler, numOfShards),
+	}
+}
+
+// TODO: Delete this once map[uint32][]*ValidatorInfo is completely replaced with new interface
+
+// CreateShardValidatorsMap creates an instance of shardValidatorsInfoMap which manages a shard validator
+// info map internally.
+func CreateShardValidatorsMap(input map[uint32][]*ValidatorInfo) *shardValidatorsInfoMap {
+	ret := &shardValidatorsInfoMap{valInfoMap: make(map[uint32][]ValidatorInfoHandler, len(input))}
```
iulianpascalau, line 31: also here, we do not need the call to len(input)

state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+		copy(validatorsCopy, validatorsInShard)
+		ret = append(ret, validatorsCopy...)
+	}
+
+	return ret
+}
+
+// GetShardValidatorsInfoMap returns a <shard, ValidatorInfoHandler> map copy of internally stored data
+func (vi *shardValidatorsInfoMap) GetShardValidatorsInfoMap() map[uint32][]ValidatorInfoHandler {
+	ret := make(map[uint32][]ValidatorInfoHandler, len(vi.valInfoMap))
```
iulianpascalau, line 60: len not required

state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+		ret[shardID] = validatorsCopy
+	}
+
+	return ret
+}
+
+// Add adds a ValidatorInfoHandler in its corresponding shardID
+func (vi *shardValidatorsInfoMap) Add(validator ValidatorInfoHandler) error {
+	if check.IfNil(validator) {
+		return ErrNilValidatorInfo
```
iulianpascalau, line 77: :+1:

state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+			new.GetShardId(),
+		)
+	}
+
+	shardID := old.GetShardId()
+
+	vi.mutex.Lock()
+	defer vi.mutex.Unlock()
+
+	for idx, validator := range vi.valInfoMap[shardID] {
```
iulianpascalau, line 124: slight optimization:
```GO
validatorSlice := vi.valInfoMap[shardID]
for idx, validator := range validatorSlice {
....
```

state/validatorsInfoMap.go
```diff
@@ -0,0 +1,224 @@
...
+
+// GetShardValidatorsInfoMap returns a <shard, ValidatorInfoHandler> map copy of internally stored data
+func (vi *shardValidatorsInfoMap) GetShardValidatorsInfoMap() map[uint32][]ValidatorInfoHandler {
+	ret := make(map[uint32][]ValidatorInfoHandler, len(vi.valInfoMap))
+
+	vi.mutex.RLock()
+	defer vi.mutex.RUnlock()
+
+	for shardID, validatorsInShard := range vi.valInfoMap {
+		validatorsCopy := make([]ValidatorInfoHandler, len(validatorsInShard))
```
iulianpascalau, line 66: :+1: by always using copies of slices you prevented the wrongly usage of this component, by directly making a slice element to nil that would have altered the inner map's components

state/validatorsInfoMap_test.go
```diff
@@ -0,0 +1,358 @@
...
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/ElrondNetwork/elrond-go-core/core"
+	"github.com/stretchr/testify/require"
+)
+
+func TestShardValidatorsInfoMap_Add_Delete_Replace_SetValidatorsInShard_NilValidators(t *testing.T) {
```
iulianpascalau, line 14: maybe stick to the rule: "one _ in test function names" ?
And use t.Run if more complex description is required

iulianpascalau (commented):
minor stuff

state/validatorsInfoMap.go
mariusmihaic: I was thinking the same, but was not sure. I'll delete it, thanks for suggestion !

state/validatorsInfoMap.go
mariusmihaic: I think here it is ok to call len to allocate memory for the map, since they will have the same size, right?

state/validatorsInfoMap.go
mariusmihaic: Right, this was the intention !

state/validatorsInfoMap_test.go
mariusmihaic: Refactored name to `TestShardValidatorsInfoMap_OperationsWithNilValidators` and added t.Run for each nil operation

state/validatorsInfoMap.go
mariusmihaic: :pray:

state/validatorsInfoMap.go
mariusmihaic: Same comment as above. This link also states something about it:

https://stackoverflow.com/questions/53388332/memory-efficient-implementation-of-go-map
> Go maps are optimized for integer keys. Optimize the map allocation by giving the exact map size as a hint.

state/validatorsInfoMap.go
mariusmihaic: Actually, it seems like this is slightly less efficient. I've run a benchmark on a map with 1,000,000 entries which
- `Iterate1` iterates using `for range vi.valInfoMap[shardID]`
- `Iterate2` iterates using:
``` go
 vSlice := vi.valInfoMap[shardID]
	for range vSlice {
}
```

Results:
```
BenchmarkShardValidatorsInfoMap_IterateMap1/iterate_1
BenchmarkShardValidatorsInfoMap_IterateMap1/iterate_1-8         	1000000000	         0.0002455 ns/op
BenchmarkShardValidatorsInfoMap_IterateMap1/iterate_2
BenchmarkShardValidatorsInfoMap_IterateMap1/iterate_2-8         	1000000000	         0.0002704 ns/op
```

iulianpascalau: ok

iulianpascalau (approved):

PR #3911, opened 2022-03-16
Rollback final block with scheduled

AdoAdoAdo:
This PR allows rollback of final block to allow re-processing if it holds scheduled mini blocks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/ElrondNetwork/concurrent-map v0.1.3
 	github.com/ElrondNetwork/covalent-indexer-go v1.0.6
 	github.com/ElrondNetwork/elastic-indexer-go v1.1.37
-	github.com/ElrondNetwork/elrond-go-core v1.1.14
+	github.com/ElrondNetwork/elrond-go-core v1.1.15-0.20220316125754-b5ee1f265897
 	github.com/ElrondNetwork/elrond-go-crypto v1.0.1
 	github.com/ElrondNetwork/elrond-go-logger v1.0.5
 	github.com/ElrondNetwork/elrond-vm-common v1.2.12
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/ElrondNetwork/concurrent-map v0.1.3
 	github.com/ElrondNetwork/covalent-indexer-go v1.0.6
 	github.com/ElrondNetwork/elastic-indexer-go v1.1.37
-	github.com/ElrondNetwork/elrond-go-core v1.1.14
+	github.com/ElrondNetwork/elrond-go-core v1.1.15-0.20220316125754-b5ee1f265897
```
iulianpascalau, line 12: let's not forget proper release

process/sync/baseSync.go
```diff
@@ -826,6 +828,24 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
 	return nil
 }
 
+func (boot *baseBootstrap) shouldAllowRollback(currHeader data.HeaderHandler) bool {
+	finalBlockNonce := boot.forkDetector.GetHighestFinalBlockNonce()
+	isRollBackBehindFinal := currHeader.GetNonce() <= finalBlockNonce
+	isFinalBlockRollBack := currHeader.GetNonce() == finalBlockNonce
```
iulianpascalau, line 834: so this will ensure that we only revert once because the highest final block nonce in the fork detector will keep the last value.
What happens if the node is slightly behind and the fork detector has a nonce slightly larger than one? The rollback will never happen and the node will still be stuck.

SebastianMarian: This scenario should not be happen, as there is no reason to roll back two blocks behind final. Fork detector never returns a final block nonce higher than your last processed and committed even if the tracker feeds it with the future blocks from the longest chain. Fork detector returns the minimum nonce between the higher received as notarized by meta (from tracker) and the higher processed and committed which match with the one added by the tracker in the longest chain.

SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #3915, opened 2022-03-18
Integrate validators info map into validators statistics

mariusmihaic:
- Replaced `map[uin32][]*ValidatorInfo` with `ShardValidatorsInfoMap` interface in `validatorStatistics`
- Deleted duplicated stubs and kept only one in `testscommon`


process/peer/validatorsProvider.go
```diff
@@ -180,7 +180,8 @@ func (vp *validatorsProvider) updateCache() {
 		return
 	}
 	allNodes, err := vp.validatorStatistics.GetValidatorInfoForRootHash(lastFinalizedRootHash)
-	if err != nil {
+	if err != nil || allNodes == nil {
```
bogdan-rosianu, line 183: I don't think there is a case when `allNodes` is `nil` but the error isn't`

bogdan-rosianu (dismissed):

ssd04 (commented):
from what i've seen, there is one more `ValidatorStatisticsProcessorStub` in consensus, which is not being used, i think we can remove that one also

sasurobert (approved):

ssd04 (approved):

bogdan-rosianu (approved):

PR #3927, opened 2022-03-21
Refactor communication on connection topic

sstanculeanu:
- added `directConnectionsProcessor` which sends a message with shard id to all connected peers on a separate goroutine. Added this component as a `heartbeatV2Component`.
- added interceptor for `ConnectionTopic` used by the above component, which saves the pair pid-shard id into peer shard mapper.
- removed `crossShardStatusProcessor` as it is not used anymore.
- now networkMessenger does not implement connectionWatcher. Not needed with the new flow. Also removed `currentPayloadProvider`.
- removed all the todos left for debug only and updated the network sharding test to not check for unknown peers


heartbeat/processor/connectionsProcessor.go
```diff
@@ -0,0 +1,147 @@
...
+
+// ArgConnectionsProcessor represents the arguments for the connections processor
+type ArgConnectionsProcessor struct {
+	Messenger                 p2p.Messenger
+	Marshaller                marshal.Marshalizer
+	ShardCoordinator          sharding.Coordinator
+	DelayBetweenNotifications time.Duration
+}
+
+type connectionsProcessor struct {
```
iulianpascalau, line 27: maybe rename to `directConnectionsProcessor` ?

heartbeat/processor/connectionsProcessor_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func Test_connectionsProcessor_computeNewPeers(t *testing.T) {
+	t.Parallel()
+
+	t.Run("no peers connected", func(t *testing.T) {
+		t.Parallel()
+
+		cp, _ := NewConnectionsProcessor(createMockArgConnectionsProcessor())
+		assert.False(t, check.IfNil(cp))
+		_ = cp.Close() // avoid data races
```
iulianpascalau, line 133: data races can happen in extreme situations: the go routine does not stop in time during high CPU usages and instead, accesses the map
In order to completely use a connection processor without the routine started, why not create a new constructor in export_test.go, something like: `NewConnectionsProcessorNoGoRoutine` 
Valid on all occurrences

heartbeat/processor/peerAuthenticationRequestsProcessor_test.go
```diff
@@ -271,6 +271,7 @@ func TestPeerAuthenticationRequestsProcessor_requestKeysChunks(t *testing.T) {
 	processor, err := NewPeerAuthenticationRequestsProcessor(args)
 	assert.Nil(t, err)
 	assert.False(t, check.IfNil(processor))
+	_ = processor.Close() // avoid data races
```
iulianpascalau, line 274: does this component "suffer" from concurrent accesses problem?

integrationTests/testHeartbeatNode.go
```diff
@@ -495,6 +495,7 @@ func (thn *TestHeartbeatNode) initInterceptors() {
 		CoreComponents: &processMock.CoreComponentsMock{
 			IntMarsh: TestMarshaller,
 		},
+		ShardCoordinator:             thn.ShardCoordinator,
```
iulianpascalau, line 498: instead of using comments on L506, L515, L526, you could have extracted the code in simple functions. 
Bonus if you apply this: all defined arguments can be called `args` as they "live" only inside those functions.

process/factory/interceptorscontainer/baseInterceptorsContainerFactory.go
```diff
@@ -687,3 +687,46 @@ func (bicf *baseInterceptorsContainerFactory) generateHeartbeatInterceptor() err
 
 	return bicf.container.Add(identifierHeartbeat, interceptor)
 }
+
+// ------- ShardValidatorInfo interceptor
+
+func (bicf *baseInterceptorsContainerFactory) generateShardValidatorInfoInterceptor() error {
+	identifier := common.ConnectionTopic
+
+	shardValidatorInfoFactory, err := interceptorFactory.NewInterceptedShardValidatorInfoFactory(*bicf.argInterceptorFactory)
```
iulianpascalau, line 696: maybe find another name here? NewInterceptedShardValidatorInfoFactory suggest that this component might be used on all "shard" nodes and we might have a "meta" implementation

process/p2p/InterceptedShardValidatorInfo.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Type returns the type of this intercepted data
+func (isvi *interceptedShardValidatorInfo) Type() string {
+	return interceptedShardValidatorInfoType
+}
+
+// Identifiers always returns an array with an empty string
+func (isvi *interceptedShardValidatorInfo) Identifiers() [][]byte {
+	return [][]byte{[]byte("")}
```
iulianpascalau, line 97: `return [][]byte{make([]byte, 0)}`?

heartbeat/processor/peerAuthenticationRequestsProcessor_test.go
sstanculeanu: not really, deleted it

process/factory/interceptorscontainer/baseInterceptorsContainerFactory.go
iulianpascalau: renaming is not complete: "shard" persists
Also we still have the `interceptedShardValidatorInfoFactory` struct & `generateShardValidatorInfoInterceptor` function

process/interceptors/processor/shardValidatorInfoInterceptorProcessor_test.go
```diff
@@ -118,7 +118,7 @@ func Test_shardValidatorInfoInterceptorProcessor_Save(t *testing.T) {
 	})
 }
 
-func Test_shardValidatorInfoInterceptorProcessor_DisabledMethods(t *testing.T) {
+func Test_validatorInfoInterceptorProcessor_DisabledMethods(t *testing.T) {
```
iulianpascalau, line 121: TestValidatorInfoInterceptorProcessor_DisabledMethod ? 
This suggestion will comply with the rest of the project's naming conventions

process/interceptors/processor/shardValidatorInfoInterceptorProcessor_test.go
```diff
@@ -15,60 +15,60 @@ import (
...
 
-		processor, err := NewShardValidatorInfoInterceptorProcessor(createMockArgShardValidatorInfoInterceptorProcessor())
+		processor, err := NewValidatorInfoInterceptorProcessor(createMockArgValidatorInfoInterceptorProcessor())
 		assert.Nil(t, err)
 		assert.False(t, check.IfNil(processor))
 	})
 }
 
-func Test_shardValidatorInfoInterceptorProcessor_Save(t *testing.T) {
+func Test_validatorInfoInterceptorProcessor_Save(t *testing.T) {
```
iulianpascalau, line 57: TestValidatorInfoInterceptorProcessor_Save ?

process/p2p/InterceptedShardValidatorInfo.go
```diff
@@ -0,0 +1,113 @@
+package p2p
+
```
ssd04, line 2: usually we start file names with lower case, i think it's better to change it, for consistency

process/p2p/InterceptedShardValidatorInfo_test.go
```diff
@@ -0,0 +1,125 @@
+package p2p
+
```
ssd04, line 2: same here, for filename

process/interceptors/processor/shardValidatorInfoInterceptorProcessor.go
```diff
@@ -0,0 +1,66 @@
...
+		return process.ErrWrongTypeAssertion
+	}
+
+	processor.peerShardMapper.PutPeerIdShardId(fromConnectedPeer, shardValidatorInfo.ShardID())
+
+	return nil
+}
+
+// RegisterHandler registers a callback function to be notified of incoming shard validator info
+func (processor *validatorInfoInterceptorProcessor) RegisterHandler(_ func(topic string, hash []byte, data interface{})) {
```
ssd04, line 59: maybe specify in the comment, also that this is not implemented

heartbeat/processor/directConnectionsProcessor.go
```diff
@@ -0,0 +1,147 @@
...
+		if !wasNotified {
+			newPeers = append(newPeers, connectedPeer)
+		}
+	}
+
+	return newPeers
+}
+
+func (dcp *directConnectionsProcessor) notifyNewPeers(newPeers []core.PeerID) {
+	dcp.notifiedPeersMap = make(map[core.PeerID]struct{})
```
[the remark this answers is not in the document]
sstanculeanu: yes, it is expected, as an optimization to avoid messages overload..the ones available in `notifiedPeersMap` were already notified and there is no need to notify them again

ssd04 (approved):

iulianpascalau (approved):
maybe remove the `shard` part from the filename:
process/p2p/interceptedShardValidatorInfo.go

PR #3932, opened 2022-03-23
Validator statistics - save shuffled out nodes in auction list

mariusmihaic:
- In `validatorStatistics`, saved all shuffled out nodes from nodes coordinator into `AuctionList`
- Added `GetAllShuffledOutValidatorsPublicKeys` getter in `NodesCoordinator`


iulianpascalau (approved):

AdoAdoAdo (approved):

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
```diff
@@ -502,6 +502,30 @@ func (ihnc *indexHashedNodesCoordinator) GetAllLeavingValidatorsPublicKeys(epoch
...
+	ihnc.mutNodesConfig.RUnlock()
+
+	if !ok {
+		return nil, fmt.Errorf("%w epoch=%v", ErrEpochNodesConfigDoesNotExist, epoch)
+	}
+
+	nodesConfig.mutNodesMaps.RLock()
+	defer nodesConfig.mutNodesMaps.RUnlock()
+
+	for shardID, shuffledOutList := range nodesConfig.shuffledOutMap {
```
bogdan-rosianu, line 520: it is a very small and extreme edge case. If, at first RLock you'll read the epoch 70 and `ok` will be `true` but until your next RLock it will be removed from the map, then you will range over an empty map, and instead of returning an error, you will return an empty map.
I would suggest that you extend the RLock - RUnlock area

bogdan-rosianu (approved):

PR #3935, opened 2022-03-24
Level DB batch removal fix

iulianpascalau:
- fixed level db removal from batch by not using a hardcoded marker

### DIFF
--- a/api/gin/webServer.go
+++ b/api/gin/webServer.go
@@ -74,7 +74,7 @@ func (ws *webServer) UpdateFacade(facade shared.FacadeHandler) error {
 	return nil
 }
 
-// CreateHttpServer will create a new instance of http.Server and populate it with all the routes
+// StartHttpServer will create a new instance of http.Server and populate it with all the routes
 func (ws *webServer) StartHttpServer() error {
 	ws.Lock()
 	defer ws.Unlock()
@@ -98,8 +98,9 @@ func (ws *webServer) StartHttpServer() error {
 		return err
 	}
 
-	for _, proc := range processors {
+	for idx, proc := range processors {
 		if check.IfNil(proc) {
+			log.Error("got nil middleware processor, skipping it...", "index", idx)
 			continue
 		}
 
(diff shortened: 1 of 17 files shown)


storage/interface.go
```diff
@@ -38,6 +38,7 @@ type Batcher interface {
 	Delete(key []byte) error
 	// Reset clears the contents of the batch
 	Reset()
+	IsRemoved(key []byte) bool
```
bogdan-rosianu, line 41: add comment, for consistency

storage/leveldb/leveldb.go
```diff
@@ -151,11 +150,13 @@ func (s *DB) Get(key []byte) ([]byte, error) {
 		return nil, storage.ErrDBIsClosed
 	}
 
+	isRemoved := s.batch.IsRemoved(key)
+	if isRemoved {
+		return nil, storage.ErrKeyNotFound
+	}
+
 	data := s.batch.Get(key)
 	if data != nil {
```
mariusmihaic, line 159: [question]

Couldn't we avoid this by making `Get` from `batch.go` return error and have ErrKeyNotFound when key is not found?

mariusmihaic (approved):

storage/interface.go
iulianpascalau: done

storage/leveldb/leveldb.go
```diff
@@ -151,11 +150,13 @@ func (s *DB) Get(key []byte) ([]byte, error) {
 		return nil, storage.ErrDBIsClosed
 	}
 
+	isRemoved := s.batch.IsRemoved(key)
+	if isRemoved {
```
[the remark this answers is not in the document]
iulianpascalau: done

storage/leveldb/leveldb.go
iulianpascalau: could have been done like that. However I do not intend to change the batcher behavior just for this. The main reason: the implication it might arise.

mariusmihaic (approved):

AdoAdoAdo (approved):
Looks good

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.

PR #3948, opened 2022-03-29
Fix topic error

sstanculeanu:
- added more checks for ConnectionTopic on netMessenger
- fixed monitor_test causing random fails

### DIFF
--- a/p2p/libp2p/netMessenger.go
+++ b/p2p/libp2p/netMessenger.go
@@ -1112,7 +1112,6 @@ func (netMes *networkMessenger) UnjoinAllTopics() error {
 
 	var errFound error
 	for topicName, t := range netMes.topics {
-
 		subscr := netMes.subscriptions[topicName]
 		if subscr != nil {
 			subscr.Cancel()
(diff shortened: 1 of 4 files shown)


p2p/libp2p/netMessenger.go
```diff
@@ -1089,9 +1093,11 @@ func (netMes *networkMessenger) UnregisterAllMessageProcessors() error {
 	defer netMes.mutTopics.Unlock()
 
 	for topic := range netMes.processors {
-		err := netMes.pb.UnregisterTopicValidator(topic)
```
iulianpascalau, line 1092: maybe refactor this way?
```GO
for topic := range netMes.processors {
    if topic == common.ConnectionTopic {
        continue
    }

    err := netMes.pb.UnregisterTopicValidator(topic)
    if err != nil {
        return err
    }

    delete(netMes.processors, topic)
}
```

iulianpascalau (dismissed):

p2p/libp2p/netMessenger.go
sstanculeanu: added this solution, as we still need to delete the processor from netMes.processors(saved for `directMessageHandler`)

```go
for topic := range netMes.processors {
	if topic == common.ConnectionTopic {
		delete(netMes.processors, topic)
		continue
	}

	err := netMes.pb.UnregisterTopicValidator(topic)
	if err != nil {
		return err
	}

	delete(netMes.processors, topic)
}
```

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #3955, opened 2022-03-31
Fix for vm query api logs

bogdan-rosianu:
fixed the situation when VM queries that generate logs didn't return anything + added unit test that covers the entire conversion function

### DIFF
--- a/facade/nodeFacade.go
+++ b/facade/nodeFacade.go
@@ -490,12 +490,12 @@ func (nf *nodeFacade) convertVmOutputToApiResponse(input *vmcommon.VMOutput) *vm
 			logAddress = ""
 		}
 
-		logs[i] = &vm.LogEntryApi{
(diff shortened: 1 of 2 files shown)


facade/nodeFacade.go
```diff
@@ -490,12 +490,12 @@ func (nf *nodeFacade) convertVmOutputToApiResponse(input *vmcommon.VMOutput) *vm
 			logAddress = ""
 		}
 
-		logs[i] = &vm.LogEntryApi{
+		logs = append(logs, &vm.LogEntryApi{
```
iulianpascalau, line 493: :+1: good catch

iulianpascalau (approved):

andreibancioiu (approved):

gabi-vuls (approved):
System test passed.

PR #3974, opened 2022-04-06
Preferred peers refactor

sstanculeanu:


p2p/peersHolder/peersHolder.go
```diff
@@ -0,0 +1,199 @@
...
+		index:                  len(ph.peerIDsPerShard[shardID]) - 1,
+	}
+
+	delete(ph.tempPeerIDsWaitingForShard, peerID)
+}
+
+// Get will return a map containing the preferred peer IDs, split by shard ID
+func (ph *peersHolder) Get() map[uint32][]core.PeerID {
+	ph.RLock()
+	peerIDsPerShardCopy := ph.peerIDsPerShard
```
bogdan-rosianu, line 109: you should create a new map because we can have race conditions this way

p2p/peersHolder/peersHolder.go
```diff
@@ -0,0 +1,199 @@
...
+		delete(ph.tempPeerIDsWaitingForShard, peerID)
+	}
+}
+
+// getKnownSlice checks if the connection address contains any of the initial preferred connection address slices
+// if true, it returns it
+// this function must be called under mutex protection
+func (ph *peersHolder) getKnownSlice(connectionAddressStr string) string {
+	for _, preferredConnAddrSlice := range ph.preferredConnAddrSlices {
+		if strings.Contains(connectionAddressStr, preferredConnAddrSlice) {
```
bogdan-rosianu, line 159: I don't like these `strings.Contains`. Can we have something like a wrapper over the connection string types?
a component that handles a `/ip4/20.200.200.200/tcp/8080/p2p/some-random-pid`, a component that handles a  `/ip4/10.100.100.100`, plus a factory/builder that knows which instance to create.

bogdan-rosianu: might leave a TODO for now, tho

p2p/peersHolder/peersHolder_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	newConnection1 := []byte("/ip4/10.100.100.100/tcp/38191/p2p/some-random-pid") // preferredPeers[0]
+	providedPid1 := core.PeerID("provided pid 1")
+	ph.PutConnectionAddress(providedPid1, newConnection1)
+	providedShardID := uint32(123)
+	ph.PutShardID(providedPid1, providedShardID)
+	assert.True(t, ph.Contains(providedPid1))
+
+	newConnection2 := []byte("/ip4/10.100.100.102/tcp/38191/p2p/16Uiu2HAm6yvbp1oZ6zjnWsn9FdRqBSaQkbhELyaThuq48ybdojvJ") // preferredPeers[1]
+	providedPid2 := core.PeerID("provided pid 1")
```
bogdan-rosianu, line 199: `providedPid2` is equal to `providedPid1`. expected?

bogdan-rosianu (changes requested):

p2p/peersHolder/peersHolder.go
```diff
@@ -0,0 +1,199 @@
...
+	sync.RWMutex
+}
+
+// NewPeersHolder returns a new instance of peersHolder
+func NewPeersHolder(preferredConnectionAddressSlices [][]byte) *peersHolder {
+	preferredConnections := make([]string, 0)
+	connAddrSliceToPeerIDs := make(map[string]*peerInfo)
+
+	for _, connAddrSlice := range preferredConnectionAddressSlices {
+		preferredConnections = append(preferredConnections, string(connAddrSlice))
```
iulianpascalau, line 36: As discussed with @bogdan-rosianu we might have here a string validator for the input strings: let's permit only peer ids or valid ips. Nothing else. This string validator should be called only here, in the constructor as to be as light as possible in processing terms.

p2p/peersHolder/peersHolder.go
```diff
@@ -0,0 +1,199 @@
...
+		ph.tempPeerIDsWaitingForShard[peerID] = knownSlice
+		ph.connAddrSliceToPeerInfo[knownSlice] = &peerInfo{
+			pid:     peerID,
+			shardID: 0, // this will be overwritten once shard is available
+		}
+
+		return
+	}
+
+	isOldData := peerID == pInfo.pid
```
iulianpascalau, line 70: how is this gonna perform if you have defined a preferred connection string as an IP but that IP has 3 peer IDs running?

p2p/peersHolder/peersHolder.go
sstanculeanu: as discussed, the check on ctor would be enough

p2p/peersHolder/peersHolder.go
sstanculeanu: updated the map to support multiple pids for the same connection addr

p2p/peersHolder/connectionStringValidator/connectionStringValidator_test.go
```diff
@@ -0,0 +1,54 @@
...
+	t.Parallel()
+
+	csv := NewConnectionStringValidator()
+	assert.False(t, csv.IsValid("invalid string"))
+
+	assert.True(t, csv.IsValid("5.22.219.242"))
+	assert.True(t, csv.IsValid("2031:0:130F:0:0:9C0:876A:130B"))
+	assert.True(t, csv.IsValid("16Uiu2HAm6yvbp1oZ6zjnWsn9FdRqBSaQkbhELyaThuq48ybdojvJ"))
+}
+func Test_connectionStringValidator_isValidIP(t *testing.T) {
```
iulianpascalau, line 19: add an empty string

p2p/peersHolder/peersHolder.go
```diff
@@ -13,80 +14,106 @@ type peerInfo struct {
...
+	connAddrToPeerIDs := make(map[string][]*peerInfo)
 
-	for _, connAddrSlice := range preferredConnectionAddressSlices {
-		preferredConnections = append(preferredConnections, string(connAddrSlice))
-		connAddrSliceToPeerIDs[string(connAddrSlice)] = nil
+	connectionValidator := connectionStringValidator.NewConnectionStringValidator()
+
+	for _, connAddr := range preferredConnectionAddresses {
+		if !connectionValidator.IsValid(connAddr) {
+			continue
```
iulianpascalau, line 40: silent error treatment. I think we should return an error in order to signal to the user that his/her input in the prefs.toml file was wrong

p2p/peersHolder/peersHolder.go
```diff
@@ -13,80 +14,106 @@ type peerInfo struct {
...
+	// if we have new peer for same connection, add it to maps
+	pInfo := ph.getPeerInfoForPeerID(peerID, peersInfo)
 	if pInfo == nil {
-		ph.tempPeerIDsWaitingForShard[peerID] = knownSlice
-		ph.connAddrSliceToPeerInfo[knownSlice] = &peerInfo{
-			pid:     peerID,
-			shardID: 0, // this will be overwritten once shard is available
-		}
+		ph.addNewPeerInfoToMaps(peerID, knownConnection)
+	}
```
iulianpascalau, line 76: missing statement as an else? what happens if pInfo != nil?

sstanculeanu: if pInfo != nil it means that we already have that specific pid saved for a preferred connection and there is nothing to update for it

p2p/peersHolder/connectionStringValidator/connectionStringValidator_test.go
sstanculeanu: you mean on `TestConnectionStringValidator_IsValid`? added

iulianpascalau (dismissed):

p2p/peersHolder/peersHolder.go
```diff
@@ -134,30 +165,49 @@ func (ph *peersHolder) Remove(peerID core.PeerID) {
...
+
+	var index int
+	var pInfo *peerInfo
+	for index, pInfo = range peersInfo {
+		if peerID == pInfo.pid {
+			break
+		}
+	}
+
+	peersInfo = append(peersInfo[:index], peersInfo[index+1:]...)
```
bogdan-rosianu, line 196: if the peer does not exist, then you will remove index 0. can you treat this case?

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4024, opened 2022-04-27
Feat/refactor peers mbs

SebastianMarian:

In the old implementation, meta chain creates after each epoch a meta block named "start of epoch meta block", in which it adds all the information regarding: rewards, validators info and so on. This block contains something different and not so optimal when we talk about PeersMiniBlocks. In our system all mini blocks hold only the tx hashes inside them, just to keep the block body smaller and to improve its transportation over the network. The only place where this rule is not respected is in the PeersMiniBlocks contained in each start of epoch meta block. Here, we have inside of each peer mini block, the marshalled ValidatorInfo DTO instead of its hash. This conducted to a bigger size of each start of epoch meta block. Right now we have 3200 x marshalled validator info DTO, inside these peers mini blocks in each start of epoch meta block. Assuming that the validators number will grow over the time, also the size of the start of epoch meta block will increase more and more. 

## Proposed Changes

The scope of this feature is to change the structure/approach of the PeersMiniBlocks in each start of epoch meta block in the way we are using it right now in each other mini block in the system. So, inside them we should keep only the hash of the marshalled validator info DTO, and all the nodes in the network will receive on separate topic these validator info txs, or they could request them on dedicated topic created, using the hashes inside each PeerMiniBlock. For this new approach to be achieved, the feature needed to implement/change a lot of stuff, from dedicated interceptors/resolvers/storages/pools but also the creating/processing procedure, which was a breaking change. For this reason a new dedicated activation flag was created, just to preserve the backward compatibility (import-db, etc).  

## Testing procedure

We should test this feature as usual: We need #all-in with UPGRADE, so we can check that all the things will work as expected and no problems will occur before, in the transition epoch (from old approach to the new approach), and also after the new feature will be activated. We can try a lot of restarts before/during/after the transition epoch, and we also need a successful import-db, just to check and validate the fully backwards compatibility of this new feature.


iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

bogdan-rosianu (approved):

PR #4042, opened 2022-05-02
Bug fixes

sstanculeanu:


bogdan-rosianu (dismissed):

process/heartbeat/interceptedPeerAuthentication.go
```diff
@@ -250,6 +250,12 @@ func verifyPropertyLen(property string, value []byte) error {
 	if len(value) > maxSizeInBytes {
 		return fmt.Errorf("%w for %s", process.ErrPropertyTooLong, property)
 	}
+
+	shouldSkipMinSizeCheck := property == identityProperty || property == nodeDisplayNameProperty
```
iulianpascalau, line 254: I do not like this if. Could have had 2 functions like this:
```GO
func verifyPropertyMaxLen(property string, value []byte) error {
	if len(value) > maxSizeInBytes {
		return fmt.Errorf("%w for %s", process.ErrPropertyTooLong, property)
	}
}

func verifyPropertyMinMaxLen(property string, value []byte) error {
	err := verifyPropertyMaxLen(property, value)
        if err != nil{
               return err
        }

        if len(value) < minSizeInBytes {
		return fmt.Errorf("%w for %s", process.ErrPropertyTooShort, property)
	}
	return nil
}
```

and use either function since the calls are done manually, anyway.

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4055, opened 2022-05-04
Fix missing epoch start prepare

AdoAdoAdo:


process/block/interceptedBlocks/interceptedMetaBlockHeader.go
```diff
@@ -91,6 +94,15 @@ func (imh *InterceptedMetaHeader) CheckValidity() error {
 		if err != nil {
 			return err
 		}
+
+		if imh.isMetaHeaderEpochOutOfRange() {
+			// TODO: remove this log after testing
```
bogdan-rosianu, line 99: I think this log can remain as it would be helpful in the future for debugging purposes

bogdan-rosianu (approved):

process/block/interceptedBlocks/interceptedMetaBlockHeader.go
iulianpascalau: We can keep it as a log.Trace because we might end up printing too much when a node will voluntary send bad data

iulianpascalau (approved):

iulianpascalau (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.

PR #4056, opened 2022-05-04
Resolver + interceptor integration

sstanculeanu:
- Integrated the newly implemented resolver and interceptor.
- Added new `ValidatorInfoDataPool`

### DIFF
--- a/dataRetriever/requestHandlers/requestHandler.go
+++ b/dataRetriever/requestHandlers/requestHandler.go
@@ -30,6 +30,7 @@ const uniqueMiniblockSuffix = "mb"
 const uniqueHeadersSuffix = "hdr"
 const uniqueMetaHeadersSuffix = "mhdr"
 const uniqueTrieNodesSuffix = "tn"
+const uniqueValidatorInfoSuffix = "vi"
 
 // TODO move the keys definitions that are whitelisted in core and use them in InterceptedData implementations, Identifiers() function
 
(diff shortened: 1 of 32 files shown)


dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -551,6 +552,36 @@ func (rrh *resolverRequestHandler) RequestMetaHeaderByNonce(nonce uint64) {
 	rrh.addRequestedItems([][]byte{key}, uniqueMetaHeadersSuffix)
 }
 
+// RequestValidatorInfo asks for the validator info associated with a specific hash from connected peers
+func (rrh *resolverRequestHandler) RequestValidatorInfo(hash []byte) {
+	log.Debug("requesting validator info messages from network",
```
SebastianMarian, line 557: You can define in the beginning of the file this identifier:

`const uniqueValidatorInfoSuffix = "vi"`

and add here this check: 

	if !rrh.testIfRequestIsNeeded(hash, uniqueValidatorInfoSuffix) {
		return
	}

then in L573 we will whitelist this hash requested:

`rrh.whiteList.Add([][]byte{hash})`
	
and before L583 we will add this request to avoid another one done with the same hash in the next 1 sec:

`rrh.addRequestedItems([][]byte{hash}, uniqueValidatorInfoSuffix)`

This way we will protect also the validator info request for many requests in a short time frame and also we will whitelist these requests to be accepted on the interceptor in any situation.

process/interceptors/processor/validatorInfoInterceptorProcessor.go
```diff
@@ -60,7 +74,21 @@ func (viip *validatorInfoInterceptorProcessor) Save(data process.InterceptedData
 
 	viip.validatorInfoPool.HasOrAdd(hash, validatorInfo, validatorInfo.Size())
 
-	return nil
+	return viip.updateStorage(hash, validatorInfo, epoch)
```
SebastianMarian, line 77: I think we don't need to put them in the storage. This decision will be taken by the proposer/validator if the received stuff will be used/included in one block. Do we use this approach to the other intercepted objects in this package? If for each intercepted object we will access also the disk the performance would decrease very much in this side, I guess. Maybe @iulianpascalau will have an opinion here.

sstanculeanu: No, this approach is not used in the package. So probably this component should not receive the storer anymore.

sstanculeanu: fixed

iulianpascalau (approved):

PR #4057, opened 2022-05-04
extracted max gas limit for vm query to config

bogdan-rosianu:
extracted the maximum gas limit to be used for vm queries inside config.toml

PR #4058, opened 2022-05-04
rc may <- extract max gas limit for vm query to config

bogdan-rosianu:
the changeset of #4057 but targeted towards `rc/2022-may`


AdoAdoAdo (dismissed):

iulianpascalau (changes requested):
unnecessary PR, we should merge the other one in the `master` branch.
Suggestion to close this one.

PR #4059, opened 2022-05-05
RequestDataFromHashArray

sstanculeanu:
Added `RequestDataFromHashArray` in order to be able to request multiple `validatorInfo` on the same request.


dataRetriever/factory/resolverscontainer/baseResolversContainerFactory.go
```diff
@@ -97,6 +100,10 @@ func (brcf *baseResolversContainerFactory) checkParams() error {
 	if brcf.numFullHistoryPeers <= 0 {
 		return fmt.Errorf("%w for numFullHistoryPeers", dataRetriever.ErrInvalidValue)
 	}
+	if brcf.maxNumOfValidatorInfoInResponse < minNumOfValidatorInfoInResponse {
```
iulianpascalau, line 103: please remove this + the const defined here. The factory should not care about the params used in the instances it creates because this will just double the code logic & defined constants

dataRetriever/resolvers/validatorInfoResolver.go
```diff
@@ -121,7 +150,81 @@ func (res *validatorInfoResolver) resolveHashRequest(hash []byte, epoch uint32,
...
+	validatorInfoForHashes, err := res.fetchValidatorInfoForHashes(hashes, epoch)
+	if err != nil {
+		return fmt.Errorf("resolveMultipleHashesRequest error %w from buff %s", err, hashesBuff)
+	}
+
+	return res.sendValidatorInfoForHashes(validatorInfoForHashes, pid)
+}
+
+func (res *validatorInfoResolver) sendValidatorInfoForHashes(validatorInfoForHashes [][]byte, pid core.PeerID) error {
+	if len(validatorInfoForHashes) > res.maxNumOfValidatorInfoInResponse {
```
iulianpascalau, line 174: instead of using this fixed size variant, why not use the dataPacker implementation as seen in the transactionResolver.go. The maxNumOfValidatorInfoInResponse + config will no longer be necessary as you can define a maximum package size const.

dataRetriever/resolvers/validatorInfoResolver.go
```diff
@@ -124,6 +154,55 @@ func (res *validatorInfoResolver) resolveHashRequest(hash []byte, epoch uint32,
...
+	b := batch.Batch{}
+	err := res.marshalizer.Unmarshal(&b, hashesBuff)
+	if err != nil {
+		return err
+	}
+	hashes := b.Data
+
+	validatorInfoForHashes, err := res.fetchValidatorInfoForHashes(hashes, epoch)
+	if err != nil {
+		return fmt.Errorf("resolveMultipleHashesRequest error %w from buff %s", err, hashesBuff)
```
bogdan-rosianu, line 168: I don't think that the print for `hashesBuff` will work as expected, because it will display the string of each hash, but you would need the hex string of each hash

dataRetriever/resolvers/validatorInfoResolver_test.go
```diff
@@ -361,6 +440,198 @@ func TestValidatorInfoResolver_ProcessReceivedMessage(t *testing.T) {
...
+	t.Run("unmarshal fails", func(t *testing.T) {
+		t.Parallel()
+
+		numCalls := 0
+		expectedErr := errors.New("expected err")
+		args := createMockArgValidatorInfoResolver()
+		args.Marshaller = &testscommon.MarshalizerStub{
+			UnmarshalCalled: func(obj interface{}, buff []byte) error {
+				marshallerMock := testscommon.MarshalizerMock{}
+				if numCalls < 1 {
```
bogdan-rosianu, line 454: so the first unmarshal is for request data which should work, and you want to simulate the unmarshalling failing for the batch, right? might add a comment

bogdan-rosianu: or you can use a type switch like this:
```golang
UnmarshalCalled: func(obj interface{}, buff []byte) error {
				switch obj.(type) {
				case *dataRetriever.RequestData:
					return testscommon.MarshalizerMock{}.Unmarshal(obj, buff)
				case *batch.Batch:
					return expectedErr
				}
				return nil
			},
```

dataRetriever/resolvers/validatorInfoResolver_test.go
```diff
@@ -361,6 +440,198 @@ func TestValidatorInfoResolver_ProcessReceivedMessage(t *testing.T) {
...
+		}
+		res, _ := resolvers.NewValidatorInfoResolver(args)
+		require.False(t, check.IfNil(res))
+
+		b := &batch.Batch{
+			Data: [][]byte{[]byte("hash")},
+		}
+		buff, _ := args.Marshaller.Marshal(b)
+		err := res.ProcessReceivedMessage(createRequestMsg(dataRetriever.HashArrayType, buff), fromConnectedPeer)
+		assert.True(t, strings.Contains(err.Error(), dataRetriever.ErrValidatorInfoNotFound.Error()))
```
bogdan-rosianu, line 490: the IDE shows an issue `Method call 'err.Error()' may lead to nil pointer dereference` . add `assert.Error(t, err)` before this line

dataRetriever/resolvers/validatorInfoResolver_test.go
```diff
@@ -361,6 +440,198 @@ func TestValidatorInfoResolver_ProcessReceivedMessage(t *testing.T) {
...
+			GetCalled: func(key []byte) (value interface{}, ok bool) {
+				val := providedData[numOfCalls]
+				numOfCalls++
+				return val, true
+			},
+		}
+		numOfCallsSend := 0
+		args.SenderResolver = &mock.TopicResolverSenderStub{
+			SendCalled: func(buff []byte, peer core.PeerID) error {
+				println(numOfCallsSend)
```
bogdan-rosianu, line 604: remove this print?

genesis/process/disabled/requestHandler.go
```diff
@@ -82,6 +82,10 @@ func (r *RequestHandler) CreateTrieNodeIdentifier(_ []byte, _ uint32) []byte {
 func (r *RequestHandler) RequestValidatorInfo(_ []byte) {
 }
 
+// RequestValidatorsInfo -
```
bogdan-rosianu, line 85: // RequestValidatorsInfo does nothing

dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -591,6 +591,51 @@ func (rrh *resolverRequestHandler) RequestValidatorInfo(hash []byte) {
 	rrh.addRequestedItems([][]byte{hash}, uniqueValidatorInfoSuffix)
 }
 
+// RequestValidatorsInfo asks for the validator info associated with the specified hashes from connected peers
```
bogdan-rosianu, line 594: validator**s'** ?

dataRetriever/resolvers/validatorInfoResolver_test.go
sstanculeanu: this is a way better looking solution. thanks, fixed

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4081, opened 2022-05-16
Fix roll back behind final edge case

SebastianMarian:
* Fixed roll back behind final edge case, when final block hash from fork detector is not the same with the header hash committed even if the nonce is the same


sstanculeanu (approved):

iulianpascalau (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #4090, opened 2022-05-19
Fix token type log for register and set all roles

bogdan-rosianu:
This PR fixes the log that resembles the token's type when calling the registerAndSetAllRoles function


iulianpascalau (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4107, opened 2022-05-24
Refactor node create transaction

ssd04:


node/node.go
```diff
@@ -676,27 +677,11 @@ func (n *Node) checkSenderIsInShard(tx *transaction.Transaction) error {
...
-	dataField []byte,
-	signatureHex string,
-	chainID string,
-	version uint32,
-	options uint32,
-	guardian string,
-	guardianSigHex string,
-) (*transaction.Transaction, []byte, error) {
-	if version == 0 {
+func (n *Node) CreateTransaction(txArgs *external.ArgsCreateTransaction) (*transaction.Transaction, []byte, error) {
```
AdoAdoAdo, line 680: can you also check for nil txArgs?
so we don't get a panic on dereferencing it below

AdoAdoAdo (changes requested):

AdoAdoAdo (approved):

node/node.go
ssd04: added nil args check

raduchis (approved):

PR #4108, opened 2022-05-24
On Transaction API, add "initially paid fee" and "is refund"

andreibancioiu:
On Transaction API, add:
 - `InitiallyPaidFee` (computed wrt. the epoch of the transaction)
 - `IsRefund`

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -33,6 +33,8 @@ github.com/ElrondNetwork/elrond-go-core v1.1.15 h1:CMGtJMsK+fBHsSrh5b3N/okYMywbf
 github.com/ElrondNetwork/elrond-go-core v1.1.15/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726 h1:f8UmoIKktGHrYvjXBquoJiejuzxJtJBQJBtEDe/yftk=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
+github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce h1:KtcRP5g71Br1rOCvIuKyzjw+6jwqBWSB0ZaA3WuUMlM=
+github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.0/go.mod h1:DGiR7/j1xv729Xg8SsjYaUzWXL5svMd44REXjWS/gAc=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.1 h1:xJUUshIZQ7h+rG7Art/9QHVyaPRV1wEjrxXYBdpmRlM=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.1/go.mod h1:uunsvweBrrhVojL8uiQSaTPsl3YIQ9iBqtYGM6xs4s0=
--- a/go.sum
+++ b/go.sum
@@ -29,12 +29,6 @@ github.com/ElrondNetwork/elrond-go-core v1.0.0/go.mod h1:FQMem7fFF4+8pQ6lVsBZq6y
 github.com/ElrondNetwork/elrond-go-core v1.1.7/go.mod h1:O9FkkTT2H9kxCzfn40TbhoCDXzGmUrRVusMomhK/Y3g=
 github.com/ElrondNetwork/elrond-go-core v1.1.13/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-core v1.1.14/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
-github.com/ElrondNetwork/elrond-go-core v1.1.15 h1:CMGtJMsK+fBHsSrh5b3N/okYMywbfH7Hhvgnf3J5hRQ=
-github.com/ElrondNetwork/elrond-go-core v1.1.15/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
-github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726 h1:f8UmoIKktGHrYvjXBquoJiejuzxJtJBQJBtEDe/yftk=
-github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
-github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce h1:KtcRP5g71Br1rOCvIuKyzjw+6jwqBWSB0ZaA3WuUMlM=
-github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220524072249-638eaf9733ee h1:wbAy2qGE0UEPufQ67Acyfbg2y1fCr+CxpHwh0nUbACg=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220524072249-638eaf9733ee/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.0/go.mod h1:DGiR7/j1xv729Xg8SsjYaUzWXL5svMd44REXjWS/gAc=
--- a/go.sum
+++ b/go.sum
@@ -35,6 +35,8 @@ github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726 h1
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce h1:KtcRP5g71Br1rOCvIuKyzjw+6jwqBWSB0ZaA3WuUMlM=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
+github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220524072249-638eaf9733ee h1:wbAy2qGE0UEPufQ67Acyfbg2y1fCr+CxpHwh0nUbACg=
+github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220524072249-638eaf9733ee/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.0/go.mod h1:DGiR7/j1xv729Xg8SsjYaUzWXL5svMd44REXjWS/gAc=
 github.com/ElrondNetwork/elrond-go-crypto v1.0.1 h1:xJUUshIZQ7h+rG7Art/9QHVyaPRV1wEjrxXYBdpmRlM=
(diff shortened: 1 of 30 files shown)


go.sum
```diff
@@ -33,6 +33,10 @@ github.com/ElrondNetwork/elrond-go-core v1.1.15 h1:CMGtJMsK+fBHsSrh5b3N/okYMywbf
 github.com/ElrondNetwork/elrond-go-core v1.1.15/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726 h1:f8UmoIKktGHrYvjXBquoJiejuzxJtJBQJBtEDe/yftk=
 github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220520142226-1f321f3d1726/go.mod h1:Yz8JK5sGBctw7+gU8j2mZHbzQ09Ek4XHJ4Uinq1N6nM=
+github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220523135140-7d93ca7217ce h1:KtcRP5g71Br1rOCvIuKyzjw+6jwqBWSB0ZaA3WuUMlM=
```
miiu96, line 36: go mod tidy

node/external/timemachine/fee/feeComputer.go
```diff
@@ -0,0 +1,110 @@
...
+	economicsConfig                *config.EconomicsConfig
+	penalizedTooMuchGasEnableEpoch uint32
+	gasPriceModifierEnableEpoch    uint32
+	economicsInstances             map[int]economicsDataWithComputeFee
+	mutex                          sync.RWMutex
+}
+
+// NewFeeComputer creates a fee computer which handles historical transactions, as well
+func NewFeeComputer(args ArgsNewFeeComputer) (*feeComputer, error) {
+	err := args.check()
```
miiu96, line 24: :+1:

node/external/timemachine/fee/feeComputer.go
```diff
@@ -0,0 +1,110 @@
...
+	computer := &feeComputer{
+		builtInFunctionsCostHandler:    args.BuiltInFunctionsCostHandler,
+		economicsConfig:                args.EconomicsConfig,
+		penalizedTooMuchGasEnableEpoch: args.PenalizedTooMuchGasEnableEpoch,
+		gasPriceModifierEnableEpoch:    args.GasPriceModifierEnableEpoch,
+		economicsInstances:             make(map[int]economicsDataWithComputeFee),
+	}
+
+	// Create some economics data instance (but do not save them) in order to validate the arguments:
+	_, err = computer.createEconomicsInstance(0)
```
miiu96, line 38: interesting approach

node/external/transactionAPI/refundDetector.go
```diff
@@ -0,0 +1,45 @@
...
+
+type refundDetector struct {
+}
+
+func newRefundDetector() *refundDetector {
+	return &refundDetector{}
+}
+
+// Also see: https://github.com/ElrondNetwork/elastic-indexer-go/blob/master/process/transactions/scrsDataToTransactions.go
+func (detector *refundDetector) isRefund(result refundDetectorInput) bool {
```
miiu96, line 25: maybe  can rename the `result` variable in something more specific like `input` or `scrInput`

node/external/timemachine/fee/feeComputer.go
```diff
@@ -0,0 +1,110 @@
...
+	if ok {
+		return instance, nil
+	}
+
+	newInstance, err := computer.createEconomicsInstance(epoch)
+	if err != nil {
+		return nil, err
+	}
+
+	computer.economicsInstances[epoch] = newInstance
```
iulianpascalau, line 84: can we get OOMs because of using maps that do not get cleared?
Why not using a cacher instead of the map?

node/external/transactionAPI/apiTransactionProcessor.go
```diff
@@ -105,12 +115,40 @@ func (atp *apiTransactionProcessor) doGetTransaction(hash []byte, withResults bo
...
 	typeOnSource, typeOnDestination := atp.txTypeHandler.ComputeTransactionType(tx.Tx)
 	tx.ProcessingTypeOnSource = typeOnSource.String()
 	tx.ProcessingTypeOnDestination = typeOnDestination.String()
 }
 
+func (atp *apiTransactionProcessor) populateComputedFieldInitiallyPaidFee(tx *transaction.ApiTransactionResult) {
+	initiallyPaidFee, err := atp.feeComputer.ComputeTransactionFee(tx.Tx, int(tx.Epoch))
+	if err != nil {
+		log.Warn("populateInitiallyPaidFee(): unexpected condition, cannot compute fee", "tx", tx.Hash)
+		return
```
iulianpascalau, line 128: so in this case the tx.InitiallyPaidFee is empty as "no fee"?
The return is no different as of the return on L133. How can we know something went wrong other than looking in the logs?

The error is propagated from this call:
feeComputer.createEconomicsInstance(epoch int) which was kind of tested in the constructor already in order to not hit "late construction errors due to programming errors"
Wondering if we can get rid of the error parameter and ease the handling results as seen in this component

node/external/transactionAPI/apiTransactionProcessor.go
```diff
@@ -105,12 +115,40 @@ func (atp *apiTransactionProcessor) doGetTransaction(hash []byte, withResults bo
...
+	}
+
+	isZero := initiallyPaidFee.Cmp(big.NewInt(0)) == 0
+	if isZero {
+		return
```
iulianpascalau, line 133: why not setting tx.InitiallyPaidFee to "0" and then return? Will be more consistent when consuming the request.

node/external/transactionAPI/refundDetector.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+type refundDetector struct {
+}
+
+func newRefundDetector() *refundDetector {
+	return &refundDetector{}
+}
+
+// Also see: https://github.com/ElrondNetwork/elastic-indexer-go/blob/master/process/transactions/scrsDataToTransactions.go
```
iulianpascalau, line 24: wondering how we can remove all these code duplicates

go.sum
andreibancioiu: Done.

node/external/transactionAPI/refundDetector.go
andreibancioiu: Renamed to `input`.

node/external/timemachine/fee/args.go
```diff
@@ -0,0 +1,26 @@
+package fee
+
+import (
+	"github.com/ElrondNetwork/elrond-go-core/core/check"
+	"github.com/ElrondNetwork/elrond-go/config"
+	"github.com/ElrondNetwork/elrond-go/process"
+	"github.com/ElrondNetwork/elrond-go/process/economics"
+)
+
+type ArgsNewFeeComputer struct {
```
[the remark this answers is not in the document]
andreibancioiu: Added.

node/external/timemachine/fee/feeComputer_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestNewFeeComputer(t *testing.T) {
+	t.Run("NilBuiltInFunctionsCostHandler", func(t *testing.T) {
+		arguments := ArgsNewFeeComputer{
+			BuiltInFunctionsCostHandler: nil,
+			EconomicsConfig:             testscommon.GetEconomicsConfig(),
+		}
+
+		_, err := NewFeeComputer(arguments)
+		require.Equal(t, process.ErrNilBuiltInFunctionsCostHandler, err)
```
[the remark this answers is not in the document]
andreibancioiu: Thanks, added.

node/external/timemachine/fee/feeComputer_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	t.Run("NilEconomicsConfig", func(t *testing.T) {
+		arguments := ArgsNewFeeComputer{
+			BuiltInFunctionsCostHandler: &testscommon.BuiltInCostHandlerStub{},
+			EconomicsConfig:             nil,
+		}
+
+		_, err := NewFeeComputer(arguments)
+		require.Equal(t, ErrNilEconomicsConfig, err)
+	})
```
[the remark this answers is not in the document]
andreibancioiu: Fixed: added `.Run("AllArgumentsProvided", ...)`, then dropped the redundant check in `TestFeeComputer_ComputeTransactionFeeShouldWorkForDifferentEpochs`.

node/external/transactionAPI/refundDetector.go
```diff
@@ -0,0 +1,45 @@
...
+func (detector *refundDetector) isRefund(result refundDetectorInput) bool {
+	hasValue := result.Value != "0" && result.Value != ""
+	hasNoGasLimit := result.GasLimit == 0
+	hasReturnCodeOK := detector.isReturnCodeOK(result.Data)
+	isRefundForRelayTxSender := result.ReturnMessage == gasRefundForRelayerMessage
```
bogdan-rosianu, line 29: maybe `strings.Contains` ? If in the future another thing is added to the message, the code will still work

node/external/transactionAPI/refundDetector.go
```diff
@@ -0,0 +1,45 @@
...
+	isSuccessful := hasReturnCodeOK || isRefundForRelayTxSender
+
+	return hasValue && hasNoGasLimit && isSuccessful
+}
+
+// Also see: https://github.com/ElrondNetwork/elastic-indexer-go/blob/master/process/transactions/checkers.go
+func (detector *refundDetector) isReturnCodeOK(resultData []byte) bool {
+	okCode := vmcommon.Ok.String()
+	okMarker := []byte("@" + hex.EncodeToString([]byte(okCode)))
```
bogdan-rosianu, line 38: could have defined a const `@6f6b` so we don't compute it each time
alternatively, you could use a var `var okMarker = []byte("@" + hex.EncodeToString([]byte(vmcommon.Ok.String())))`

node/external/transactionAPI/refundDetector.go
```diff
@@ -0,0 +1,45 @@
...
+	okMarkerBackwardsCompatible := []byte("@" + okCode)
```
bogdan-rosianu, line 39: also, you could have used a const `@ok` or a var

node/external/transactionAPI/refundDetector_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}))
+
+	require.False(t, detector.isRefund(refundDetectorInput{
+		Value:    "1000",
+		Data:     []byte("foobar"),
+		GasLimit: 0,
+	}))
+
+	require.False(t, detector.isRefund(refundDetectorInput{
+		Value:    "1000",
```
bogdan-rosianu, line 44: please add a test for `6f6b` as well

node/external/timemachine/fee/feeComputer_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type dummyTx struct {
+	gasLimit uint64
+	gasPrice uint64
+	data     []byte
+	receiver []byte
+}
+
+func (tx *dummyTx) GetGasLimit() uint64 {
```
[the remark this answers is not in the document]
andreibancioiu: Fixed by using the actual `transaction.Transaction` instead (though indeed we should probably define a `transactionHandlerMock` in `testscommon`).

node/external/transactionAPI/apiTransactionProcessor.go
andreibancioiu: Fixed to not return an error in `ComputeTransactionFee`, indeed.

Also fixed the outdated log message.

node/external/transactionAPI/apiTransactionProcessor.go
andreibancioiu: Previously, I wanted to not have the field in the response if that's zero ("0" will not be omitted by the serializer). However, on a second thought, changed to:

```
func (atp *apiTransactionProcessor) populateComputedFieldInitiallyPaidFee(tx *transaction.ApiTransactionResult) {
	// Only user-initiated transactions will present an initially paid fee.
	if tx.Type == string(transaction.TxTypeNormal) || tx.Type == string(transaction.TxTypeInvalid) {
		fee := atp.feeComputer.ComputeTransactionFee(tx.Tx, int(tx.Epoch))
		tx.InitiallyPaidFee = fee.String()
	}
}
```

@iulianpascalau, @bogdan-rosianu, @miiu96, is this all right?

node/external/transactionAPI/refundDetector.go
andreibancioiu: One option would be to create a new repository (and go package), such as `transaction-decoder` or `transaction-classifier`, independent of any other package. Then, we would reference this package in `elrond-go` and `elastic-indexer-go`.

The newly created repository / package would be quite stable in practice.

node/external/transactionAPI/refundDetector.go
andreibancioiu: Fixed.

node/external/transactionAPI/refundDetector.go
andreibancioiu: Good idea :bulb: 

Fixed, added constants.

node/external/transactionAPI/refundDetector.go
andreibancioiu: Fixed, added a constant.

node/external/transactionAPI/refundDetector_test.go
andreibancioiu: Added.

node/external/timemachine/fee/feeComputer.go
andreibancioiu: Using a cache is indeed desirable. Currently, our LRU / time-based caches are non-generic. Though we can work around their functionality and use them in this context, as well, do you think it's OK to postpone until, perhaps, upgrading the go version & support generics (added a TODO).

Howerver, also added a test for analyzing the memory footprint of the fee computer: `~12 MB for 10 000 epochs`.

bogdan-rosianu (approved):

node/external/transactionAPI/apiTransactionProcessor.go
iulianpascalau: actually, for 0 it will still return "" because it's a `big.Int` :man_facepalming:

iulianpascalau (approved):

miiu96 (approved):

PR #4119, opened 2022-05-26
En 12302 cosigned set guardian

AdoAdoAdo:


cmd/node/config/config.toml
```diff
@@ -66,6 +70,7 @@
    # it is a good idea to increase the maximum number of opened files allowed by the operating system
    FullArchiveNumActivePersisters = 10
 
+
```
iulianpascalau, line 73: remove extra line?

process/guardian/guardedAccount.go
```diff
@@ -101,6 +112,43 @@ func (agc *guardedAccount) setAccountGuardian(uah state.UserAccountHandler, guar
...
+	accountGuardians.Slice = []*guardians.Guardian{guardian}
+	accHandler, ok := uah.(vmcommon.UserAccountHandler)
+	if !ok {
+		return process.ErrWrongTypeAssertion
+	}
+
+	return agc.saveAccountGuardians(accHandler, accountGuardians)
+}
+
+// TODO: add constraints on not co-signed txs on interceptor, for setGuardian
```
iulianpascalau, line 149: task for TODO?

Also, question: is there a check as to not send value on a setGuardian transaction call? (in builtin function implementation)

AdoAdoAdo: Created a task: [EN-12365]

Good point, will add the check on the builtin function in vmcommon, currently there is no check for the value.

process/guardian/guardedAccount.go
```diff
@@ -69,17 +75,21 @@ func (agc *guardedAccount) GetActiveGuardian(uah vmcommon.UserAccountHandler) ([
...
 	if !ok {
 		return process.ErrWrongTypeAssertion
 	}
 
+	if len(txGuardianAddress) > 0 {
+		return agc.instantSetGuardian(stateUserAccount, guardianAddress, txGuardianAddress)
+	}
+
+	guardian := &guardians.Guardian{
+		Address:         guardianAddress,
```
raduchis, line 89: is there a check on the newGuardianAddress? at leat for 32 bytes

process/guardian/guardedAccount.go
```diff
@@ -15,29 +16,34 @@ import (
...
+	epochNotifier vmcommon.EpochNotifier,
+	setGuardianEpochsDelay uint32,
+) (*guardedAccount, error) {
 	if check.IfNil(marshaller) {
 		return nil, process.ErrNilMarshalizer
 	}
 	if check.IfNil(epochNotifier) {
 		return nil, process.ErrNilEpochNotifier
 	}
+	if setGuardianEpochsDelay == 0 {
```
raduchis, line 39: should there be a check with greater than unbondingPeriod? or the fact that someone set it to a value greater than zero should suffice?

cmd/node/config/config.toml
AdoAdoAdo: done on #4159

process/guardian/guardedAccount.go
AdoAdoAdo: In elrond-vm-common/builtInFunctions/setGuardian.go -> checkSetGuardianArgs
there is already a check for:
isGuardianAddrLenOk := len(vmInput.Arguments[0]) == len(senderAddr)

the only place in production code where SetGuardian is called (and the only place where it should be called) is from `func (sg *setGuardian) ProcessBuiltinFunction` which does this check.

process/guardian/guardedAccount.go
AdoAdoAdo: I have added for now a comment in config, otherwise the system smart contracts config should be passed to the bootstrap component and checked there upon creation.

iulianpascalau (approved):

process/guardian/guardedAccount.go
raduchis: ok

process/guardian/guardedAccount.go
raduchis: ok

raduchis (approved):

PR #4135, opened 2022-05-30
Fixed the init call in transaction coordinator

iulianpascalau:
- fixed init call on transaction coordinator when processing a new miniblock
- removed some weak warnings

### DIFF
--- a/process/coordinator/process.go
+++ b/process/coordinator/process.go
@@ -149,9 +149,9 @@ func NewTransactionCoordinator(args ArgTransactionCoordinator) (*transactionCoor
 		return tc.keysTxPreProcs[i] < tc.keysTxPreProcs[j]
 	})
 	for _, value := range tc.keysTxPreProcs {
-		preProc, err := args.PreProcessors.Get(value)
-		if err != nil {
-			return nil, err
+		preProc, errGet := args.PreProcessors.Get(value)
+		if errGet != nil {
+			return nil, errGet
 		}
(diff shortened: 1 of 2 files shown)


process/coordinator/process.go
```diff
@@ -588,7 +588,7 @@ func (tc *transactionCoordinator) CreateMbsAndProcessCrossShardTransactionsDstMe
 
 	headerHash, err := core.CalculateHash(tc.marshalizer, tc.hasher, hdr)
 	if err != nil {
-		return createMBDestMeExecutionInfo.miniBlocks, createMBDestMeExecutionInfo.numTxAdded, false, nil
+		return createMBDestMeExecutionInfo.miniBlocks, createMBDestMeExecutionInfo.numTxAdded, false, nil // TODO: check if we should return err here
```
SebastianMarian, line 591: If we returns err here all the processing will be dropped. If we let it like this the processing will be stop here as we return false as allProcessed so the proposer can propose at least what it has been processed until now. It is hard to believe that this core.CalculateHash method will work with the first header but not for the second one. Anyway, if somehow we will have this case at least the proposer could propose something here without reverting everything.

SebastianMarian (approved):

AdoAdoAdo (dismissed):

process/coordinator/process.go
iulianpascalau: added comment + log.Warn print

SebastianMarian (approved):

PR #4138, opened 2022-05-31
Fix isRelayedTransactionV2

sstanculeanu:
- `isRelayedTransactionV2` does not check the flag anymore, only returns if the transaction is of type `RelayedTxV2` and the flag check is done on `processRelayedTxV2`. With this check on, `ComputeTransactionType` returns type `MoveBalance` for a `RelayedTxV2` with the flag disabled.

### DIFF
--- a/process/coordinator/transactionType.go
+++ b/process/coordinator/transactionType.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 
 	"github.com/ElrondNetwork/elrond-go-core/core"
-	"github.com/ElrondNetwork/elrond-go-core/core/atomic"
 	"github.com/ElrondNetwork/elrond-go-core/core/check"
 	"github.com/ElrondNetwork/elrond-go-core/data"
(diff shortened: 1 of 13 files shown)


ccorcoveanu (approved):

dragos-rebegea (approved):

process/coordinator/transactionType.go
```diff
@@ -183,9 +171,6 @@ func (tth *txTypeHandler) isRelayedTransactionV1(functionName string) bool {
 }
 
 func (tth *txTypeHandler) isRelayedTransactionV2(functionName string) bool {
-	if !tth.flagRelayedTxV2.IsSet() {
```
iulianpascalau, line 186: we need an import-db operation to consider this change backwards compatible

iulianpascalau (commented):
not sure this PR is backwards compatible, we need an extra check

iulianpascalau (approved):
We merge this at is and test later in an import-db process

gabi-vuls (approved):
System test passed.

PR #4146, opened 2022-06-01
Track "previous-to-final" block info in ChainHandler

andreibancioiu:
In `shardBlock` and `metaBlock`, call `SetPreviousToFinalBlockInfo` of `chain handler`.


dataRetriever/blockchain/baseBlockchain.go
```diff
@@ -70,3 +71,25 @@ func (bbc *baseBlockChain) SetCurrentBlockHeaderHash(hash []byte) {
 	bbc.currentBlockHeaderHash = hash
 	bbc.mut.Unlock()
 }
+
+func (bbc *baseBlockChain) SetHighestFinalBlockCoordinates(nonce uint64, headerHash []byte, rootHash []byte) {
+	bbc.mut.Lock()
+
+	bbc.finalCoordinates.blockNonce = nonce
+	bbc.finalCoordinates.blockHash = headerHash
+	bbc.finalCoordinates.blockRootHash = rootHash
```
AdoAdoAdo, line 80: can we call this: committed root hash? instead of root hash?
on shard this will be the scheduled root hash, and on meta as we do not have yet scheduled txs support, it is the root hash

dataRetriever/blockchain/blockchain.go
```diff
@@ -51,6 +52,7 @@ func (bc *blockChain) SetGenesisHeader(genesisBlock data.HeaderHandler) error {
 }
 
 // SetCurrentBlockHeaderAndRootHash sets current block header pointer and the root hash
+// Question for review: perhaps also receive the blockHash as a parameter (since SetCurrentBlockHeaderAndRootHash and SetCurrentBlockHeaderHash are always called in pair, so that we set all of them in the same critical section)?
```
AdoAdoAdo, line 55: :+1:

dataRetriever/blockchain/metachain.go
```diff
@@ -53,6 +54,7 @@ func (mc *metaChain) SetGenesisHeader(header data.HeaderHandler) error {
 }
 
 // SetCurrentBlockHeaderAndRootHash sets current block header pointer and the root hash
+// Question for review: perhaps also receive the blockHash as a parameter (since SetCurrentBlockHeaderAndRootHash and SetCurrentBlockHeaderHash are always called in pair, so that we set all of them in the same critical section)?
```
AdoAdoAdo, line 57: :+1:

process/block/metablock.go
```diff
@@ -1487,6 +1487,7 @@ func (mp *metaProcessor) updateState(lastMetaBlock data.MetaHeaderHandler, _ []b
 	)
 
 	mp.setFinalizedHeaderHashInIndexer(lastMetaBlock.GetPrevHash())
+	mp.blockChain.SetHighestFinalBlockCoordinates(prevMetaBlock.GetNonce(), prevMetaBlockHash, prevMetaBlock.GetRootHash())
```
AdoAdoAdo, line 1490: this is not exactly the final block it is the prev to final block.

process/block/shardblock.go
```diff
@@ -1152,6 +1152,7 @@ func (sp *shardProcessor) updateState(headers []data.HeaderHandler, currentHeade
 		)
 
 		sp.setFinalizedHeaderHashInIndexer(header.GetPrevHash())
+		sp.blockChain.SetHighestFinalBlockCoordinates(prevHeader.GetNonce(), prevHeaderHash, prevHeaderRootHashForPruning)
```
AdoAdoAdo, line 1155: `prevHeaderRootHashForPruning` is the root hash from which the previous to final block started processing. The roothash that was committed after previous to final block was processed should be `scheduledPrevHeaderRootHash`

dataRetriever/blockchain/baseBlockchain.go
```diff
@@ -70,3 +71,25 @@ func (bbc *baseBlockChain) SetCurrentBlockHeaderHash(hash []byte) {
 	bbc.currentBlockHeaderHash = hash
 	bbc.mut.Unlock()
 }
+
+func (bbc *baseBlockChain) SetHighestFinalBlockCoordinates(nonce uint64, headerHash []byte, rootHash []byte) {
```
iulianpascalau, line 75: Missing comments.
I would not call these "coordinates"
Let's rename to SetHighestFinalBlockInfo(...)

dataRetriever/blockchain/baseBlockchain.go
```diff
@@ -70,3 +71,25 @@ func (bbc *baseBlockChain) SetCurrentBlockHeaderHash(hash []byte) {
...
+	bbc.mut.Lock()
+
+	bbc.finalCoordinates.blockNonce = nonce
+	bbc.finalCoordinates.blockHash = headerHash
+	bbc.finalCoordinates.blockRootHash = rootHash
+
+	bbc.mut.Unlock()
+}
+
+func (bbc *baseBlockChain) GetHighestFinalBlockCoordinates() (nonce uint64, hash []byte, rootHash []byte) {
```
iulianpascalau, line 85: missing comments.
Let's find another way to return those []byte slices as that is dangerous (we can swap them by mistake where we call this function)
Also, we stopped using named output parameters. The code will be prone to future errors.

dataRetriever/blockchain/blockchain.go
iulianpascalau: can be done

dataRetriever/blockchain/finalCoordinates.go
```diff
@@ -0,0 +1,7 @@
+package blockchain
+
+type finalCoordinates struct {
```
iulianpascalau, line 3: this is a simple dto, can be placed inside blockchain.go

process/block/metablock.go
iulianpascalau: should use `lastMetaBlock`

iulianpascalau (changes requested):

process/block/metablock.go
AdoAdoAdo: so should set then lastMetaBlock.GetNonce(), lastMetaBlockHash can be used from param list (un-ignore) and lastMetaBlock.GetRootHash()

process/block/shardblock.go
AdoAdoAdo: if we want to set the last finalized Header data then header.GetNonce, headerHash, and scheduledHeaderRootHash

dataRetriever/blockchain/baseBlockchain.go
andreibancioiu: Renamed as follows:

```
type blockInfo struct {
	nonce             uint64
	hash              []byte
	committedRootHash []byte
}
```

dataRetriever/blockchain/blockchain.go
andreibancioiu: Even though they were always called in pair in production code, they weren't in some tests. In order to avoid breaking the tests, I did not apply this change at the moment. Is this all right?

dataRetriever/blockchain/metachain.go
andreibancioiu: Even though they were always called in pair in production code, they weren't in some tests. In order to avoid breaking the tests, I did not apply this change at the moment. Is this all right?

process/block/metablock.go
andreibancioiu: Fixed.

process/block/shardblock.go
andreibancioiu: Fixed.

dataRetriever/blockchain/baseBlockchain.go
andreibancioiu: Added comments. As discussed, left the signature of the function as it was.

dataRetriever/blockchain/finalCoordinates.go
andreibancioiu: Moved and renamed to `blockInfo`.

process/block/shardblock.go
```diff
@@ -1152,7 +1152,7 @@ func (sp *shardProcessor) updateState(headers []data.HeaderHandler, currentHeade
 		)
 
 		sp.setFinalizedHeaderHashInIndexer(header.GetPrevHash())
-		sp.blockChain.SetHighestFinalBlockCoordinates(prevHeader.GetNonce(), prevHeaderHash, prevHeaderRootHashForPruning)
+		sp.blockChain.SetPreviousToFinalBlockInfo(header.GetNonce(), headerHash, scheduledHeaderRootHash)
```
iulianpascalau, line 1155: we should check this

iulianpascalau (approved):

AdoAdoAdo (approved):

PR #4164, opened 2022-06-06
EnableRoundsHandler implementation + integration

iulianpascalau:
- refactored the enableRoundsHandler


common/enableRounds/enableRoundsHandler.go
```diff
@@ -0,0 +1,52 @@
...
+
+	return &enableRoundsHandler{
+		flagsHolder: &flagsHolder{
+			checkValueOnExecByCaller: checkValueOnExecByCaller,
+		},
+	}, nil
+}
+
+func getRoundConfig(args config.RoundConfig, configName string) (*roundFlag, error) {
+	for _, activationRound := range args.RoundActivations {
```
sstanculeanu, line 31: perhaps `RoundActivations` from `RoundConfig` can be a `map[string]ActivationRound` with key `activationRound.Name`.. this way we avoid this for loop

mariusmihaic: :+1:

common/enableRounds/flags.go
```diff
@@ -0,0 +1,10 @@
+package enableRounds
+
+type flagsHolder struct {
+	checkValueOnExecByCaller *roundFlag
+}
+
+// IsCheckValueOnExecByCallerEnabled returns true if checkValueOnExecByCaller is enabled
+func (fh *flagsHolder) IsCheckValueOnExecByCallerEnabled() bool {
```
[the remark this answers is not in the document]
iulianpascalau: it can be redefined in any other package, say, arwen v1.4 since it contains only a bool return type. In this way we have maximum decoupling.

common/enableRounds/enableRoundsHandler.go
```diff
@@ -0,0 +1,52 @@
...
+				options: activationRound.Options,
+			}, nil
+		}
+	}
+
+	return nil, fmt.Errorf("%w for config %s", errMissingRoundActivation, configName)
+}
+
+// CheckRound should be called whenever a new round is known. It will trigger the updating of all containing round flags
+func (handler *enableRoundsHandler) CheckRound(round uint64) {
```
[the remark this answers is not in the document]
iulianpascalau: I would prefer CheckRound since it is clear enough and resembles the EpochNotifier.CheckEpoch
This method will be called once when the node starts to process a received block or whenever it wants to propose a new one. The method will update **all current and future flags** (it should be slight optimal this approach as opposed the old handler and for loop implementation -> we are phasing that implementation out, also)

common/enableRounds/enableRoundsHandler.go
iulianpascalau: good suggestion, implemented

PR #4193, opened 2022-06-16
Merge dev into feat refactor enable epochs 2022.06.16

sstanculeanu:


sharding/mock/enableEpochsHandlerMock.go
```diff
@@ -389,6 +389,16 @@ func (mock *EnableEpochsHandlerMock) IsMiniBlockPartialExecutionFlagEnabled() bo
 	return false
 }
 
+// IsManagedCryptoAPIsFlagEnabled returns false
```
iulianpascalau, line 392: Question: why do we have this mock in the sharding package? Why not moving this in the testscommon package instead (can be done in a new PR) as it might be used elsewhere.

common/enableEpochs/flags.go
```diff
@@ -504,3 +508,13 @@ func (fh *flagsHolder) IsHeartbeatDisableFlagEnabled() bool {
 func (fh *flagsHolder) IsMiniBlockPartialExecutionFlagEnabled() bool {
 	return fh.isMiniBlockPartialExecutionFlag.IsSet()
 }
+
+// IsManagedCryptoAPIsFlagEnabled returns true if managedCryptoAPIsFlag is enabled
+func (fh *flagsHolder) IsManagedCryptoAPIsFlagEnabled() bool {
```
iulianpascalau, line 513: shouldn't we use these flags in the VM factory?

sharding/mock/enableEpochsHandlerMock.go
sstanculeanu: we already have the stub in testscommon, but using it into nodesCoordinator package was leading to cyclic imports.

I decided that creating a new mock(altought this means duplicate code), with the needed functionality for this package was a smaller change than moving all the tests from package nodesCoordinator to a new nodesCoordinator_test package, as they directly use a lot of unexported methods

common/enableEpochs/flags.go
sstanculeanu: that's right, but enableEpochsHandler will be integrated into arwen as well

iulianpascalau (approved):

mariusmihaic (approved):

PR #4195, opened 2022-06-16
Vm query different config for meta

bogdan-rosianu:
separate the configurations for shards and meta for the maximum gas to be used for VM Queries. 

**How to test**
There are 2 ways of validating this PR:
1. when running a shard/meta node, a vm query that consumes more than what is set in config will return error
2. search in the logs for `maximum gas per VM Query`. for shard it should be 1.5b and for meta it should be 0

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -699,9 +699,10 @@
         ]
 
     [VirtualMachine.GasConfig]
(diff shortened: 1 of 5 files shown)


iulianpascalau (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed. 
##                            vm-query-different-config-for-meta           ##

PR #4205, opened 2022-06-21
Update self to self intermediate mini block after scheduled execution

SebastianMarian:
* Updated self to self intermediate mini block after scheduled execution. Needed to be saved into storage to resolve some requests only by some APIs, not needed by the protocol which always recreates it by the time of block execution. (these mini blocks are also deleted from body as they are self to self receipts or scrs)


andreibancioiu (approved):
Tested on epochs 684 - 685, checked intra-shard SCRs of transactions in scheduled miniblocks.

Import for epoch 683 failed (`num miniblocks does not match`).

andreibancioiu (approved):

popenta (approved):
##                            update-intermediate-self-to-self-miniblock   ##

PR #4206, opened 2022-06-21
Health records changes

iulianpascalau:
- added a new flag to set a custom value when the health records start triggering pprof files
- removed the custom changes whenever an import-db process starts

Testing scenario: start a testnet, take a node, stop it, and start it with these 2 custom flags: `-use-health-service -memory-usage-to-create-profiles 1073741824` which translates to: create health records, but alter the value to 1GB.
grep the log to the following string: `setting a new value for the memoryUsageToCreateProfiles option` and you should be able to check that the new value was set correctly (the default one from the config is ~2.25GB)
Also, check that health records are to be created in the `health-records` directory and contains files with the filenames in the following format: `mem__20220621124307__1_xx_GB.pprof`

### DIFF
--- a/cmd/node/flags.go
+++ b/cmd/node/flags.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"runtime"
 
-	"github.com/ElrondNetwork/elrond-go-core/core"
 	logger "github.com/ElrondNetwork/elrond-go-logger"
 	"github.com/ElrondNetwork/elrond-go/common"
 	"github.com/ElrondNetwork/elrond-go/config"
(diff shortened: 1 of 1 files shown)


cmd/node/flags.go
```diff
@@ -499,14 +510,6 @@ func processConfigImportDBMode(log logger.Logger, configs *config.Configs) error
 	p2pConfigs.Node.ThresholdMinConnectedPeers = 0
 	p2pConfigs.KadDhtPeerDiscovery.Enabled = false
 
-	defaultSecondsToCheckHealth := 300                   // 5minutes
-	defaultDiagnoseMemoryLimit := 6 * 1024 * 1024 * 1024 // 6GB
-
-	generalConfigs.Health.IntervalDiagnoseComponentsDeeplyInSeconds = defaultSecondsToCheckHealth
-	generalConfigs.Health.IntervalDiagnoseComponentsInSeconds = defaultSecondsToCheckHealth
-	generalConfigs.Health.IntervalVerifyMemoryInSeconds = defaultSecondsToCheckHealth
```
ssd04, line 507: how about these? it looks like there are not being set anymore

andreibancioiu (approved):

cmd/node/flags.go
andreibancioiu: Normally, they default to what it's written in `config.toml` (no CLI flags for them).

ssd04: ah, yes :+1:

ssd04 (approved):

gabi-vuls (approved):
##                            health-records-changes                       ##
- Known Warnings              8

PR #4215, opened 2022-06-22
Merge dev into liquid staking 22 06 2022

mariusmihaic:
- Merge dev into liquid staking 22 06 2022


miiu96 (dismissed):

ssd04 (approved):

process/peer/validatorsProvider_test.go
```diff
@@ -732,96 +732,97 @@ func TestValidatorsProvider_GetAuctionList(t *testing.T) {
 		response := vp.GetAuctionList()
 		require.Empty(t, response)
 	})
-
```
miiu96, line 735: Add a todo here to FIX the test or remove it

mariusmihaic: There is no fix for this actually.
It is the first API version; which will be completely removed.
I think it used a map while processing causing non-deterministic results

PR #4229, opened 2022-06-27
Added integration test for relayed tx v2

bogdan-rosianu:
added an integration test for relayed tx v2, which illustrates how such a transaction should be made + ensure the functionality

### DIFF
--- a/integrationTests/multiShard/relayedTx/common.go
+++ b/integrationTests/multiShard/relayedTx/common.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/integrationTests"
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/state"
+	"github.com/ElrondNetwork/elrond-go/testscommon/txDataBuilder"
 )
 
 // CreateGeneralSetupForRelayTxTest will create the general setup for relayed transactions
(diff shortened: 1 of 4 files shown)


integrationTests/multiShard/relayedTx/common.go
```diff
@@ -73,6 +74,29 @@ func CreateAndSendRelayedAndUserTx(
...
+	gasLimit uint64,
+	txData []byte,
+) *transaction.Transaction {
+	txDispatcherNode := getNodeWithinSameShardAsPlayer(nodes, relayer.Address)
+
+	userTx := createUserTx(player, rcvAddr, value, 0, txData)
+	relayedTx := createRelayedTxV2(txDispatcherNode.EconomicsData, relayer, userTx, gasLimit)
+
+	_, err := txDispatcherNode.SendTransaction(relayedTx)
+	if err != nil {
```
BeniaminDrasovean, line 93: Isn't an assert for nil useful here?

BeniaminDrasovean (approved):

sstanculeanu (approved):

integrationTests/multiShard/relayedTx/common.go
bogdan-rosianu: this is the same behavior as for the relayed tx integration test. I assume it is enough to only print the error if any. the test would fail anyway if this doesn't work

sstanculeanu (approved):

raduchis (approved):

PR #4233, opened 2022-06-28
In API, fix error propagation, handle transactions without SCRs

andreibancioiu:
- Regression introduced in feat/rosetta, fixed now
  - Updated some tests, added some TODOs related to errors & existing functions with side-effects

### DIFF
--- a/dblookupext/errors.go
+++ b/dblookupext/errors.go
@@ -4,7 +4,17 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 )
+
+// ErrNotFoundInStorage signals that an item was not found in storage
+var ErrNotFoundInStorage = errors.New("not found in storage")
+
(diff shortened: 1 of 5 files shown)


miiu96 (approved):

dblookupext/errors.go
```diff
@@ -4,8 +4,18 @@ import (
...
 	"fmt"
+	"strings"
 )
 
+// ErrNotFoundInStorage signals that an item was not found in storage
+var ErrNotFoundInStorage = errors.New("not found in storage")
+
+// Currently, "item not found" storage errors are untyped (thus not distinguishable from others). E.g. see "pruningStorer.go".
+// As a workaround, we test the error message for a match.
+func isNotFoundInStorageErr(err error) bool {
```
bogdan-rosianu, line 15: might add an `if err == nil { return false }` to avoid any lead to null pointers

andreibancioiu: Fixed.

bogdan-rosianu (approved):

PR #4237, opened 2022-06-28
On API, populate field "miniblock.isFromReceiptsStorage".

andreibancioiu:
On API, inform clients when the miniblock was loaded from the receipts storage (as opposed to being loaded from the block body).

### DIFF
--- a/node/external/blockAPI/shardBlock.go
+++ b/node/external/blockAPI/shardBlock.go
@@ -118,12 +118,14 @@ func (sbp *shardAPIBlockProcessor) convertShardBlockBytesToAPIBlock(hash []byte,
 		miniblocks = append(miniblocks, miniblockAPI)
 	}
 
-	intraMb, err := sbp.getIntraMiniblocks(blockHeader.GetReceiptsHash(), headerEpoch, options)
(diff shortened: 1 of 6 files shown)


node/external/blockAPI/shardBlock.go
```diff
@@ -118,12 +118,14 @@ func (sbp *shardAPIBlockProcessor) convertShardBlockBytesToAPIBlock(hash []byte,
...
 	}
 
-	intraMb, err := sbp.getIntraMiniblocks(blockHeader.GetReceiptsHash(), headerEpoch, options)
+	intraMb, err := sbp.getIntrashardMiniblocksFromReceiptsStorage(blockHeader.GetReceiptsHash(), headerEpoch, options)
 	if err != nil {
 		return nil, err
 	}
 
 	miniblocks = append(miniblocks, intraMb...)
+	// QUESTION FOR REVIEW: perhaps do not do any deduplication here, and let clients do it?
```
bogdan-rosianu, line 127: I think it is better to filter them here

bogdan-rosianu (approved):

PR #4251, opened 2022-07-01
Arrange flags 2022.07.01

iulianpascalau:


BeniaminDrasovean (approved):

BeniaminDrasovean (approved):

bogdan-rosianu (approved):
confirmed that toml's order doesn't count and the config is still parsed correctly :+1:

PR #4254, opened 2022-07-01
Fixed the node stuck at epoch start on rc/2022-july

iulianpascalau:
- fixed the updateIndexesOfProcessedTxs function that will trigger the correct marshalling of the reserved structure


sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4265, opened 2022-07-05
Trie fixes

BeniaminDrasovean:
When syncing, save the trie nodes to both current and previous epoch storer.


state/syncer/baseAccountsSyncer.go
```diff
@@ -214,6 +214,30 @@ func (b *baseAccountsSyncer) GetSyncedTries() map[string]common.Trie {
...
+	if err != nil {
+		log.Error("getLatestStorageEpoch error", "error", err)
+	}
+
+	err = trieStorageManager.Put([]byte(common.TrieSyncedKey), []byte(common.TrieSyncedVal))
+	if err != nil {
+		log.Error("error while putting trieSynced value into main storer after sync", "error", err)
+	}
+
+	lastEpoch := epoch - 1
```
bogdan-rosianu, line 230: if `trieStorageManager.GetLatestStorageEpoch()` returns error, the `epoch` will be set to `0`. therefore, `lastEpoch` will be 0 - 1 so max uint 32, and the check will pass

state/snapshotStatistics.go
```diff
@@ -55,3 +57,20 @@ func (ss *snapshotStatistics) NewDataTrie() {
...
+
+	ss.mutex.RLock()
+	defer ss.mutex.RUnlock()
+
+	log.Debug("snapshot statistics",
+		"type", identifier,
+		"duration", time.Since(ss.startTime).Truncate(time.Second),
+		"num of nodes copied", ss.numNodes,
+		"total size copied", core.ConvertBytes(ss.trieSize),
+		"num data tries copied", ss.numDataTries,
```
bogdan-rosianu, line 73: can we also add a function that will better display high numbers? for example, 12.53M will be displayed as 12530000, and it would be better if we display it as 12,530,000

trie/syncTrieStorageManager.go
```diff
@@ -0,0 +1,35 @@
...
+import "github.com/ElrondNetwork/elrond-go/common"
+
+type syncTrieStorageManager struct {
+	common.StorageManager
+	epoch uint32
+}
+
+// NewSyncTrieStorageManager creates a new instance of syncTrieStorageManager
+func NewSyncTrieStorageManager(tsm common.StorageManager) (*syncTrieStorageManager, error) {
+	epoch, err := tsm.GetLatestStorageEpoch()
```
bogdan-rosianu, line 12: check `tsm` for nil value?

trie/syncTrieStorageManager_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func TestNewSyncTrieStorageManager(t *testing.T) {
+	_, trieStorage := newEmptyTrie()
+	trieStorage.mainStorer = &trie.SnapshotPruningStorerStub{}
+	stsm, err := NewSyncTrieStorageManager(trieStorage)
+	assert.Nil(t, err)
+	assert.NotNil(t, stsm)
+}
+
+func TestNewSyncTrieStorageManager_PutInFirstEpoch(t *testing.T) {
```
iulianpascalau, line 27: please change this test. This does not properly test that we do only 1 put. (change putInEpochCalled from bool to int)

storage/pruning/triePruningStorer.go
```diff
@@ -12,7 +12,7 @@ import (
 const (
 	lastEpochIndex             = 1
 	currentEpochIndex          = 0
-	minNumOfActiveDBsNecessary = 2
+	minNumOfActiveDBsNecessary = 1
```
iulianpascalau, line 15: why this change?

state/syncer/validatorAccountsSyncer.go
```diff
@@ -77,10 +77,7 @@ func (v *validatorAccountsSyncer) SyncAccounts(rootHash []byte) error {
 		return err
 	}
 
-	err = mainTrie.GetStorageManager().Put([]byte(common.TrieSyncedKey), []byte(common.TrieSyncedVal))
-	if err != nil {
-		log.Warn("error while putting trieSynced value into main storer after sync", "error", err)
-	}
+	markStorerAsSyncedAndActive(mainTrie)
```
iulianpascalau, line 80: can we move this function in the trie? So we can have: `mainTrie.MarkStorerAsSyncedAndActive()`

AdoAdoAdo (commented):
No new findings, only what was already mentioned in comments.

state/syncer/baseAccountsSyncer.go
BeniaminDrasovean: If `epoch == 0` , `lastEpoch` will be overwritten with 0, so it will work as expected.

state/snapshotStatistics.go
BeniaminDrasovean: I think it will be better if we add that func to the logger.

trie/syncTrieStorageManager_test.go
BeniaminDrasovean: Done.

storage/pruning/triePruningStorer.go
BeniaminDrasovean: Only 1 activeDB is needed for a complete snapshot. If it is set to 2, more DBs will remain open, even if it is not necessary.

gabi-vuls (approved):
System test passed.

trie/patriciaMerkleTrie.go
```diff
@@ -621,6 +621,29 @@ func (tr *patriciaMerkleTrie) Close() error {
...
+	if err != nil {
+		log.Error("error while putting trieSynced value into main storer after sync", "error", err)
+	}
+
+	lastEpoch := epoch - 1
+	if epoch == 0 {
+		lastEpoch = 0
+	}
+
+	err = tr.trieStorage.PutInEpoch([]byte(common.ActiveDBKey), []byte(common.ActiveDBVal), lastEpoch)
```
iulianpascalau, line 641: when we finish the snapshot we call PutInEpochWithoutCache for writing on the ActiveDBKey. Is this correct?

BeniaminDrasovean: When we get the value for ActiveDBKey we ignore the cache, so it does not really matter which Put is used. Will change for consistency.

PR #4266, opened 2022-07-05
Multisigner: keys holder redundancy support

iulianpascalau:
- added redundancy support for keysHolder


keysManagement/errors.go
```diff
@@ -6,3 +6,4 @@ var errDuplicatedKey = errors.New("duplicated key found")
 var errMissingPublicKeyDefinition = errors.New("missing public key definition")
 var errNilKeyGenerator = errors.New("nil key generator")
 var errNilP2PIdentityGenerator = errors.New("nil p2p identity generator")
+var errInvalidValue = errors.New("invalid value")
```
bogdan-rosianu, line 9: could have grouped them in a `var (  )` group

keysManagement/keysHolder.go
```diff
@@ -153,6 +161,89 @@ func (holder *virtualPeersHolder) GetMachineID(pkBytes []byte) (string, error) {
...
+			errMissingPublicKeyDefinition, hex.EncodeToString(pkBytes))
+	}
+
+	pInfo.resetRoundsWithoutReceivedMessages()
+
+	return nil
+}
+
+// GetAllManagedKeys returns all keys that will be managed by this node
+func (holder *virtualPeersHolder) GetAllManagedKeys() map[string]crypto.PrivateKey {
```
bogdan-rosianu, line 199: the function name includes an `All` but you filter the output. Might rename?

keysManagement/keysHolder.go
```diff
@@ -153,6 +161,89 @@ func (holder *virtualPeersHolder) GetMachineID(pkBytes []byte) (string, error) {
...
+	holder.mut.RLock()
+	defer holder.mut.RUnlock()
+
+	allManagedKeys := make(map[string]crypto.PrivateKey)
+	for pk, pInfo := range holder.data {
```
bogdan-rosianu, line 204: `holder.isMainMachine` can be extracted in a var outside the for loop - slight improvement

keysManagement/errors.go
iulianpascalau: done

keysManagement/keysHolder.go
iulianpascalau: not quite an improvement as it will be also a var. `holder.isMainMachine` is already a bool var

raduchis (approved):

PR #4281, opened 2022-07-11
[FIX]Added exception when marking activeDB after ending the snapshot

iulianpascalau:
- added exception when marking activeDB after ending the snapshot


BeniaminDrasovean (commented):
Also add the IsClosed check to peerAccountsDB.go on SnapshotState()

trie/trieStorageManager.go
```diff
@@ -518,7 +518,8 @@ func (tsm *trieStorageManager) Remove(hash []byte) error {
 	return storer.RemoveFromCurrentEpoch(hash)
 }
 
-func (tsm *trieStorageManager) isClosed() bool {
+// IsClosed returns true if the trie storage manager signals that it has been closed
```
mariusmihaic, line 521: [comment]
Could have simplified to :

```suggestion
// IsClosed returns true if the trie storage has been closed
```

BeniaminDrasovean (dismissed):

state/accountsDB.go
```diff
@@ -1113,17 +1113,25 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
 
 	adb.waitForCompletionIfRunningInImportDB(stats)
 }
 
+func (adb *AccountsDB) markActiveDBAfterSnapshot(stats *snapshotStatistics, rootHash []byte, message string, epoch uint32) {
+	printStats(stats, message, rootHash)
+
+	trieStorageManager := adb.mainTrie.GetStorageManager()
+	if trieStorageManager.IsClosed() {
+		log.Debug("will not set activeDB in epoch as the snapshot might be incomplete", "epoch", epoch)
```
mariusmihaic, line 1126: [comment]
Could've added extra explanation, like: 

```suggestion
		log.Debug("will not set activeDB in epoch as the snapshot might be incomplete, since trie storage manager is closed", "epoch", epoch)
```

state/export_test.go
```diff
@@ -6,6 +6,9 @@ import (
 	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 )
 
+// LastSnapshotStarted -
+const LastSnapshotStarted = lastSnapshotStarted
```
mariusmihaic, line 10: [question]

Why create an extra exported variable(`LastSnapshotStarted`)  instead of exporting the existing one(`lastSnapshotStarted`) ?

state/accountsDB_test.go
```diff
@@ -922,6 +922,60 @@ func TestAccountsDB_SnapshotState(t *testing.T) {
...
+				},
+			}
+		},
+	}
+	adb := generateAccountDBFromTrie(trieStub)
+	adb.SnapshotState([]byte("roothash"))
+	time.Sleep(time.Second)
+
+	mut.RLock()
+	defer mut.RUnlock()
```
mariusmihaic, line 974: [question]
Are these necessary since we have a `time.Sleep(time.Second)` above?

state/accountsDB_test.go
```diff
@@ -922,6 +922,60 @@ func TestAccountsDB_SnapshotState(t *testing.T) {
...
+				IsClosedCalled: func() bool {
+					return true
+				},
+				PutCalled: func(key []byte, val []byte) error {
+					mut.Lock()
+					defer mut.Unlock()
+
+					if string(key) == state.LastSnapshotStarted {
+						lastSnapshotStartedWasPut = true
+						return nil
```
mariusmihaic, line 950: Could delete this return, as well as the one from `PutInEpochCalled` inside `if`, since it will return nil anyway.
Also applies for `state/peerAccountsDB_test.go`

mariusmihaic (commented):
Small comments

state/export_test.go
iulianpascalau: I wanted exported only in tests. Will be exported on the package only if required elsewhere

state/accountsDB_test.go
iulianpascalau: The test involves go routines. Wherever this occurs, I use mutex or other concurrent safe operations in order to stay away of "seldom failing tests" or race conditions during execution

state/accountsDB_test.go
iulianpascalau: right, simplified :+1:

state/accountsDB.go
iulianpascalau: added extra explanation

trie/trieStorageManager.go
iulianpascalau: done

state/export_test.go
mariusmihaic: Right, my bad, I thought `lastSnapshotStarted` is from a test package as well

state/accountsDB_test.go
mariusmihaic: Right, good point

mariusmihaic (dismissed):

BeniaminDrasovean (approved):

trie/trieStorageManager.go
```diff
@@ -371,8 +380,20 @@ func (tsm *trieStorageManager) TakeSnapshot(
...
-	if tsm.isClosed() {
+func (tsm *trieStorageManager) SetCheckpoint(
+	rootHash []byte,
+	mainTrieRootHash []byte,
+	leavesChan chan core.KeyValueHolder,
+	errChan chan error,
+	stats common.SnapshotStatisticsHandler,
+) {
+	if errChan == nil {
+		log.Error("programming error in trieStorageManager.SetCheckpoint, cannot set checkpoint because errChan is nil")
```
iulianpascalau, line 391: write the error in the errChan? Valid for all error branches

trie/trieStorageManager.go
```diff
@@ -333,10 +334,17 @@ func (tsm *trieStorageManager) TakeSnapshot(
 	rootHash []byte,
 	mainTrieRootHash []byte,
 	leavesChan chan core.KeyValueHolder,
+	errChan chan error,
 	stats common.SnapshotStatisticsHandler,
 	epoch uint32,
 ) {
-	if tsm.isClosed() {
+	if errChan == nil {
+		log.Error("programming error in trieStorageManager.TakeSnapshot, cannot take snapshot because errChan is nil")
```
iulianpascalau, line 342: write the error in the errChan? Valid for all error branches

trie/trieStorageManager.go
BeniaminDrasovean: Do not write the err in errChan, because errChan is nil :sweat_smile:

state/accountsDB.go
```diff
@@ -1102,56 +1102,58 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
 
 	trieStorageManager.EnterPruningBufferingMode()
 
+	errChan := make(chan error, 1)
```
BeniaminDrasovean, line 1105: Shouldn't this channel be closed somewhere?

trie/trieStorageManager.go
iulianpascalau: yes, in this case you are right. But the other cases?

state/accountsDB.go
iulianpascalau: not required

BeniaminDrasovean (dismissed):

sasurobert (approved):

trie/trieStorageManager_test.go
```diff
@@ -85,13 +89,63 @@ func TestTrieCheckpoint(t *testing.T) {
...
 	assert.NotNil(t, val)
+	assert.Equal(t, 0, len(errChan))
+}
+
+func TestTrieStorageManager_SetCheckpointEmptyErrorChan(t *testing.T) {
+	t.Parallel()
+
+	args := getNewTrieStorageManagerArgs()
+	ts, _ := trie.NewTrieStorageManager(args)
+	_ = ts.Close()
```
mariusmihaic, line 108: Can we remove this line from this test?
Otherwise, if someone changes the code to not check for nil channel, this test would still pass (wrong behavior), because `TrieStorageManager` is closed

trie/trieStorageManager_test.go
```diff
@@ -246,6 +302,21 @@ func TestTrieStorageManager_GetLatestStorageEpoch(t *testing.T) {
 	assert.True(t, getLatestSorageCalled)
 }
 
+func TestTrieStorageManager_TakeSnapshotEmptyErrorChan(t *testing.T) {
+	t.Parallel()
+
+	args := getNewTrieStorageManagerArgs()
+	ts, _ := trie.NewTrieStorageManager(args)
+	_ = ts.Close()
```
mariusmihaic, line 310: Same remark as above

trie/trieStorageManager_test.go
```diff
@@ -85,13 +89,63 @@ func TestTrieCheckpoint(t *testing.T) {
...
+	trieStorage.SetCheckpoint(rootHash, []byte{}, nil, errChan, &trieMock.MockStatistics{})
 	trie.WaitForOperationToComplete(trieStorage)
 
 	val, err = trieStorage.GetFromCheckpoint(rootHash)
 	assert.Nil(t, err)
 	assert.NotNil(t, val)
+	assert.Equal(t, 0, len(errChan))
+}
+
+func TestTrieStorageManager_SetCheckpointEmptyErrorChan(t *testing.T) {
```
mariusmihaic, line 103:
```suggestion
func TestTrieStorageManager_SetCheckpointNilErrorChan(t *testing.T) {
```

trie/trieStorageManager_test.go
```diff
@@ -269,10 +342,12 @@ func TestTrieStorageManager_TakeSnapshotEmptyTrieRootHash(t *testing.T) {
...
-	ts.TakeSnapshot(rootHash, rootHash, leavesChan, &trieMock.MockStatistics{}, 0)
+	errChan := make(chan error, 1)
+	ts.TakeSnapshot(rootHash, rootHash, leavesChan, errChan, &trieMock.MockStatistics{}, 0)
 
 	_, ok := <-leavesChan
 	assert.False(t, ok)
+	assert.Equal(t, 0, len(errChan))
 }
 
 func TestTrieStorageManager_TakeSnapshot(t *testing.T) {
```
mariusmihaic, line 353: [comment]

Perhaps rename the test to highlight that an error is expected?

trie/trieStorageManager_test.go
```diff
@@ -315,3 +395,101 @@ func TestNewSnapshotTrieStorageManager_GetFromCurrentEpoch(t *testing.T) {
 	assert.Nil(t, err)
 	assert.True(t, getFromCurrentEpochCalled)
 }
+
+func TestWriteInChanNonBlocking(t *testing.T) {
```
mariusmihaic, line 399: :+1:

state/accountsDB.go
```diff
@@ -1102,56 +1102,58 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
-		"total size", core.ConvertBytes(stats.trieSize),
-		"num data tries", stats.numDataTries,
-		"rootHash", rootHash,
-	)
+func emptyErrChanReturningHadContained(errChan chan error) bool {
+	contained := false
+	for {
+		select {
+		case <-errChan:
+			contained = true
```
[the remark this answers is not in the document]
iulianpascalau: if we log them here we will end up logging twice. I have not changed the existing behavior since the logging is done in other parts and might contain more context information. I would leave this as it is now.
emptyErrChanReturningHadContained specify what it does: emptying the channel and returning true if the chan had contained an element. Do you have a better naming for this?

trie/trieStorageManager_test.go
iulianpascalau: right, moved the `_ = ts.Close()` as the last instruction in the test to free the resources involved

trie/trieStorageManager_test.go
iulianpascalau: done, moved

trie/trieStorageManager_test.go
iulianpascalau: changed

state/accountsDB.go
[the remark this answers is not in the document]
mariusmihaic: I really can't come up with smth "decent" (errChanHadErrorWhileEmptying? -> sounds wrong). Could leave it as it is

mariusmihaic (approved):

sasurobert (approved):

BeniaminDrasovean (approved):

popenta (approved):
##                            activeDB-markup-exception                    ##
- Known Warnings              22
- New Warnings                10
- New Errors                  3

PR #4293, opened 2022-07-14
[FIX]Active db markup exception on master

iulianpascalau:
## Description of the pull request (what is new / what has changed / what is fixed)
Fixed the writing of the `activeDB` key in the trie DB. Before this PR the marker was written also when the snapshotting procedure ended prematurely (user closed the node) or an error occurred in the snapshotting process.
  
## Proposed Changes
Signal any type of errors in a `chan error` variable and treat the error to decide if the marker should be written or not 

## Testing procedure
Intermittently stop a nod while in the middle of the snapshotting process. The following log line should appear:
```
will not set activeDB in epoch as the snapshot might be incomplete
```
After the node restarts, it will start again to create its snapshot since the `activeDB` was not written.


BeniaminDrasovean (approved):

sasurobert (approved):

mariusmihaic (approved):

popenta (approved):
##                            activeDB-markup-exception-master             ##
- Known Warnings              13

PR #4294, opened 2022-07-14
Test nfts async wrong args

sasurobert:
vmcommon integration and test


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/ElrondNetwork/elrond-go-core v1.1.16-0.20220708085217-ccc0c5ac9076
 	github.com/ElrondNetwork/elrond-go-crypto v1.0.1
 	github.com/ElrondNetwork/elrond-go-logger v1.0.7
-	github.com/ElrondNetwork/elrond-vm-common v1.3.13
+	github.com/ElrondNetwork/elrond-vm-common v1.3.14-0.20220715122033-be46f4a725c5
```
iulianpascalau, line 15: let's not forget about proper releases

integrationTests/vm/testInitializer.go
```diff
@@ -563,7 +564,7 @@ func CreateVMAndBlockchainHookAndDataPool(
 		AutomaticCrawlerAddress:   bytes.Repeat([]byte{1}, 32),
 		MaxNumNodesInTransferRole: 100,
 	}
-	builtInFuncs, nftStorageHandler, globalSettingsHandler, _ := builtInFunctions.CreateBuiltInFuncContainerAndNFTStorageHandler(argsBuiltIn)
+	builtInFuncFactory, _ := builtInFunctions.CreateBuiltInFunctionsFactory(argsBuiltIn)
```
iulianpascalau, line 567: naming: this is rather a handler, not a factory :D

integrationTests/vm/txsFee/multiShard/asyncCall_test.go
```diff
@@ -122,8 +122,8 @@ func TestAsyncCallShouldWork(t *testing.T) {
 	scr = intermediateTxs[0]
 	utils.ProcessSCRResult(t, testContextSecondContract, scr, vmcommon.Ok, nil)
 
-	require.Equal(t, big.NewInt(49993150), testContextSecondContract.TxFeeHandler.GetAccumulatedFees())
-	require.Equal(t, big.NewInt(4999315), testContextSecondContract.TxFeeHandler.GetDeveloperFees())
+	require.Equal(t, big.NewInt(3680), testContextSecondContract.TxFeeHandler.GetAccumulatedFees())
```
iulianpascalau, line 125: why these changes?
If these changes are required because the activation flag is set for the transferAndAsyncCallbackFixEnableEpoch, then, I think we should double the tests, one with the active flag and one with the disabled flag in order to easily assure backwards compatibility.

integrationTests/vm/txsFee/multiShard/asyncESDT_test.go
```diff
@@ -102,10 +102,10 @@ func TestAsyncESDTTransferWithSCCallShouldWork(t *testing.T) {
 	_, err = testContextSender.Accounts.Commit()
 	require.Nil(t, err)
 
-	expectedAccumulatedFees = big.NewInt(189880)
+	expectedAccumulatedFees = big.NewInt(189890)
```
iulianpascalau, line 105: here also. What changed in terms of fees?
If these changes are required because the activation flag is set for the transferAndAsyncCallbackFixEnableEpoch, then, I think we should double the tests, one with the active flag and one with the disabled flag in order to easily assure backwards compatibility.

process/coordinator/transactionType.go
```diff
@@ -2,6 +2,7 @@ package coordinator
 
 import (
 	"bytes"
+	"github.com/ElrondNetwork/elrond-go-core/core/atomic"
```
iulianpascalau, line 5: go imports

process/coordinator/transactionType_test.go
```diff
@@ -3,6 +3,7 @@ package coordinator
 import (
 	"bytes"
 	"encoding/hex"
+	"github.com/ElrondNetwork/elrond-go/testscommon/epochNotifier"
```
iulianpascalau, line 6: go imports

process/smartContract/process.go
```diff
@@ -300,6 +304,9 @@ func (sc *scProcessor) GasScheduleChange(gasSchedule map[string]map[string]uint6
 	}
 
 	sc.builtInGasCosts = builtInFuncCost
+	if sc.flagFixAsyncCallBackArgumentsParser.IsSet() {
+		sc.builtInGasCosts[core.BuiltInFunctionMultiESDTNFTTransfer] = builtInFuncCost["ESDTNFTMultiTransfer"]
```
iulianpascalau, line 308: this will be pretty hard to keep track. Why not create a new gas map that will be activated exactly when the flagFixAsyncCallBackArgumentsParser is set?

process/smartContract/process.go
```diff
@@ -1234,13 +1246,20 @@ func (sc *scProcessor) createVMInputWithAsyncCallBackAfterBuiltIn(
...
-		function, args, err := sc.argsParser.ParseCallData(string(outAcc.OutputTransfers[0].Data))
-		log.LogIfError(err, "function", "createVMInputWithAsyncCallBackAfterBuiltIn.ParseCallData")
-		if len(function) > 0 {
-			arguments = append(arguments, []byte(function))
-		}
 
-		arguments = append(arguments, args...)
+		if sc.flagFixAsyncCallBackArgumentsParser.IsSet() {
+			args, err := sc.argsParser.ParseArguments(string(outAcc.OutputTransfers[0].Data))
+			log.LogIfError(err, "function", "createVMInputWithAsyncCallBackAfterBuiltIn.ParseArguments")
```
iulianpascalau, line 1252: why not returning error here? Otherwise, the code execution will simply append a nil args slice and jump to L1265

process/smartContract/process.go
```diff
@@ -1957,6 +1976,9 @@ func (sc *scProcessor) penalizeUserIfNeeded(
 	if callType == vmData.AsynchronousCall {
 		return
 	}
+	if callType == vmData.AsynchronousCallBack {
```
iulianpascalau, line 1979: I think we need a flag for this

integrationTests/vm/txsFee/multiShard/asyncESDT_test.go
sasurobert: because the consumption was wrong. it was completely wrong for different reasons - like 2 fixes in process.go

integrationTests/vm/txsFee/multiShard/asyncCall_test.go
sasurobert: because the consumption was wrong. it was completely wrong for different reasons - like 2 fixes in process.go. Backward compatibility is ensured by flags.

process/coordinator/transactionType.go
sasurobert: done

process/coordinator/transactionType_test.go
sasurobert: done

process/smartContract/process.go
sasurobert: Here we have an issue of naming in gasSchedule vs builtInFunctions name and sc.bultInGasCosts[function] returned 0. Which was wrong. This is the case of different string constants between core.BuiltInFunctionMultiESDTNFTTransfer and the one from gas schedule. If I add it into the gasSchedule it will not be backward compatible.

process/smartContract/process.go
sasurobert: would leave as it is - as it calls back the contract (asyncCallBack) with empty args and that's it.

integrationTests/vm/txsFee/multiShard/asyncCall_test.go
sasurobert: We can duplicate one of these tests if needed.

integrationTests/vm/testInitializer.go
sasurobert: FactoryHandler to be precise 📦

process/smartContract/process.go
iulianpascalau: It will be, only in a v7 of the gas map that will be active in the same epoch with the fix

integrationTests/vm/testInitializer.go
iulianpascalau: no, handler, as it stores components. Factory just creates pointers but do not hold them

integrationTests/vm/txsFee/multiShard/asyncCall_test.go
iulianpascalau: ok, got it

process/smartContract/process.go
iulianpascalau: Understood the problem, I do not have a better way to fix this since all variants will involve a flag like this

iulianpascalau (dismissed):

andreibancioiu (approved):
Tests passed.

popenta (dismissed):
##                            test-nfts-async-wrong-args                   ##
- Known Warnings              16
- New Warnings                5

AdoAdoAdo (approved):

integrationTests/vm/txsFee/multiShard/asyncESDT_test.go
```diff
@@ -136,7 +136,7 @@ func TestAsyncESDTTransferWithSCCallShouldWork(t *testing.T) {
 
 	utils.ProcessSCRResult(t, testContextFirstContract, intermediateTxs[1], vmcommon.Ok, nil)
 
-	require.Equal(t, big.NewInt(278300), testContextFirstContract.TxFeeHandler.GetAccumulatedFees())
+	require.Equal(t, big.NewInt(4936720), testContextFirstContract.TxFeeHandler.GetAccumulatedFees())
```
AdoAdoAdo, line 139: There is quite a large change on the expected values.
how can we check that this value is correct?

sasurobert: checked manually the code and we get the new values. It was a problem on builtInFunction costs.

popenta (approved):
system test passed

andreibancioiu (approved):

PR #4305, opened 2022-07-19
[FIX]Heartbeat v2 integration with the hardfork trigger

iulianpascalau:
Fixed the heartbeat v2 integration with the hardfork trigger by adding an extra blocking time.sleep when the trigger is activated. This will prevent the continuous re-entry on the `Execute` function which will raise havoc at the p2p level (high amount of messages sent)

### DIFF
--- a/heartbeat/sender/routineHandler.go
+++ b/heartbeat/sender/routineHandler.go
@@ -2,6 +2,7 @@ package sender
 
 import (
 	"context"
+	"time"
 
 	logger "github.com/ElrondNetwork/elrond-go-logger"
 )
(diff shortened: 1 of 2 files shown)


heartbeat/sender/routineHandler.go
```diff
@@ -49,6 +52,7 @@ func (handler *routineHandler) processLoop(ctx context.Context) {
 			handler.heartbeatSender.Execute()
 		case <-handler.hardforkSender.ShouldTriggerHardfork():
 			handler.hardforkSender.Execute()
+			time.Sleep(handler.delayAfterHardforkMessageBroadcast)
```
iulianpascalau, line 55: Do not use time.Sleep here but rather a select that takes into account the context to properly finish the go routine on this case.

iulianpascalau: done, added an unit test for this.

popenta (approved):
##                            fix-hbv2-hardfork-trigger                    ##

BeniaminDrasovean (approved):

raduchis (approved):

PR #4311, opened 2022-07-20
[FIX]EN-12757 - Fix ERROR "bootstrap process has failed: time is out" when requesting a miniblock

iulianpascalau:
- rename the PendingMiniBlocksSyncHandler to MiniBlocksSyncHandler
- requesting & saving all involved miniblocks from shard headers


epochStart/bootstrap/process.go
```diff
@@ -898,9 +892,15 @@ func (e *epochStartBootstrap) requestAndProcessForShard(peerMiniBlocks []*block.
 		return err
 	}
 
-	for hash, hdr := range dts.additionalHeaders {
```
iulianpascalau, line 901: Put back the code here!

iulianpascalau: done, thanks

PR #4314, opened 2022-07-21
Feat/kosk bls multisigner

AdoAdoAdo:
This PR brings performance improvements to block signature verification time/cpu usage.
By relying on the staking mechanism, where the knowledge of the secret key(kosk) for every registered BLS public key is verified by providing also a valid signature for it, we can remove some operations for the signatures aggregation and the aggregated signature verification, that were previously used to avoid rogue keys, since the same is achieved through KOSK.

The removal of these operations save ~100ms CPU time for the verification of every metachain block, and ~20ms for shard blocks.


gabi-vuls (dismissed):

ssd04 (approved):

factory/crypto/multiSignerContainer.go
```diff
@@ -0,0 +1,156 @@
...
+		multiSigners: make([]*epochMultiSigner, len(multiSignerConfig)),
+	}
+
+	sortedMultiSignerConfig := sortMultiSignerConfig(multiSignerConfig)
+	if sortedMultiSignerConfig[0].EnableEpoch != 0 {
+		return nil, errors.ErrMissingEpochZeroMultiSignerConfig
+	}
+
+	for i, mConfig := range sortedMultiSignerConfig {
+		multiSigner, err := createMultiSigner(mConfig.Type, args)
```
iulianpascalau, line 59: :+1:

factory/crypto/multiSignerContainer.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return c, nil
+}
+
+// GetMultiSigner returns the multiSigner configured for the given epoch
+func (c *container) GetMultiSigner(epoch uint32) (crypto.MultiSigner, error) {
+	c.mutSigners.RLock()
+	defer c.mutSigners.RUnlock()
+
+	for i := len(c.multiSigners) - 1; i >= 0; i-- {
```
iulianpascalau, line 78: :+1:

update/factory/fullSyncInterceptors.go
```diff
@@ -246,7 +246,11 @@ func checkBaseParams(
 	if check.IfNil(cryptoComponents.BlockSigner()) {
 		return process.ErrNilSingleSigner
 	}
-	if check.IfNil(cryptoComponents.MultiSigner()) {
+	multiSigner, err := cryptoComponents.GetMultiSigner(0)
```
iulianpascalau, line 249: is this ok?

AdoAdoAdo: this one checks that we at least have a configured multisigner (for epoch 0)
we need minimum one multisigner.
If we have more, the change is that we will have epoch ranges where different multisigners are used, but the minimum required is to have at least one.

iulianpascalau: Still doesn't feel right. We are using this only to test the next error. This will always work as we will always a component for the epoch 0 (otherwise we will not have the integration & system tests pass). Suggestion to completely remove L249-L255

ssd04 (approved):

iulianpascalau (approved):

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/ElrondNetwork/covalent-indexer-go v1.0.6
 	github.com/ElrondNetwork/elastic-indexer-go v1.2.41
 	github.com/ElrondNetwork/elrond-go-core v1.1.19
-	github.com/ElrondNetwork/elrond-go-crypto v1.0.1
+	github.com/ElrondNetwork/elrond-go-crypto v1.1.1-0.20220725093518-7a5126e0f2e3
```
ssd04, line 12: we should not forget about this

iulianpascalau: either a push here or in an upcoming PR as we have another PR waiting there

PR #4325, opened 2022-07-26
[FIX]Added a disabled peer authentication resolver in import-db mode

iulianpascalau:
- added a disabled peer authentication resolver when the node is running in import-db mode to avoid the benign error messages like:
```
ERROR[2022-07-26 13:51:57.318]   RequestPeerAuthenticationsChunk.MetaChainResolver error = element does not exist in container in resolvers container for key peerAuthentication topic = peerAuthentication shard = 0 chunk = 13 epoch = 301
```
  
## Proposed Changes
- added a new disabled resolver in order to avoid the error signaled by the requestHandler component 

## Testing procedure
- standard internal testnet procedure and re-run the import-db process and validate that the aforementioned error does not appear


sstanculeanu (approved):

andreibancioiu (approved):

popenta (approved):
##                            fix-import-db-resolvers                      ##
- Known Warnings              18

PR #4327, opened 2022-07-27
Fixed import-db

iulianpascalau:
- there was a bug in the import-db process if we have used the old version of DBs that did not contain the `ActiveDB` marker. The versions <=1.3.36 did not care about it as it would have re-written anyway that marker and the snapshotting process worked as expected. With the rc/2022-july branch, a new condition appeared as to write the marker only on epoch 0;
- some prints regarding the wrong epoch data were not in line with the rest of the project, refactored the prints for better debugging experience.
  
## Proposed Changes
- add a composed condition in which, we take for granted in `import-db` process that the DB is correctly constructed as to reuse the old trie databases.

## Testing procedure
- `import-db` process should work as expected, without ERRORs or WARNs;
- internal testing scenarios should work.

### DIFF
--- a/epochStart/metachain/epochStartData.go
+++ b/epochStart/metachain/epochStartData.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/epochStart"
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/sharding"
+	"github.com/davecgh/go-spew/spew"
 )
 
 var _ process.EpochStartDataCreator = (*epochStartData)(nil)
@@ -106,11 +107,8 @@ func (e *epochStartData) VerifyEpochStartDataForMetablock(metaBlock *block.MetaB
 	}
 
 	if !bytes.Equal(receivedEpochStartHash, createdEpochStartHash) {
-		log.Warn("RECEIVED epoch start data")
-		displayEpochStartData(&metaBlock.EpochStart)
-
-		log.Warn("CREATED epoch start data")
-		displayEpochStartData(startData)
+		displayEpochStartData("received", receivedEpochStartHash, &metaBlock.EpochStart)
+		displayEpochStartData("created", createdEpochStartHash, startData)
 
 		return process.ErrEpochStartDataDoesNotMatch
 	}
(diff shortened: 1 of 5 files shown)


epochStart/metachain/epochStartData.go
```diff
@@ -106,21 +107,21 @@ func (e *epochStartData) VerifyEpochStartDataForMetablock(metaBlock *block.MetaB
...
 }
 
-func displayEpochStartData(startData *block.EpochStart) {
+func displayEpochStartData(mode string, hash []byte, startData *block.EpochStart) {
+	log.Warn(mode+" epoch start data",
+		"hash", hash, "startData", spew.Sdump(startData))
+
 	for _, shardData := range startData.LastFinalizedHeaders {
-		log.Debug("epoch start shard data",
+		log.Warn("epoch start shard data",
```
bogdan-rosianu, line 124: change back to DEBUG ?

state/accountsDB.go
```diff
@@ -156,12 +156,16 @@ func checkArgsAccountsDB(args ArgsAccountsDB) error {
...
-func startSnapshotAfterRestart(adb AccountsAdapter, tsm common.StorageManager) {
+func startSnapshotAfterRestart(adb AccountsAdapter, args ArgsAccountsDB) {
+	tsm := args.Trie.GetStorageManager()
 	epoch, err := tsm.GetLatestStorageEpoch()
 	if err != nil {
 		log.Error("could not get latest storage epoch")
 	}
-	if epoch == 0 && err == nil {
+	putActiveDBMarker := epoch == 0 && err == nil
+	putActiveDBMarker = putActiveDBMarker || args.ProcessingMode == common.ImportDb
```
bogdan-rosianu, line 166: might add a new variable. for example `isInImportDBMode` and then check if either of them is true

BeniaminDrasovean (dismissed):

epochStart/metachain/epochStartData.go
iulianpascalau: no, as the message that indicate a serious problem is on WARN level, the details are on debug level

state/accountsDB.go
iulianpascalau: done

BeniaminDrasovean (approved):

bogdan-rosianu (approved):

popenta (approved):
System test passed

PR #4332, opened 2022-08-01
new SCResults from functions

mihaieremia:
-  Liquid Staking will require more information in the callbacks return from the withdraw/reDelegateRewards/delegate functions.
- Getting the reDelegatedRewards amount in the callback will remove the complexity in Rust contracts from computing the  rewards in the local memory in order to know exactly how much will be reDelegated and update the rust virtual balance.
- With this implementation we can avoid one claimRewards and then reDelegateRewards in liquid staking SC.
  
## Proposed Changes
- For reDelegateRewards and delegate we should return the amount that was reDelegated or delegated so we can handle it better in the Rust smart contracts
- For withdraw because of the transfer and execute method used in the delegation contract we are not receiving the amount in our callback, instead we could just return the amount we sent as a SCResult which will be received in our Rust callback
- 

## Testing procedure
- Added extra checks in the unit tests
- 
-


vm/systemSmartContracts/delegation.go
```diff
@@ -1498,6 +1498,7 @@ func (d *delegation) finishDelegateUser(
 		return vmcommon.UserError
 	}
 
+	d.eei.Finish(delegateValue.Bytes())
```
sasurobert, line 1501: these changes are not backwards compatible. It should not be targeted to RC-JULY as that RC is closed from features perspective. Backward compatibility flag has to be added. There in no need to add these information for normal users, only for SC calling, any information like this has to be limited.
For liquid staking - it is almost mandatory to claim rewards and delegate after, as redelegation can fail for various reasons. Separate delegation means more control of funds. For staking phase 4 we have a series of changes on staking, delegation, delegation manager, we might introduce a set of things there. feat/liquid-staking is the branch.

PR #4338, opened 2022-08-02
En-11896 move constants to config

sstanculeanu:
- `MaxRoundsToKeepUnprocessedMiniBlocks` and `MaxRoundsToKeepUnprocessedTransactions` constants needed to be moved to config as per [EN-11896](https://elrondnetwork.atlassian.net/browse/EN-11896)
  
## Proposed Changes
- move constants to configs
- small refactor, moved common code to `basePoolsCleaner`

## Testing procedure
- system test


process/block/poolsCleaner/miniBlocksPoolsCleaner.go
```diff
@@ -29,44 +34,42 @@ type mbInfo struct {
...
+	if check.IfNil(args.MiniblocksPool) {
+		return process.ErrNilMiniBlockPool
+	}
+
+	return err
+}
+
 // StartCleaning actually starts the pools cleaning mechanism
 func (mbpc *miniBlocksPoolsCleaner) StartCleaning() {
 	var ctx context.Context
```
iulianpascalau, line 75: as discussed, we might add a guard check here in order to not be able to call this method twice. The guard can be made up with a bool & mutex

sstanculeanu: added

process/block/poolsCleaner/basePoolsCleaner.go
```diff
@@ -22,13 +24,16 @@ type basePoolsCleaner struct {
 	shardCoordinator               sharding.Coordinator
 	maxRoundsToKeepUnprocessedData int64
 	cancelFunc                     func()
+	isCleaningRoutineRunning       atomic.Flag
```
iulianpascalau, line 27: since you have a mutex protection, could have been a plain old bool var. Can be left as it is

iulianpascalau (dismissed):

process/block/poolsCleaner/basePoolsCleaner.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Close will close the endless running go routine
+func (base *basePoolsCleaner) Close() error {
+	base.mut.Lock()
+	defer base.mut.Unlock()
+
+	if base.cancelFunc != nil {
+		base.cancelFunc()
+		base.isCleaningRoutineRunning.Reset()
```
ssd04, line 62: can't we remove the mutex if it's for an atomic flag?

sstanculeanu: made the flag bool and kept mutex

process/block/poolsCleaner/basePoolsCleaner.go
sstanculeanu: pushed

iulianpascalau (approved):

ssd04 (approved):

popenta (approved):
##                            test_constants_to_config                     ##

PR #4341, opened 2022-08-03
Fix flag in import-db

iulianpascalau:
- there was a bug when the node ran in import-db that automatically reset the flag `SnapshotsEnabled` causing the node to not start due to the fact it didn't found the main trie data
  
## Proposed Changes
- remove the auto-reset operation in import-db mode

## Testing procedure
- import-db process should work when starting on a non-zero epoch and regular system test should pass


andreibancioiu (approved):

BeniaminDrasovean (approved):

gabi-vuls (approved):
System test passed.

PR #4343, opened 2022-08-03
Always keep a minimum of 2 active dbs open

BeniaminDrasovean:

- The pruning storer will keep only 3 db's open. When a snapshot takes longer than an epoch, it will close a db that is needed. To fix this case, always keep a minimum of 2 db's that are marked as active, opened.

  
## Proposed Changes

-  Modify `minNumOfActiveDBsNecessary` from 1 to 2

## Testing procedure

- Add a time.Sleep() on Snapshot(), so that it will take more than 1 epoch, and check that the storer life was extended. After the snapshot is completed, verify that the correct db's are closed - (already tested this case)

### DIFF
--- a/storage/pruning/triePruningStorer.go
+++ b/storage/pruning/triePruningStorer.go
@@ -10,9 +10,10 @@ import (
 )
 
 const (
-	lastEpochIndex             = 1
-	currentEpochIndex          = 0
-	minNumOfActiveDBsNecessary = 1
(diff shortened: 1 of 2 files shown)


iulianpascalau (approved):

mariusmihaic (approved):

popenta (approved):
##                            snapshot-takes-too-long-fix                  ##
- New Warnings                6

PR #4345, opened 2022-08-03
Disabled storer on SaveInStorageEnabled false

sstanculeanu:
- when `SaveInStorageEnabled` was false, the `TxLogsUnit` storer was nil, leading to panic on different api calls
  
## Proposed Changes
- added disabled storer for this case

## Testing procedure
- system test, using `/block/by-nonce/15?withTxs=true&withLogs=true` should not panic no matter what block is asked for, with config values `LogsAndEvents.SaveInStorageEnabled = false` and `DbLookupExtensions.Enabled = false`

### DIFF
--- a/node/external/logs/logsRepository.go
+++ b/node/external/logs/logsRepository.go
@@ -4,10 +4,12 @@ import (
 	"encoding/hex"
 	"fmt"
 
+	"github.com/ElrondNetwork/elrond-go-core/core/check"
 	"github.com/ElrondNetwork/elrond-go-core/data/transaction"
 	"github.com/ElrondNetwork/elrond-go-core/marshal"
 	"github.com/ElrondNetwork/elrond-go/dataRetriever"
 	"github.com/ElrondNetwork/elrond-go/storage"
+	"github.com/ElrondNetwork/elrond-go/storage/disabled"
 )
 
 type logsRepository struct {
@@ -17,6 +19,10 @@ type logsRepository struct {
 
 func newLogsRepository(storageService dataRetriever.StorageService, marshaller marshal.Marshalizer) *logsRepository {
 	storer := storageService.GetStorer(dataRetriever.TxLogsUnit)
+	if check.IfNil(storer) {
+		log.Debug("could not find TxLogsUnit, using a disabled storer instead...")
+		storer = disabled.NewStorer()
+	}
 
 	return &logsRepository{
 		storer:     storer,
--- a/node/external/logs/logsRepository.go
+++ b/node/external/logs/logsRepository.go
@@ -4,12 +4,10 @@ import (
 	"encoding/hex"
 	"fmt"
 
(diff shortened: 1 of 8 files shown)


node/external/logs/logsRepository_test.go
```diff
@@ -61,3 +61,17 @@ func TestLogsRepository_GetLogsShouldErr(t *testing.T) {
 	require.ErrorIs(t, err, errCannotUnmarshalLog)
 	require.Nil(t, badLog)
 }
+
+func TestLogsRepository_GetLogsShouldUseDisabled(t *testing.T) {
```
andreibancioiu, line 65: :+1:

storage/disabled/storer.go
```diff
@@ -0,0 +1,85 @@
+package disabled
+
+import (
+	storageCore "github.com/ElrondNetwork/elrond-go-core/storage"
+)
+
+type storer struct{}
+
+// NewStorer returns a new instance of this disabled storer
```
andreibancioiu, line 9: Oh, we didn't have an existing "generic" disabled storer, all the other ones were defined in their client packages. :+1: for adding it here.

andreibancioiu (approved):

node/external/logs/logsRepository.go
```diff
@@ -17,6 +19,10 @@ type logsRepository struct {
 
 func newLogsRepository(storageService dataRetriever.StorageService, marshaller marshal.Marshalizer) *logsRepository {
 	storer := storageService.GetStorer(dataRetriever.TxLogsUnit)
+	if check.IfNil(storer) {
```
andreibancioiu, line 22: It's a bit debatable if we should have used a `disabledLogsFacade` at a higher level, instead, when logs saving is disabled. But it's all right this way, as well :pray:

andreibancioiu (approved):

node/external/logs/logsRepository.go
iulianpascalau, line 22: instead of this approach, I would make the `storageService.GetStorer(dataRetriever.TxLogsUnit)` call return always a not-nil pointer, on all conditions. So, in StorageServiceFactory.setupLogsAndEventsStorer function, on L556 I would add a disabled storer.
This will simplify all involved code and this new if branch will no longer be necessary. 
For a next PR: we might return an error on the GetStorer function whenever we've provided a key that does not contain a valid storer as to align the code to the rest of the project.

sstanculeanu: pushed Iulian's solution which fixes this

sstanculeanu: pushed

iulianpascalau (approved):

andreibancioiu (approved):

popenta (approved):
##                            fix_logs_repository                          ##
- New Warnings                3

PR #4365, opened 2022-08-11
Enable snapshot even if pruning is disabled

BeniaminDrasovean:
- If pruning was disabled, the snapshotting process was automatically disabled. Snapshotting is needed to copy from an older db to a new one (because db pruning is enabled even if trie pruning is disabled)
  
## Proposed Changes
- Snapshot the state even if trie pruning is disabled

## Testing procedure
- Test with `AccountsStatePruningEnabled = false` and with `AccountsStatePruningEnabled = true` and both scenarios should work


raduchis (approved):

AdoAdoAdo (approved):

iulianpascalau (approved):

popenta (approved):
##                            enable-snapshot-if-pruning-disabled          ##
- Known Warnings              15
- New Warnings                4

PR #4370, opened 2022-08-12
Do not start state sync if the node is closing

BeniaminDrasovean:
- When the node is closing, the sync block will fail because it can not get from storage, so it starts a state sync process. 
  
## Proposed Changes
- If the node is closing, do not start the state sync process

## Testing procedure
- We should not have any `base sync: started syncUserAccountsState` WARNs

### DIFF
--- a/process/sync/shardblock.go
+++ b/process/sync/shardblock.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go-core/data/block"
 	"github.com/ElrondNetwork/elrond-go/common"
 	"github.com/ElrondNetwork/elrond-go/dataRetriever"
+	"github.com/ElrondNetwork/elrond-go/errors"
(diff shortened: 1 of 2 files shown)


sstanculeanu (approved):

process/sync/shardblock.go
```diff
@@ -149,6 +149,26 @@ func (boot *ShardBootstrap) SyncBlock(ctx context.Context) error {
 	return err
 }
 
+func isErrGetNodeFromDB(err error) bool {
```
iulianpascalau, line 152: perhaps we should rename this function: "isKnownClosingOrMissingError" ?

BeniaminDrasovean: If it is closingErr, this func will return false. So naming the func `isKnownClosing...`, but if it is closing we return false is misleading.

iulianpascalau (approved):

popenta (approved):
System test passed

PR #4381, opened 2022-08-19
Integrate named identities

sstanculeanu:
- integrate changes on prefs config, which now provides identities that runs nodes on the multisig
  
## Proposed Changes
- extend `peerInfo` structure to hold node name and identity as well
- use bls keys + name and identity from prefs config on `keysHolder`

## Testing procedure
- system test

### DIFF
--- a/cmd/node/config/prefs.toml
+++ b/cmd/node/config/prefs.toml
@@ -31,3 +31,16 @@
    #  - "disabled" - no connection watching should be made
    #  - "print" - new connection found will be printed in the log file
    ConnectionWatcherType = "disabled"
+
+# NamedIdentity represents an identity that runs nodes on the multisig.
+# There can be multiple identities set just by dupplicating the NamedIdentity
+[[NamedIdentity]]
+   # Identity represents the keybase identity for the current NamedIdentity
+   Identity = ""
+   # NodeName represents the name that will be given to the names of the current identity
+   NodeName = ""
+   # BLSKeys represents the BLS keys assigned to the current NamedIdentity
+   BLSKeys = [
+      "",
+      ""
+   ]
--- a/cmd/node/config/prefs.toml
+++ b/cmd/node/config/prefs.toml
@@ -4,10 +4,12 @@
    # if "disabled" is provided then the node will start in the corresponding shard for its public key or 0 otherwise
    DestinationShardAsObserver = "disabled"
 
-   # NodeDisplayName represents the friendly name a user can pick for his node in the status monitor
+   # NodeDisplayName represents the friendly name a user can pick for his node in the status monitor when the node does not run in multikey mode
+   # In multikey mode, all bls keys not mentioned in NamedIdentity section will use this one as default
    NodeDisplayName = ""
 
-   # Identity represents the keybase's identity
+   # Identity represents the keybase's identity when the node does not run in multikey mode
+   # In multikey mode, all bls keys not mentioned in NamedIdentity section will use this one as default
    Identity = ""
 
    # RedundancyLevel represents the level of redundancy used by the node (-1 = disabled, 0 = main instance (default),
@@ -32,8 +34,8 @@
    #  - "print" - new connection found will be printed in the log file
    ConnectionWatcherType = "disabled"
 
-# NamedIdentity represents an identity that runs nodes on the multisig.
-# There can be multiple identities set just by dupplicating the NamedIdentity
+# NamedIdentity represents an identity that runs nodes on the multikey
+# There can be multiple identities set on the same node, each one of them having different bls keys, just by duplicating the NamedIdentity
 [[NamedIdentity]]
    # Identity represents the keybase identity for the current NamedIdentity
    Identity = ""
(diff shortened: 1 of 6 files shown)


keysManagement/keysHolder.go
```diff
@@ -63,10 +69,40 @@ func checkVirtualPeersHolderArgs(args ArgsVirtualPeersHolder) error {
 		return fmt.Errorf("%w for MaxRoundsWithoutReceivedMessages, minimum %d, got %d",
 			errInvalidValue, minRoundsWithoutReceivedMessages, args.MaxRoundsWithoutReceivedMessages)
 	}
+	if len(args.NamedIdentities) == 0 {
+		return errMissingNamedIdentity
```
iulianpascalau, line 73: we should not consider this a critical error. Not even a log.Debug is required. If we did not provide some named identities, use the default ones on all managed keys.

keysManagement/keysHolder.go
```diff
@@ -42,14 +45,17 @@ func NewVirtualPeersHolder(args ArgsVirtualPeersHolder) (*virtualPeersHolder, er
...
-	return &virtualPeersHolder{
+	holder := &virtualPeersHolder{
 		data:                             make(map[string]*peerInfo),
 		pids:                             make(map[core.PeerID]struct{}),
 		keyGenerator:                     args.KeyGenerator,
 		p2pIdentityGenerator:             args.P2PIdentityGenerator,
 		isMainMachine:                    args.IsMainMachine,
 		maxRoundsWithoutReceivedMessages: args.MaxRoundsWithoutReceivedMessages,
-	}, nil
+		namedIdentitiesMap:               createNamedIdentitiesMap(args.NamedIdentities),
```
iulianpascalau, line 55: we should not have a different map for this: instead, place all the data in the data member of type make(map[string]*peerInfo), add new fields in the peerInfo struct
If we do not found a provided named identity for a certain BLS pk, take the ones provided in the prefs.toml L8 & L11

keysManagement/peerInfo.go
```diff
@@ -7,11 +7,17 @@ import (
...
+	name     string
+	identity string
+}
+
 type peerInfo struct {
 	pid                core.PeerID
 	p2pPrivateKeyBytes []byte
 	privateKey         crypto.PrivateKey
 	machineID          string
+	namedIdentity      namedIdentity
```
iulianpascalau, line 20: we do not need a different struct here, place directly the name & identity members

keysManagement/keysHolder.go
```diff
@@ -67,6 +75,28 @@ func checkVirtualPeersHolderArgs(args ArgsVirtualPeersHolder) error {
...
 }
 
+func createDataMap(namedIdentities []config.NamedIdentity) map[string]*peerInfo {
+	dataMap := make(map[string]*peerInfo)
+
+	for _, identity := range namedIdentities {
+		for _, blsKey := range identity.BLSKeys {
+			bls, err := hex.DecodeString(blsKey)
+			if err != nil {
+				continue
```
iulianpascalau, line 85: maybe return error here? The node should stop prematurely if the config is not done properly

keysManagement/keysHolder.go
```diff
@@ -87,29 +117,35 @@ func (holder *virtualPeersHolder) AddVirtualPeer(privateKeyBytes []byte) error {
...
-		machineID:          generateRandomMachineID(),
-	}
-
 	holder.mut.Lock()
 	defer holder.mut.Unlock()
 
-	_, found := holder.data[string(publicKeyBytes)]
-	if found {
+	pInfo, found := holder.data[string(publicKeyBytes)]
+	if found && len(pInfo.pid.Bytes()) != 0 {
```
iulianpascalau, line 124: :+1: we can live with this condition for "check if value is unset"

iulianpascalau (approved):

iulianpascalau (dismissed):

cmd/node/config/prefs.toml
```diff
@@ -31,3 +31,16 @@
    #  - "disabled" - no connection watching should be made
    #  - "print" - new connection found will be printed in the log file
    ConnectionWatcherType = "disabled"
+
+# NamedIdentity represents an identity that runs nodes on the multisig.
+# There can be multiple identities set just by dupplicating the NamedIdentity
```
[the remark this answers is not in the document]
bogdan-rosianu: we still have the `Identity` field as well. This is misleading. explain in a comment above `Identity = ""` that is to be used only by a regular node (1 key)

cmd/node/config/prefs.toml
```diff
@@ -31,3 +31,16 @@
    #  - "disabled" - no connection watching should be made
    #  - "print" - new connection found will be printed in the log file
    ConnectionWatcherType = "disabled"
+
+# NamedIdentity represents an identity that runs nodes on the multisig.
```
bogdan-rosianu, line 35: multikey*?

sstanculeanu: pushed

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #4384, opened 2022-08-19
Fix import-db flags (NumActivePersisters)

andreibancioiu:
- When starting an **import-db** flow, the `config.StoragePruning.NumActivePersisters` was overwritten, and set to `config.StoragePruning.NumEpochsToKeep`. When `NumEpochsToKeep` is set to a large number (e.g. 50), a large number of files would be open (without being necessary, though).
  
## Proposed Changes
- On import-db, do not overwrite `config.StoragePruning.NumActivePersisters` anymore

## Testing procedure
- In config, set a large number to `NumEpochsToKeep`, then start an import-db flow. Once the process starts, look for the log details around `the node is in import mode`. Also look for RAM usage (compare it to an import-db flow without this fix).


iulianpascalau (approved):

bogdan-rosianu (dismissed):

gabi-vuls (approved):
System test passed.

cmd/node/flags.go
```diff
@@ -511,7 +511,6 @@ func processConfigImportDBMode(log logger.Logger, configs *config.Configs) error
 		generalConfigs.GeneralSettings.StartInEpochEnabled = false
 	}
 
-	generalConfigs.StoragePruning.NumActivePersisters = generalConfigs.StoragePruning.NumEpochsToKeep
```
iulianpascalau, line 514: we need this in a form or another:
```GO
// we need to increment this in order to make the storage resolvers work since they open 2 epochs in advance
generalConfigs.StoragePruning.NumActivePersisters++
```

iulianpascalau (changes requested):

iulianpascalau (dismissed):

gabi-vuls (approved):
System test passed.

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #4385, opened 2022-08-20
Switch heartbeat v2 to single data interceptor as multi data is not needed

sstanculeanu:
- heartbeat v2 message is only sent one message at a time, so using multi data interceptor leads to an extra marshal/unmarshal operation on each message
  
## Proposed Changes
- switch to use singleDataInterceptor

## Testing procedure
- system test


heartbeat/sender/heartbeatSender.go
```diff
@@ -106,16 +105,7 @@ func (sender *heartbeatSender) execute() error {
...
-		Data: make([][]byte, 1),
-	}
-	b.Data[0] = msgBytes
-	data, err := sender.marshaller.Marshal(b)
-	if err != nil {
-		return err
-	}
-
-	sender.messenger.Broadcast(sender.topic, data)
+	sender.messenger.Broadcast(sender.topic, msgBytes)
```
iulianpascalau, line 108: :+1:

iulianpascalau (approved):

bogdan-rosianu (approved):

schimih (approved):

gabi-vuls (approved):
System test passed.

PR #4391, opened 2022-08-22
Add code-hash endpoint to address group

lcswillems:

In [Arda](https://arda.run), we are going to add a service that tracks if smart contract binary codes change.

Currently, it is impossible to only request the code hash. The only way is to download all the account data (including the binary code that can be huge).
  
## Proposed Changes

Add a code-hash endpoint to the address group.

## Testing procedure
-


api/groups/addressGroup.go
```diff
@@ -20,6 +20,7 @@ const (
 	getAccountPath            = "/:address"
```
iulianpascalau, line 20: I do not think we need this new route because on /address/:address route we return the api.AccountResponse struct that contains the codehash already.

lcswillems: Yes but it will return all the account data, including the hex code, so a lot of data is going to be returned whereas I just want the code hash.

As an example, on a page, I may need to request 10 code hashes of 10 different smart contracts. Currently, the data sent over the wire is 100x superior to what could be achieved if only the code hash was sent.

Furthermore, if we following the reasoning "we don't need this route because /address/:address returns everything", why having added "/address/:address/balance" or "/address/:address/username"?

facade/nodeFacade.go
```diff
@@ -175,6 +175,11 @@ func (nf *nodeFacade) GetUsername(address string, options apiData.AccountQueryOp
 	return nf.node.GetUsername(address, options)
 }
 
+// GetCodeHash gets the username for a specified address
```
bogdan-rosianu, line 178: wrong comment

node/node.go
```diff
@@ -179,6 +179,17 @@ func (n *Node) GetUsername(address string, options api.AccountQueryOptions) (str
...
 
+// GetCodeHash gets the code hash for a specific address
+func (n *Node) GetCodeHash(address string, options api.AccountQueryOptions) (string, api.BlockInfo, error) {
+	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
+	if err != nil {
+		return "", api.BlockInfo{}, err
+	}
+
+	codeHash := userAccount.GetCodeHash()
+	return string(codeHash), blockInfo, nil
```
bogdan-rosianu, line 190: this isn't right. Please have a look over `nodeFacade.GetAccount()` function. 
You will see this code block:
```
        codeHash := accountResponse.CodeHash
	code, _ := nf.node.GetCode(codeHash, options)
	accountResponse.Code = hex.EncodeToString(code)
	return accountResponse, blockInfo, nil
```
Therefore, I suggest that you refactor your code, by calling `n.GetCode` inside this function (`n.GetCodeHash`)

In the end, the code hash should be hexified

api/groups/addressGroup.go
```diff
@@ -20,6 +20,7 @@ const (
...
 	getBalancePath            = "/:address/balance"
 	getUsernamePath           = "/:address/username"
+	getCodeHashPath						= "/:address/code-hash"
```
bogdan-rosianu, line 23: wrong alignment

bogdan-rosianu (changes requested):
Thank you for opening this PR. I have left a few comments

api/groups/addressGroup.go
bogdan-rosianu: agree, it is ok to add this endpoint. But keep in mind that `development` branch is used for future releases, therefore your addition won't go live too soon. However, it is the only way we can accept your PR, because the release candidate branch is in final testing procedures.
Also, please use signed commits. Otherwise, we won't be able to merge your changes

iulianpascalau: ok, will review the code then

api/groups/addressGroup.go
```diff
@@ -215,6 +222,29 @@ func (ag *addressGroup) getUsername(c *gin.Context) {
 	shared.RespondWithSuccess(c, gin.H{"username": userName, "blockInfo": blockInfo})
 }
 
+// getCodeHash returns the code hash for the address parameter
+func (ag *addressGroup) getCodeHash(c *gin.Context) {
+	addr := c.Param("address")
+	if addr == "" {
+		shared.RespondWithValidationError(c, errors.ErrGetUsername, errors.ErrEmptyAddress)
```
iulianpascalau, line 229: wrong error var returned, please create another one

api/groups/addressGroup.go
```diff
@@ -215,6 +222,29 @@ func (ag *addressGroup) getUsername(c *gin.Context) {
 	shared.RespondWithSuccess(c, gin.H{"username": userName, "blockInfo": blockInfo})
 }
 
+// getCodeHash returns the code hash for the address parameter
```
iulianpascalau, line 225: missing unit tests

node/node.go
iulianpascalau: As @bogdan-rosianu suggested, you should not return the code hash as a string as it might contain special characters that will not be correctly handled by the webserver. So, you should use:
```GO
return hex.EncodeToString(codeHash), blockInfo, nil
```

node/node.go
```diff
@@ -179,6 +179,17 @@ func (n *Node) GetUsername(address string, options api.AccountQueryOptions) (str
 	return string(username), blockInfo, nil
 }
 
+// GetCodeHash gets the code hash for a specific address
+func (n *Node) GetCodeHash(address string, options api.AccountQueryOptions) (string, api.BlockInfo, error) {
```
iulianpascalau, line 183: please add an unit test for this

iulianpascalau (changes requested):

api/groups/addressGroup.go
lcswillems: You are right. I indented with tabs instead of spaces and didn't remark it in my editor

facade/nodeFacade.go
lcswillems: Yes sorry

api/groups/addressGroup.go
lcswillems: Yes, you are right. Line 241, I also used:

```
shared.RespondWithInternalError(c, errors.ErrCouldNotGetAccount, err)
```

And I believe it is not the correct error too, so I changed it to be:

```
shared.RespondWithInternalError(c, errors.ErrGetCodeHash, err)
```

node/node.go
lcswillems: Added

api/groups/addressGroup.go
lcswillems: Added

lcswillems:
@bogdan-rosianu @iulianpascalau I have forced pushed to add changes you requested and have a signed commit.

I think I fixed all the issues you reported except one about the following code in `node.go`:

```
}

codeHash := userAccount.GetCodeHash()
return string(codeHash), blockInfo, nil
```

1. I don't understand what is wrong with this. I checked the code in `nodeFacade.GetAccount()` but don't understand why `GetCode` should appear in `GetCodeHash`.

2. You told me to return an hex version of the code hash. I did the modification. However, it seems I should rather return a base64 version of the code hash, isn't it? Because in the endpoint /address/:address, the code hash is base64 encoded.

iulianpascalau:
> @bogdan-rosianu @iulianpascalau I have forced pushed to add changes you requested and get the commit signed.
> 
> I think I fixed all the issues you reported except one about the following code in `node.go`:
> 
> ```
> }
> 
> codeHash := userAccount.GetCodeHash()
> return string(codeHash), blockInfo, nil
> ```
> 
> 1. I don't understand what is wrong with this. I checked the code in `nodeFacade.GetAccount()` but don't understand why `GetCode` should appear in `GetCodeHash`.
> 2. You told me to return an hex version of the code hash. I did the modification. However, it seems I should rather return a base64 version of the code hash, isn't it? Because in the endpoint /address/:address, the code hash is base64 encoded.

Let's see: the userAccount.GetCodeHash() returns the plain []byte slice (usually will be here 32 bytes). These will almost surely contain unreadable characters. Like `/0` or `/1` and so forth. If you just cast it to string, the webserver will think that this is a valid string and, I think, will not try to encode it even further. The suggestion to return the hex encoded code hash was one suggestion, but of course it can be encoded into a base64 form. 
To comply with the "clean code" guidance, indeed, you should do the encoding only on the most outer layer of the plugin involved and pass the code hash as []byte slice everywhere else (node.go, the facade, interfaces involved and so on).
So, in addressGroup.go L245 you should put:
```GO
shared.RespondWithSuccess(c, gin.H{"codeHash": base64.StdEncoding.EncodeToString(codeHash), "blockInfo": blockInfo}) 
```
Or you can use hex. We mostly return hex encoded stuff as it is easier to handle on some frontends.

lcswillems:
Thanks a lot @iulianpascalau for the time you took to explain me! I added a new commit that should address the remaining issue

Thanks a lot @bogdan-rosianu for your review and time too!

api/groups/addressGroup_test.go
```diff
@@ -395,6 +405,58 @@ func TestGetUsername_ShouldWork(t *testing.T) {
...
+	ws := startWebServer(addrGroup, "address", getAddressRoutesConfig())
+
+	req, _ := http.NewRequest("GET", fmt.Sprintf("/address/%s/code-hash", testAddress), nil)
+	resp := httptest.NewRecorder()
+	ws.ServeHTTP(resp, req)
+
+	codeHashResponseObj := codeHashResponse{}
+	loadResponse(resp.Body, &codeHashResponseObj)
+	assert.Equal(t, http.StatusOK, resp.Code)
+	assert.Equal(t, testCodeHash, codeHashResponseObj.Data.CodeHash)
```
iulianpascalau, line 457: this test fail, maybe:
```GO
assert.Equal(t, base64.StdEncoding.EncodeToString(testCodeHash), codeHashResponseObj.Data.CodeHash)
```
?

lcswillems: Two tests were failing. Sorry for that! I think I fixed them now

lcswillems:
@iulianpascalau Two tests were failing. Sorry for that! I think I fixed them now.

iulianpascalau (dismissed):

api/groups/addressGroup.go
```diff
@@ -215,6 +223,29 @@ func (ag *addressGroup) getUsername(c *gin.Context) {
...
+		return
+	}
+
+	codeHash, blockInfo, err := ag.getFacade().GetCodeHash(addr, options)
+	if err != nil {
+		shared.RespondWithInternalError(c, errors.ErrGetCodeHash, err)
+		return
+	}
+
+	shared.RespondWithSuccess(c, gin.H{"codeHash": base64.StdEncoding.EncodeToString(codeHash), "blockInfo": blockInfo})
```
bogdan-rosianu, line 246: no need to encode it. you can leave the field as byte array since json will automatically convert it to base64 string

bogdan-rosianu (changes requested):

lcswillems:
@bogdan-rosianu Thanks for the review. I removed the useless encoding

bogdan-rosianu (approved):

lcswillems:
Thank you!!

lcswillems:
Hi @iulianpascalau @bogdan-rosianu 
Has this PR merged in the codebase somewhere? The branch external-codehash-endpoint doesn't exist anymore. And the code of the PR doesn't seem to be in development or master.

lcswillems:
By the way, the product that would benefit from this endpoint has been launched. It is Arda Audits: https://arda.run/audits

iulianpascalau:
Hello. We've planned to phase out the `development` branch. The docs will be updated soon to specify where the external PRs should be started from. 
However, before phasing out the `development` branch, we've merged everything to `rc/v1.4.0` since the release we've just rolled out (v1.3.42) was feature-frozen. 
TL;DR, the changes can be found on `rc/v1.4.0` -> the upcoming planned release
https://github.com/ElrondNetwork/elrond-go/blob/894a3d7c29c8b4f24030679ca46f5339466cadce/api/groups/addressGroup.go#L23

PR #4392, opened 2022-08-23
* Fixed set of indexes for processed txs in pending mini blocks, rece…

SebastianMarian:
- Fixed set of indexes for processed txs in pending mini blocks, received in epoch start meta block, when a node starts in epoch before mini block partial execution feature is activated
- 
- 
  
## Proposed Changes
- 
- 
- 

## Testing procedure
- 
- 
-

### DIFF
--- a/epochStart/bootstrap/baseStorageHandler.go
+++ b/epochStart/bootstrap/baseStorageHandler.go
@@ -18,7 +18,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/sharding/nodesCoordinator"
 )
 
-type miniBlockInfo struct {
+type miniBlocksInfo struct {
 	miniBlockHashes              [][]byte
 	fullyProcessed               []bool
 	indexOfLastTxProcessed       []int32
--- a/epochStart/bootstrap/shardStorageHandler.go
+++ b/epochStart/bootstrap/shardStorageHandler.go
@@ -569,8 +569,8 @@ func (ssh *shardStorageHandler) getProcessedAndPendingMiniBlocks(
 		return nil, nil, nil, err
 	}
 
-	mbInfo := getMiniBlocksInfo(epochShardData, neededMeta, ssh.shardCoordinator.SelfId())
-	processedMiniBlocks, pendingMiniBlocks := createProcessedAndPendingMiniBlocks(mbInfo, epochShardData)
+	mbsInfo := getMiniBlocksInfo(epochShardData, neededMeta, ssh.shardCoordinator.SelfId())
+	processedMiniBlocks, pendingMiniBlocks := createProcessedAndPendingMiniBlocks(mbsInfo, epochShardData)
 
 	return processedMiniBlocks, pendingMiniBlocks, epochShardData.GetFirstPendingMetaBlock(), nil
 }
@@ -604,8 +604,8 @@ func getEpochShardDataAndNeededMetaBlock(
 	return epochShardData, neededMeta, nil
 }
 
-func getMiniBlocksInfo(epochShardData data.EpochStartShardDataHandler, neededMeta *block.MetaBlock, shardID uint32) *miniBlockInfo {
-	mbInfo := &miniBlockInfo{
+func getMiniBlocksInfo(epochShardData data.EpochStartShardDataHandler, neededMeta *block.MetaBlock, shardID uint32) *miniBlocksInfo {
+	mbsInfo := &miniBlocksInfo{
 		miniBlockHashes:              make([][]byte, 0),
 		fullyProcessed:               make([]bool, 0),
 		indexOfLastTxProcessed:       make([]int32, 0),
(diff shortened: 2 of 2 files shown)


epochStart/bootstrap/baseStorageHandler.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/sharding/nodesCoordinator"
 )
 
-type miniBlockInfo struct {
+type miniBlocksInfo struct {
```
iulianpascalau, line 21: :+1:

epochStart/bootstrap/shardStorageHandler.go
```diff
@@ -604,75 +604,76 @@ func getEpochShardDataAndNeededMetaBlock(
...
-		mbInfo.pendingMiniBlocksMap[string(mbHeader.GetHash())] = struct{}{}
-
-		if mbHeader.GetIndexOfLastTxProcessed() > -1 {
-			mbInfo.miniBlockHashes = append(mbInfo.miniBlockHashes, mbHeader.GetHash())
-			mbInfo.fullyProcessed = append(mbInfo.fullyProcessed, false)
-			mbInfo.indexOfLastTxProcessed = append(mbInfo.indexOfLastTxProcessed, mbHeader.GetIndexOfLastTxProcessed())
+		mbsInfo.pendingMiniBlocksPerShardMap[receiverShardID] = append(mbsInfo.pendingMiniBlocksPerShardMap[receiverShardID], mbHeader.GetHash())
+		mbsInfo.pendingMiniBlocksMap[string(mbHeader.GetHash())] = struct{}{}
+
+		isPendingMiniBlockPartiallyExecuted := mbHeader.GetIndexOfLastTxProcessed() > -1 && mbHeader.GetIndexOfLastTxProcessed() < int32(mbHeader.GetTxCount())-1
```
iulianpascalau, line 636: so this composed condition will prevent the current edge case. Can we still have the scenario in which both fields are 0 (due to the unmarshaling of the empty slice) in which the if branch will still be executed?

iulianpascalau, line 636: an unit test for the exact encountered scenario would have been great

SebastianMarian: No, because if reserved will be nil, no matter of reason, the first line in the method returns TxCount-1

```
// GetIndexOfLastTxProcessed returns index of the last transaction processed in the miniBlock
func (m *MiniBlockHeader) GetIndexOfLastTxProcessed() int32 {
	miniBlockHeaderReserved, err := m.getMiniBlockHeaderReserved()
	if err != nil || miniBlockHeaderReserved == nil {
		return int32(m.TxCount) - 1
	}
```

SebastianMarian: And here we check only with GetIndexOfLastTxProcessed twice and not with both fields. GetIndexOfFirstTxProcessed does not matter when we try to determine if a block is partially executed. Only the index of the last tx executed will say if a miniblock is not yet fully executed.

SebastianMarian: A mini block referenced in a shard header with first index executed set to 0 or 5 could be partially executed but also fully executed, depending of the value of the last index executed.

SebastianMarian: Done

iulianpascalau: there is a difference between reserved field nil or empty. Should we change the condition to:
```GO
if err != nil || len(miniBlockHeaderReserved) == 0 {
```
The marhaller will output an empty string for a struct with default values as can be seen in this test:
```GO
func Test2(t *testing.T){
    m := &marshal.GogoProtoMarshalizer{}

    b := &block.MiniBlockReserved{}

    buff, _ := m.Marshal(b)

    fmt.Println(buff)
}
```

raduchis: could be changed to len() to accommodate both the nil and empty

raduchis (approved):

epochStart/bootstrap/shardStorageHandler.go
SebastianMarian: Actually miniBlockHeaderReserved could be only nil or NOT empty if you check the method m.getMiniBlockHeaderReserved() . There is no marshal but only unmarshal inside the method, so is not the case you underscored above. Anyways if we want this change, a new PR in elrond-go-core should be done. And also this change should be tested very well, as it will affect many parts of the code where the method is called.

```
// GetMiniBlockHeaderReserved returns the MiniBlockHeader Reserved field as a MiniBlockHeaderReserved
func (m *MiniBlockHeader) getMiniBlockHeaderReserved() (*MiniBlockHeaderReserved, error) {
	if len(m.Reserved) > 0 {
		mbhr := &MiniBlockHeaderReserved{}
		err := mbhr.Unmarshal(m.Reserved)
		if err != nil {
			return nil, err
		}

		return mbhr, nil
	}
	return nil, nil
}
```

iulianpascalau: return nil, nil. That is another red flag

SebastianMarian: This is for backwards compatibility as reserved was not used before scheduled miniblocks and here we can not pass the epoch notifier

iulianpascalau: still not agree. It can easily raise panics as we got used to use the code

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #4393, opened 2022-08-23
remove destroy persisters calls

bogdan-rosianu:
- if a persister could have not been created (for example, when extra tweaking the configuration and the system's ulimit was too low), all the opened DBs would have been destroyed, which should only be valid if the databases are created (not opened)
  
## Proposed Changes
- remove the code block that called `DestroyUnit()`

## Testing procedure
- specific scenario already tested and validated
- these changes don't affect other storage operations


andreibancioiu (approved):

AdoAdoAdo (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #4394, opened 2022-08-24
Multikey peer authentication sender integration

sstanculeanu:
- peer authentication sender is not compatible with the multikey system
  
## Proposed Changes
- integrate keysHolder into newly added component multikeyPeerAuthenticationSender in order to support multikey
- updated keysHolder with the needed methods
- factory which will create the proper one(multikeyPeerAuthenticationSender or peerAuthenticationSender) will come in a further PR

## Testing procedure
- system test hbv2


heartbeat/sender/multikeyPeerAuthenticationSender.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	msg := &heartbeat.PeerAuthentication{
+		Pid:    pid.Bytes(),
+		Pubkey: pkBytes,
+	}
+
+	hardforkPayload, isTriggered := sender.getHardforkPayload()
```
iulianpascalau, line 217: this function contains a lot of duplicated code with the original peerAuthenticationSender implementation. Can we extract duplicated code as seen in the heartbeat sender future implementation?

heartbeat/sender/multikeyPeerAuthenticationSender_test.go
```diff
@@ -0,0 +1,727 @@
...
+			buffResulted = append(buffResulted, buff)
+			pids = append(pids, pid)
+			skBytesBroadcast = append(skBytesBroadcast, skBytes)
+			mutData.Unlock()
+		}
+
+		senderInstance, _ := newMultikeyPeerAuthenticationSender(args)
+		senderInstance.Execute()
+
+		time.Sleep(time.Second * 2) // allow the go routines to finish
```
iulianpascalau, line 351: if we remove the go routines all these waits should be removed.

heartbeat/interface.go
```diff
@@ -16,8 +16,10 @@ import (
 // P2PMessenger defines a subset of the p2p.Messenger interface
 type P2PMessenger interface {
 	Broadcast(topic string, buff []byte)
+	BroadcastWithSk(topic string, buff []byte, pid core.PeerID, skBytes []byte)
```
ccorcoveanu, line 19: Can we raname this into `BroadcastUsingSk` (or `BroadcastSigningWithSk`)? I know that the naming convention follows libp2p's `PublishWithSk` but I find it very misleading, I really thought we were sending the sk over the wire. (`SendableData` having an SK member doesn't help either :)) @iulianpascalau thoughts?)

heartbeat/interface.go
```diff
@@ -16,8 +16,10 @@ import (
...
 	ID() core.PeerID
 	Sign(payload []byte) ([]byte, error)
+	SignWithPrivateKey(skBytes []byte, payload []byte) ([]byte, error)
```
ccorcoveanu, line 22: Maybe also follow the naming convention above (either abbreviation of secret key, or private key)

heartbeat/interface.go
iulianpascalau: right, let's rename to `BroadcastWithPrivateKey`

sstanculeanu: pushed

iulianpascalau (dismissed):

ccorcoveanu (approved):

ccorcoveanu (approved):

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #4400, opened 2022-08-26
add guardian data REST api

AdoAdoAdo:
This PR adds a rest API for getting the guardian data of an address
  
## Proposed Changes
The guardianData returned on the API will give several useful information such as
 - the active guardian (bech32 if any) and its activation epoch
 - the pending guardian (bech32 if any) and its activation epoch
- frozen state of the account - bool

## Testing procedure
- add guardian on an account and use the API to check the guardian data (one pending guardian)
- wait the expected time for the guardian to become active, then use the API to check the guardian data is as expected (one active guardian)
- add another guardian and use the API to check the guardian data (one active one pending)
- set account in frozen state then check the guardian data (one active guardian, the pending was cleared by freeze account action, and account is in frozen state)
- set another guardian (through an ungarded tx) and check the guardian data (one active, one pending, account frozen)

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go-core/data/api"
 	"github.com/ElrondNetwork/elrond-go-core/data/endProcess"
 	"github.com/ElrondNetwork/elrond-go-core/data/esdt"
+	"github.com/ElrondNetwork/elrond-go-core/data/guardians"
 	"github.com/ElrondNetwork/elrond-go-core/data/transaction"
 	disabledSig "github.com/ElrondNetwork/elrond-go-crypto/signing/disabled/singlesig"
 	logger "github.com/ElrondNetwork/elrond-go-logger"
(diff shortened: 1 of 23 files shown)


node/node.go
```diff
@@ -318,6 +319,51 @@ func (n *Node) GetValueForKey(address string, key string, options api.AccountQue
 	return hex.EncodeToString(valueBytes), blockInfo, nil
 }
 
+// GetGuardianData returns the guardian data for given account
+func (n *Node) GetGuardianData(address string, options api.AccountQueryOptions) (api.GuardianData, api.BlockInfo, error) {
```
bogdan-rosianu, line 323: can we add a unit test for this as well?

process/guardian/guardedAccount.go
```diff
@@ -287,6 +287,19 @@ func (agc *guardedAccount) getActiveGuardian(gs *guardians.Guardians) (*guardian
...
 }
 
+// GetConfiguredGuardians returns the configured guardians for an account
+func (agc *guardedAccount) GetConfiguredGuardians(uah state.UserAccountHandler) (active *guardians.Guardian, pending *guardians.Guardian, err error) {
+	configuredGuardians, err := agc.getConfiguredGuardians(uah)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	active, _ = agc.getActiveGuardian(configuredGuardians)
```
bogdan-rosianu, line 297: is it ok to ignore the error?

bogdan-rosianu (commented):
good implementation, small stuff only

node/node.go
AdoAdoAdo: added

process/guardian/guardedAccount.go
AdoAdoAdo: This is a tough question
Is not having an active guardian an error?
Or not having a pending guardian an error?
or not having both?

Here I didn't consider the above errors, as these are configurations so preferences shouldn't be considered errors.
I am only returning an error in case the saved data is wrong.

node/node_test.go
```diff
@@ -4048,6 +4051,308 @@ func TestNode_setTxGuardianData(t *testing.T) {
...
+			ActiveGuardian:  apiG1,
+			PendingGuardian: apiG2,
+			Frozen:          true,
+		}, guardianData)
+		require.Equal(t, api.BlockInfo{}, blockInfo)
+		require.Nil(t, err)
+	})
+}
+
+func TestNode_getPendingAndActiveGuardians(t *testing.T) {
```
iulianpascalau, line 4248: kind of duplicated tests. Can we keep only the first set of tests? TestNode_GetGuardianData

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4404, opened 2022-08-29
Prevent parallel snapshots

BeniaminDrasovean:
- If a snapshot takes more than one epoch, another snapshot will start in parallel. In extreme cases, this will lead to an IO bottleneck

  
## Proposed Changes
- Add `isSnapshotInProgress` flag in order to prevent this kind of behaviour, and skip a snapshot if another one is running.
- Remove duplicated code from AccountsDB and PeerAccountsDB

## Testing procedure
- Create another branch, and add a time.Sleep(1h) to simulate a snapshot that takes 1h. No other snapshot should start during this time.
- Another normal testing procedure to check that nothing has changed in normal cases.


state/accountsDB.go
```diff
@@ -1089,53 +1105,89 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
+		trieStorageManager.TakeSnapshot(rootHash, rootHash, leavesChannel, errChan, stats, epoch)
+		adb.snapshotUserAccountDataTrie(true, rootHash, leavesChannel, errChan, stats, epoch)
+		trieStorageManager.ExitPruningBufferingMode()
+
+		stats.wg.Done()
+	}()
+
+	go adb.markActiveDBAfterSnapshot(stats, errChan, rootHash, "snapshotState user trie", epoch)
+
+	adb.waitForCompletionIfRunningInImportDB(stats)
```
andreibancioiu, line 1128: Perhaps, in the future, we can also `waitForCompletionIfHeavilySyncing()` (when out of sync). From my understanding of the logs (if correct), that would greatly reduce the snapshot time. Does this make sense?

state/accountsDB.go
```diff
@@ -107,7 +109,19 @@ func NewAccountsDB(args ArgsAccountsDB) (*AccountsDB, error) {
...
+	trieStorageManager := adb.mainTrie.GetStorageManager()
+	val, err := trieStorageManager.GetFromCurrentEpoch([]byte(common.ActiveDBKey))
+	if err != nil || !bytes.Equal(val, []byte(common.ActiveDBVal)) {
+		startSnapshotAfterRestart(adb, args)
+	}
+
+	return adb, nil
+}
+
+func getAccountsDb(args ArgsAccountsDB) *AccountsDB {
```
iulianpascalau, line 123: I think createAccountsDb will be a better suitable name

state/accountsDB.go
```diff
@@ -1089,53 +1105,89 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
+	}()
+
+	go adb.markActiveDBAfterSnapshot(stats, errChan, rootHash, "snapshotState user trie", epoch)
+
+	adb.waitForCompletionIfRunningInImportDB(stats)
+}
+
+func (adb *AccountsDB) prepareSnapshot(rootHash []byte) (common.StorageManager, uint32, bool) {
+	trieStorageManager, epoch, err := adb.getTrieStorageManagerAndLatestEpoch()
+	if err != nil {
```
andreibancioiu, line 1133: Perhaps also log the results of this call (e.g. the returned epoch?) - for debugging purposes.

state/accountsDB.go
```diff
@@ -1089,53 +1105,89 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
+		return false
+	}
+
+	return true
 }
 
 func (adb *AccountsDB) markActiveDBAfterSnapshot(stats *snapshotStatistics, errChan chan error, rootHash []byte, message string, epoch uint32) {
 	stats.PrintStats(message, rootHash)
 
+	defer adb.isSnapshotInProgress.Reset()
```
andreibancioiu, line 1189: So the function `markActiveDBAfterSnapshot()` is now more like a `doStuffOnSnapshotCompletion`.

state/accountsDB.go
```diff
@@ -1089,53 +1105,89 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
+	snapshotAlreadyTaken := bytes.Equal(adb.lastSnapshot.rootHash, rootHash) && adb.lastSnapshot.epoch == epoch
+	if snapshotAlreadyTaken {
+		return false
+	}
+
+	if adb.isSnapshotInProgress.IsSet() {
+		return false
+	}
+
+	if !trieStorageManager.ShouldTakeSnapshot() {
```
iulianpascalau, line 1179: return directly here? like 
```GO
return trieStorageManager.ShouldTakeSnapshot()
```

state/accountsDB.go
andreibancioiu: E.g. import db finishes snapshot in 2-3 minutes.

state/peerAccountsDB.go
```diff
@@ -60,44 +43,20 @@ func (adb *PeerAccountsDB) MarkSnapshotDone() {
...
-		log.Error("SnapshotState error", "err", err.Error())
-		return
-	}
+	adb.mutOp.Lock()
+	defer adb.mutOp.Unlock()
 
-	if !trieStorageManager.ShouldTakeSnapshot() {
-		log.Debug("skipping snapshot for rootHash", "hash", rootHash)
+	trieStorageManager, epoch, shouldTakeSnapshot := adb.prepareSnapshot(rootHash)
+	if !shouldTakeSnapshot {
```
andreibancioiu, line 52: Perhaps log a message for not taking the snapshot?

andreibancioiu (approved):

state/accountsDB.go
BeniaminDrasovean: You're right, changed

state/accountsDB.go
BeniaminDrasovean: Done.

state/accountsDB.go
```diff
@@ -107,7 +109,19 @@ func NewAccountsDB(args ArgsAccountsDB) (*AccountsDB, error) {
 		return nil, err
 	}
 
-	adb := &AccountsDB{
+	adb := getAccountsDb(args)
+
+	trieStorageManager := adb.mainTrie.GetStorageManager()
+	val, err := trieStorageManager.GetFromCurrentEpoch([]byte(common.ActiveDBKey))
+	if err != nil || !bytes.Equal(val, []byte(common.ActiveDBVal)) {
+		startSnapshotAfterRestart(adb, args)
```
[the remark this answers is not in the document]
BeniaminDrasovean: Extracted in a separate variable. 

`startSnapshotAfterRestart` is moved from the constructor in https://github.com/ElrondNetwork/elrond-go/pull/4367

state/accountsDB.go
```diff
@@ -1089,53 +1105,89 @@ func (adb *AccountsDB) SnapshotState(rootHash []byte) {
...
+
+	go adb.markActiveDBAfterSnapshot(stats, errChan, rootHash, "snapshotState user trie", epoch)
+
+	adb.waitForCompletionIfRunningInImportDB(stats)
+}
+
+func (adb *AccountsDB) prepareSnapshot(rootHash []byte) (common.StorageManager, uint32, bool) {
+	trieStorageManager, epoch, err := adb.getTrieStorageManagerAndLatestEpoch()
+	if err != nil {
+		log.Error("snapshot user state error", "err", err.Error())
```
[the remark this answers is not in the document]
BeniaminDrasovean: There might be a chance that the next snapshot will succeed, so the node can continue processing. What do you say @iulianpascalau ?

state/accountsDB.go
BeniaminDrasovean: The snapshot process is started on a goroutine. During import-db, epochs are processed much faster than an actual epoch time, so there is a chance that an epoch is processed faster than it will take the snapshot to finish. `adb.waitForCompletionIfRunningInImportDB(stats)` is added to assure that the processing will wait for the snapshot to finish.

state/accountsDB.go
BeniaminDrasovean: Right. Renamed to `processSnapshotCompletion`. I look forward to a better naming idea :sweat_smile:

state/peerAccountsDB.go
BeniaminDrasovean: It is logged inside `prepareSnapshot`

state/accountsDB.go
```diff
@@ -1005,14 +1012,11 @@ func (adb *AccountsDB) RecreateAllTries(rootHash []byte) (map[string]common.Trie
 		return nil, err
 	}
 
-	recreatedTrie, err := adb.mainTrie.Recreate(rootHash)
+	allTries, err := adb.recreateMainTrie(rootHash)
```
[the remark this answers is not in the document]
BeniaminDrasovean: Not having any great idea

iulianpascalau (approved):

andreibancioiu (dismissed):

iulianpascalau (dismissed):

gabi-vuls (approved):
System test passed.

andreibancioiu (approved):

iulianpascalau (approved):

PR #4410, opened 2022-08-30
Fix heartbeat monitor to properly set peer type

sstanculeanu:
- heartbeat v2 only responds with validator/observer as peer type on `/node/heartbeatstatus`
  
## Proposed Changes
- integrate `peerTypeProvider` into heartbeat v2 monitor in order to properly compute the peer type

## Testing procedure
- system test + proxy call to `/node/heartbeatstatus` after epoch 3(when heartbeat v1 is disabled) in order to check if eligible/waiting is displayed for validators


iulianpascalau (approved):

schimih (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #4414, opened 2022-08-30
Peer authentication sender factory

sstanculeanu:
- 2 types of peer authentication available, one for multikey and the old one
  
## Proposed Changes
- add a factory which automatically creates the proper peer authentication based on the configuration

## Testing procedure
- system test


heartbeat/sender/peerAuthenticationSenderFactory.go
```diff
@@ -0,0 +1,129 @@
...
+		hardforkTimeBetweenSends: pasf.hardforkTimeBetweenSends,
+		hardforkTriggerPubKey:    pasf.hardforkTriggerPubKey,
+	}
+
+	return newPeerAuthenticationSender(args)
+}
+
+func (pasf *peerAuthenticationSenderFactory) createMultikeyPeerAuthenticationSender() (*multikeyPeerAuthenticationSender, error) {
+	args := argMultikeyPeerAuthenticationSender{
+		argBaseSender: argBaseSender{
```
bogdan-rosianu, line 110: could have extracted a function `createBaseSenderArgs` and use it in the both functions

heartbeat/sender/peerAuthenticationSenderFactory.go
```diff
@@ -0,0 +1,129 @@
...
+	hardforkTimeBetweenSends  time.Duration
+	hardforkTriggerPubKey     []byte
+	keysHolder                heartbeat.KeysHolder
+	timeBetweenChecks         time.Duration
+	shardCoordinator          process.ShardCoordinator
+	privKey                   crypto.PrivateKey
+	redundancyHandler         heartbeat.NodeRedundancyHandler
+}
+
+func newPeerAuthenticationSenderFactory(args argPeerAuthenticationSenderFactory) (*peerAuthenticationSenderFactory, error) {
```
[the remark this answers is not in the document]
sstanculeanu: updated

heartbeat/sender/peerAuthenticationSenderFactory.go
sstanculeanu: not needed anymore with latest push

heartbeat/sender/peerAuthenticationSenderFactory.go
```diff
@@ -25,105 +22,53 @@ type argPeerAuthenticationSenderFactory struct {
...
+	pkBytes, err := pk.ToByteArray()
+	if err != nil {
+		return nil, err
 	}
 
-	if !isMultikey && !isRegular {
-		return nil, fmt.Errorf("%w for peerAuthenticationSenderFactory, could not determine node's type", heartbeat.ErrInvalidConfiguration)
+	_, _, err = args.nodesCoordinator.GetValidatorWithPublicKey(pkBytes)
+	if err == nil {
+		return createRegularSender(args)
```
iulianpascalau, line 34: please add a check here for  `len(keysMap)` and allow to create the regular peer only if that len is 0. Otherwise we should output error

factory/heartbeatV2Components.go
```diff
@@ -142,6 +143,9 @@ func (hcf *heartbeatV2ComponentsFactory) Create() (*heartbeatV2Components, error
 		HardforkTrigger:                             hcf.processComponents.HardforkTrigger(),
 		HardforkTimeBetweenSends:                    time.Second * time.Duration(cfg.HardforkTimeBetweenSendsInSec),
 		HardforkTriggerPubKey:                       hcf.coreComponents.HardforkTriggerPubKey(),
+		KeysHolder:                                  &testscommon.KeysHolderStub{}, // TODO: use the real component once it is completely integrated
```
bogdan-rosianu, line 146: don't use mock / test packages in production code. Please create a `disabled` package somewhere, alongside with a `disabledKeysHolder` and use this instead.

bogdan-rosianu (changes requested):

factory/heartbeatV2Components.go
iulianpascalau: Agree, let's pass the real component here. Should be created & stored in crypto components

sstanculeanu: pushed the integration into crypto components

cmd/node/config/config.toml
```diff
@@ -486,6 +486,10 @@
     MaxRoundsToKeepUnprocessedMiniBlocks = 300   # max number of rounds unprocessed miniblocks are kept in pool
     MaxRoundsToKeepUnprocessedTransactions = 300 # max number of rounds unprocessed transactions are kept in pool
 
+[KeysHolderConfig]
+    IsMainMachine = false
```
bogdan-rosianu, line 490: why do we have a separate configuration? Couldn't this be determined based on the redundancy level?

factory/cryptoComponents.go
```diff
@@ -167,6 +174,19 @@ func (ccf *cryptoComponentsFactory) Create() (*cryptoComponents, error) {
 		return nil, err
 	}
 
+	blsKeyGenerator := signing.NewKeyGenerator(mcl.NewSuiteBLS12())
```
bogdan-rosianu, line 177: shouldn't it be created based on a config? similar to `factory/cryptoComponents.getSuite()`

cmd/node/config/config.toml
iulianpascalau: we have this in prefs.toml file

iulianpascalau: and the MaxRoundsWithoutReceivedMessages is actually the RedundancyLevel value

factory/cryptoComponents.go
iulianpascalau, line 177: instead of this, we should use the `blockSignKeyGen`

keysManagement/keysHolder.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/ElrondNetwork/elrond-go/config"
 )
 
-const minRoundsWithoutReceivedMessages = 1
+const minRoundsWithoutReceivedMessages = -1
```
iulianpascalau, line 18: we might refactor this in a future PR and re-check the logic.

iulianpascalau (approved):

bogdan-rosianu (approved):

PR #4420, opened 2022-08-31
Optimized peer authentication messages management

iulianpascalau:
- optimized peer authentication messages management by indexing the messages in the pool by their public key. In this way unnecessary get operations are not done on the resolvers.
- simplified interceptedHeartbeat.go
- fixed timeCacher implementation to use a `put` function instead of `upsert` as to properly override existing values with newer ones 
- removed unnecessary function from peerShardMapper that was previously used just in the peer authentication resolver that was simplified

## Proposed Changes
- code simplification as explained above

## Testing procedure
- standard testing procedure


sstanculeanu (dismissed):

storage/timecache/timeCacheCore.go
```diff
@@ -57,6 +57,49 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
 	return found, nil
 }
 
+// put will add the key, value and provided duration, overriding values if the data already existed
+// It returns if the value existed before this call. It also operates on the locker so the call is concurrent safe
```
mariusmihaic, line 61:
```suggestion
// It returns true if the value existed before this call. It also operates on the locker so the call is concurrent safe
```

storage/timecache/timeCacheCore.go
```diff
@@ -57,6 +57,49 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
...
+	tcc.data[key] = &entry{
+		timestamp: time.Now(),
+		span:      duration,
+		value:     value,
+	}
+	return found, nil
+}
+
+// hasOrAdd will add the key, value and provided duration, if the key is not found
+// It returns if the value existed before this call and if it has been added or not. It also operates on the locker so the call is concurrent safe
```
mariusmihaic, line 81:
```suggestion
// It returns true if the value existed before this call and if it has been added or not. It also operates on the locker so the call is concurrent safe
```

storage/timecache/timeCacheCore.go
```diff
@@ -57,6 +57,49 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
...
+func (tcc *timeCacheCore) hasOrAdd(key string, value interface{}, duration time.Duration) (bool, bool) {
+	if len(key) == 0 {
+		return false, false
```
mariusmihaic, line 84: [comment]

Could also add a log here, just like the one from `put` :

storage/timecache/timeCacheCore.go
iulianpascalau: done

storage/timecache/timeCacheCore.go
iulianpascalau: done

storage/timecache/timeCacheCore.go
iulianpascalau: refactored + added unit tests

storage/timecache/timeCacher.go
```diff
@@ -115,7 +115,15 @@ func (tc *timeCacher) Peek(key []byte) (value interface{}, ok bool) {
...
-	return tc.timeCache.hasOrAdd(string(key), value, tc.timeCache.defaultSpan)
+	var err error
+	has, added, err = tc.timeCache.hasOrAdd(string(key), value, tc.timeCache.defaultSpan)
+	if err != nil {
+		log.Error("mapTimeCacher.HasOrAdd", "error", key)
+
+		return
+	}
+
+	return
```
mariusmihaic, line 126: This func always returns nothing. Is this intended?

mariusmihaic: Also, could've used logIfError

storage/timecache/timeCacher.go
```diff
@@ -81,7 +81,7 @@ func (tc *timeCacher) Clear() {
 
 // Put adds a value to the cache. It will always return false since the eviction did not occur
 func (tc *timeCacher) Put(key []byte, value interface{}, _ int) (evicted bool) {
-	_, err := tc.timeCache.upsert(string(key), value, tc.timeCache.defaultSpan)
+	_, err := tc.timeCache.put(string(key), value, tc.timeCache.defaultSpan)
```
andreibancioiu, line 84: So all callers of `timeCacher.Put()` will receive this fix. Should be only positive impact, correct?

storage/timecache/timeCacher.go
```diff
@@ -115,14 +115,15 @@ func (tc *timeCacher) Peek(key []byte) (value interface{}, ok bool) {
 // HasOrAdd checks if a key is in the cache.
 // If key exists, does not update the value. Otherwise, adds the key-value in the cache
 func (tc *timeCacher) HasOrAdd(key []byte, value interface{}, _ int) (has, added bool) {
-	existed, err := tc.timeCache.upsert(string(key), value, tc.timeCache.defaultSpan)
+	var err error
+	has, added, err = tc.timeCache.hasOrAdd(string(key), value, tc.timeCache.defaultSpan)
```
andreibancioiu, line 119: So all callers of `timeCacher.HasOrAdd()` will receive this fix. Should be only positive impact, correct?

storage/timecache/timeCacheCore.go
```diff
@@ -58,7 +58,7 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
 }
 
 // put will add the key, value and provided duration, overriding values if the data already existed
-// It returns if the value existed before this call. It also operates on the locker so the call is concurrent safe
+// It returns true if the value existed before this call. It also operates on the locker so the call is concurrent safe
 func (tcc *timeCacheCore) put(key string, value interface{}, duration time.Duration) (bool, error) {
```
mariusmihaic, line 62: Why have this func return true if the value existed, since in `func (tc *timeCacher) Put` you always return false and ignore if this one returns true?

storage/timecache/timeCacheCore.go
```diff
@@ -57,6 +57,49 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
...
+	if len(key) == 0 {
+		return false, false, storage.ErrEmptyKey
+	}
+
+	tcc.Lock()
+	defer tcc.Unlock()
+
+	_, found := tcc.data[key]
+	if found {
+		return found, false, nil
```
andreibancioiu, line 92: Could have been `true` (for symmetry).

storage/timecache/timeCacher.go
andreibancioiu, line 119: `has` and `added` are mutually exclusive, right? Bit of redundancy therefore, but it's good to be explicit :+1:

heartbeat/sender/peerAuthenticationSender.go
```diff
@@ -170,6 +170,9 @@ func (sender *peerAuthenticationSender) execute() (error, bool) {
 		return err, isTriggered
 	}
 
+	log.Debug("sending peer authentication message",
```
SebastianMarian, line 173: Do we keep this on Debug level?

andreibancioiu (approved):

storage/timecache/timeCacher.go
iulianpascalau: please check that the function has named returned parameters. This is go idiomatic. log.LogIfError does not add sufficient context. 
Removed one return, thus

storage/timecache/timeCacher.go
iulianpascalau: right, this was the actual fix, a complete rewrite of the existing data

storage/timecache/timeCacheCore.go
iulianpascalau: removed & optimized

storage/timecache/timeCacher.go
iulianpascalau: right

storage/timecache/timeCacheCore.go
iulianpascalau: done

storage/timecache/timeCacher.go
iulianpascalau: yes, they are exclusive. Since I have to implement the already existing and used interface, I had to return something.
Also, beside (true, false) and (false, true) we also have (false, false) when the provided key is nil or empty, so we had to encode 3 states.

storage/timecache/timeCacheCore.go
andreibancioiu: `timeCacheCore` should normally not care about the existence, nor the logic of its caller (`timeCacher`). Therefore, it's alright to provide the information to the caller - even if that piece of information isn't used.

andreibancioiu (dismissed):

mariusmihaic (approved):

storage/timecache/timeCacheCore.go
SebastianMarian: `return true, false, nil` or at L100 `return found, true, nil` just to keep the same logic ?

storage/timecache/timeCacheCore.go
```diff
@@ -57,6 +57,47 @@ func (tcc *timeCacheCore) upsert(key string, value interface{}, duration time.Du
 	return found, nil
 }
 
+// put will add the key, value and provided duration, overriding values if the data already existed
+// It returns true if the value existed before this call. It also operates on the locker so the call is concurrent safe
```
sstanculeanu, line 61:
```suggestion
```

heartbeat/sender/peerAuthenticationSender.go
iulianpascalau: yes, one message at ~2hours. Helps the debugging a lot.

storage/timecache/timeCacheCore.go
iulianpascalau: already done in last commit.

storage/timecache/timeCacheCore.go
iulianpascalau: removed part of the comment

sstanculeanu (approved):

andreibancioiu (approved):

mariusmihaic (approved):

SebastianMarian (dismissed):

mariusmihaic (approved):

sstanculeanu (approved):

andreibancioiu (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #4421, opened 2022-08-31
Send public key on heartbeatv2 message

sstanculeanu:
- missing public key for observers on `/node/heartbeatstatus`
  
## Proposed Changes
- send public key on heartbeat v2 message

## Testing procedure
- system test:
- scenario: 
   - restart a validator from a testnet running `rc/2022-july`(make sure epoch is >= 3)
   - save its response for `/node/heartbeatstatus`. At this point, the public key of the observers in the same shard with this validator should be empty
   - after starting a testnet with this branch, **after epoch 3**, restart one of the validators to clean its cache and get another response from it
   - then compare the responses. With this branch only, the public keys for the observers in the same shard with the validator should be filled


iulianpascalau (approved):

schimih (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #4422, opened 2022-08-31
fix meta genesis block

miiu96:
- Fix the problem with the hash of the `metachain` genesis block.
  
## Proposed Changes
- Changed the order when the `validatorStatsRootHash` is set on the `metchain` genesis block.

## Testing procedure
- Fetch from `gateway` or `Elasticsearch` the genesis block ( block with nonce 0 from metachain ) and check if the block's hash is equal with the `previousBlockHash` of the block with the nonce 1.


AdoAdoAdo (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #4424, opened 2022-08-31
Fix duplicated txs into tx pool api response

sstanculeanu:
- same tx is displayed multiple times in the response of `/transaction/pool?by-sender=erd1...` as `ShardDataStore` returns the same data storer for all ids "0", "0_1", "0_2", "0_META"
  
## Proposed Changes
- removed the iteration through all data stores with ids `sourceShard_destinationShard` as they all point to the same one, leading to duplicates

## Testing procedure
- scenario:
   - get one or 2 txs in pool(save the sender if possible..if not, get `/transaction/pool` on the observer from the shard you sent txsf and copy a sender address)
   - running `/transaction/pool?by-sender=sender_addr_from_above_point` on either proxy or intra shard observer, tyou should not the duplicates among tx hashes received in response

### DIFF
--- a/node/external/transactionAPI/apiTransactionProcessor.go
+++ b/node/external/transactionAPI/apiTransactionProcessor.go
@@ -21,7 +21,6 @@ import (
 	"github.com/ElrondNetwork/elrond-go/process"
 	"github.com/ElrondNetwork/elrond-go/process/txstatus"
 	"github.com/ElrondNetwork/elrond-go/sharding"
-	"github.com/ElrondNetwork/elrond-go/storage"
 	"github.com/ElrondNetwork/elrond-go/storage/txcache"
 )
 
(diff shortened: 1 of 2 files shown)


node/external/transactionAPI/apiTransactionProcessor.go
```diff
@@ -351,35 +350,17 @@ func (atp *apiTransactionProcessor) extractRequestedTxInfo(wrappedTx *txcache.Wr
...
-	cachers := atp.getDataStoresForSender(senderShard)
-	for _, cache := range cachers {
-		txCache, ok := cache.(*txcache.TxCache)
-		if !ok {
-			continue
-		}
-
-		txs := txCache.GetTransactionsPoolForSender(sender)
-		txsForSender = append(txsForSender, txs...)
+	cacheId := process.ShardCacherIdentifier(senderShard, senderShard)
```
iulianpascalau, line 354: so this will work despite the fact that a transaction might be cross shard? On both nodes: sender & destination?

iulianpascalau (approved):

node/external/transactionAPI/apiTransactionProcessor.go
sstanculeanu: yes, as verified with @andreibancioiu and can be seen [here](https://github.com/ElrondNetwork/elrond-go/blob/8e402fa6d7e91e779980122d3798b2bf50892945/dataRetriever/txpool/shardedTxPool.go#L361), no matter the destination shard, there will always be returned the source id.

the problem was observed on a transaction from shard 0 to shard 1, when the same transaction was returned 4 times from a shard 0 observer.

also, @gabi-vuls confirmed that testing was made with cross shard txs.

PR #4427, opened 2022-08-31
Added and integrated peerAuthenticationPayloadValidator component

iulianpascalau:
- sending one invalid peerAuthenticationMessage in a batch will make that entire batch being discarded on the destination node. Since this is a security feature on a node (not to consider a batch containing wrong data just to discourage nodes acting maliciously) might prevent large packets of these type of messages being discarded just because one validator gets offline and making its last message to expire on all nodes that still contained that message in their pools. 
  
## Proposed Changes
- to prevent this happening, 2 solutions are available: make the node accepts large batches of messages even if some are invalid (defeats the purpose of the aforementioned security feature) or, second, make the nodes that respond to requests to re-check their stored messages before sending them. In order to avoid duplicated code (in the interceptor and in the resolver) the implementation was made by having a common component that validates the expiration timestamp of a payload. 

## Testing procedure
- standard testing procedure
- run a system testnet with no transactions and with the log level `*:DEBUG,p2p:TRACE,intercept:TRACE,etriever:TRACE,consensus:TRACE` and turn off one validator. After ~10minutes restart a few other nodes or let the testnet change a few epochs. A log print like `found peer authentication payload with invalid value, will not send it` should appear on the nodes that served requests. The testnet should run smoothly.


AdoAdoAdo (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4435, opened 2022-09-02
Adapt components for side chain

SebastianMarian:

* Adapted BlockTracker and ForkDetector component for side chain approach
- 
- 
  
## Proposed Changes
- 
- 
- 

## Testing procedure
- 
- 
-

### DIFF
--- a/process/sync/metaForkDetector.go
+++ b/process/sync/metaForkDetector.go
@@ -66,6 +66,8 @@ func NewMetaForkDetector(
 		baseForkDetector: bfd,
 	}
 
+	mfd.doJobOnBHProcessedFunc = mfd.doJobOnBHProcessed
+
 	bfd.forkDetector = &mfd
 
 	return &mfd, nil
@@ -85,7 +87,7 @@ func (mfd *metaForkDetector) AddHeader(
 		state,
 		selfNotarizedHeaders,
 		selfNotarizedHeadersHashes,
-		mfd.doJobOnBHProcessed,
+		mfd.doJobOnBHProcessedFunc,
 	)
 }
 
--- a/process/sync/metaForkDetector.go
+++ b/process/sync/metaForkDetector.go
@@ -15,7 +15,7 @@ var _ process.ForkDetector = (*metaForkDetector)(nil)
 // metaForkDetector implements the meta fork detector mechanism
 type metaForkDetector struct {
 	*baseForkDetector
-	doJobOnBHProcessedFunc func(_ data.HeaderHandler, _ []byte, _ []data.HeaderHandler, _ [][]byte)
+	doJobOnBHProcessedFunc func(header data.HeaderHandler, headerHash []byte, selfNotarizedHeaders []data.HeaderHandler, selfNotarizedHeadersHashes [][]byte)
 }
 
 // NewMetaForkDetector method creates a new metaForkDetector object
--- a/process/sync/metaForkDetector.go
+++ b/process/sync/metaForkDetector.go
@@ -15,6 +15,7 @@ var _ process.ForkDetector = (*metaForkDetector)(nil)
 // metaForkDetector implements the meta fork detector mechanism
 type metaForkDetector struct {
 	*baseForkDetector
+	doJobOnBHProcessedFunc func(_ data.HeaderHandler, _ []byte, _ []data.HeaderHandler, _ [][]byte)
 }
 
 // NewMetaForkDetector method creates a new metaForkDetector object
--- a/process/sync/shardForkDetector.go
+++ b/process/sync/shardForkDetector.go
@@ -68,6 +68,7 @@ func NewShardForkDetector(
 	}
 
 	sfd.blockTracker.RegisterSelfNotarizedFromCrossHeadersHandler(sfd.ReceivedSelfNotarizedFromCrossHeaders)
+	sfd.doJobOnBHProcessedFunc = sfd.doJobOnBHProcessed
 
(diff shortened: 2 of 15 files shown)


sasurobert (dismissed):

process/sync/sideChainShardForkDetector.go
```diff
@@ -0,0 +1,32 @@
...
+	if shardForkDetector == nil {
+		return nil, process.ErrNilForkDetector
+	}
+
+	return &sideChainShardForkDetector{
+		shardForkDetector,
+	}, nil
+}
+
+func (scsfd *sideChainShardForkDetector) doJobOnBHProcessed(
```
iulianpascalau, line 23: this does not work as expected

process/track/sideChainBlockProcessor.go
```diff
@@ -0,0 +1,33 @@
...
+	if blockProcessor == nil {
+		return nil, process.ErrNilBlockProcessor
+	}
+
+	return &sideChainBlockProcessor{
+		blockProcessor,
+	}, nil
+}
+
+func (scbp *sideChainBlockProcessor) shouldProcessReceivedHeader(headerHandler data.HeaderHandler) bool {
```
iulianpascalau, line 23: also here

process/track/sideChainShardBlockTrack.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	sideChainBlockTrack.blockProcessor = newBlockProcessor
+	return sideChainBlockTrack, nil
+}
+
+// ComputeLongestSelfChain computes the longest chain from self shard
+func (scsbt *sideChainShardBlockTrack) ComputeLongestSelfChain() (data.HeaderHandler, []byte, []data.HeaderHandler, [][]byte) {
```
iulianpascalau, line 37: this will work as expected if it is only called from exterior. Also the next one

iulianpascalau (changes requested):

process/sync/sideChainShardForkDetector.go
SebastianMarian: Fixed

process/track/sideChainBlockProcessor.go
SebastianMarian: Fixed

process/track/sideChainShardBlockTrack.go
SebastianMarian: Fixed

process/sync/metaForkDetector.go
```diff
@@ -15,6 +15,7 @@ var _ process.ForkDetector = (*metaForkDetector)(nil)
 // metaForkDetector implements the meta fork detector mechanism
 type metaForkDetector struct {
 	*baseForkDetector
+	doJobOnBHProcessedFunc func(_ data.HeaderHandler, _ []byte, _ []data.HeaderHandler, _ [][]byte)
```
iulianpascalau, line 18: could have kept the parameters names as it provides better context

sasurobert (dismissed):

process/sync/shardForkDetector.go
```diff
@@ -16,6 +16,7 @@ var _ process.ForkDetector = (*shardForkDetector)(nil)
 // shardForkDetector implements the shard fork detector mechanism
 type shardForkDetector struct {
 	*baseForkDetector
+	doJobOnBHProcessedFunc func(_ data.HeaderHandler, _ []byte, _ []data.HeaderHandler, _ [][]byte)
```
iulianpascalau, line 19: also here, could have kept the parameter names

process/track/baseBlockTrack.go
```diff
@@ -51,6 +51,8 @@ type baseBlockTrack struct {
 	mutHeaders                  sync.RWMutex
 	headers                     map[uint32]map[uint64][]*HeaderInfo
 	maxNumHeadersToKeepPerShard int
+	ComputeLongestSelfChainFunc func() (data.HeaderHandler, []byte, []data.HeaderHandler, [][]byte)
```
iulianpascalau, line 54: Let's not use these exported function pointers. Reason: I can take the component, set this function pointer to nil and then the component will break in panic

process/track/blockProcessor.go
```diff
@@ -25,7 +25,8 @@ type blockProcessor struct {
 	finalMetachainHeadersNotifier         blockNotifierHandler
 	roundHandler                          process.RoundHandler
 
-	blockFinality uint64
+	blockFinality                   uint64
+	shouldProcessReceivedHeaderFunc func(_ data.HeaderHandler) bool
```
iulianpascalau, line 29: Also here, could have kept the original param names

process/track/baseBlockTrack.go
SebastianMarian: Removed, as it is used only from outside of the component

process/track/blockProcessor.go
SebastianMarian: Removed, as it is used only from outside of the component and is not necessary to be overwritten

SebastianMarian: Done

process/sync/shardForkDetector.go
SebastianMarian: Done

iulianpascalau (approved):

sasurobert (approved):

PR #4437, opened 2022-09-02
Fix heartbeat v2 metrics update

sstanculeanu:
- connections metrics are not updated on heartbeat v2 monitor
- node type metrics are not updated on heartbeat v2 sender
  
## Proposed Changes
- add metrics update

## Testing procedure
- system test
- scenario: after epoch 3, check metrics on `erd_live_validator_nodes` and `erd_connected_nodes`, `erd_node_type`, `erd_peer_type`, `erd_peer_subtype`. They should display the actual values, not 0.


heartbeat/monitor/monitor.go
```diff
@@ -213,6 +223,29 @@ func (monitor *heartbeatV2Monitor) shouldSkipMessage(messageAge time.Duration, p
 	return false
 }
 
+func (monitor *heartbeatV2Monitor) updateMetrics(_ []byte, _ interface{}) {
+	heartbeats := monitor.GetHeartbeats()
+
+	counterActiveValidators := 0
+	counterConnectedNodes := 0
+	for _, heartbeatMessage := range heartbeats {
```
iulianpascalau, line 235: this is so sub-optimal, do not have a better idea, this is how the v1 was implemented.

iulianpascalau: Wondering if we can have a dedicated go routine that will call this each 30 seconds or so.

sstanculeanu: indeed this would be lighter..pushed

factory/heartbeatV2Components.go
```diff
@@ -230,6 +233,10 @@ func (hc *heartbeatV2Components) Close() error {
 		log.LogIfError(hc.directConnectionsProcessor.Close())
 	}
 
+	if !check.IfNil(hc.monitor) {
+		log.LogIfError(hc.monitor.Close())
```
iulianpascalau, line 237: :+1:

testscommon/cacherMock.go
```diff
@@ -102,6 +102,9 @@ func (cacher *CacherMock) Remove(key []byte) {
 
 // Keys -
 func (cacher *CacherMock) Keys() [][]byte {
+	cacher.mut.RLock()
```
iulianpascalau, line 105: :+1:

iulianpascalau (approved):

PR #4440, opened 2022-09-03
Add new flags (useful for some observers): "serialize-snapshots", "disable-consensus-watchdog"

andreibancioiu:
 - During import-db, _snapshotting procedures_ and _processing procedures_ are serialized, which leads to **fast** snapshotting.
 - When an observer is fully synchronized and in it's in the _normal_ processing mode, snapshotting is **fast**, as well (at the beginning of a new epoch).
 - Syncing a large number of epochs for a full history devnet observer leads to snapshots that are concurrent with the syncing procedures. In this case (syncing from the past, plus concurrent snapshots), snapshotting happens **slowly** (which can often be problematic).
  
## Proposed Changes
 - Add new CLI flag: `--serialize-snapshots` - this will force the state snapshotting to be serialized with processing.
 - Add new CLI flag: `--disable-consensus-watchdog` - necessary to allow snapshotting that takes more than 10 rounds (which is always the time), in conjunction with `--serialize-snapshots`.

## Testing procedure
 - Synchronize a full history observer on devnet (e.g. shard 0), which starts with a large number of epochs (e.g. 20, 30) behind blockchain tip. Look for snapshotting logs. No processing happens while snapshotting, no concurrent snapshots take place. Tested in rosetta scenarios :heavy_check_mark: 
 - Regular internal testnet, but having observers started with the flags: `--serialize-snapshots --disable-consensus-watchdog`. When they will perform the snapshots, they will be out of sync for a few minutes (that is expected).

### DIFF
--- a/cmd/node/flags.go
+++ b/cmd/node/flags.go
@@ -327,6 +327,11 @@ var (
 		Name:  "force-start-from-network",
 		Usage: "Flag that will force the start from network bootstrap process",
 	}
+	// disableConsensusWatchdog defines a flag that will disable the consensus watchdog
+	disableConsensusWatchdog = cli.BoolFlag{
+		Name:  "disable-consensus-watchdog",
+		Usage: "Flag that will disable the consensus watchdog",
+	}
 )
 
 func getFlags() []cli.Flag {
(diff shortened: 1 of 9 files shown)


iulianpascalau (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #4450, opened 2022-09-09
Fix missing epoch on peer authentication messages requests

sstanculeanu:
- epoch is missing from `RequestData` object, while requesting peer authentication messages, leading to always requesting from epoch 0, which sometimes therefore leads to querying nodes from full history list, which may be empty
  
## Proposed Changes
- add epoch on `RequestData` object

## Testing procedure
- standard system test, looking for `peer list is empty or errors during the sending, topic: peerAuthentication` **not to be present**


iulianpascalau (approved):

andreibancioiu (approved):

gabi-vuls (approved):
System test passed.

PR #4459, opened 2022-09-13
State snapshot improvement

BeniaminDrasovean:
- There are some edgecases in which a certain trie node is not saved in the proper storer. The problem is that we will have storers that do not have the complete state.
  
## Proposed Changes
- During the snapshot process, if a trie node is not present in the previous storer (which should have had the entire state), it will also be saved there.

## Testing procedure
- Normal test, but with "*:DEBUG,trie:TRACE" log level, and we will see in the logs that the missing nodes are also saved in the  previous storer.

### DIFF
--- a/trie/snapshotTrieStorageManager.go
+++ b/trie/snapshotTrieStorageManager.go
@@ -67,6 +67,7 @@ func (stsm *snapshotTrieStorageManager) putInPreviousStorerIfAbsent(key []byte, 
 		return
 	}
 
+	log.Trace("put missing hash in snapshot storer", "hash", key, "epoch", stsm.epoch-1)
 	err := stsm.mainSnapshotStorer.PutInEpoch(key, val, stsm.epoch-1)
 	if err != nil {
 		log.Warn("can not put in epoch",
--- a/trie/snapshotTrieStorageManager.go
+++ b/trie/snapshotTrieStorageManager.go
@@ -55,6 +55,10 @@ func (stsm *snapshotTrieStorageManager) putInPreviousStorerIfAbsent(key []byte, 
 		return
 	}
 
+	if stsm.epoch == 0 {
+		return
+	}
+
 	if epoch.Value >= stsm.epoch-1 {
 		return
 	}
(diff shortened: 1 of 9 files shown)


trie/snapshotTrieStorageManager.go
```diff
@@ -35,17 +38,40 @@ func (stsm *snapshotTrieStorageManager) Get(key []byte) ([]byte, error) {
 		return nil, errors.ErrContextClosing
 	}
 
-	val, err := stsm.mainSnapshotStorer.GetFromOldEpochsWithoutAddingToCache(key)
+	val, epoch, err := stsm.mainSnapshotStorer.GetFromOldEpochsWithoutAddingToCache(key)
 	if errors.IsClosingError(err) {
 		return nil, err
 	}
 	if len(val) != 0 {
+		stsm.putInPreviousStorerIfAbsent(key, val, epoch)
```
iulianpascalau, line 46: putIfAbsent :D 
brings back so many memories

trie/snapshotTrieStorageManager.go
```diff
@@ -35,17 +38,40 @@ func (stsm *snapshotTrieStorageManager) Get(key []byte) ([]byte, error) {
...
 
 	return stsm.getFromOtherStorers(key)
 }
 
+func (stsm *snapshotTrieStorageManager) putInPreviousStorerIfAbsent(key []byte, val []byte, epoch core.OptionalUint32) {
+	if !epoch.HasValue {
+		return
+	}
+
+	if epoch.Value >= stsm.epoch-1 {
```
iulianpascalau, line 58: so basically, if we are in current epoch 500, epoch will be 499 and we test for epoch 498?
Is this >= correct?

BeniaminDrasovean: The idea is that if the value was not retrieved from the current storer or previous storer, it needs to be saved to the previous storer.

iulianpascalau (dismissed):

sstanculeanu (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

iulianpascalau (approved):

PR #4470, opened 2022-09-15
async callback arguments inconsitency fix

sasurobert:
Inconsistency  of arguments when creating asynchronous callback SCR from an output transfer

## Proposed Changes
Added missing argument. Only after activation of special flag (backward compatibility issues).

## Testing procedure
Deploy a contract, issue a token from that contract through asnycCall, check one of the resulting SCRs which is with type AsyncchronousCallback. It should start with "@00" after activation. Before activation this one is missing.


iulianpascalau (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #4477, opened 2022-09-16
Create static storer for meta if snapshots are disabled

BeniaminDrasovean:
- If snapshots are disabled, the trie should use a static storer instead of a pruning storer (that creates a db for each epoch). The meta is using a pruning storer even if snapshots are disabled.
  
## Proposed Changes
- Use static storer for meta if snapshots are disabled.

## Testing procedure
- Normal test
- Test with snapshots disabled

### DIFF
--- a/storage/factory/pruningStorerFactory.go
+++ b/storage/factory/pruningStorerFactory.go
@@ -328,14 +328,12 @@ func (psf *StorageServiceFactory) CreateForMeta() (dataRetriever.StorageService,
 		return nil, err
 	}
 
-	userAccountsUnitArgs := psf.createPruningStorerArgs(psf.generalConfig.AccountsTrieStorage, customDatabaseRemover)
-	userAccountsUnit, err = psf.createTriePruningPersister(userAccountsUnitArgs)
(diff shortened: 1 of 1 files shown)


storage/factory/pruningStorerFactory.go
```diff
@@ -328,14 +328,12 @@ func (psf *StorageServiceFactory) CreateForMeta() (dataRetriever.StorageService,
 		return nil, err
 	}
 
-	userAccountsUnitArgs := psf.createPruningStorerArgs(psf.generalConfig.AccountsTrieStorage, customDatabaseRemover)
-	userAccountsUnit, err = psf.createTriePruningPersister(userAccountsUnitArgs)
+	userAccountsUnit, err = psf.createTriePersister(psf.generalConfig.AccountsTrieStorage, psf.generalConfig.StateTriesConfig, customDatabaseRemover)
```
iulianpascalau, line 331: "the duplicated code is the root of all evil" - Uncle Bob

iulianpascalau (approved):

AdoAdoAdo (approved):

LucianMincu (approved):

PR #4513, opened 2022-09-26
properly treat commitSnapshot error

BeniaminDrasovean:
- If the node is closing during a trie snapshot, some nodes will not be retrieved from storage because the dbs are closed, thus automatically triggering the syncing process for those nodes.
  
## Proposed Changes
- If the reason for which the nodes can not be retrieved from storage is because the node is closing, do not try to sync those nodes.

## Testing procedure
- Normal testing procedure


SebastianMarian (approved):

andreibancioiu (approved):

gabi-vuls (approved):
System test passed.

PR #4515, opened 2022-09-26
Improve nil checks

bogdan-rosianu:
- There were some structs that extended a base structure but didn't overload the `IsInterfaceNil() bool` func, which could have caused issues when calling `check.IfNil()` with the inner structure being nil
- In pruning storer, the arguments were sent as pointer
  
## Proposed Changes
- Add `IsInterfaceNil() bool` to needed places
- Switch pruning storer arguments to be passed as a value

## Testing procedure
- regular testing scenario


iulianpascalau (commented):
Looks good, all implementations changed could have been tested in their unit tests like
```GO
assert.True(t, check.IfNil(instance))
```
rather than
```GO
assert.Nil(t, instance)
```

BeniaminDrasovean (dismissed):

iulianpascalau (approved):

BeniaminDrasovean (approved):

PR #4520, opened 2022-09-27
fix accounts trie remover config value

bogdan-rosianu:
- The config value `AccountsTrieCleanOldEpochsData` was not used 
  
## Proposed Changes
- Added a factory function that creates either a custom or a disabled db remover, based on the above config value

## Testing procedure
1. should remove
- set an observer's configuration: `AccountsTrieCleanOldEpochsData` to `true` and `AccountsTrieSkipRemovalCustomPattern` to `%2`. When looking in the `db` directory, for epoch older than current - 4 (so if in epoch 8, from epoch 0 to epoch 4)  databases that are not divisible by 2 should not have `AccountsTrie` db for that epoch. Therefore: Epoch_0: should have AccountsTrie, Epoch_1: should not have AccountsTrie, and so on
2. should not remove
- set an observer's configuration `AccountsTrieCleanOldEpochsData` to `false`. In epoch 8, when checking the database dir, no `AccountsTrie` should have been removed

### DIFF
--- a/storage/databaseremover/factory/customDatabaseRemoverCreator_test.go
+++ b/storage/databaseremover/factory/customDatabaseRemoverCreator_test.go
@@ -1,0 +1,41 @@
+package factory
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/ElrondNetwork/elrond-go/config"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateCustomDatabaseRemover(t *testing.T) {
+	t.Parallel()
+
+	t.Run("should create real custom database remover", func(t *testing.T) {
+		t.Parallel()
+
+		storagePruningArgs := config.StoragePruningConfig{
+			AccountsTrieCleanOldEpochsData:       true,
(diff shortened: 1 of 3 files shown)


storage/databaseremover/factory/customDatabaseRemoverCreator_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+		storagePruningArgs := config.StoragePruningConfig{
+			AccountsTrieCleanOldEpochsData:       true,
+			AccountsTrieSkipRemovalCustomPattern: "%1",
+		}
+
+		removerInstance, err := CreateCustomDatabaseRemover(storagePruningArgs)
+		require.NoError(t, err)
+
+		require.Equal(t, "*databaseremover.customDatabaseRemover", fmt.Sprintf("%T", removerInstance))
```
iulianpascalau, line 25: :+1:

iulianpascalau (approved):

andreibancioiu (approved):

gabi-vuls (approved):
System test passed.

PR #4534, opened 2022-09-29
Workaround: get logs on a best-effort basis (handle epoch-change edge-case)

andreibancioiu:
- In some rare cases, right at the beginning of an epoch, `logs & events` of a transaction can we mistakenly saved in the storage associated with the previous epoch.
  
## Proposed Changes
- In this PR, we attempt to load `logs & events` both from the requested epoch `N`, and from the epoch `N-1`.

## Testing procedure
- Regular testing
- `rosetta-cli` checker (with ESDT) against devnet (e.g. shard 0)

### DIFF
--- a/node/external/logs/logsRepository.go
+++ b/node/external/logs/logsRepository.go
@@ -28,7 +28,7 @@ func newLogsRepository(storageService dataRetriever.StorageService, marshaller m
 // It searches for logs both in epoch N and epoch N-1, in order to overcome epoch-changing edge-cases.
 func (repository *logsRepository) getLog(logKey []byte, epoch uint32) (*transaction.Log, error) {
 	txLog, err := repository.doGetLog(logKey, epoch)
-	if err != nil {
+	if err != nil && epoch > 0 {
 		txLog, errPreviousEpoch := repository.doGetLog(logKey, epoch-1)
 		if errPreviousEpoch == nil {
 			return txLog, nil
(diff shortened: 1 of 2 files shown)


iulianpascalau (approved):

node/external/logs/logsRepository.go
```diff
@@ -24,7 +24,23 @@ func newLogsRepository(storageService dataRetriever.StorageService, marshaller m
 	}
 }
 
+// getLog loads event & logs on a best-effort basis.
+// It searches for logs both in epoch N and epoch N-1, in order to overcome epoch-changing edge-cases.
 func (repository *logsRepository) getLog(logKey []byte, epoch uint32) (*transaction.Log, error) {
+	txLog, err := repository.doGetLog(logKey, epoch)
+	if err != nil {
+		txLog, errPreviousEpoch := repository.doGetLog(logKey, epoch-1)
```
bogdan-rosianu, line 32: might check that epoch is not 0 before subtracting 1 from it

andreibancioiu: Good catch :rocket: Fixed.

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed.

node/external/logs/logsRepository.go
```diff
@@ -28,7 +28,7 @@ func newLogsRepository(storageService dataRetriever.StorageService, marshaller m
 // It searches for logs both in epoch N and epoch N-1, in order to overcome epoch-changing edge-cases.
 func (repository *logsRepository) getLog(logKey []byte, epoch uint32) (*transaction.Log, error) {
 	txLog, err := repository.doGetLog(logKey, epoch)
-	if err != nil {
+	if err != nil && epoch > 0 {
```
iulianpascalau, line 31: :+1:

iulianpascalau (approved):

PR #4535, opened 2022-10-02
Remove unnecessary gasComputation.Reset call

iulianpascalau:
- gasComputation component started to consume large amounts of time when the calls to `SetGasProvided` were done on a several hundreds of transactions and up. The time consumed was caused that each transaction hash from the miniblock was added as record in the map that was latter used to append each executed hash, causing exponential time of execution consumed.
  
## Proposed Changes
- removed unnecessary gasComputation.Reset call from the basePreProcess.handleProcessTransactionInit call

## Testing procedure
- standard system tests
- half network tests, before and after the partial miniblock execution enable epoch


dragos-rebegea (approved):

AdoAdoAdo (approved):

sasurobert (approved):

bogdan-rosianu (approved):

SebastianMarian (approved):

gabi-vuls (approved):
System test passed.

PR #4536, opened 2022-10-02
[DO NOT USE/MERGE THIS]Always leader node - for debugging purposes

iulianpascalau:
- debugging only branch. 
WARNING: if used in production, this variant will most probably made its associated key be jailed. 
  
## Proposed Changes
- removed the p2p broadcast functionality
- made the consensus components "think" that the leader is always the provided public key
- made the block syncer "think" it is always synchronized

## Testing procedure
- get the DB/configs up until this node will need to propose the next block. Add breakpoints or prints wherever necessary to study how the processing is done from the proposer's perspective


sasurobert (changes requested):
Not

PR #4549, opened 2022-10-03
Fix old database cleaner

bogdan-rosianu:
- oldDatabaseRemover relied on the fact that all storers return the correct oldest epoch and ones that don't have an epoch approach, they would have returned an error. However, a disabled Storer has been introduced recently and returned `0` as that value, with no error
  
## Proposed Changes
- return `0` for the disabled storer

## Testing procedure
- db cleanup

### DIFF
--- a/storage/clean/oldDatabaseCleaner.go
+++ b/storage/clean/oldDatabaseCleaner.go
@@ -192,10 +192,6 @@ func (odc *oldDatabaseCleaner) cleanOldEpochs(currentEpoch uint32) error {
 		return nil
 	}
 
-	if len(sortedEpochs) == 0 {
-		return nil
-	}
-
 	for idx, epoch := range sortedEpochs {
 		if epoch >= epochToDeleteTo {
 			break
--- a/storage/clean/oldDatabaseCleaner.go
+++ b/storage/clean/oldDatabaseCleaner.go
@@ -101,7 +101,7 @@ func (odc *oldDatabaseCleaner) handleEpochChangeAction(epoch uint32) error {
 	odc.Unlock()
 
 	shouldClean := odc.shouldCleanOldData(epoch, newOldestEpoch)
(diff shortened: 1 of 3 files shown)


storage/clean/oldDatabaseCleaner.go
```diff
@@ -192,7 +192,9 @@ func (odc *oldDatabaseCleaner) cleanOldEpochs(currentEpoch uint32) error {
 		return nil
 	}
 
+	wasCleaned := false
```
SebastianMarian: You can replace this line with:
```
if len(sortedEpochs) == 0 {
    return nil
}
```
and remove L197, L210 and L213

storage/clean/oldDatabaseCleaner.go
```diff
@@ -192,7 +192,9 @@ func (odc *oldDatabaseCleaner) cleanOldEpochs(currentEpoch uint32) error {
...
 	for idx, epoch := range sortedEpochs {
+		wasCleaned = true
```
SebastianMarian: What I mentioned above will not apply if this line, actually should be moved after L207 when we are sure that something has been cleaned and only If err == nil. With current approach this flag will be set always if we have something in sortedEpoch and not only if something was cleaned because if epoch >= epochToDeleteTo then the for will break here but the flag will be set and the L212 will be executed.. Which was the real intention?

storage/clean/oldDatabaseCleaner.go
bogdan-rosianu: as discussed, will remove this, since it's ok to clean the map for the case when db deletion is set to false, then the map won't grow continuously

SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.

PR #4551, opened 2022-10-04
Start snapshot after restart fix

BeniaminDrasovean:
- If a snapshot is interrupted, when the node restarts it needs to start that snapshot again. The code that triggered that snapshot was on the constructor, thus triggering the same snapshot multiple times (each time the NewAccountsDB constructor is called)
  
## Proposed Changes
- Start the snapshot after restart on a separate method that is called only when is needed.

## Testing procedure
- Restart the node during a snapshot, and the snapshot should start only once.


node/nodeRunner.go
```diff
@@ -384,6 +384,14 @@ func (nr *nodeRunner) executeOneComponentCreationCycle(
 		return true, err
 	}
 
+	selfId := managedBootstrapComponents.ShardCoordinator().SelfId()
```
iulianpascalau, line 387: please re-check this. On rc/v1.4.0 we do not have this call here

state/accountsDB.go
```diff
@@ -160,17 +154,16 @@ func checkArgsAccountsDB(args ArgsAccountsDB) error {
 	return nil
 }
 
-func startSnapshotAfterRestart(adb AccountsAdapter, args ArgsAccountsDB) {
-	tsm := args.Trie.GetStorageManager()
+func startSnapshotAfterRestart(adb AccountsAdapter, tsm common.StorageManager, processingMode common.NodeProcessingMode) {
```
iulianpascalau, line 157: checked the file with v1.4.0, looks good :+1:

state/peerAccountsDB.go
```diff
@@ -39,15 +38,14 @@ func NewPeerAccountsDB(args ArgsAccountsDB) (*PeerAccountsDB, error) {
...
-	trieStorageManager := adb.mainTrie.GetStorageManager()
-	val, err := trieStorageManager.GetFromCurrentEpoch([]byte(common.ActiveDBKey))
-	if err != nil || !bytes.Equal(val, []byte(common.ActiveDBVal)) {
-		startSnapshotAfterRestart(adb, args)
-	}
-
 	return adb, nil
 }
 
+// StartSnapshotIfNeeded starts the snapshot if the previous snapshot process was not fully completed
```
iulianpascalau, line 44: checked with the rc/v1.4.0, looks good

iulianpascalau (changes requested):

node/nodeRunner.go
iulianpascalau: in rc/v1.4.0 these are called in the `addSyncersToAccountsDB` function after setting the trie syncers.

iulianpascalau (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
System test passed.

PR #4578, opened 2022-10-11
Unit tests for GetAllLeavesOnChannel refactorings

ssd04:
- More unit test for GetAllLeavesOnChannel refactoring

## Testing procedure
- N/A

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -118,17 +118,23 @@ func TestNewNode(t *testing.T) {
 }
 
 func TestNewNode_NilOptionShouldError(t *testing.T) {
+	t.Parallel()
+
 	_, err := node.NewNode(node.WithCoreComponents(nil))
 	assert.NotNil(t, err)
 }
 
 func TestNewNode_ApplyNilOptionShouldError(t *testing.T) {
+	t.Parallel()
+
 	n, _ := node.NewNode()
 	err := n.ApplyOptions(node.WithCoreComponents(nil))
 	assert.NotNil(t, err)
 }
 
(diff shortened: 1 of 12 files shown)


node/node_test.go
```diff
@@ -342,7 +358,64 @@ func TestNode_GetKeyValuePairs(t *testing.T) {
...
+	acc, _ := state.NewUserAccount([]byte("newaddress"))
+
+	accDB := &stateMock.AccountsStub{}
+
+	expectedErr := errors.New("expected err")
+	acc.SetDataTrie(
+		&trieMock.TrieStub{
+			GetAllLeavesOnChannelCalled: func(leavesChannels *common.TrieIteratorChannels, ctx context.Context, rootHash []byte, _ common.KeyBuilder) error {
+				go func() {
+					close(leavesChannels.LeavesChan)
```
iulianpascalau, line 373: this is a flaky test (might fail from time to time)
Reason? You close the leaves chan before writing the error in the error chan. Since this is done on a go routine, we might hit the case when the LeavesChan is closed but no error written yet in ErrChan. 
```GO
assert.Equal(t, expectedErr, err)
```
might fail

Suggestion here and perhaps in other tests (if applicable) is to revert the instructions order from L373-L343
The production implementation is done correctly :+1:

BeniaminDrasovean (approved):

node/node_test.go
ssd04: thanks :+1: changed in all places

iulianpascalau (approved):

PR #4579, opened 2022-10-11
Fix the heartbeat v2 monitor when the redundancy feature was activated

iulianpascalau:
- after the upgrade to v1.3.44, when some validators activated the redundancy feature, we observed that on the affected public key, the `numInstances` value from the heartbeatStatus route wrongly indicated the value 2
  
## Proposed Changes
- fixed heartbeat v2 monitor for when a validator used the redundancy feature

## Testing procedure
- internal system test
- redundancy feature specific tests


raduchis (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4582, opened 2022-10-12
added separate cli flag for db and logs path

ssd04:
- It will be useful to have a separate dir path for logs and db
  
## Proposed Changes
- added separate cli flags for logs and db paths

## Testing procedure
- Standard system test
- Default cli values should work as before, it should be backwards compatible
- Check booting from storage with custom db path

### DIFF
--- a/cmd/node/flags.go
+++ b/cmd/node/flags.go
@@ -249,6 +249,18 @@ var (
 		Usage: "This flag specifies the `directory` where the node will store databases, logs and statistics.",
 		Value: "",
 	}
+	// dbDirectory defines a flag for the path for the db directory.
+	dbDirectory = cli.StringFlag{
+		Name:  "db-path",
+		Usage: "This flag specifies the `directory` where the node will store databases and statistics.",
+		Value: "",
+	}
+	// logsDirectory defines a flag for the path for the logs directory.
+	logsDirectory = cli.StringFlag{
+		Name:  "logs-path",
+		Usage: "This flag specifies the `directory` where the node will store logs.",
+		Value: "",
+	}
 
 	// destinationShardAsObserver defines a flag for the prefered shard to be assigned to as an observer.
 	destinationShardAsObserver = cli.StringFlag{
@@ -405,6 +417,8 @@ func getFlags() []cli.Flag {
 		serializeSnapshots,
 		noKey,
 		p2pKeyPemFile,
+		dbDirectory,
+		logsDirectory,
 	}
 }
 
@@ -413,6 +427,8 @@ func getFlagsConfig(ctx *cli.Context, log logger.Logger) *config.ContextFlagsCon
 
 	workingDir := ctx.GlobalString(workingDirectory.Name)
 	flagsConfig.WorkingDir = getWorkingDir(workingDir, log)
+	flagsConfig.DbDir = getWorkingDir(ctx.GlobalString(dbDirectory.Name), log)
+	flagsConfig.LogsDir = getWorkingDir(ctx.GlobalString(logsDirectory.Name), log)
 	flagsConfig.EnableGops = ctx.GlobalBool(gopsEn.Name)
 	flagsConfig.SaveLogFile = ctx.GlobalBool(logSaveFile.Name)
 	flagsConfig.EnableLogCorrelation = ctx.GlobalBool(logWithCorrelation.Name)
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -247,7 +247,7 @@ func attachFileLogger(log logger.Logger, flagsConfig *config.ContextFlagsConfig)
 	var err error
 	if flagsConfig.SaveLogFile {
 		args := logging.ArgsFileLogging{
-			WorkingDir:      flagsConfig.WorkingDir,
+			WorkingDir:      flagsConfig.LogsDir,
 			DefaultLogsPath: defaultLogsPath,
 			LogFilePrefix:   logFilePrefix,
 		}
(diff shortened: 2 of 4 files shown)


cmd/node/flags.go
```diff
@@ -246,7 +246,19 @@ var (
 	// workingDirectory defines a flag for the path for the working directory.
 	workingDirectory = cli.StringFlag{
 		Name:  "working-directory",
-		Usage: "This flag specifies the `directory` where the node will store databases, logs and statistics.",
+		Usage: "This flag specifies the `directory` where the node will store statistics.",
```
andreibancioiu, line 249: This is now more like `statsDirectory`? Or is the `config` also loaded from here, by default?

cmd/node/flags.go
```diff
@@ -405,14 +417,17 @@ func getFlags() []cli.Flag {
...
 }
 
 func getFlagsConfig(ctx *cli.Context, log logger.Logger) *config.ContextFlagsConfig {
 	flagsConfig := &config.ContextFlagsConfig{}
 
-	workingDir := ctx.GlobalString(workingDirectory.Name)
-	flagsConfig.WorkingDir = getWorkingDir(workingDir, log)
+	flagsConfig.WorkingDir = getWorkingDir(ctx, workingDirectory, log)
+	flagsConfig.DbDir = getWorkingDir(ctx, dbDirectory, log)
+	flagsConfig.LogsDir = getWorkingDir(ctx, logsDirectory, log)
```
bogdan-rosianu, line 430: what happens if? 
`--working-directory new-path`
`--db-path` not set
`--log-path` not set

it would save only the statistics on `new-path` but `db` and `logs` will stain near the node binary, while they should be inside `new-path` as well. so, only if the new flags are set they should change the behaviour in some way
This is not backward compatible

integrationTests/factory/componentsHelper.go
```diff
@@ -67,6 +67,8 @@ func CreateDefaultConfig() *config.Configs {
 	configs.RoundConfig = roundConfig
 	configs.FlagsConfig = &config.ContextFlagsConfig{
 		WorkingDir: "workingDir",
+		DbDir:      "workingDir",
```
andreibancioiu, line 70: Not sure, should we use separate folders for tests, as well, instead?

bogdan-rosianu (commented):
also update the testing scenarios:

1. start the node with `./node --working-directory new-path`. `stats`, `db` and `logs` should be all placed inside the `new-path` directory
2. start the node with `./node` . `stats`, `db` and `logs` should be near the node
3. start the node with `./node --db-path new-db-path`. `db` should be saved inside the `new-db-path` directory while `stats` and `logs` should be near the node
4. start the node with `./node --logs-path new-logs-path`. `logs` should be saved inside the `new-logs-path` directory
and so on (cover all cases: default values, custom values, and so son)

cmd/node/flags.go
ssd04: right, thanks

cmd/node/flags.go
ssd04: in the default working directory we actually have `health-records`, `stats`, and `compiledSCStorage` from what I see, so I can change the description a bit

ssd04: changed the description

integrationTests/factory/componentsHelper.go
ssd04: changed to use separate folders, not sure if it has a relevant impact in the tests though

cmd/node/flags.go
```diff
@@ -523,6 +523,19 @@ func getWorkingDir(ctx *cli.Context, cliFlag cli.StringFlag, log logger.Logger)
 	return workingDir
 }
 
+func getCustomDirIfSet(ctx *cli.Context, cliFlag cli.StringFlag, log logger.Logger) string {
+	dbDir := ctx.GlobalString(cliFlag.Name)
+
+	var dirStr string
+	if len(dbDir) == 0 {
```
bogdan-rosianu, line 530: [optional]
could have been simplified to:
```
if len(dbDir) == 0 {
   return getWorkingDir(ctxm workingDirectory, log)
}

return getWorkingDir(ctx, cliFlag, log)
```

ssd04: done

bogdan-rosianu (approved):

cmd/node/flags.go
```diff
@@ -246,7 +246,7 @@ var (
 	// workingDirectory defines a flag for the path for the working directory.
 	workingDirectory = cli.StringFlag{
 		Name:  "working-directory",
-		Usage: "This flag specifies the `directory` where the node will store statistics.",
+		Usage: "This flag specifies the `directory` where the node will store databases, logs and statistics if no other related flags are set.",
```
andreibancioiu, line 249: :+1:

cmd/node/flags.go
```diff
@@ -523,6 +523,16 @@ func getWorkingDir(ctx *cli.Context, cliFlag cli.StringFlag, log logger.Logger)
 	return workingDir
 }
 
+func getCustomDirIfSet(ctx *cli.Context, cliFlag cli.StringFlag, log logger.Logger) string {
+	dbDir := ctx.GlobalString(cliFlag.Name)
```
andreibancioiu, line 527: Isn't `dbDir` a bit too specific (naming)?

andreibancioiu (dismissed):

cmd/node/flags.go
ssd04: renamed :+1:

bogdan-rosianu (dismissed):

iulianpascalau (approved):

iulianpascalau (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed

PR #4585, opened 2022-10-12
moved snapshots enabled field from config to cli params

ssd04:
- We need to be able to specify from cli params if snapshot is enabled 
  
## Proposed Changes
- add cli flag for specifying if snapshots are enabled
- remove SnapshotsEnabled from config

## Testing procedure
1. Standard testing procedure with snapshots enabled (default case)
2. Scenario with snapshots disabled on observer nodes, it has to be disabled before restart (in config file) and also after restart, via newly added CLI parameter

- !!! testing scripts has to be aligned with this change
- Change scripts in order to specify snasphot enabled from CLI
    - test snapshot scenarios based on CLI param
    - by default snapshots are enabled
    - if `--snapshots-enabled` snapshots will remain enabled
    - if `--snapshots-enabled=false` snapshots will have to be disabled


BeniaminDrasovean (approved):

schimih (dismissed):

cmd/node/flags.go
```diff
@@ -365,6 +365,12 @@ var (
 		Value: "./config/p2pKey.pem",
 	}
 
+	// snapshotsEnabled is used to enable snapshots, if it is not set it defaults to true, it will be set to false if it is set specifically
+	snapshotsEnabled = cli.BoolTFlag{
+		Name:  "snapshots-enabled",
+		Usage: "Boolean option for enabling state snapshots. If it is not set it defaults to true, it will be set to false if it is set specifically as --snapshots-enabled=false",
```
iulianpascalau, line 371: Add `Value: true` so it will be enabled by default as stated in the Usage string?

ssd04: it will be set to true by default

iulianpascalau (dismissed):

gabi-vuls (approved):
System test passed

BeniaminDrasovean (approved):

iulianpascalau (approved):

PR #4587, opened 2022-10-12
Block tracker fix on metablocks for the forced rollback event

iulianpascalau:
- there was an edge case if a node started from storage on an exact block that contained scheduled miniblock and a forced rollback event was triggered. The block tracker contained only one element and could not remove as it would have rendered the block tracker unusable. This "leftover" metablock would have prevented itself to be re-executed and thus, the node would get stuck.
  
## Proposed Changes
- fixed the block tracker by adding the previous metachain block so a forced rollback will work

## Testing procedure
- standard system tests with a lot of restarts


process/sync/storageBootstrap/shardStorageBootstrapper_test.go
```diff
@@ -252,3 +259,129 @@ func TestShardStorageBootstrapper_GetCrossNotarizedHeaderNonceShouldWork(t *test
...
+		assert.Equal(t, currentMeta.metablock, addedCrossNotarized[string(currentMeta.hash)])
+		assert.Equal(t, currentMeta.metablock, addedTrackedHeaders[string(currentMeta.hash)])
+
+		assert.Equal(t, prevMeta.metablock, addedCrossNotarized[string(prevMeta.hash)])
+		assert.Equal(t, prevMeta.metablock, addedTrackedHeaders[string(prevMeta.hash)])
+	})
+}
+
+func setupForApplyCrossNotarizedHeadersTests(
+	addedCrossNotarized map[string]data.HeaderHandler,
```
andreibancioiu, line 314: `addedCrossNotarizedHeaders` / `crossNotarizedHeadersAccumulator` (optional renaming) - and above / below, if the second option is applied.

process/sync/storageBootstrap/shardStorageBootstrapper_test.go
```diff
@@ -252,3 +259,129 @@ func TestShardStorageBootstrapper_GetCrossNotarizedHeaderNonceShouldWork(t *test
...
+	t.Run("missing current meta should error", func(t *testing.T) {
+		addedCrossNotarized := make(map[string]data.HeaderHandler)
+		addedTrackedHeaders := make(map[string]data.HeaderHandler)
+
+		bootstrapper, _, currentMeta, crossNotarizedHeaders := setupForApplyCrossNotarizedHeadersTests(addedCrossNotarized, addedTrackedHeaders)
+		_ = bootstrapper.store.GetStorer(dataRetriever.MetaBlockUnit).Remove(currentMeta.hash)
+
+		err := bootstrapper.applyCrossNotarizedHeaders(crossNotarizedHeaders)
+		assert.NotNil(t, err)
+		assert.True(t, strings.Contains(err.Error(), "missing header : GetMarshalizedHeaderFromStorage"))
```
andreibancioiu, line 275: `assert.ErrorContains()`? (optional)

process/sync/storageBootstrap/shardStorageBootstrapper_test.go
```diff
@@ -252,3 +259,129 @@ func TestShardStorageBootstrapper_GetCrossNotarizedHeaderNonceShouldWork(t *test
...
+
+func TestShardStorageBootstrapper_applyCrossNotarizedHeaders(t *testing.T) {
+	t.Parallel()
+
+	t.Run("missing current meta should error", func(t *testing.T) {
+		addedCrossNotarized := make(map[string]data.HeaderHandler)
+		addedTrackedHeaders := make(map[string]data.HeaderHandler)
+
+		bootstrapper, _, currentMeta, crossNotarizedHeaders := setupForApplyCrossNotarizedHeadersTests(addedCrossNotarized, addedTrackedHeaders)
+		_ = bootstrapper.store.GetStorer(dataRetriever.MetaBlockUnit).Remove(currentMeta.hash)
```
andreibancioiu, line 271: Optionally add comment: "Let's remove current meta" (and below).

andreibancioiu (approved):

SebastianMarian (approved):

process/sync/storageBootstrap/shardStorageBootstrapper_test.go
iulianpascalau: renamed to `crossNotarizedHeadersAccumulator` and `trackedHeadersAccumulator`

process/sync/storageBootstrap/shardStorageBootstrapper_test.go
iulianpascalau: :1st_place_medal:

sasurobert (approved):

process/sync/storageBootstrap/shardStorageBootstrapper_test.go
iulianpascalau: done

SebastianMarian (approved):

process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -206,6 +212,44 @@ func (st *storageBootstrapper) loadBlocks() error {
...
+	if header.ShardId == common.MetachainShardId {
+		return nil
+	}
+
+	hdr, err := st.bootstrapper.getHeader(header.Hash)
+	if err != nil {
+		return err
+	}
+
+	shardHeader, ok := hdr.(data.ShardHeaderHandler)
```
SebastianMarian, line 225: I think we should replace L225-L228 with:

```
	prevHdr, err := st.bootstrapper.getHeader(hdr.GetPrevHash())
	if err != nil {
		return err
	}
	
	shardHeader, ok := prevHdr.(data.ShardHeaderHandler)
	if !ok {
	}

```

As we need to add the metablocks and metablocks hashes contained in the previous header from the one we will started. All the info regarding the headerInfo.LastHeader is already added in the last call of applyBootInfos

iulianpascalau: forgot about previous header :man_facepalming:

SebastianMarian (approved):

iulianpascalau:
This is not the fix, will open another PR

PR #4590, opened 2022-10-13
EN-13278: move webserver antiflood configuration

bogdan-rosianu:
- The `Antiflood` enabled config value was not taken into consideration for the webserver section
  
## Proposed Changes
- Move the webserver antiflood values to a new configuration structure
- Don't use any limitation if the web server antiflood is disabled

## Testing procedure
1) general throttlers
- set config.toml `WebServerAntifloodEnabled` to `true`, `SameSourceRequests` to `2` and `SameSourceResetIntervalInSec` to `10` and query the endpoint `<node-ip>:<node:port"/network/config` few times. you should receive a "too many requests" response
- for the above scenario, set `WebServerAntifloodEnabled` to `false` and try the queries again. This time, the "too many requests" message should not be returned

2) trie operations
- set config.toml `WebServerAntifloodEnabled` to `true`, and`TrieOperationsDeadlineMilliseconds` to `1`. Query the endpoint `<node-ip>:<node:port"/network/esdts` on a testnet where there are a lot of tokens. It should fail with trie timeout
- for the above scenario, set  `WebServerAntifloodEnabled` to `false` and try the query again. This time, it should work


cmd/node/config/config.toml
```diff
@@ -591,6 +575,25 @@
         # clutter the network exactly in the same moment
         MaxDeviationTimeInMilliseconds = 25
 
+[WebServerAntiflood]
+    WebServerAntifloodEnabled = true
```
mariusmihaic, line 579: [comment-optional]

Could've simply named this one `Enabled`, since it is already part of `WebServerAntiflood`.
Also, shouldn't this have a comm?

facade/nodeFacade.go
```diff
@@ -428,6 +445,10 @@ func (nf *nodeFacade) GetPeerInfo(pid string) ([]core.QueryP2PPeerInfo, error) {
 
 // GetThrottlerForEndpoint returns the throttler for a given endpoint if found
 func (nf *nodeFacade) GetThrottlerForEndpoint(endpoint string) (core.Throttler, bool) {
+	if !nf.wsAntifloodConfig.WebServerAntifloodEnabled {
```
mariusmihaic, line 448: [comment]

I think we could've avoided this if and the one above by using some sort of polymorphic func.
However, i think the overhead is too big for too small reward

mariusmihaic (approved):

cmd/node/config/config.toml
bogdan-rosianu: The name is meant to be this way, since it will make the devops process easier

facade/nodeFacade.go
bogdan-rosianu: I would leave it as it is

api/gin/webServer.go
```diff
@@ -126,11 +126,15 @@ func (ws *webServer) StartHttpServer() error {
 		return err
 	}
 
-	log.Debug("starting web server",
-		"SimultaneousRequests", ws.antiFloodConfig.SimultaneousRequests,
-		"SameSourceRequests", ws.antiFloodConfig.SameSourceRequests,
-		"SameSourceResetIntervalInSec", ws.antiFloodConfig.SameSourceResetIntervalInSec,
-	)
+	if !ws.antiFloodConfig.WebServerAntifloodEnabled {
```
sstanculeanu, line 129: why not replace L129-137 with:
```go
log.Debug("starting web server",
	"SimultaneousRequests", ws.antiFloodConfig.SimultaneousRequests,
	"SameSourceRequests", ws.antiFloodConfig.SameSourceRequests,
	"SameSourceResetIntervalInSec", ws.antiFloodConfig.SameSourceResetIntervalInSec,
	"with throttle middleware", ws.antiFloodConfig.WebServerAntifloodEnabled,
)
```

bogdan-rosianu: It makes no point in printing SimultaneousRequests and so on if they are not effective

sstanculeanu: oh, right

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4594, opened 2022-10-14
error processing refactor

matei-p:
Existing tx processing failure mechanism is encompassed in ProcessIfError(). This code was scattered everywhere and there was also the risk of double calling it.
  
## Proposed Changes
Refactor of the failure processing calling code, by moving it as high as possible in the calling hierarchy.

## Testing procedure
Existing unit tests cover the error scenarios and their pass guarantees a correct refactor.

### DIFF
--- a/process/smartContract/processorV2/failureContext.go
+++ b/process/smartContract/processorV2/failureContext.go
@@ -1,0 +1,58 @@
+package processorV2
+
+import vmcommon "github.com/ElrondNetwork/elrond-vm-common"
+
+type failureContext struct {
+	sc            *scProcessor
+	processFail   bool
+	txHash        []byte
+	errorMessage  string
+	returnMessage []byte
+	snapshot      int
+	gasLocked     uint64
+	logs          []*vmcommon.LogEntry
+}
+
+func NewFailureContext(sc *scProcessor) *failureContext {
+	return &failureContext{
+		sc:          sc,
+		processFail: false,
+	}
+}
+
+func (failCtx *failureContext) makeSnaphot() *failureContext {
+	failCtx.snapshot = failCtx.sc.accounts.JournalLen()
+	return failCtx
+}
+
+func (failCtx *failureContext) setTxHash(txHash []byte) *failureContext {
+	failCtx.txHash = txHash
+	return failCtx
+}
+
+func (failCtx *failureContext) setGasLocked(gasLocked uint64) *failureContext {
+	failCtx.gasLocked = gasLocked
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessage(errorMessage string) *failureContext {
+	failCtx.setMessages(errorMessage, []byte(errorMessage))
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessagesFromError(err error) *failureContext {
+	failCtx.setMessage(err.Error())
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessages(errorMessage string, returnMessage []byte) *failureContext {
+	failCtx.processFail = true
+	failCtx.errorMessage = errorMessage
+	failCtx.returnMessage = returnMessage
+	return failCtx
+}
+
+func (failCtx *failureContext) setLogs(logs []*vmcommon.LogEntry) *failureContext {
+	failCtx.logs = logs
+	return failCtx
+}
--- a/process/smartContract/processorV2/failureContext.go
+++ b/process/smartContract/processorV2/failureContext.go
@@ -3,7 +3,6 @@ package processorV2
 import vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 
 type failureContext struct {
-	sc            *scProcessor
 	processFail   bool
 	txHash        []byte
 	errorMessage  string
@@ -13,46 +12,38 @@ type failureContext struct {
 	logs          []*vmcommon.LogEntry
 }
 
-func NewFailureContext(sc *scProcessor) *failureContext {
+func NewFailureContext() *failureContext {
 	return &failureContext{
-		sc:          sc,
 		processFail: false,
 	}
 }
 
-func (failCtx *failureContext) makeSnaphot() *failureContext {
-	failCtx.snapshot = failCtx.sc.accounts.JournalLen()
-	return failCtx
+func (failCtx *failureContext) makeSnaphot(snapshot int) {
+	failCtx.snapshot = snapshot
 }
 
-func (failCtx *failureContext) setTxHash(txHash []byte) *failureContext {
+func (failCtx *failureContext) setTxHash(txHash []byte) {
 	failCtx.txHash = txHash
-	return failCtx
 }
 
-func (failCtx *failureContext) setGasLocked(gasLocked uint64) *failureContext {
+func (failCtx *failureContext) setGasLocked(gasLocked uint64) {
 	failCtx.gasLocked = gasLocked
-	return failCtx
 }
 
-func (failCtx *failureContext) setMessage(errorMessage string) *failureContext {
+func (failCtx *failureContext) setMessage(errorMessage string) {
 	failCtx.setMessages(errorMessage, []byte(errorMessage))
-	return failCtx
 }
 
-func (failCtx *failureContext) setMessagesFromError(err error) *failureContext {
+func (failCtx *failureContext) setMessagesFromError(err error) {
 	failCtx.setMessage(err.Error())
-	return failCtx
 }
 
-func (failCtx *failureContext) setMessages(errorMessage string, returnMessage []byte) *failureContext {
+func (failCtx *failureContext) setMessages(errorMessage string, returnMessage []byte) {
 	failCtx.processFail = true
 	failCtx.errorMessage = errorMessage
 	failCtx.returnMessage = returnMessage
-	return failCtx
 }
 
-func (failCtx *failureContext) setLogs(logs []*vmcommon.LogEntry) *failureContext {
+func (failCtx *failureContext) setLogs(logs []*vmcommon.LogEntry) {
 	failCtx.logs = logs
-	return failCtx
 }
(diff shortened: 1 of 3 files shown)


sasurobert (commented):
I like the idea - resolve TODOs

process/smartContract/processorV2/failureContext.go
```diff
@@ -0,0 +1,58 @@
...
+	processFail   bool
+	txHash        []byte
+	errorMessage  string
+	returnMessage []byte
+	snapshot      int
+	gasLocked     uint64
+	logs          []*vmcommon.LogEntry
+}
+
+func NewFailureContext(sc *scProcessor) *failureContext {
```
iulianpascalau, line 16: providing the sc processor pointer here suggest that the failureContext and scProcessor are closely tight together.
In this case, I suggest making the scProcessorV2 keep a pointer of type failure context inside it. In this way you will not need to pass it around on all methods. You will need to reset it whenever the sc processor needs to start processing something.

process/smartContract/processorV2/failureContext.go
```diff
@@ -0,0 +1,58 @@
...
+		processFail: false,
+	}
+}
+
+func (failCtx *failureContext) makeSnaphot() *failureContext {
+	failCtx.snapshot = failCtx.sc.accounts.JournalLen()
+	return failCtx
+}
+
+func (failCtx *failureContext) setTxHash(txHash []byte) *failureContext {
```
iulianpascalau, line 28: daisy chaining that does not quite add clarity but makes the calls a one-liner.
I always prefer multiple calls on multiple lines as long as they are clear enough.

process/smartContract/processorV2/processV2.go
```diff
@@ -825,27 +846,50 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
...
+	acntSnd, acntDst state.UserAccountHandler,
+	failureContext *failureContext,
+) (vmcommon.ReturnCode, error) {
+	returnCode, vmInput, txHash, err := sc.prepareExecution(tx, acntSnd, acntDst, true, failureContext)
 	if err != nil || returnCode != vmcommon.Ok {
 		return returnCode, err
 	}
 
-	snapshot := sc.accounts.JournalLen()
+	failureContext.makeSnaphot()
```
iulianpascalau, line 873: this is the second usage of these 3 functions. Suggestion to add a single function that set all these 3 values.

process/smartContract/processorV2/processV2.go
```diff
@@ -825,27 +846,50 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
...
 
 	acntSnd, err = sc.reloadLocalAccount(acntSnd)
 	if err != nil {
-		return 0, err
+		return vmcommon.Ok, err
 	}
 
 	if vmInput.ReturnCallAfterError && vmInput.CallType != vmData.AsynchronousCallBack {
-		return sc.finishSCExecution(make([]data.TransactionHandler, 0), txHash, tx, vmOutput, 0)
+		returnCode, err := sc.finishSCExecution(make([]data.TransactionHandler, 0), txHash, tx, vmOutput, 0)
```
iulianpascalau, line 891: why this change?

process/smartContract/processorV2/processV2.go
```diff
@@ -939,20 +993,22 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
...
-	if errCheck != nil {
-		return vmcommon.UserError, sc.ProcessIfError(acntSnd, txHash, tx, errCheck.Error(), []byte(errCheck.Error()), snapshot, vmInput.GasLocked)
+	err = sc.checkSCRSizeInvariant(scrResults)
+	if err != nil {
+		failureContext.setMessagesFromError(err)
+		return vmcommon.UserError, nil
 	}
 
-	return sc.finishSCExecution(scrResults, txHash, tx, newVMOutput, builtInFuncGasUsed)
+	returnCode, err = sc.finishSCExecution(scrResults, txHash, tx, newVMOutput, builtInFuncGasUsed)
```
iulianpascalau, line 1010: why these changes?

process/smartContract/processorV2/processV2.go
SebastianMarian: +1

process/smartContract/processorV2/processV2.go
```diff
@@ -939,20 +993,22 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
 				tx:                   tx,
 				scrTxs:               scrResults,
 				acntSnd:              acntSnd,
-				snapshot:             snapshot,
 				createdAsyncCallback: createdAsyncCallback,
+				failureContext:       failureContext,
 			})
 		if errReturnCode != vmcommon.Ok || err != nil {
-			return errReturnCode, sc.ProcessIfError(acntSnd, txHash, tx, err.Error(), []byte(err.Error()), snapshot, vmInput.GasLocked)
+			return errReturnCode, nil
```
SebastianMarian, line 1000: failureContext.setMessagesFromError(err) before return?

process/smartContract/processorV2/processV2.go
SebastianMarian: +1

process/smartContract/processorV2/processV2.go
matei-p: In my opinion would complicate a little bit the function signatures and in one of the cases we should set the gas locked only after an error checks is made so it's possible they are not linked.

process/smartContract/processorV2/failureContext.go
sasurobert: In one sense you are right, however we do not have any accumulator in this sense at SCProcessorLevel - I think it is better to keep it that way. dependency on scProcessor can be taken out if instead on makeSnapshot - we have setSnapshot(sc.accounts.journallen)

process/smartContract/processorV2/processV2.go
```diff
@@ -899,23 +952,24 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
...
 		if err != nil {
-			return vmcommon.ExecutionFailed, sc.ProcessIfError(acntSnd, txHash, tx, err.Error(), []byte(err.Error()), snapshot, vmInput.GasLocked)
+			return vmcommon.ExecutionFailed, nil
 		}
 
-		newVMOutput, errReturnCode, err = sc.executeSmartContractCallAndCheckGas(newVMInput, tx, newVMInput.CurrentTxHash, snapshot, acntSnd, newDestSC)
+		newVMOutput, errReturnCode, err = sc.executeSmartContractCallAndCheckGas(newVMInput, tx, newVMInput.CurrentTxHash, acntSnd, newDestSC, failureContext)
 		if err != nil {
-			return vmcommon.ExecutionFailed, sc.ProcessIfError(acntSnd, txHash, tx, err.Error(), []byte(err.Error()), snapshot, vmInput.GasLocked)
+			return vmcommon.ExecutionFailed, nil
```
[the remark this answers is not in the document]
matei-p: good catch

process/smartContract/processorV2/processV2.go
matei-p: good catch

process/smartContract/processorV2/processV2.go
```diff
@@ -899,23 +952,24 @@ func (sc *scProcessor) doExecuteBuiltInFunction(
 				scExecuteOutTransfer: executionDataAfterBuiltIn.scExecuteOutTransfer,
 				tx:                   tx,
 				acntSnd:              acntSnd,
-				snapshot:             snapshot,
+				failureContext:       failureContext,
 			})
 		if err != nil {
-			return vmcommon.ExecutionFailed, sc.ProcessIfError(acntSnd, txHash, tx, err.Error(), []byte(err.Error()), snapshot, vmInput.GasLocked)
+			return vmcommon.ExecutionFailed, nil
```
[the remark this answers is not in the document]
matei-p: good catch

process/smartContract/processorV2/failureContext.go
```diff
@@ -0,0 +1,58 @@
...
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessage(errorMessage string) *failureContext {
+	failCtx.setMessages(errorMessage, []byte(errorMessage))
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessagesFromError(err error) *failureContext {
+	failCtx.setMessage(err.Error())
```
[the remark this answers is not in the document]
matei-p: no longer the case

process/smartContract/processorV2/failureContext.go
```diff
@@ -0,0 +1,58 @@
...
+	return failCtx
+}
+
+func (failCtx *failureContext) setGasLocked(gasLocked uint64) *failureContext {
+	failCtx.gasLocked = gasLocked
+	return failCtx
+}
+
+func (failCtx *failureContext) setMessage(errorMessage string) *failureContext {
+	failCtx.setMessages(errorMessage, []byte(errorMessage))
```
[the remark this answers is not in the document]
matei-p: no longer the case

process/smartContract/processorV2/processV2.go
matei-p: at some point there was some code between that and return :) ofc, fixed

process/smartContract/processorV2/processV2.go
matei-p: at some point there was some code between that and return :) ofc, fixed

process/smartContract/processorV2/failureContext.go
matei-p: done

process/smartContract/processorV2/failureContext.go
matei-p: done

SebastianMarian (approved):

PR #4596, opened 2022-10-17
Protocol tx fields

mariusmihaic:
- When querying for `ApiTransactionResult` (either in block/proxy hyperblock), some protocol transaction+scr fields were missing. Some of them were mandatory for external clients to re-create protocol transaction(e.g. validate transaction signature based on transaction version + options). 
## Proposed Changes
- Added  missing fields support for API calls
## Testing procedure
`ApiTransactionResult` holds all type of txs in a single struct (move balance, scr, rewards, invalid, etc).
To test it, generate different types of transactions and check the validity of the following:
- `ChainID + Options` : should be available and **correct** for all tx types 
- `Version`: should be available only for normal txs, otherwise it should be missing
- `CallType` + `RelayerAddress` + `RelayedValue`: for scr txs, should be correctly computed, otherwise these should be missing

An observer node from each shard should be individually queried for these information, as well as proxy(hyperblock for example). For proxy, one should use this branch: `EN-13246-protocol-tx-fields`


andreibancioiu (commented):
Just the comment related to a proper tag, if appropriate.

factory/api/apiResolverFactory.go
```diff
@@ -234,6 +234,7 @@ func CreateApiResolver(args *ApiResolverArgs) (facade.ApiResolver, error) {
 		TxTypeHandler:            txTypeHandler,
 		LogsFacade:               logsFacade,
 		DataFieldParser:          dataFieldParser,
+		ChainID:                  args.Configs.GeneralConfig.GeneralSettings.ChainID,
```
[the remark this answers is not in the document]
mariusmihaic: I honestly didn't know it can change to a different value during on-chain execution, but not sure where else to get this field from(for all tx types).
Any suggestion? Otherwise, would leave it as it is

node/external/transactionAPI/unmarshaller.go
```diff
@@ -159,10 +166,27 @@ func (tu *txUnmarshaller) prepareUnsignedTx(tx *smartContractResult.SmartContrac
...
+		RelayerAddress:          tu.getEncodedAddress(tx.GetRelayerAddr()),
+		RelayedValue:            bigIntToStr(tx.GetRelayedValue()),
+		OriginalSender:          tu.getEncodedAddress(tx.GetOriginalSender()),
 	}
 
 	return txResult, nil
 }
+
+func (tu *txUnmarshaller) getEncodedAddress(address []byte) string {
+	if len(address) == tu.addressPubKeyConverter.Len() {
```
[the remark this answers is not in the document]
mariusmihaic: > `len(address) == 0 then return ""`

Right, this is also valid, will add an extra check for this

> len(address) != addressPubKeyConverter.Len() then return "bad
I think there might be some edge case for metachain address here?

andreibancioiu (dismissed):

factory/api/apiResolverFactory.go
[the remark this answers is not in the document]
raduchis: do you actually need the ChainID here?

node/external/transactionAPI/unmarshaller.go
```diff
@@ -110,6 +115,8 @@ func (tu *txUnmarshaller) prepareNormalTx(tx *transaction.Transaction) (*transac
 		GasLimit:         tx.GasLimit,
 		Data:             tx.Data,
 		Signature:        hex.EncodeToString(tx.Signature),
+		Options:          tx.Options,
+		Version:          tx.Version,
```
raduchis, line 119: the chainID should be filled also here from the internalTx if it exists.

node/external/transactionAPI/unmarshaller.go
```diff
@@ -92,6 +96,7 @@ func (tu *txUnmarshaller) unmarshalTransaction(txBytes []byte, txType transactio
 	apiTx.Receivers = datafield.EncodeBytesSlice(tu.addressPubKeyConverter.Encode, res.Receivers)
 	apiTx.ReceiversShardIDs = res.ReceiversShardID
 	apiTx.IsRelayed = res.IsRelayed
+	apiTx.ChainID = tu.chainID
```
raduchis, line 99: this should be on the initial transaction.

raduchis (changes requested):

node/external/transactionAPI/unmarshaller.go
mariusmihaic: Right, only use for normal txs. Refactored

node/external/transactionAPI/unmarshaller.go
mariusmihaic: Refactored

andreibancioiu (approved):

node/external/transactionAPI/unmarshaller.go
raduchis: ok

raduchis (approved):

PR #4599, opened 2022-10-17
Sync statistics on status core comp

sstanculeanu:
- same `trieSyncStatistics` should pe used for both trie sync and bootstrap heartbeat sender for the newly added field on heartbeat v2 message
  
## Proposed Changes
- moved `trieSyncStatistics` component on `statusCoreComponents`, **but only the one used for user account** which takes most of the sync time
- integrate this on monitor, in order to further send the info through api
- moved num of nodes on the heartbeat v2 message instead of payload

## Testing procedure
- proper test scenario will be defined on the feat branch with all components done


heartbeat/sender/peerAuthenticationSender.go
```diff
@@ -116,8 +116,9 @@ func (sender *peerAuthenticationSender) execute() (error, bool) {
 
 	hardforkPayload, isTriggered := sender.getHardforkPayload()
 	payload := &heartbeat.Payload{
-		Timestamp:       time.Now().Unix(),
-		HardforkMessage: string(hardforkPayload),
+		Timestamp:          time.Now().Unix(),
+		HardforkMessage:    string(hardforkPayload),
+		NumTrieNodesSynced: 0, // sent through heartbeat v2 message
```
iulianpascalau, line 121: no need for this

factory/consensus/consensusComponents.go
```diff
@@ -502,6 +503,7 @@ func (ccf *consensusComponentsFactory) createArgsBaseAccountsSyncer(trieStorageM
 		TrieSyncerVersion:         ccf.config.TrieSync.TrieSyncerVersion,
 		CheckNodesOnDisk:          ccf.config.TrieSync.CheckNodesOnDisk,
 		StorageMarker:             storageMarker.NewTrieStorageMarker(),
+		SyncStatisticsHandler:     statistics.NewTrieSyncStatistics(),
```
iulianpascalau, line 506: by providing a new pointer here we assure that no trie sync information will be sent on the heartbeat, if, for example, the node syncs blocks but starts an ad-hoc trie sync process because it finds a missing trie node.

state/syncer/baseAccountsSyncer.go
```diff
@@ -72,6 +74,9 @@ func checkArgs(args ArgsNewBaseAccountsSyncer) error {
 	if check.IfNil(args.Cacher) {
 		return state.ErrNilCacher
 	}
+	if check.IfNil(args.SyncStatisticsHandler) {
```
iulianpascalau, line 77: Let's rename this to UserAccountsSyncStatisticsHandler as it only applies to the user accounts trie

state/syncer/userAccountsSyncer.go
```diff
@@ -129,7 +128,7 @@ func (u *userAccountsSyncer) SyncAccounts(rootHash []byte) error {
 
 	log.Debug("main trie synced, starting to sync data tries", "num data tries", len(u.dataTries))
 
-	err = u.syncAccountDataTries(mainTrie, tss, ctx)
+	err = u.syncAccountDataTries(mainTrie, ctx)
```
iulianpascalau, line 131: after we finished the trie sync, on L137, for example, we need to somehow call a reset method on the syncStatisticsHandler in order to not broadcast the last value from the last trie sync.

factory/bootstrap/bootstrapComponents.go
```diff
@@ -180,6 +189,7 @@ func (bcf *bootstrapComponentsFactory) Create() (*bootstrapComponents, error) {
 		HeaderIntegrityVerifier:    headerIntegrityVerifier,
 		DataSyncerCreator:          dataSyncerFactory,
 		ScheduledSCRsStorer:        nil, // will be updated after sync from network
+		TrieSyncStatisticsProvider: bcf.statusCoreComponents.TrieSyncStatistics(),
```
iulianpascalau, line 192: I think we should increment by one the values for the processed nodes just before we start the heartbeat sender during bootstrap.
Reason: when the sender is instantiated, the very first heartbeat message should contain a non-0 value on the new `NumTrieNodesSynced` field so it will immediately be marked correctly in the explorer

BeniaminDrasovean (approved):

factory/bootstrap/bootstrapComponents.go
```diff
@@ -168,6 +168,11 @@ func (bcf *bootstrapComponentsFactory) Create() (*bootstrapComponents, error) {
 
 	dataSyncerFactory := bootstrap.NewScheduledDataSyncerFactory()
 
+	// increment num received to make sure that first heartbeat message
+	// will have value 1, thus explorer will display status in progress
+	tss := bcf.statusCoreComponents.TrieSyncStatistics()
+	tss.AddNumReceived(1)
```
iulianpascalau, line 174: I think we should use processed instead of received.

heartbeat/sender/heartbeatSender.go
```diff
@@ -122,14 +120,16 @@ func (sender *heartbeatSender) execute() error {
 		return err
 	}
 
+	trieNodesReceived := sender.trieSyncStatisticsProvider.NumReceived()
```
iulianpascalau, line 123: Also here, let's use processed, not received

factory/bootstrap/bootstrapComponents.go
iulianpascalau: wrong naming

heartbeat/sender/heartbeatSender.go
iulianpascalau: wrong naming

iulianpascalau (approved):

factory/bootstrap/bootstrapComponents.go
sstanculeanu: as discussed, will change the naming of the methods in a separate PR

sstanculeanu: updated into #4614

PR #4600, opened 2022-10-18
Feat trie nodes synced on heartbeat

sstanculeanu:
- there was no option to display the sync status on explorer, leading to a lot of confusion about node's state
  
## Proposed Changes
- created a new heartbeat sender that sends the number of trie nodes synced during bootstrap through the newly added field on heartbeat message, `NumTrieNodesSynced`
- added a new metric `erd_accounts_snapshot_num_nodes` which is filled right after snapshot with the total number of nodes

## Testing procedure
- standard system test
- special scenario:
   - start a squad of observers on testnet/mainnet and let them sync, then wait at least one epoch change so one snapshot is done and the metric is filled
   - start a proxy from branch `new_route_trie_statistics`, with this squad and query `/network/trie-statistics/:shard` for each shard. It should not be 0.
   - start another node with a specific identity in prefs.toml. If the test is done on testnet, for this new node update [config.toml](https://github.com/ElrondNetwork/elrond-config-testnet/blob/T1.3.48.0/config.toml#L920) to 20seconds.
   - while this node is in trie sync, query the proxy on `/node/heartbeatstatus`. On the response, there should be an entry with this identity + a new field, `numTrieNodesReceived` which should not be 0.


gabi-vuls (approved):
System test passed.

bogdan-rosianu (dismissed):

schimih (dismissed):

schimih (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):

PR #4605, opened 2022-10-18
Added operation mode cli flag

bogdan-rosianu:
- There were some use cases when some configuration values had to be altered in order to run the node in a certain mode
  
## Proposed Changes
- Added a CLI flag `--operation-mode` which can contain an array of possible values that will alter the node's configuration accordingly

## Testing procedure
- `cd elrond-go/cmd/node`
- `go build`
- `./node --help`
- test all possible combinations for the `--operation-mode` flag


cmd/node/operationModes.go
```diff
@@ -0,0 +1,49 @@
...
+
+const (
+	operationModeImportDb           = "import-db"
+	operationModeFullArchive        = "full-archive"
+	operationModeDbLookupExtension  = "db-lookup-extension"
+	operationModeHistoricalBalances = "historical-balances"
+)
+
+func checkOperationModes(modes []string) error {
+	// db lookup extension and historical balances
```
bogdan-rosianu, line 13: For reviewers: think about other combinations of modes that can lead to unexpected configuration. For now, I've checked modes that altered the same fields

cmd/node/operationModes.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import "fmt"
+
+const (
+	operationModeImportDb           = "import-db"
+	operationModeFullArchive        = "full-archive"
+	operationModeDbLookupExtension  = "db-lookup-extension"
+	operationModeHistoricalBalances = "historical-balances"
```
bogdan-rosianu, line 9: For reviewers: are other operation modes that have to be included for now?

cmd/node/flags.go
```diff
@@ -507,26 +517,66 @@ func getWorkingDir(workingDir string, log logger.Logger) string {
...
+	isInDbLookupExtensionMode := sliceContainsElement(operationModes, operationModeDbLookupExtension)
+	if isInDbLookupExtensionMode {
+		processDbLookupExtensionMode(log, configs)
 	}
 
 	return nil
 }
 
+func processHistoricalBalancesMode(log logger.Logger, configs *config.Configs) {
+	configs.GeneralConfig.StoragePruning.ObserverCleanOldEpochsData = false
```
andreibancioiu, line 558: For validators, as well, perhaps? (in case there are validators started with this operation mode - only if we can recommend this, otherwise not).

cmd/node/operationModes.go
```diff
@@ -0,0 +1,49 @@
...
+	// db lookup extension and historical balances
+	isInvalid := sliceContainsBothElements(modes, operationModeHistoricalBalances, operationModeDbLookupExtension)
+	if isInvalid {
+		return fmt.Errorf("operation-mode flag cannot contain both db-lookup-extension and historical-balances")
+	}
+
+	return nil
+}
+
+func sliceContainsBothElements(elements []string, first string, second string) bool {
```
andreibancioiu, line 22: Here, it's all right to call `sliceContainsElement` twice - shorter implementation (no actual performance issue on short arrays).

andreibancioiu: This function doesn't have a unit test.

andreibancioiu (commented):
Nice feature :rocket: 

(a few questions / comments)

cmd/node/flags.go
```diff
@@ -507,26 +517,66 @@ func getWorkingDir(workingDir string, log logger.Logger) string {
...
 
-	if importDbFlags.IsImportDBMode {
-		return processConfigImportDBMode(log, configs)
+	operationModes := strings.Split(configs.FlagsConfig.OperationMode, ",")
+	err := checkOperationModes(operationModes)
+	if err != nil {
+		return err
+	}
+
+	isInImportDBMode := configs.ImportDbConfig.IsImportDBMode || sliceContainsElement(operationModes, operationModeImportDb)
```
iulianpascalau, line 530: I will not mix the import-db mode here as we specifically need to provide the DB path for the import operation

cmd/node/flags.go
```diff
@@ -351,6 +352,13 @@ var (
 		Usage: "The `filepath` for the PEM file which contains the secret keys for the p2p key. If this is not specified a new key will be generated (internally) by default.",
 		Value: "./config/p2pKey.pem",
 	}
+
+	// operationMode defines the flag for specifying how configs should be altered depending on the node's intent
+	operationMode = cli.StringFlag{
+		Name:  "operation-mode",
+		Usage: "String flag for specifying the desired `operation mode`(s) of the node, resulting in altering some configuration values accordingly. Values are: import-db, full-archive, db-lookup-extension, historical-balances or `\"\"` (empty). Multiple values can be separated via ,",
```
iulianpascalau, line 359: maybe add the operation mode: "observer" in which we have the `no-key` flag set?

cmd/node/flags.go
```diff
@@ -507,26 +517,66 @@ func getWorkingDir(workingDir string, log logger.Logger) string {
 }
 
 func applyCompatibleConfigs(log logger.Logger, configs *config.Configs) error {
-	importDbFlags := configs.ImportDbConfig
-	importDbFlags.ImportDbNoSigCheckFlag = importDbFlags.ImportDbNoSigCheckFlag && importDbFlags.IsImportDBMode
-	importDbFlags.ImportDbSaveTrieEpochRootHash = importDbFlags.ImportDbSaveTrieEpochRootHash && importDbFlags.IsImportDBMode
+	if configs.FlagsConfig.EnablePprof {
+		runtime.SetMutexProfileFraction(5)
```
iulianpascalau, line 521: what is this?

iulianpascalau (changes requested):

cmd/node/flags.go
```diff
@@ -507,26 +517,66 @@ func getWorkingDir(workingDir string, log logger.Logger) string {
 }
 
 func applyCompatibleConfigs(log logger.Logger, configs *config.Configs) error {
-	importDbFlags := configs.ImportDbConfig
-	importDbFlags.ImportDbNoSigCheckFlag = importDbFlags.ImportDbNoSigCheckFlag && importDbFlags.IsImportDBMode
```
iulianpascalau, line 511: thinking again, we need to keep this in a form or another as we might specify the no sig check flag but not enable the import-db main flag, yielding in a very very bad configuration

iulianpascalau (changes requested):

cmd/node/flags.go
bogdan-rosianu: the `import-db` mode is optional and a not breaking change, since the old flag is perfectly valid

cmd/node/operationModes.go
bogdan-rosianu: added lite observer mode. it won't be compatible with historical balances and full archive, since the old data is removed

cmd/node/operationModes.go
bogdan-rosianu: added

cmd/node/flags.go
bogdan-rosianu: done

cmd/node/flags.go
bogdan-rosianu: I thought about it, but if I do so, the node operators won't be able to use their preferred key. There aren't other config flags that should be changed, so I didn't include this mode. Can come in the future

cmd/node/flags.go
bogdan-rosianu: it is the setting for allowing the mutex profiling. It's only set when the profile mode is enabled.
It's old code, just moved

cmd/node/operationModes.go
bogdan-rosianu: refactored and added unit tests

cmd/node/flags.go
iulianpascalau: But it won't work if you just specify the operation mode.
Why? Because the `import-db` flag will hold the empty string, meaning that the source directory will be the same as the current directory and then the source directory will be equal to the destination directory leading to a failed operation.

cmd/node/flags.go
iulianpascalau: ok

cmd/node/flags.go
iulianpascalau: right, did not see that you've moved the code

cmd/node/flags.go
iulianpascalau: Great. Now the node will signal a critical error and stop. :+1:

cmd/node/flags.go
bogdan-rosianu: ok, removed the `import-db` from the operation modes list

common/operationmodes/operationmodes.go
```diff
@@ -12,6 +11,10 @@ const (
 
 // CheckOperationModes will check the compatibility of the provided operation modes and return an error if any
 func CheckOperationModes(modes []string) error {
+	if len(modes) == 0 {
```
iulianpascalau, line 14: :+1:

iulianpascalau (approved):

andreibancioiu (approved):

gabi-vuls (approved):
system test passed.

andreibancioiu (approved):

iulianpascalau (approved):

PR #4624, opened 2022-10-21
EN-13328: overridable config fields

bogdan-rosianu:
- During network upgrades, the entire `config.toml` should be replaced. This could create issues for node operators that perform changes to the `config.toml`, for example setting the `DbLookupExtension` to `true`. 
  
## Proposed Changes
- Added a section in `prefs.toml` where new values can be updated by simply specifying the path and the new value

## Testing procedure
- try to update as many paths as possible, alongside invalid combinations that should error


common/reflectcommon/structFieldsUpdate.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	splitStr := strings.Split(path, ".")
+
+	value := reflect.ValueOf(structure)
+	if value.Kind() != reflect.Ptr {
+		return errCannotUpdateValueStructure
+	}
+
+	value = value.Elem().FieldByName(splitStr[0])
```
BeniaminDrasovean, line 56: Check len of `splitStr` before this.

common/reflectcommon/structFieldsUpdate.go
```diff
@@ -0,0 +1,139 @@
...
+		err = fmt.Errorf("%w. field name=%s", errCannotSetValue, splitStr[len(splitStr)-1])
+		return
+	}
+
+	return trySetTheNewValue(&value, newValue)
+}
+
+func trySetTheNewValue(value *reflect.Value, newValue string) error {
+	valueKind := value.Kind()
+	fmt.Println(valueKind)
```
BeniaminDrasovean, line 75: log?

BeniaminDrasovean (changes requested):

common/reflectcommon/structFieldsUpdate.go
```diff
@@ -0,0 +1,139 @@
...
+	if !value.IsValid() {
+		value = reflect.Value{}
+		err = fmt.Errorf("%w: %s", errInvalidNestedStructure, fieldName)
+		return
+	}
+
+	return
+}
+
+// AdaptStructureValueBasedOnPath will try to update the value specified at the given path in any structure
```
iulianpascalau, line 30: :+1: great comment

common/reflectcommon/structFieldsUpdate.go
iulianpascalau: right, will need a log here

common/reflectcommon/structFieldsUpdate_test.go
```diff
@@ -0,0 +1,404 @@
...
+		require.ErrorContains(t, err, "cannot cast field <invalid uint32> to kind <uint32>")
+	})
+
+	t.Run("should error when setting invalid uint64", func(t *testing.T) {
+		t.Parallel()
+
+		path := "TrieSyncStorage.SizeInBytes"
+		expectedNewValue := "invalid uint64"
+		cfg := &config.Config{}
+		cfg.TrieSyncStorage.SizeInBytes = 37
```
iulianpascalau, line 79: :+1:

iulianpascalau (commented):
GJ :+1: 
small stuff

common/reflectcommon/structFieldsUpdate.go
bogdan-rosianu: removed print, it was just used for testing. no need for the log, as it will be printed on the higher level (even if the overloading is successful or not)

common/reflectcommon/structFieldsUpdate.go
BeniaminDrasovean: Ok

BeniaminDrasovean (approved):

iulianpascalau (approved):

PR #4626, opened 2022-10-21
Heartbeat v2 monitor better filtering

iulianpascalau:
- after some **mainnet** node upgrades, the observed heartbeat status response randomly showed active or inactive nodes in the network. This was caused by the fact that the node returned all messages (inactive or active) regardless that the public key is now behind another pid
  
## Proposed Changes
- made the heartbeat v2 monitor struct better filter-out inactive messages by using the following algorithm:

suppose that, for the same BLS public key we got `n` messages from `n` different pids:
1. if all stored messages show inactive nodes, only keep the latest message, remove the older ones, and put the value `0` on the `NumInstances` field from the `data.PubKeyHeartbeat` struct;
2. if all stored messages show only active nodes, sort the messages based on the `pid` (so that the response will be consistent on all requests), take the first active heartbeat message, and set the `NumInstances` field from `data.PubKeyHeartbeat` struct to how many active messages the node has;
3. if some stored messages show inactive nodes and others show active nodes, remove the stored messages that show inactive nodes, and for the rest of the messages that show active nodes, apply point 2. 

## Testing procedure
- standard system test
- mainnet-specific tests that will prove that the new code will better filter-out inactive messages

### DIFF
--- a/heartbeat/monitor/heartbeatMessage.go
+++ b/heartbeat/monitor/heartbeatMessage.go
@@ -1,0 +1,62 @@
+package monitor
+
+import (
+	"sort"
+
+	"github.com/ElrondNetwork/elrond-go-core/core"
+	"github.com/ElrondNetwork/elrond-go/heartbeat/data"
+)
+
+type heartbeatMessages struct {
+	activePids         []core.PeerID
+	activeHeartbeats   map[core.PeerID]*data.PubKeyHeartbeat
+	inactiveHeartbeats map[core.PeerID]*data.PubKeyHeartbeat
+}
+
+func newHeartbeatMessages() *heartbeatMessages {
+	return &heartbeatMessages{
+		activePids:         make([]core.PeerID, 0),
+		activeHeartbeats:   make(map[core.PeerID]*data.PubKeyHeartbeat),
+		inactiveHeartbeats: make(map[core.PeerID]*data.PubKeyHeartbeat),
+	}
+}
+
+func (instance *heartbeatMessages) addMessage(pid core.PeerID, message *data.PubKeyHeartbeat) {
+	if message.IsActive {
+		instance.activePids = append(instance.activePids, pid)
+		instance.activeHeartbeats[pid] = message
+
+		return
+	}
+
+	instance.inactiveHeartbeats[pid] = message
+}
+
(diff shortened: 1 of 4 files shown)


heartbeat/monitor/heartbeatMessage.go
```diff
@@ -0,0 +1,62 @@
...
+		activePids:         make([]core.PeerID, 0),
+		activeHeartbeats:   make(map[core.PeerID]*data.PubKeyHeartbeat),
+		inactiveHeartbeats: make(map[core.PeerID]*data.PubKeyHeartbeat),
+	}
+}
+
+func (instance *heartbeatMessages) addMessage(pid core.PeerID, message *data.PubKeyHeartbeat) {
+	if message.IsActive {
+		instance.activePids = append(instance.activePids, pid)
+		instance.activeHeartbeats[pid] = message
```
dragos-rebegea, line 27:
```suggestion
                if (pid < instance.activePid) {
		    instance.activePid = pid
		    instance.activeHeartbeat = message
		}
```
this way you just store the message for the lowerst peerID, so you dont have to sort to find it after that

sstanculeanu (commented):
no other comments

heartbeat/monitor/heartbeatMessage.go
iulianpascalau: refactored, now the `heartbeatMessages` struct does not contain any maps. Removed several for... loops in the process. 
Good suggestion :+1:

dragos-rebegea (approved):

heartbeat/monitor/heartbeatMessage.go
```diff
@@ -0,0 +1,86 @@
+package monitor
+
+import (
+	"github.com/ElrondNetwork/elrond-go-core/core"
+	"github.com/ElrondNetwork/elrond-go/heartbeat/data"
+)
+
+type heartbeatMessages struct {
+	firstActivePid          core.PeerID
```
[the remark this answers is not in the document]
iulianpascalau: yes, but we need a decode for that as it is already base58 encoded. I would keep this as it is for simplicity and ease of understanding

sstanculeanu (approved):

gabi-vuls (approved):
System test passed.

PR #4630, opened 2022-10-24
Pseudo slashing fixes

ssd04:
- Error on creating the aggregated signature on leader was not treated accordingly. Since the sig share is not checked by each node separately, it can happen that an invalid format sig share is provided, so signature aggregation will fail
  
## Proposed changes
- handler invalid signers similar as in the verification case

## Testing procedure
- N/A

## Pre-requisites

Based on the [Contributing Guidelines](CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:

### DIFF
--- a/consensus/spos/bls/subroundEndRound.go
+++ b/consensus/spos/bls/subroundEndRound.go
@@ -299,20 +299,16 @@ func (sr *subroundEndRound) doEndRoundJobByLeader() bool {
 	}
 
 	// Aggregate sig and add it to the block
-	aggSigVerificationHandled := false
 	sig, err := sr.SignatureHandler().AggregateSigs(bitmap, sr.Header.GetEpoch())
 	if err != nil {
 		log.Debug("doEndRoundJobByLeader.AggregateSigs", "error", err.Error())
 
-		aggSigVerificationHandled = true
 		bitmap, sig, err = sr.handleInvalidSignersOnAggSigFail()
 		if err != nil {
-			log.Debug("doEndRoundJobByLeader.computeAggSigOnValidNodes", "error", err.Error())
+			log.Debug("doEndRoundJobByLeader.handleInvalidSignersOnAggSigFail", "error", err.Error())
 			return false
 		}
-	}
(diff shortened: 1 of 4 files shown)


consensus/spos/bls/subroundEndRound.go
```diff
@@ -302,39 +302,28 @@ func (sr *subroundEndRound) doEndRoundJobByLeader() bool {
 	sig, err := sr.SignatureHandler().AggregateSigs(bitmap, sr.Header.GetEpoch())
```
SebastianMarian, line 302: I think you can extract L301-L328 in a separate method which will return bitmap, sig and err, actually you can directly return from there sr.handleInvalidSignersOnAggSigFail(), and make the code easier to read there without so much if else levels. Here you should have only something like this:

```
bitmap, sig, err := sr.aggregateSigsAndHandleInvalidSigners(bitmap)
if err != nil {
	log.Debug("doEndRoundJobByLeader.aggregateSigsAndHandleInvalidSigners", "error", err.Error())
	return false
}

```

consensus/spos/bls/subroundEndRound.go
```diff
@@ -491,6 +477,32 @@ func (sr *subroundEndRound) getFullMessagesForInvalidSigners(invalidPubKeys []st
 	return invalidSigners, nil
 }
 
+func (sr *subroundEndRound) handleInvalidSignersOnAggSigFail() ([]byte, []byte, error) {
+	invalidPubKeys, err := sr.verifyNodesOnAggSigFail()
+	if err != nil {
+		log.Debug("doEndRoundJobByLeader.verifyNodesOnAggSigVerificationFail", "error", err.Error())
```
SebastianMarian, line 483: .verifyNodesOnAggSigFail

AdoAdoAdo (commented):
no additional comments

AdoAdoAdo (approved):

SebastianMarian (approved):

PR #4634, opened 2022-10-24
New metric erd_accounts_snapshot_num_nodes

sstanculeanu:
- new metric is needed in order to provide the max number of nodes for trie sync
  
## Proposed changes
- added new metric `erd_accounts_snapshot_num_nodes`

## Testing procedure
- proper tests on feat branch with all components done

## Pre-requisites

Based on the [Contributing Guidelines](CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:


trie/statistics/trieStatisticsCollector.go
```diff
@@ -55,6 +55,11 @@ func (tsc *trieStatisticsCollector) Print() {
 	)
 }
 
+// GetNumNodes returns the number of nodes
```
iulianpascalau, line 58: I know this is a problem not related to your PR but we need to fix this:
L36 can easily panic if I send a nil object. Either nil check or change the definition of the Add function to pass the argument by value

sstanculeanu: pushed, passing by value now

state/snapshotStatistics.go
```diff
@@ -52,7 +52,7 @@ func (ss *snapshotStatistics) AddTrieStats(trieStats *statistics.TrieStatsDTO) {
 	ss.mutex.Lock()
 	defer ss.mutex.Unlock()
 
-	ss.trieStatisticsCollector.Add(trieStats)
+	ss.trieStatisticsCollector.Add(*trieStats)
```
iulianpascalau, line 55: this can easily panic if I send the trie stats as nil. We need to change here also

sstanculeanu: pushed, kept using pointer but added check if nil

BeniaminDrasovean (approved):

iulianpascalau (approved):

PR #4686, opened 2022-11-11
Data trie value

BeniaminDrasovean:
- In order to parse the value from a data trie leaf differentely (based on the activation of AutoBalanceDataTriesFlag), a new object that will handle the parsing needs to be passed to `GetAllLeavesOnChannel()`. 
  
## Proposed changes
- Pass a `disabledTrieLeafParser()` to `GetAllLeavesOnChannel()`. This will be changed to a new implementation of `trieLeafParser` in a future PR.

## Pre-requisites


state/dataTrieValue/dataTrieValue.pb.go
```diff
@@ -0,0 +1,499 @@
...
+var _ = fmt.Errorf
+var _ = math.Inf
+
+// This is a compile-time assertion to ensure that this generated file
+// is compatible with the proto package it is being compiled against.
+// A compilation error at this line likely means your copy of the
+// proto package needs to be updated.
+const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
+
+type TrieLeafData struct {
```
ssd04, line 29: `TrieLeafData` is not being used yet, it will be used when returning parsing the leafs, or elsewhere? because in the interface it returns core.KeyValueHolder right now

BeniaminDrasovean: Yes, it will be used in the next PR, in which a real `trieLeafParser` will be implemented.

ssd04 (dismissed):

state/disabled/disabledTrieLeafParser.go
```diff
@@ -0,0 +1,24 @@
+package disabled
+
+import (
+	"github.com/ElrondNetwork/elrond-go-core/core"
+	"github.com/ElrondNetwork/elrond-go-core/core/keyValStorage"
+)
+
+type disabledTrieLeafParser struct {
```
iulianpascalau, line 8: maybe rename this to `trieLeafParserV1` ? 
As this will be the first version we have implemented

And the package should be `parsers` ?

BeniaminDrasovean: Refactored.

iulianpascalau (approved):

ssd04 (approved):

PR #4699, opened 2022-11-14
Bootstrap from storage or from epoch start for sovereign chain

SebastianMarian:

- To be able to support the sovereign chain implementation the EpochStartBootstrap, StorageEpochStartBootstrap and StorageBootstrap components should be adapted to the new approach by making a new version and use as much as we can the common code used also in the regular chain (main chain)
  
## Proposed changes

- Added support for sovereign chain bootstrapper from storage and from epoch start
- Added support for all the factories needed for sovereign chain components (EpochStartBootstrap, StorageEpochStartBootstrap, StorageBootstrap, etc.)

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/sync/storageBootstrap/baseStorageBootstrapper.go
+++ b/process/sync/storageBootstrap/baseStorageBootstrapper.go
@@ -79,6 +79,8 @@ type storageBootstrapper struct {
 	epochNotifier                process.EpochNotifier
 	processedMiniBlocksTracker   process.ProcessedMiniBlocksTracker
 	appStatusHandler             core.AppStatusHandler
+	getScheduledRootHashMethod   func(headerFromStorage data.HeaderHandler, headerHash []byte) []byte
+	setScheduledInfoMethod       func(headerInfo bootstrapStorage.BootstrapData)
 }
 
 func (st *storageBootstrapper) loadBlocks() error {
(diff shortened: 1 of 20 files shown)


process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -295,6 +296,19 @@ func (st *storageBootstrapper) applyHeaderInfo(hdrInfo bootstrapStorage.Bootstra
 	return nil
 }
 
+func (st *storageBootstrapper) getScheduledRootHash(headerFromStorage data.HeaderHandler, headerHash []byte) []byte {
+	rootHash := headerFromStorage.GetRootHash()
+
+	scheduledRootHash, err := st.scheduledTxsExecutionHandler.GetScheduledRootHashForHeader(headerHash)
+	if err == nil {
```
sasurobert, line 303: a log.Warn in case of error ? This is old code, but still.

sasurobert (approved):

process/sync/storageBootstrap/baseStorageBootstrapper.go
SebastianMarian: This fi executes when there is not an error -> err == nil

iulianpascalau (approved):

PR #4703, opened 2022-11-15
fix account query options for altered accounts

bogdan-rosianu:
- account query options were not set correctly when fetching the altered accounts
  
## Proposed changes
- set all available options, since the data is there

## Testing procedure
full system test + test altered accounts as follows:
- find a block that has altered accounts (let's say nonce 50)
- query `<node-ip>:<node-port>/block/altered-accounts/50`
- choose a random address and query `<node-ip>:<node-port>/address/<address from the previous step>?blockNonce=50` - the balance should be identical to the one obtained in the previous step
- search a block where the same address had another transaction and query the same 2 endpoints, looking for the exact same balance

## Pre-requisites


node/external/blockAPI/baseBlock.go
```diff
@@ -409,19 +409,24 @@ func shouldIncludeAllTokens(tokensFilter string) bool {
...
+
 	alteredAccountsOptions := shared.AlteredAccountsOptions{
 		WithCustomAccountsRepository: true,
 		AccountsRepository:           bap.accountsRepository,
-		// TODO: AccountQueryOptions could be used like options.WithBlockNonce(..) instead of thinking what to provide
-
-		// send the block nonce as it guarantees the opening of the storer for the right epoch. Sending the block root hash
-		// would be more optimal, but there is no link between a root hash and a block, which can result in the endpoint
-		// not working
 		AccountQueryOptions: api.AccountQueryOptions{
```
mariusmihaic, line 425: Shouldn't we fill scheduled rootHash? Or don't we have this info?

mariusmihaic: Or isn't it needed?

node/external/blockAPI/baseBlock.go
```diff
@@ -409,19 +411,28 @@ func shouldIncludeAllTokens(tokensFilter string) bool {
...
-			BlockNonce: core.OptionalUint64{
-				HasValue: true,
-				Value:    apiBlock.Nonce,
-			},
```
mariusmihaic, line 424: Are these all the needed/required/critical field?

factory/api/apiResolverFactory.go
```diff
@@ -529,19 +529,20 @@ func createAPIBlockProcessorArgs(args *ApiResolverArgs, apiTransactionHandler ex
...
+		HistoryRepo:                  args.ProcessComponents.HistoryRepository(),
+		APITransactionHandler:        apiTransactionHandler,
+		StatusComputer:               statusComputer,
+		AddressPubkeyConverter:       args.CoreComponents.AddressPubKeyConverter(),
+		Hasher:                       args.CoreComponents.Hasher(),
+		LogsFacade:                   logsFacade,
+		ReceiptsRepository:           args.ProcessComponents.ReceiptsRepository(),
+		AlteredAccountsProvider:      alteredAccountsProvider,
+		AccountsRepository:           args.StateComponents.AccountsRepository(),
+		ScheduledTxsExecutionHandler: args.ProcessComponents.ScheduledTxsExecutionHandler(),
```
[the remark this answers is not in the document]
bogdan-rosianu: It shouldn't be a problem, since I only use a function used for reading. Anyway, I've followed the same mechanism found in `nodeLoadAccounts.go`

node/external/blockAPI/baseBlock.go
bogdan-rosianu: I've followed the same mechanism found in `nodeLoadAccounts.go`. Right now, no, but in the future, they might be used. It can remain like this since I don't have to perform any additional computation, all the data is already there

factory/api/apiResolverFactory.go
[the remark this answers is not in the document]
bogdan-rosianu: :eyes:

mariusmihaic (dismissed):

AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

gabi-vuls (approved):

mariusmihaic (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):

PR #4713, opened 2022-11-18
use bech32 address in trie statistics

BeniaminDrasovean:
- When printing the trie statistics, the account address is displayed in hex format instead of bech 32
  
## Proposed changes
- Use bech32 for address when printing trie statistics

## Testing procedure
- Normal testing

## Pre-requisites


trie/statistics/trieStatistics.go
```diff
@@ -143,8 +146,18 @@ type TrieStatsDTO struct {
 
 // ToString returns the collected statistics as a string array
 func (tsd *TrieStatsDTO) ToString() []string {
+	addressConverter, err := pubkeyConverter.NewBech32PubkeyConverter(addressLength, log)
```
iulianpascalau, line 149: we should not create a bech32 pubkey converter each time we call this String method. Suggestion to pass to this String method a converter interface

iulianpascalau (changes requested):

trie/statistics/trieStatistics.go
```diff
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/ElrondNetwork/elrond-go-core/core"
-	"github.com/ElrondNetwork/elrond-go-core/core/pubkeyConverter"
 )
 
 const addressLength = 32
```
iulianpascalau: from linter: unused const

trie/statistics/trieStatistics.go
```diff
@@ -146,18 +145,8 @@ type TrieStatsDTO struct {
...
-	}
-
-	address := ""
-	if len(tsd.Address) == addressLength {
-		address = addressConverter.Encode(tsd.Address)
-	}
-
 	stats := make([]string, 0)
-	stats = append(stats, fmt.Sprintf("address %v,", address))
+	stats = append(stats, fmt.Sprintf("address %v,", tsd.Address))
```
iulianpascalau, line 147: do we have tsd.Address as string or bytes? Because if we have as bytes, %v will print something different than a bech32-string

state/accountsDB.go
```diff
@@ -1339,7 +1345,8 @@ func (adb *AccountsDB) snapshotUserAccountDataTrie(
 			ErrChan:    iteratorChannels.ErrChan,
 		}
 		if isSnapshot {
-			adb.mainTrie.GetStorageManager().TakeSnapshot(account.Address, account.RootHash, mainTrieRootHash, iteratorChannelsForDataTries, missingNodesChannel, stats, epoch)
+			address := adb.addressConverter.Encode(account.Address)
```
iulianpascalau, line 1348: does this impact the performance?

trie/statistics/trieStatistics.go
BeniaminDrasovean: Solved.

trie/statistics/trieStatistics.go
BeniaminDrasovean: Good point. It was stored as `[]byte`, will change to `string`

state/accountsDB.go
BeniaminDrasovean: One `Encode` operation takes ~800 ns/op

iulianpascalau (approved):

sstanculeanu (approved):

gabi-vuls (approved):
System test passed

sstanculeanu (approved):

iulianpascalau (approved):

PR #4725, opened 2022-11-22
index only non empty genesis miniblocks

ssd04:
- Fix warn for index of last tx processed, this is an edge case for genesis block on api request
  
## Proposed changes
- Index only non empty miniblocks

## Testing procedure
- Check the logs, there should be no warns like `extractExecutedTxHashes encountered invalid indices` when fetching genesis block on api. The api should be fetched soon after the testnet is started, because after several epoch, the genesis data will not be in storage anymore, so it will not be available on the api
- Check if the genesis data is indexed correctly

## Pre-requisites


bogdan-rosianu (approved):

gabi-vuls (approved):
system test passed

AdoAdoAdo (approved):

PR #4728, opened 2022-11-22
EN-13476: Make esdt supply endpoint work for collections

dragos-rebegea:

- Currently, the esdt supply endpoint only works for a nonce of an SFT/MetaESDT. for example, you can know the total supply of LKMEX-aab910-01, but you cannot know the supply of LKMEX-aab910.
  
## Proposed changes

- Extend the esdtSuppliesProcessor to also save the total amount for a collection. For example, when there is a local mint for LKMEX-aab910-01, reflect the same change on the LKMEX-aab910 as well 

## Testing procedure
- Issue a non-fungible token (ABC-123456)
- Mint some tokens with nonce 1 (ABC-123456-01)
- Mint some tokens with nonce 2 (ABC-123456-02)
- Check the correct supply & minted values for both nonce [1](esdt/supply/ABC-123456-01) and [2](esdt/supply/ABC-123456-02)
- Check the correct supply & minted for the [whole collection](esdt/supply/ABC-123456) (the values are the sum for each nonce of the token) 
- Burn some tokens with nonce 1
- Burn some tokens with nonce 2
- Check the correct burned values for both nonce [1](esdt/supply/ABC-123456-01) and [2](esdt/supply/ABC-123456-02)
- Check the correct burned for the [whole collection](esdt/supply/ABC-123456) (the values are the sum for each nonce of the token)

## Pre-requisites


dblookupext/esdtSupply/logsProcessor.go
```diff
@@ -127,20 +129,39 @@ func (lp *logsProcessor) processEvent(txLog *transaction.Event, supplies map[str
...
+	err := lp.updateOrCreateTokenSupply(tokenIdentifier, valueFromEvent, string(txLog.Identifier), supplies, isRevert)
+	if err != nil {
+		return err
+	}
+
+	if isESDTFungible {
+		return nil
+	}
+
+	collectionIdentifier := txLog.Topics[0]
```
bogdan-rosianu, line 141: same as `tokenIdentifier` declared on L120

dragos-rebegea: tokenIdentifier is updated `if len(txLog.Topics[1]) != 0`

bogdan-rosianu: right

bogdan-rosianu (approved):

PR #4750, opened 2022-11-30
Load private keys

iulianpascalau:
- added the possibility to load multiple private keys to be handled by the node
  
## Proposed changes
- added new .pem file flag
- modified the crypto factory to take into account this new flag and apply the newly handled keys. In this scenario, the node will autogenerate a public key for use in the normal operation (just as it does in the backup operation)

## Testing procedure
- standard system test
- system test with 4 observers, each on a different shard, allValidatorsKey.pem containing all private keys registered on the network, the chain should advance normally

## Pre-requisites

### DIFF
--- a/factory/crypto/cryptoComponents.go
+++ b/factory/crypto/cryptoComponents.go
@@ -73,11 +73,12 @@ type cryptoComponentsFactory struct {
 
 // cryptoParams holds the node public/private key data
 type cryptoParams struct {
-	publicKey       crypto.PublicKey
-	privateKey      crypto.PrivateKey
-	publicKeyString string
-	publicKeyBytes  []byte
-	privateKeyBytes []byte
+	publicKey          crypto.PublicKey
+	privateKey         crypto.PrivateKey
+	publicKeyString    string
+	publicKeyBytes     []byte
+	privateKeyBytes    []byte
+	handledPrivateKeys [][]byte
 }
 
 // cryptoComponents struct holds the crypto components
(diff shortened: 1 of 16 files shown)


factory/crypto/cryptoComponents.go
```diff
@@ -197,6 +201,13 @@ func (ccf *cryptoComponentsFactory) Create() (*cryptoComponents, error) {
 		return nil, err
 	}
 
+	for _, skBytes := range cp.handledPrivateKeys {
```
sstanculeanu, line 204: I think this for loop can be avoided :thinking: 

what if we move L143-L146 on L204 and pass `managedPeersHolder` as parameter on `createCryptoParams` and further on `processAllHandledKeys`, this `managedPeersHolder.AddManagedPeer` can be done on L399

factory/crypto/cryptoComponents.go
```diff
@@ -352,6 +375,64 @@ func (ccf *cryptoComponentsFactory) getSkPk() ([]byte, []byte, error) {
...
+	return handledPrivateKeys, nil
+}
+
+func (ccf *cryptoComponentsFactory) processPrivatePublicKey(keygen crypto.KeyGenerator, encodedSk []byte, pkString string, index int) ([]byte, error) {
+	skBytes, err := hex.DecodeString(string(encodedSk))
+	if err != nil {
+		return nil, fmt.Errorf("%w for encoded secret key, key index %d", err, index)
+	}
+
+	pkBytes, err := ccf.coreComponentsHolder.ValidatorPubKeyConverter().Decode(pkString)
```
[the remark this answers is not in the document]
iulianpascalau: done, refactored code. Should be safer to use now. Added unit test

factory/crypto/cryptoComponents.go
iulianpascalau: It can be avoided but then, the createCryptoParams function will also add the handled key in the managedPeersHolder. This will mean that the createCryptoParams will do more than one thing and I believe the code will lose its readability. Since this operation is only done once, when the node starts, I do not think performance degradation will be an issue.
I rather keep the code as it is for this consideration.

sstanculeanu (approved):

SebastianMarian (approved):

PR #4751, opened 2022-12-02
HardforkExclusionHandler into consensus

sstanculeanu:
- round should be excluded at consensus level as well
  
## Proposed changes
- integrated hardforkExclusionHandler into subroundStartRound

## Testing procedure
- will be tested on feat branch

## Pre-requisites

### DIFF
--- a/consensus/spos/bls/subroundBlock.go
+++ b/consensus/spos/bls/subroundBlock.go
@@ -20,6 +20,7 @@ type subroundBlock struct {
 	*spos.Subround
 
 	processingThresholdPercentage int
+	hardforkExclusionHandler      common.HardforkExclusionHandler
 }
 
 // NewSubroundBlock creates a subroundBlock object
@@ -27,15 +28,20 @@ func NewSubroundBlock(
 	baseSubround *spos.Subround,
 	extend func(subroundId int),
 	processingThresholdPercentage int,
+	hardforkExclusionHandler common.HardforkExclusionHandler,
 ) (*subroundBlock, error) {
 	err := checkNewSubroundBlockParams(baseSubround)
 	if err != nil {
 		return nil, err
 	}
+	if check.IfNil(hardforkExclusionHandler) {
+		return nil, spos.ErrNilHardforkExclusionHandler
+	}
 
 	srBlock := subroundBlock{
 		Subround:                      baseSubround,
 		processingThresholdPercentage: processingThresholdPercentage,
+		hardforkExclusionHandler:      hardforkExclusionHandler,
 	}
 
 	srBlock.Job = srBlock.doBlockJob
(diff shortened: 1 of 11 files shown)


consensus/spos/bls/subroundBlock.go
```diff
@@ -79,6 +87,10 @@ func (sr *subroundBlock) doBlockJob(ctx context.Context) bool {
 		return false
 	}
 
+	if sr.hardforkExclusionHandler.IsRoundExcluded(uint64(currentRound)) {
```
iulianpascalau, line 90: can we add a similar condition in the `receivedBlockBodyAndHeader` function just after we decoded the header? We need to prevent the validators from signing a received block if it was decided it is excluded.
```GO
sr.Header = sr.BlockProcessor().DecodeBlockHeader(cnsDta.Header)
```

cc @SebastianMarian do you have a better suggestion?

SebastianMarian: I think here the code will have no effect, only if we want to set an exclusion round in the future, as here the round will be always the current round and not the past/synced round. Also in the validator side, I think we should put this check in processReceivedBlock as this method si called either by receivedBlockBody, receivedBlockHeader or receivedBlockBodyAndHeader, depending on how the proposer has been sent the info. But again, only if the purpose of this code is to avoid something in the future rounds.

sstanculeanu: pushed

consensus/spos/bls/subroundStartRound.go
```diff
@@ -121,6 +127,12 @@ func (sr *subroundStartRound) initCurrentRound() bool {
 		return false
 	}
 
+	currentRound := sr.RoundHandler().Index()
+	if sr.hardforkExclusionHandler.IsRoundExcluded(uint64(currentRound)) {
+		log.Debug("initCurrentRound round excluded in hardfork handler", "round", currentRound)
```
iulianpascalau, line 132: :+1:

iulianpascalau (approved):

consensus/spos/bls/subroundStartRound.go
```diff
@@ -121,6 +127,12 @@ func (sr *subroundStartRound) initCurrentRound() bool {
 		return false
 	}
 
+	currentRound := sr.RoundHandler().Index()
```
SebastianMarian, line 130: There are more 4 more places below in this method where you can replace `sr.RoundHandler().Index()` with `currentRound` variable.

SebastianMarian (approved):

consensus/spos/bls/subroundStartRound.go
sstanculeanu: right, will be replaced in the next PR

PR #4757, opened 2022-12-02
Update testnet scripts for multikey

sstanculeanu:
- update testnet scripts to support multikey
  
## Proposed changes
- updated scripts

## Testing procedure
- multikey testnet

## Pre-requisites

### DIFF
--- a/factory/crypto/cryptoComponents.go
+++ b/factory/crypto/cryptoComponents.go
@@ -294,7 +294,7 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
 			return nil, fmt.Errorf("invalid node configuration: import-db mode and allValidatorsKeys.pem file provided")
 		}
 
-		return ccf.generateCryptoParams(keygen, "in import mode", make([][]byte, 0))
+		return ccf.generateCryptoParams(keygen, "in import mode", handledPrivateKeys)
 	}
 	if ccf.noKeyProvided {
 		return ccf.generateCryptoParams(keygen, "with no-key flag enabled", make([][]byte, 0))
--- a/factory/crypto/cryptoComponents.go
+++ b/factory/crypto/cryptoComponents.go
@@ -294,10 +294,10 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
 			return nil, fmt.Errorf("invalid node configuration: import-db mode and allValidatorsKeys.pem file provided")
 		}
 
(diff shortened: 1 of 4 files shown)


factory/crypto/cryptoComponents.go
```diff
@@ -294,10 +294,10 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
 			return nil, fmt.Errorf("invalid node configuration: import-db mode and allValidatorsKeys.pem file provided")
 		}
 
-		return ccf.generateCryptoParams(keygen, "in import mode", handledPrivateKeys)
+		return ccf.generateCryptoParams(keygen, "in import mode", make([][]byte, 0))
```
iulianpascalau, line 297: because of L293 we are sure this line will be called with an empty slice. Can be either reverted or left as it is.

factory/crypto/cryptoComponents.go
```diff
@@ -294,10 +294,10 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
...
 	}
 	if ccf.noKeyProvided {
-		return ccf.generateCryptoParams(keygen, "with no-key flag enabled", handledPrivateKeys)
+		return ccf.generateCryptoParams(keygen, "with no-key flag enabled", make([][]byte, 0))
```
iulianpascalau, line 300: When I first wrote the code I thought that I can provide a `-no-key` flag but will still use the allValidatorsKeys file, if existed. 
No, I think it is better to not use any of the provided keys if the flag was set to `-no-key`. So yes, it will remain like this :+1:

scripts/testnet/include/validators.sh
```diff
@@ -46,6 +50,29 @@ iterateOverValidators() {
   done
 }
 
+iterateOverValidatorsMultiKey() {
+  local callback=$1
+  local VALIDATOR_INDEX=0
+
+  # Iterate over shards and start validators
+  (( max_shard_id=$SHARDCOUNT - 1 ))
+  for SHARD in $(seq 0 1 $max_shard_id); do
```
iulianpascalau, line 59: :+1:

iulianpascalau (commented):
Looks good, will test this

iulianpascalau (dismissed):
Tested, works as expected

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4760, opened 2022-12-05
use numeric instead of bool for snapshots in progress

ssd04:
- Set snapshot status metrics as integers instead of bool strings
  
## Proposed changes
- Update the metric for snaphost status in progress, from `false/true` to `0/1`

## Testing procedure
- check the updated metrics in node/status route when snapshot is in progress:
     - erd_accounts_snapshot_in_progress: 1 if the snapshot is in progress, 0 if it's not
     - erd_peers_snapshot_in_progress: 1 if the snapshot is in progress, 0 if it's not

## Pre-requisites

### DIFF
--- a/node/metrics/metrics.go
+++ b/node/metrics/metrics.go
@@ -48,6 +48,7 @@ func InitBaseMetrics(appStatusHandler core.AppStatusHandler) error {
 	appStatusHandler.SetUInt64Value(common.MetricTrieSyncNumProcessedNodes, initUint)
 	appStatusHandler.SetUInt64Value(common.MetricTrieSyncNumReceivedBytes, initUint)
 	appStatusHandler.SetUInt64Value(common.MetricAccountsSnapshotInProgress, initUint)
+	appStatusHandler.SetUInt64Value(common.MetricPeersSnapshotInProgress, initUint)
 
 	appStatusHandler.SetInt64Value(common.MetricLastAccountsSnapshotDurationSec, initInt)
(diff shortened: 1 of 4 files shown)


node/metrics/metrics.go
```diff
@@ -47,6 +47,8 @@ func InitBaseMetrics(appStatusHandler core.AppStatusHandler) error {
 	appStatusHandler.SetUInt64Value(common.MetricAccountsSnapshotNumNodes, initUint)
 	appStatusHandler.SetUInt64Value(common.MetricTrieSyncNumProcessedNodes, initUint)
 	appStatusHandler.SetUInt64Value(common.MetricTrieSyncNumReceivedBytes, initUint)
+	appStatusHandler.SetUInt64Value(common.MetricAccountsSnapshotInProgress, initUint)
```
iulianpascalau, line 50: Why do we require this change?

ssd04: it was requested by Bogdan and Liviu, to be easier for parsing, not sure if it's a hard requirement though

SebastianMarian (dismissed):

SebastianMarian (approved):

node/metrics/metrics.go
bogdan-rosianu: because it's easier in our monitoring tools to parse these values as int (0-false, 1-true) instead of booleans

bogdan-rosianu (approved):

iulianpascalau (approved):

gabi-vuls (approved):
system test passed

PR #4777, opened 2022-12-08
En 13511 add firehose enabled flag

mariusmihaic:
- Added config flag for enabling firehose

## Proposed changes
- Added config flag + initial implementation. This initial implementation is compatible with [first firehose version](https://github.com/ElrondNetwork/firehose-multiversx) 

## Testing procedure
- Tested with local testnet + firehose blocks ingestion process

## Pre-requisites

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -938,3 +938,6 @@
         Capacity = 1000
         Type = "SizeLRU"
         SizeInBytes = 314572800 #300MB
+
+[FireHoseConfig]
+    Enabled = true
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -938,6 +938,3 @@
         Capacity = 1000
         Type = "SizeLRU"
         SizeInBytes = 314572800 #300MB
-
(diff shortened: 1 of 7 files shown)


cmd/node/config/config.toml
```diff
@@ -938,3 +938,6 @@
         Capacity = 1000
         Type = "SizeLRU"
         SizeInBytes = 314572800 #300MB
+
+[FireHoseConfig]
+    Enabled = true
```
AdoAdoAdo, line 943: by default I think this should be false.
Instead of a config parameter I would add a build flag to enable firehose

AdoAdoAdo (changes requested):

SebastianMarian (dismissed):

cmd/node/config/config.toml
mariusmihaic: Right, good suggestion. I should've started with flag implementation, not config. 
Refactored

AdoAdoAdo (approved):

SebastianMarian (approved):

PR #4791, opened 2022-12-12
Disable in-epoch trie pruning 

iulianpascalau:
- Now, the node does a trie pruning operation on each new committed block. This makes the trie history unavailable after ~10 consecutive final blocks are committed
   
## Proposed changes
- disable in-epoch trie pruning

## Testing procedure
- standard system test in which, after the upgrade, you pick an old block (>50 rounds old block) and query a get operation as follows:
```
http://<proxy>/address/<address>?blockNonce=<block_nonce>
```

## Pre-requisites

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -668,7 +668,7 @@
     CheckpointRoundsModulus = 100
     CheckpointsEnabled = false
     SnapshotsEnabled = true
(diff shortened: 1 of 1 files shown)


SebastianMarian (approved):

schimih (approved):

gabi-vuls (approved):
System test passed.

PR #4821, opened 2022-12-21
Trie leaf data unmarshal

BeniaminDrasovean:
- When parsing a data trie leaf, if the leaf value was not migrated to the new version, the unmarshal should err. There are some cases in which the unmarshall does not err, but the data is still nil.
  
## Proposed changes
- After unmarshal, also check that the trieData is not empty.

## Testing procedure
- added unit test

## Pre-requisites

### DIFF
--- a/state/parsers/dataTrieLeafParser.go
+++ b/state/parsers/dataTrieLeafParser.go
@@ -37,7 +37,7 @@ func (tlp *dataTrieLeafParser) ParseLeaf(trieKey []byte, trieVal []byte) (core.K
 	if tlp.enableEpochsHandler.IsAutoBalanceDataTriesEnabled() {
 		data := &dataTrieValue.TrieLeafData{}
 		err := tlp.marshaller.Unmarshal(data, trieVal)
-		if err == nil {
+		if err == nil && !isEmptyTrieData(data) {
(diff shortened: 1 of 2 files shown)


AdoAdoAdo (dismissed):

state/parsers/dataTrieLeafParser.go
```diff
@@ -37,7 +37,7 @@ func (tlp *dataTrieLeafParser) ParseLeaf(trieKey []byte, trieVal []byte) (core.K
 	if tlp.enableEpochsHandler.IsAutoBalanceDataTriesEnabled() {
 		data := &dataTrieValue.TrieLeafData{}
 		err := tlp.marshaller.Unmarshal(data, trieVal)
-		if err == nil {
+		if err == nil && !isEmptyTrieData(data) {
```
iulianpascalau, line 40: Please add a TODO to remove this after we have a proper fix on the marshaller.

AdoAdoAdo (approved):

state/parsers/dataTrieLeafParser.go
BeniaminDrasovean: Done.

iulianpascalau (approved):

PR #4832, opened 2022-12-22
Firehose block body integration

mariusmihaic:
- Integrate block body and altered accounts into firehose block
  
## Proposed changes
- Integrate block body and altered accounts into firehose block

## Testing procedure
- Local testnet + firehose ingestion process + scripts and tools

## Pre-requisites

### DIFF
--- a/outport/firehose/firehoseIndexer.go
+++ b/outport/firehose/firehoseIndexer.go
@@ -109,6 +109,28 @@ func (fi *firehoseIndexer) SaveBlock(args *outportcore.ArgsSaveBlockData) error 
 	return nil
 }
 
+func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]byte, core.HeaderType, error) {
+	var err error
+	var headerBytes []byte
+	var headerType core.HeaderType
+
+	switch header := headerHandler.(type) {
+	case *block.MetaBlock:
+		headerType = core.MetaHeader
+		headerBytes, err = fi.marshaller.Marshal(header)
+	case *block.Header:
+		headerType = core.ShardHeaderV1
+		headerBytes, err = fi.marshaller.Marshal(header)
+	case *block.HeaderV2:
+		headerType = core.ShardHeaderV2
+		headerBytes, err = fi.marshaller.Marshal(header)
+	default:
+		return nil, "", errInvalidHeaderType
+	}
+
+	return headerBytes, headerType, err
+}
+
 func getBody(bodyHandler data.BodyHandler) (*block.Body, error) {
 	if check.IfNil(bodyHandler) {
 		return &block.Body{}, nil
@@ -158,28 +180,6 @@ func getTokens(tokens []*outportcore.AccountTokenData) []*alteredAccount.Account
 	}
 
 	return ret
-}
-
-func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]byte, core.HeaderType, error) {
-	var err error
-	var headerBytes []byte
-	var headerType core.HeaderType
-
-	switch header := headerHandler.(type) {
-	case *block.MetaBlock:
-		headerType = core.MetaHeader
-		headerBytes, err = fi.marshaller.Marshal(header)
-	case *block.Header:
-		headerType = core.ShardHeaderV1
-		headerBytes, err = fi.marshaller.Marshal(header)
-	case *block.HeaderV2:
-		headerType = core.ShardHeaderV2
-		headerBytes, err = fi.marshaller.Marshal(header)
-	default:
-		return nil, "", errInvalidHeaderType
-	}
-
-	return headerBytes, headerType, err
 }
 
 // RevertIndexedBlock does nothing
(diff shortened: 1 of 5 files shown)


outport/firehose/firehoseIndexer.go
```diff
@@ -122,6 +131,57 @@ func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]b
 	return headerBytes, headerType, err
 }
 
+func getBody(bodyHandler data.BodyHandler) (*block.Body, error) {
+	if check.IfNil(bodyHandler) {
+		return &block.Body{}, nil
```
SebastianMarian, line 136: So if bodyHandler is nil we will return an empty block body ?

outport/firehose/firehoseIndexer.go
```diff
@@ -122,6 +131,57 @@ func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]b
...
+func getBodyOrNilIfEmpty(body *block.Body) *block.Body {
+	if len(body.MiniBlocks) == 0 {
+		return nil
+	}
+
+	return body
+}
+
+func getAlteredAccounts(accounts map[string]*outportcore.AlteredAccount) []*alteredAccount.AlteredAccount {
+	ret := make([]*alteredAccount.AlteredAccount, 0, len(accounts))
```
SebastianMarian, line 156: alteredAccounts instead ret ?

outport/firehose/firehoseIndexer.go
```diff
@@ -122,6 +131,57 @@ func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]b
...
+			Balance: acc.Balance,
+			Tokens:  getTokens(acc.Tokens),
+		})
+	}
+
+	return ret
+}
+
+func getTokens(tokens []*outportcore.AccountTokenData) []*alteredAccount.AccountTokenData {
+	ret := make([]*alteredAccount.AccountTokenData, len(tokens))
```
SebastianMarian, line 171: alteredAccounts instead ret ?

schimih (approved):

schimih (approved):

outport/firehose/firehoseIndexer.go
```diff
@@ -68,16 +69,24 @@ func (fi *firehoseIndexer) SaveBlock(args *outportcore.ArgsSaveBlockData) error
 		return err
 	}
 
+	body, err := getBody(args.Body)
```
AdoAdoAdo, line 72: if I understood correctly, you return empty for nil, and then with getBodyOrNilIfEmpty you return nil for empty.
Are these two methods used in other places? 
If these are only used here I would simply return nil for nil and remove getBodyOrNilIfEmpty.

outport/firehose/firehoseIndexer.go
```diff
@@ -122,6 +131,57 @@ func (fi *firehoseIndexer) getHeaderBytes(headerHandler data.HeaderHandler) ([]b
...
+	}
+
+	return body
+}
+
+func getAlteredAccounts(accounts map[string]*outportcore.AlteredAccount) []*alteredAccount.AlteredAccount {
+	ret := make([]*alteredAccount.AlteredAccount, 0, len(accounts))
+
+	for _, acc := range accounts {
+		ret = append(ret, &alteredAccount.AlteredAccount{
```
AdoAdoAdo, line 159: It seems you always have exactly len(accounts) elements, so I would do simple assignment instead of append in this case, like you did with tokens.

`alteredAccounts[i] = *alteredAccount.AlteredAccount{...}`
and also change the initialization of alteredAccount to `make([]*alteredAccount.AlteredAccount, len(accounts))`

AdoAdoAdo (changes requested):

outport/firehose/firehoseIndexer.go
mariusmihaic: Refactored the code to return `nil`, `errNilBlockBody` in case of nil block body.
In the caller code, however, I check if the error is `errNilBlockBody` and, if so, I fill the body field directly with nil

outport/firehose/firehoseIndexer.go
mariusmihaic: Both input and output would have the same name and it would be confusing. 
This applies for all "converting funcs" such as tokens, altered accounts, etc.
Would keep it as it is.

outport/firehose/firehoseIndexer.go
mariusmihaic: Both input and output would have the same name and it would be confusing. 
This applies for all "converting funcs" such as tokens, altered accounts, etc.
Would keep it as it is.

outport/firehose/firehoseIndexer.go
mariusmihaic: Refactored the code to return `nil`, `errNilBlockBody` in case of nil block body.
In the caller code, however, I check if the error is `errNilBlockBody` and, if so, I fill the body field directly with nil

outport/firehose/firehoseIndexer.go
mariusmihaic: Right, good suggestion. Applied it.

AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

schimih (approved):

PR #4847, opened 2023-01-06
Guarded txs relayed

AdoAdoAdo:
- relay txs V1 and V2 did not consider the guardian use-case
- possibility of network split due to checking the state of an account guardian on the interceptor instead of on execution
  
## Proposed changes
- Take care to consider also guardian signature check on the relay v1 and v2
- Move the checking of active guardian (on account) from interceptor to execution

## Testing procedure
- use relayed txs with guarded and unguarded transactions scenarios and verify the results
- run all previous guardian scenarios and verify results

## Pre-requisites


sstanculeanu (dismissed):

process/transaction/baseProcess.go
```diff
@@ -218,16 +218,40 @@ func (txProc *baseTxProcessor) VerifyTransaction(tx *transaction.Transaction) er
...
 }
 
+// Setting a guardian is allowed with regular transactions on a guarded account
+// but in this case is set with the default epochs delay
+func checkOperationAllowedToBypassGuardian(txData []byte) error {
+	if process.IsSetGuardianCall(txData) {
+		return nil
+	}
+
+	return process.ErrOperationNotPermitted
```
iulianpascalau, line 228: by returning this error here, do we consider this transaction as invalid (included in an invalid block) and consumed all the gas? Or it is completely removed from the pool as it was never broadcast?

process/transaction/interceptedTransaction.go
```diff
@@ -152,17 +152,19 @@ func createRelayedV2(relayedTx *transaction.Transaction, args [][]byte) (*transa
...
+		RcvAddr:           args[0],
+		SndAddr:           relayedTx.RcvAddr,
+		GasPrice:          relayedTx.GasPrice,
+		GasLimit:          0, // the user had to sign a transaction with 0 gasLimit - as all gasLimit is coming from the relayer
+		Data:              args[2],
+		ChainID:           relayedTx.ChainID,
+		Version:           relayedTx.Version,
+		Signature:         args[3],
+		Options:           relayedTx.Options,
+		GuardianAddr:      nil,
```
iulianpascalau, line 166: :+1:

process/transaction/interceptedTransaction.go
```diff
@@ -266,6 +278,11 @@ func (inTx *InterceptedTransaction) verifyIfRelayedTx(tx *transaction.Transactio
 		return err
 	}
 
+	err = inTx.VerifyGuardianSig(userTx)
```
iulianpascalau, line 281: :+1:

process/transaction/interceptedTransaction.go
```diff
@@ -221,6 +228,11 @@ func (inTx *InterceptedTransaction) verifyIfRelayedTxV2(tx *transaction.Transact
...
```
iulianpascalau, line 231: :+1:

process/transaction/interceptedTransaction.go
```diff
@@ -364,6 +381,32 @@ func (inTx *InterceptedTransaction) verifySig(tx *transaction.Transaction) error
...
 
+// VerifyGuardianSig verifies if the guardian signature is valid
+func (inTx *InterceptedTransaction) VerifyGuardianSig(tx *transaction.Transaction) error {
+	txMessageForSigVerification, err := inTx.getTxMessageForGivenTx(tx)
+	if err != nil {
+		return err
+	}
+
+	if !inTx.txVersionChecker.IsGuardedTransaction(tx) {
+		if len(tx.GetGuardianAddr()) > 0 {
```
iulianpascalau, line 392: L392-L399 could have been a separate function named "verifyConsistencyForNotGuardedTx" or something similar.

process/transaction/interceptedTransaction.go
```diff
@@ -364,6 +381,32 @@ func (inTx *InterceptedTransaction) verifySig(tx *transaction.Transaction) error
...
+			return fmt.Errorf("%w without guardian signature", process.ErrGuardianAddressNotExpected)
```
iulianpascalau, line 393: might return the errors directly here and on L396 as they are describing what they do. 
Otherwise we would have something like: "guardian signature not expected without guardian signature"

iulianpascalau (commented):
:+1:  
Small stuff.

process/transaction/baseProcess.go
raduchis: same also for line 267?
if the scammer signs transactions with a different guardian he could pay the fees. 

`if !bytes.Equal(guardian, tx.GuardianAddr) {
		return process.ErrTransactionAndAccountGuardianMismatch
	}`

raduchis (commented):
just a small response to @iulianpascalau

process/transaction/baseProcess.go
iulianpascalau: wondering if this can be an attack vector as the scammer increments the account's nonce

process/transaction/interceptedTransaction.go
AdoAdoAdo: done

process/transaction/interceptedTransaction.go
AdoAdoAdo: done

process/transaction/baseProcess.go
AdoAdoAdo: It should not be possible to spend the gas, as this is a bad transaction and does not consume fee.
Did some change to remove these transactions also from the txpool.
If a guarded account has a pending guardian already set, then transactions to replace the pending guardian will be ignored.

AdoAdoAdo: moved the verification to (txProc *baseTxProcessor) checkTxValues so that it covers also scheduled txs

raduchis (approved):

iulianpascalau (approved):

process/transaction/baseProcess.go
```diff
@@ -218,30 +221,54 @@ func (txProc *baseTxProcessor) VerifyTransaction(tx *transaction.Transaction) er
...
+
+func (txProc *baseTxProcessor) checkGuardedAccountUnguardedTxPermission(txData []byte, account state.UserAccountHandler) error {
+	err := checkOperationAllowedToBypassGuardian(txData)
+	if err != nil {
+		return err
+	}
+
+	// block non guarded setGuardian Txs if there is a pending guardian
+	hasPendingGuardian := txProc.guardianChecker.HasPendingGuardian(account)
+	if process.IsSetGuardianCall(txData) && hasPendingGuardian {
```
SebastianMarian, line 242: I think the condition `process.IsSetGuardianCall(txData)` is not really necessary, as it is already checked in checkOperationAllowedToBypassGuardian method called in L235. In this point, L242, this condition is always true.

process/transaction/baseProcess_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/ElrondNetwork/elrond-go/state"
+	"github.com/ElrondNetwork/elrond-go/testscommon"
+	"github.com/ElrondNetwork/elrond-go/testscommon/economicsmocks"
+	"github.com/ElrondNetwork/elrond-go/testscommon/guardianMocks"
+	"github.com/ElrondNetwork/elrond-go/testscommon/hashingMocks"
+	stateMock "github.com/ElrondNetwork/elrond-go/testscommon/state"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_checkOperationAllowedToBypassGuardian(t *testing.T) {
```
SebastianMarian, line 21: t.Parallel() here and in each t.Run func ?

process/transaction/baseProcess_test.go
```diff
@@ -0,0 +1,97 @@
...
+		txData := []byte("setGuardian")
+		require.True(t, errors.Is(checkOperationAllowedToBypassGuardian(txData), process.ErrTransactionNotExecutable))
+	})
+	t.Run("set guardian builtin call allowed to bypass", func(t *testing.T) {
+		txData := []byte("SetGuardian@")
+		require.Nil(t, checkOperationAllowedToBypassGuardian(txData))
+	})
+}
+
+func Test_checkGuardedAccountUnguardedTxPermission(t *testing.T) {
```
SebastianMarian, line 40: t.Parallel() here and in each t.Run func ?

process/transaction/interceptedTransaction_test.go
```diff
@@ -1774,3 +1774,89 @@ func TestInterceptedTransaction_checkMaxGasPrice(t *testing.T) {
...
+			return 1000
+		},
+	}
+	tx := dataTransaction.Transaction{
+		Data:              []byte("some data"),
+		GuardianAddr:      []byte("guardian addr"),
+		GuardianSignature: []byte("guardian signature"),
+	}
+
+	t.Run("get data for signing with error", func(t *testing.T) {
```
SebastianMarian, line 1797: t.Parallel() in each t.Run func ?

PR #4866, opened 2023-01-11
get last epoch from all boot infos

ssd04:
- Get last epoch from most recent boot infos data

## Testing procedure
- Standard system test

## Pre-requisites


process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -395,7 +395,8 @@ func (st *storageBootstrapper) applyBootInfos(bootInfos []bootstrapStorage.Boots
 		st.forkDetector.SetFinalToLastCheckpoint()
 	}
 
-	err = st.nodesCoordinator.LoadState(bootInfos[0].NodesCoordinatorConfigKey, bootInfos[0].LastHeader.GetEpoch())
+	lastEpoch := getBootInfosLastEpoch(bootInfos)
+	err = st.nodesCoordinator.LoadState(bootInfos[0].NodesCoordinatorConfigKey, lastEpoch)
```
AdoAdoAdo, line 399: can you rename getBootInfosLastEpoch into something like `initNodesCoordinator` and move the call
err = st.nodesCoordinator.LoadState(bootInfos[0].NodesCoordinatorConfigKey, lastEpoch)
inside after checking len(bootInfos) ?

AdoAdoAdo (changes requested):

process/sync/storageBootstrap/baseStorageBootstrapper.go
ssd04: remove that function since it was not relevant to have that check, since here we get only if we have at least some entry in boot infos

AdoAdoAdo (approved):

AdoAdoAdo (approved):

sharding/nodesCoordinator/indexHashedNodesCoordinatorRegistry.go
```diff
@@ -70,17 +70,9 @@ func GetNodesCoordinatorRegistry(
...
-		epochConfigBytes, err := storer.SearchFirst(ncInternalkey)
+		err := setEpochConfigPerEpoch(epoch, epochsConfig, storer)
 		if err != nil {
-			return nil, err
-		}
-
-		err = updateEpochsConfig(epochsConfig, epochConfigBytes)
-		if err != nil {
-			return nil, err
+			log.Debug("failed to get nodes coordinator config for epoch", "epoch", epoch)
```
iulianpascalau, line 75: this was considered a critical error (in the previous version of the function)

ssd04: later on we considered that if we found the configuration for the most recent epoch, it is fine to continue, even if for the previous one it was not found; we should not return err if we can continue; this could happen at the upgrade

sharding/nodesCoordinator/indexHashedNodesCoordinatorRegistry_test.go
```diff
@@ -252,23 +252,92 @@ func TestIndexHashedNodesCoordinator_GetNodesCoordinatorRegistry(t *testing.T) {
...
+				case bytes.Equal(append([]byte(common.NodesCoordinatorRegistryKeyPrefix), []byte(fmt.Sprint(10))...), key):
+					return nil, errors.New("first get error")
+				default:
+					nodesConfigRegistryBytes, _ := json.Marshal(nodesConfigRegistry)
+					return nodesConfigRegistryBytes, nil
+				}
+			},
+			PutCalled: func(key, data []byte) error {
+				switch {
+				case bytes.Equal(append([]byte(common.NodesCoordinatorRegistryKeyPrefix), []byte(fmt.Sprint(10))...), key):
```
AdoAdoAdo, line 320: can we create some variables for the cases here? it is hard to read

AdoAdoAdo (changes requested):

sharding/nodesCoordinator/indexHashedNodesCoordinatorRegistry_test.go
ssd04: right, added variables

AdoAdoAdo (approved):

iulianpascalau (approved):

PR #4891, opened 2023-01-20
Transaction ordering component

AdoAdoAdo:
Currently the transactions execution order is re-computed after the actual execution and fed to the outport driver, which serves clients such as indexers or notifiers.
In some cases, e.g multiple smart contract results generated by the same transaction, that need to be executed cross shard in the same destination shard, the execution order for these individual SCRs cannot be correctly determined after execution and all will be treated as a batch and receive the same order.
  
## Proposed changes
Create an ordered collection component that can be used during the Txs execution and collect the transactions and scrs in their correct execution order. 
This PR introduces this component.

## Testing procedure
Created unit tests that fully cover the component.
Regular system test, the component is not integrated with the processors in this PR so there is no relevant scenario that would use the added functionality. Custom scenarios possible after the integration, in a follow up PR.

## Pre-requisites


common/ordering/orederedCollection.go
```diff
@@ -0,0 +1,109 @@
...
+		oc.itemsMap[oc.itemsArray[i]]--
+	}
+}
+
+// GetItems returns the items in the order they were added
+func (oc *orderedCollection) GetItems() []string {
+	oc.mut.RLock()
+	defer oc.mut.RUnlock()
+
+	cpItems := append([]string{}, oc.itemsArray...)
```
iulianpascalau, line 79: I think this is equivalent to:
```GO
cpItems := make([]string, len(oc.itemsArray))
copy(cpItems, oc.itemsArray)
```

common/ordering/orederedCollection_test.go
```diff
@@ -0,0 +1,185 @@
...
+	contains = oc.Contains("two")
+	require.True(t, contains)
+
+	contains = oc.Contains("three")
+	require.False(t, contains)
+}
+
+func TestOrderedCollection_IsInterfaceNil(t *testing.T) {
+	oc := ordering.GetNilOrderedCollection()
+	require.True(t, oc.IsInterfaceNil())
```
iulianpascalau, line 184: this test can be omitted and instead replace L12 with:
```GO
assert.False(t, check.IfNil(oc))
```
also, a concurrency test on all method would be nice just to test that the mutex locking works

common/ordering/orederedCollection.go
AdoAdoAdo: yes, changed.

raduchis (dismissed):

common/ordering/orederedCollection_test.go
AdoAdoAdo: done

iulianpascalau (approved):

PR #4896, opened 2023-01-23
Fix VM memory leak

camilbancioiu:
This PR updates the dependency on the VM v1.4 with a fixed version.


iulianpascalau (approved):

sasurobert (approved):

gabi-vuls (approved):
system test passed

PR #4897, opened 2023-01-23
Time between heartbeats at bootstrap

sstanculeanu:
- during bootstrap, heartbeats will be used to keep track of trie sync progress. Currently, the time between sends is the same as regular heartbeats, 5 min, which may be too much for a proper progress bar
  
## Proposed changes
- added new config value

## Testing procedure
- start a node on testnet. Either check logs for `heartbeat message sent` to be ~1 min, either check https://testnet-gateway.multiversx.com/node/heartbeatstatus and make sure the `numTrieNodesReceived` for your node increases each ~minute during **trie sync**
- after bootstrap, the same log should be after ~5min, and `numTrieNodesReceived` 0

## Pre-requisites


cmd/node/config/config.toml
```diff
@@ -917,6 +917,7 @@
     PeerAuthenticationTimeBetweenSendsWhenErrorInSec = 60   # 1min
     PeerAuthenticationThresholdBetweenSends          = 0.1  # 10%
     HeartbeatTimeBetweenSendsInSec                   = 60   # 1min
+    HeartbeatTimeBetweenSendsDuringBootstrapInSec    = 60   # 1min
```
iulianpascalau, line 920: do we need to have this set to 1 minute on the public testnets & mainnet? The original code used `HeartbeatTimeBetweenSendsInSec` which was set to 5 minutes.

sstanculeanu: exactly, this was the reason behind the PR. While testing on public testnet, I realized that 5 min with no feedback would be too much

iulianpascalau (approved):

SebastianMarian (approved):

gabi-vuls (approved):

PR #4910, opened 2023-01-26
View funcs governance

sasurobert:
Added new view functions for governance
  
## Pre-requisites

### DIFF
--- a/vm/systemSmartContracts/governance.go
+++ b/vm/systemSmartContracts/governance.go
@@ -762,10 +762,6 @@ func (g *governanceContract) checkViewFuncArguments(
 	if args.CallValue.Cmp(zero) != 0 {
 		return vm.ErrCallValueMustBeZero
 	}
-	err := g.eei.UseGas(g.gasCost.MetaChainSystemSCsCost.Vote)
-	if err != nil {
-		return vm.ErrNotEnoughGas
-	}
 	if len(args.Arguments) != numArgs {
 		return vm.ErrInvalidNumOfArguments
 	}
--- a/vm/systemSmartContracts/governance.go
+++ b/vm/systemSmartContracts/governance.go
@@ -146,8 +146,14 @@ func (g *governanceContract) Execute(args *vmcommon.ContractCallInput) vmcommon.
 		return g.changeConfig(args)
 	case "closeProposal":
 		return g.closeProposal(args)
-	case "getVotingPower":
-		return g.getVotingPower(args)
+	case "viewVotingPower":
+		return g.viewVotingPower(args)
+	case "getConfig":
+		return g.viewConfig(args)
+	case "getUserVoteHistory":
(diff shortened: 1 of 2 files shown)


vm/systemSmartContracts/governance.go
```diff
@@ -633,16 +657,132 @@ func (g *governanceContract) getVotingPower(args *vmcommon.ContractCallInput) vm
...
+	args *vmcommon.ContractCallInput,
+	numArgs int,
+) error {
+	if !bytes.Equal(args.CallerAddr, args.RecipientAddr) {
+		return vm.ErrInvalidCaller
+	}
+	if args.CallValue.Cmp(zero) != 0 {
+		return vm.ErrCallValueMustBeZero
+	}
+	err := g.eei.UseGas(g.gasCost.MetaChainSystemSCsCost.Vote)
```
iulianpascalau, line 765: no gas usage on view functions :D

vm/systemSmartContracts/governance.go
```diff
@@ -633,16 +657,132 @@ func (g *governanceContract) getVotingPower(args *vmcommon.ContractCallInput) vm
...
+	if err != nil {
+		g.eei.AddReturnMessage(err.Error())
+		return vmcommon.UserError
+	}
+
+	g.eei.Finish([]byte(gConfig.ProposalFee.String()))
+	g.eei.Finish([]byte(big.NewFloat(float64(gConfig.MinQuorum)).String()))
+	g.eei.Finish([]byte(big.NewFloat(float64(gConfig.MinPassThreshold)).String()))
+	g.eei.Finish([]byte(big.NewFloat(float64(gConfig.MinVetoThreshold)).String()))
+	g.eei.Finish([]byte(big.NewInt(int64(gConfig.LastProposalNonce)).String()))
```
iulianpascalau, line 677: :+1: with this one we will know how many proposals were made

vm/systemSmartContracts/governance.go
sasurobert: true

iulianpascalau (approved):

vm/systemSmartContracts/governance.go
```diff
@@ -146,8 +146,16 @@ func (g *governanceContract) Execute(args *vmcommon.ContractCallInput) vmcommon.
 		return g.changeConfig(args)
 	case "closeProposal":
 		return g.closeProposal(args)
-	case "getVotingPower":
-		return g.getVotingPower(args)
+	case "viewVotingPower":
```
ccorcoveanu, line 149: Why did we switch from `get` to `view`? Will we do this everywhere? Otherwise let's stick with one so it would be easier for users to "guess" what endpoints could exist

sasurobert: there were conflicts with internal get functions. adding view is better. but I would not change it for other contracts.

PR #4943, opened 2023-02-01
fix leader fees for last block in epoch

AdoAdoAdo:
The leader fees for the meta-chain block are added in validator statistics with a one block delay which cause at the end of epoch negative dust when the last metachain block before a change of epoch had  transactions.
  
## Proposed changes
- register the leader fees for the meta-chain block in the validator statistics for that specific meta-block
- the change needs activation epoch, as it is not backwards compatible 

## Testing procedure
- generate transactions for meta-chain e.g ESDT issue/ Delegation/Stake and check the added logs that the leader fees are considered for the current block validator statistics instead of the following block
- run an import-db on metachain to validate that the change is backwards compatible.

## Pre-requisites


AdoAdoAdo:
different approach will be done on another PR

PR #4944, opened 2023-02-01
MX-13760: improve handlers sorting on epoch change subscriber

bogdan-rosianu:
- previously, the epoch start subscription handler sorted the handlers each time when notifying.
  
## Proposed changes
- when adding a new handler, sort the slice right after, so the handlers are always sorted

## Testing procedure
- regular testing procedure + check in logs that the NotifyOrder is followed

## Pre-requisites


epochStart/notifier/epochStartSubscriptionHandler.go
```diff
@@ -42,6 +42,9 @@ func (essh *epochStartSubscriptionHandler) RegisterHandler(handler epochStart.Ac
 	if handler != nil {
```
iulianpascalau, line 42: please refactor here:
```GO
func (essh *epochStartSubscriptionHandler) RegisterHandler(handler epochStart.ActionHandler) {
    if check.IfNil(handler) {
        return
    }
....
```

epochStart/notifier/epochStartSubscriptionHandler.go
```diff
@@ -62,11 +65,6 @@ func (essh *epochStartSubscriptionHandler) UnregisterHandler(handlerToUnregister
 // NotifyAll will call all the subscribed functions from the internal slice
 func (essh *epochStartSubscriptionHandler) NotifyAll(hdr data.HeaderHandler) {
 	essh.mutEpochStartHandler.RLock()
-
-	sort.Slice(essh.epochStartHandlers, func(i, j int) bool {
```
iulianpascalau, line 66: :+1:

epochStart/notifier/epochStartSubscriptionHandler.go
bogdan-rosianu: done by check if nil reflect

state/syncer/userAccountsSyncer.go
```diff
@@ -274,8 +274,8 @@ func (u *userAccountsSyncer) syncAccountDataTries(
 }
 
 func (u *userAccountsSyncer) printDataTrieStatistics() {
-	u.mutStatistics.RLock()
-	defer u.mutStatistics.RUnlock()
+	u.mutStatistics.Lock()
```
SebastianMarian, line 277: Why this change?

bogdan-rosianu: because there is a `sort.Slice(u.largeTries, func(i, j int) bool ` that happened under RLock/RUnlock. It's not really an issue since the `printDataTrieStatistics` function is called sequentially, but it's better to have the fix

SebastianMarian: ok

SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed 
##                            MX-13760-epoch-subscriber-improvements       ##
- block hash does not match                  9767
- wrong nonce in block                       3278

PR #4947, opened 2023-02-01
Fix storage pruning node type selection

sstanculeanu:
- when creating the old data cleaner provider, it used to use the observer config value for nodes running in multikey mode
  
## Proposed changes
- added IsMultiKeyMode method to ManagedPeersHolder interface
- added few more unittests

## Testing procedure
- will be tested with feat branch, old data should be cleaned on multikey nodes

## Pre-requisites


keysManagement/keysHandler_test.go
```diff
@@ -260,14 +261,14 @@ func TestKeysHandler_UpdatePublicKeyLiveness(t *testing.T) {
...
+	handler, _ := keysManagement.NewKeysHandler(args)
 
-	t.Run("same pid should not call reset", func(t *testing.T) {
+	t.Run("same Pid() should not call reset", func(t *testing.T) {
 		handler.UpdatePublicKeyLiveness(randomPublicKeyBytes, pid)
 		assert.Zero(t, len(mapResetCalled))
 	})
-	t.Run("another pid should call reset", func(t *testing.T) {
-		randomPid := core.PeerID("random pid")
+	t.Run("another Pid() should call reset", func(t *testing.T) {
```
iulianpascalau, line 270: `pid` instead of `Pid()`? Also on L271

keysManagement/interface.go
```diff
@@ -1,17 +1,29 @@
...
 
 import (
+	"time"
+
 	"github.com/multiversx/mx-chain-core-go/core"
 	crypto "github.com/multiversx/mx-chain-crypto-go"
 )
 
 // ManagedPeersHolder defines the operations of an entity that holds managed identities for a node
 type ManagedPeersHolder interface {
```
iulianpascalau, line 11: so now we have the interface in the same package as the implementation. Because we need it in several packages.

node/external/timemachine/fee/memoryFootprint/memory_test.go
```diff
@@ -15,7 +15,7 @@ import (
 // keep this test in a separate package as to not be influenced by other the tests from the same package
 func TestFeeComputer_MemoryFootprint(t *testing.T) {
 	numEpochs := 10000
-	maxFootprintNumBytes := 20_000_000
+	maxFootprintNumBytes := 25_000_000
```
iulianpascalau, line 18: this indicates something is off with the memory foot print. Possible resources leak?

keysManagement/managedPeersHolder.go
```diff
@@ -339,6 +339,11 @@ func (holder *managedPeersHolder) SetNextPeerAuthenticationTime(pkBytes []byte,
 	pInfo.setNextPeerAuthenticationTime(nextTime)
 }
 
+// IsMultiKeyMode returns true if the key has at least one managed key
```
bogdan-rosianu, line 342: if the node* ?

bogdan-rosianu (approved):

iulianpascalau (approved):

PR #4961, opened 2023-02-06
Add PreviousList to peersAccount/validatorInfo/shardValidatorInfo

mariusmihaic:
- There is no easy way to find a peersAccount/validatorInfo/shardValidatorInfo previous list, especially for a node that is starting in epoch or restarting without db.
- There are currently two bugs in `indexHashedNodesCoordinator` when computing the new nodes config from current epoch.
1. In case of a node restart without db/start in epoch, `previousConfig` was actually using the nodes config from epoch 0.
2. In case of a **leaving/unstaked** node, which **would be forced to remain**, the wrongful `previousConfig` would be used to make nodes remain eligibile (wrongful nodes)

## Proposed changes
- Added `PreviousList` in peersAccount/validatorInfo/shardValidatorInfo and changed their used interfaces + regenerated proto files.
- In order to maintain backwards compatibility, when setting a validator list, added an extra flag `updatePreviousList`, e.g.: `SetList(list string, updatePreviousList bool)`, which will be called with `StakingV4StartedFlag` - meaning this fix will be enabled only with stakingV4
- Removed `previousConfig` from `indexHashedNodesCoordinator` and used `PreviousList` to compute new nodes config

## Testing procedure
- System test scenario with unstaked eligible/waiting/auction nodes with restarts without db.

## Pre-requisites


state/peerAccount.go
```diff
@@ -108,7 +108,11 @@ func (pa *peerAccount) SetTempRating(rating uint32) {
 }
 
 // SetListAndIndex will update the peer's list (eligible, waiting) and the index inside it with journal
-func (pa *peerAccount) SetListAndIndex(shardID uint32, list string, index uint32) {
+func (pa *peerAccount) SetListAndIndex(shardID uint32, list string, index uint32, updatePreviousList bool) {
+	if updatePreviousList {
```
sasurobert, line 112: I think it is better to have a peerAccountV2 - where this update is true - and leave the old code as it is.

sasurobert: And you do not need to push the boolean in the argument list.

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
```diff
@@ -612,33 +613,18 @@ func (ihnc *indexHashedNodesCoordinator) EpochStartPrepare(metaHdr data.HeaderHa
...
-	copiedPrevious := &epochNodesConfig{}
-	copiedPrevious.eligibleMap = copyValidatorMap(previousConfig.eligibleMap)
-	copiedPrevious.waitingMap = copyValidatorMap(previousConfig.waitingMap)
-	copiedPrevious.nbShards = previousConfig.nbShards
-
-	ihnc.mutNodesConfig.RUnlock()
-
 	// TODO: compare with previous nodesConfig if exists
-	newNodesConfig, err := ihnc.computeNodesConfigFromList(copiedPrevious, allValidatorInfo)
+	newNodesConfig, err := ihnc.computeNodesConfigFromList(allValidatorInfo)
```
sasurobert, line 617: how can this be backward compatible ? it you have no information ?

state/peerAccountData.proto
```diff
@@ -52,4 +52,5 @@ message PeerAccountData {
     uint32      TotalValidatorIgnoredSignaturesRate = 16 [(gogoproto.jsontag) = "totalValidatorIgnoredSignaturesRate"];
     uint64      Nonce                               = 17 [(gogoproto.jsontag) = "nonce"];
     uint32      UnStakedEpoch                       = 18 [(gogoproto.jsontag) = "unStakedEpoch"];
+    string      PreviousList                        = 19 [(gogoproto.jsontag) = "previousList,omitempty"];
```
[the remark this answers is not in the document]
mariusmihaic: I did some searching and testing and it seems it is backwards compatible.
According to [this source](https://developers.google.com/protocol-buffers/docs/overview):
> maintaining backwards compatibility is crucial. Protocol buffers allow for the seamless support of changes, including the addition of new fields and the deletion of existing fields, to any protocol buffer without breaking existing services

> New code will also transparently read old messages. New fields will not be present in old messages; in these cases protocol buffers provide a reasonable default value.

Also, I did a unit test which confirms this:

```go
func TestDsa(t *testing.T) {
	str1 := &ShardValidatorInfo{
		PublicKey:    []byte("pk"),
		ShardId:      4,
		List:         "list",
		Index:        5,
		TempRating:   6,
		PreviousList: "",
	}

	str2 := &ShardValidatorInfo2{
		PublicKey:  []byte("pk"),
		ShardId:    4,
		List:       "list",
		Index:      5,
		TempRating: 6,
	}

	marshaller := &marshal.GogoProtoMarshalizer{}
	bytesStr1, err := marshaller.Marshal(str1)
	require.Nil(t, err)

	bytesStr2, err := marshaller.Marshal(str2)
	require.Nil(t, err)

	require.Equal(t, bytesStr1, bytesStr2)
}

```

Both versions return the same marshalled byte array:
1. `ShardValidatorInfo` has `PreviousList` field
2. `ShardValidatorInfo2` is the old version, without added field

state/peerAccount.go
mariusmihaic: I would leave the existing structure with extension field instead of creating a new struct. This is also related to my comment above.

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
mariusmihaic: It is backwards compatible, since the `previousConfig` was never used on mainnet. It was supposed to be enabled with `FixWaitingListEnableEpoch`, which I've deleted in a previous PR.

Old version compatibility is maintained in `addValidatorToPreviousMap` :
```go
	if !ihnc.flagStakingV4Started.IsSet() {
		eligibleMap[shardId] = append(eligibleMap[shardId], currentValidator)
		return
	}
``` 

Here, `eligibleMap` **is not** the one from `previousEpoch`, is actually the received list from peerMiniBlocks

SebastianMarian (approved):

state/peerAccount.go
```diff
@@ -108,9 +108,10 @@ func (pa *peerAccount) SetTempRating(rating uint32) {
 }
 
 // SetListAndIndex will update the peer's list (eligible, waiting) and the index inside it with journal
-func (pa *peerAccount) SetListAndIndex(shardID uint32, list string, index uint32, updatePreviousList bool) {
-	if updatePreviousList {
+func (pa *peerAccount) SetListAndIndex(shardID uint32, list string, index uint32, updatePreviousValues bool) {
+	if updatePreviousValues {
 		pa.PreviousList = pa.List
+		pa.PreviousIndexInList = pa.IndexInList
```
SebastianMarian, line 114: We don't need also a variable to hold PreviousShardID ?

mariusmihaic: Apparently, every time SetListAndIndex (jailed/leaving) is called, it uses the same shardID, so that doesn't change.
Also, for our use-case (leaving), I don't think there will be any future use-case where a leaving unstaked validators would be set with a different shard ID.

sasurobert (dismissed):

state/peerAccount.go
SebastianMarian: Ok

SebastianMarian (dismissed):

AdoAdoAdo (approved):

epochStart/metachain/legacySystemSCs.go
```diff
@@ -290,7 +290,7 @@ func (s *legacySystemSCProcessor) unStakeNodesWithNotEnoughFunds(
 		}
 
 		validatorLeaving := validatorInfo.ShallowClone()
-		validatorLeaving.SetList(string(common.LeavingList))
+		validatorLeaving.SetList(string(common.LeavingList), s.enableEpochsHandler.IsStakingV4Started())
```
AdoAdoAdo, line 293: you don't need to set the index as well?

SebastianMarian (approved):

epochStart/metachain/legacySystemSCs.go
mariusmihaic: Yes, very good catch.
Updated index as well

AdoAdoAdo (approved):

AdoAdoAdo (approved):

sasurobert (approved):

PR #4977, opened 2023-02-09
Use custom enable epochs handler for shuffler and nodes coordinator

sstanculeanu:
#### Note for review
One who reviews should check that:
- the old conditions on `.SetValue` call match the conditions that set the same flags on `enableEpochHandler` implementation
- the custom `enableEpochsHandler` is updated properly(call to `EpochConfirmed`)

- shuffler and nodes coordinator used to receive both enableEpochsHandler and enable epochs config values
  
## Proposed changes
- pass a separate enableEpochsConfig to these components which will manage it, calling its EpochConfirmed method with the desired epoch, thus updating its flags

## Testing procedure

## Pre-requisites


integrationTests/nodesCoordinatorFactory.go
```diff
@@ -112,6 +114,8 @@ func (ihncrf *IndexHashedNodesCoordinatorWithRaterFactory) CreateNodesCoordinato
 		MaxNodesEnableConfig: nil,
 		EnableEpochsHandler: &testscommon.EnableEpochsHandlerStub{
 			IsBalanceWaitingListsFlagEnabledField: true,
+			IsStakingV4Step2FlagEnabledField:      true,
+			IsStakingV4Step3FlagEnabledField:      true,
```
mariusmihaic, line 118: [question]

Were tests not passing because of this?

sharding/nodesCoordinator/hashValidatorShuffler.go
```diff
@@ -82,8 +73,6 @@ func NewHashValidatorsShuffler(args *NodesShufflerArgs) (*randHashShuffler, erro
 	var configs []config.MaxNodesChangeConfig
 
 	log.Debug("hashValidatorShuffler: enable epoch for max nodes change", "epoch", args.MaxNodesEnableConfig)
```
mariusmihaic, line 75: This debug is wrong. `args.MaxNodesEnableConfig` is a slice of configs. Not sure how debugger prints it.
If we want to keep this line and log.Debug prints it correctly, we can change it to:

```suggestion
	log.Debug("hashValidatorShuffler: enable epoch for max nodes change", "MaxNodesEnableConfig", args.MaxNodesEnableConfig)
```

sharding/nodesCoordinator/hashValidatorShuffler.go
```diff
@@ -747,14 +732,12 @@ func (rhs *randHashShuffler) UpdateShufflerConfig(epoch uint32) {
...
-	log.Debug("staking v4 step3", "enabled", rhs.flagStakingV4Step3.IsSet())
-
-	rhs.flagStakingV4Step2.SetValue(epoch >= rhs.stakingV4Step2EnableEpoch)
-	log.Debug("staking v4 step2", "enabled", rhs.flagStakingV4Step2.IsSet())
+	epochSubscriberHandler, ok := rhs.enableEpochsHandler.(core.EpochSubscriberHandler)
+	if !ok {
+		log.Error("UpdateShufflerConfig: could not cast enableEpochsHandler to EpochSubscriberHandler")
+		return
+	}
+	epochSubscriberHandler.EpochConfirmed(epoch, 0)
```
mariusmihaic, line 740: This is **not ok**.
Under no circumstance we keep computational work with wrong flags in this component and correct flags in other components.

This code + the one you changed in nodes coordinator are time critical at epoch change and we should not do cast s every time.
Why not use an interface which extends `EnableEpochsHandler` with added `EpochConfirmed` if this what you are expecting to use?

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
```diff
@@ -149,10 +145,14 @@ func NewIndexHashedNodesCoordinator(arguments ArgNodesCoordinator) (*indexHashed
 		isFullArchive:                   arguments.IsFullArchive,
 		enableEpochsHandler:             arguments.EnableEpochsHandler,
 		validatorInfoCacher:             arguments.ValidatorInfoCacher,
-		stakingV4Step2EnableEpoch:       arguments.StakingV4Step2EnableEpoch,
 		nodesCoordinatorRegistryFactory: arguments.NodesCoordinatorRegistryFactory,
 	}
 
+	err = ihnc.updateEnableEpochsHandler(ihnc.currentEpoch)
```
mariusmihaic, line 151: I'm honestly not sure if this call in constructor is ok. Should ask @AdoAdoAdo's opinion as well

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
```diff
@@ -1288,10 +1292,11 @@ func (ihnc *indexHashedNodesCoordinator) getShardValidatorInfoData(txHash []byte
 	return shardValidatorInfo, nil
 }
 
-func (ihnc *indexHashedNodesCoordinator) updateEpochFlags(epoch uint32) {
-	ihnc.flagStakingV4Started.SetValue(epoch >= ihnc.enableEpochsHandler.StakingV4Step1EnableEpoch())
-	log.Debug("indexHashedNodesCoordinator: flagStakingV4Started", "enabled", ihnc.flagStakingV4Started.IsSet())
-
-	ihnc.flagStakingV4Step2.SetValue(epoch >= ihnc.stakingV4Step2EnableEpoch)
-	log.Debug("indexHashedNodesCoordinator: flagStakingV4Step2", "enabled", ihnc.flagStakingV4Step2.IsSet())
+func (ihnc *indexHashedNodesCoordinator) updateEnableEpochsHandler(epoch uint32) error {
```
mariusmihaic, line 1295: Same comment as the one from shuffler. Pass the expected interface here, no cast and no return error.

sharding/nodesCoordinator/indexHashedNodesCoordinator_test.go
```diff
@@ -2321,7 +2334,7 @@ func TestIndexHashedNodesCoordinator_computeNodesConfigFromListValidatorsNoFix(t
 			shardMetaLeaving1,
 		}
 
-	ihnc.flagStakingV4Started.Reset()
+	enableEpochsHandlerMock.StakingV4Step1EnableEpochField = uint32(1000000)
```
mariusmihaic, line 2337:
```suggestion
	enableEpochsHandlerMock.StakingV4Step1EnableEpochField = uint32(stakingV4Epoch)
```

mariusmihaic (changes requested):
We should request a system test on this branch with lots of restart in epoch without db before merging

epochStart/bootstrap/process.go
```diff
@@ -1023,7 +1026,7 @@ func (e *epochStartBootstrap) updateDataForScheduled(
 		HeadersSyncer:        e.headersSyncer,
 		MiniBlocksSyncer:     e.miniBlocksSyncer,
 		TxSyncer:             e.txSyncerForScheduled,
-		ScheduledEnableEpoch: e.coreComponentsHolder.EnableEpochsHandler().ScheduledMiniBlocksEnableEpoch(),
```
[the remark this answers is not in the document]
sstanculeanu: indeed not linked, but the methods returning uint32 epochs from enableEpochHandler will all rely on the same config, thus returning same thing.. can be changed tho, the main reason behind the change was to use the same component which is tested for nil on ctor

epochStart/bootstrap/process_test.go
```diff
@@ -227,6 +227,9 @@ func createMockEpochStartBootstrapArgs(
 			ForceStartFromNetwork: false,
 		},
 		TrieSyncStatisticsProvider: &testscommon.SizeSyncStatisticsHandlerStub{},
+		EnableEpochsHandler: &testscommon.EnableEpochsHandlerStub{
+			StakingV4Step2EnableEpochField: 99999,
```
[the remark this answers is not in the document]
sstanculeanu: yep, it was causing cyclic dependencies

epochStart/bootstrap/storageProcess.go
```diff
@@ -412,7 +412,7 @@ func (sesb *storageEpochStartBootstrap) processNodesConfig(pubKey []byte) error
 		ChanNodeStop:                    sesb.coreComponentsHolder.ChanStopNodeProcess(),
 		NodeTypeProvider:                sesb.coreComponentsHolder.NodeTypeProvider(),
 		IsFullArchive:                   sesb.prefsConfig.FullArchive,
-		EnableEpochsHandler:             sesb.coreComponentsHolder.EnableEpochsHandler(),
+		EnableEpochsHandler:             sesb.enableEpochsHandler,
```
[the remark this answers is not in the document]
sstanculeanu: same reason here for the change, only using an epoch getter which basically returns the value from config in both components.. but I agree with keeping the specific component for nodes coordinator only. updated

sharding/nodesCoordinator/hashValidatorShuffler.go
sstanculeanu: indeed, updated

sharding/nodesCoordinator/indexHashedNodesCoordinator_test.go
sstanculeanu: updated

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
sstanculeanu: the reason behind this was that considering we reach this point with a fresh instance of `enableEpochsHandler`, that never got `EpochConfirmed` called, will have all flags set to false, even though they have epoch 0 for activation. With this call we make sure that this component starts with the flags preset for current epoch

sharding/nodesCoordinator/hashValidatorShuffler.go
sstanculeanu: good suggestion, updated to use a different interface

sharding/nodesCoordinator/indexHashedNodesCoordinator.go
sstanculeanu: updated to different interface

integrationTests/nodesCoordinatorFactory.go
sstanculeanu: rechecked, not needed, removed

epochStart/bootstrap/storageProcess.go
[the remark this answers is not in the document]
sstanculeanu: as discussed, updated: syncValidatorStatus uses the real component

mariusmihaic (approved):
We should request a system test on this + clarify if .EpochConfirmed on nodes coordinator constructor is ok

epochStart/bootstrap/syncValidatorStatus.go
```diff
@@ -111,6 +115,11 @@ func NewSyncValidatorStatus(args ArgsNewSyncValidatorStatus) (*syncValidatorStat
 
 	s.memDB = disabled.CreateMemUnit()
 
+	enableEpochsHandlerNodesCoordinator, err := enablers.NewEnableEpochsHandler(args.EnableEpochsConfig, &timemachine.DisabledEpochNotifier{})
```
AdoAdoAdo, line 118: I think we need to change this pattern as it is very prone to error.
In case someone passes the enableEpochHandler to a subcomponent, it will be difficult to spot in the review that this is a timemachine enableEpochHandler, and it could be changed to view a different epoch than the current one.

AdoAdoAdo (changes requested):

sstanculeanu:
Closing the PR, as the new implementation of enableEpochsHandler, already merged into rc/v1.7.0 won't have this issue.

PR #4984, opened 2023-02-14
Trie mutex refactoring

iulianpascalau:
- trie mutex refactoring
  
## Proposed changes
- trie mutex refactoring

## Testing procedure
- standard system tests: should not trigger `WARN` logs messages whenever the node is closing.

## Pre-requisites


BeniaminDrasovean (approved):

trie/patriciaMerkleTrie_test.go
```diff
@@ -1067,6 +1073,71 @@ func TestPatriciaMerkleTrie_ConcurrentOperations(t *testing.T) {
...
+
+	for i := 0; i < numGoRoutines; i++ {
+		if i%2 == 0 {
+			go func() {
+				time.Sleep(time.Millisecond * 100)
+				wgStart.Done()
+
+				_, _, _ = tr.GetSerializedNodes([]byte("dog"), 1024)
+				wgEnd.Done()
+				fmt.Println("finished")
```
ssd04, line 1104: so we don't want to check the error here?

ssd04 (approved):

BeniaminDrasovean (dismissed):

trie/patriciaMerkleTrie_test.go
iulianpascalau: forgotten prints :( 
No, not quite, the storer is a stub anyway. GetSerializedNodes will error as it won't find some hashes, GetSerializedNode would have returned what the mock storer returned

BeniaminDrasovean (dismissed):

ssd04 (dismissed):

trie/patriciaMerkleTrie.go
```diff
@@ -301,6 +295,9 @@ func (tr *patriciaMerkleTrie) recreate(root []byte, tsm common.StorageManager) (
 
 // String outputs a graphical view of the trie. Mainly used in tests/debugging
 func (tr *patriciaMerkleTrie) String() string {
+	tr.mutOperation.Lock()
```
AdoAdoAdo, line 298: why do we need here Lock?
you can use RLock

same for Get(), GetObsoleteHashes(), VerifyProof(), GetNumNodes(), GetOldRoot()

AdoAdoAdo (changes requested):

trie/patriciaMerkleTrie.go
iulianpascalau: Unfortunately, we need the Lock/Unlock pair here and on all methods that are doing only get operations.
Reason: if the trie is collapsed/partially collapsed (as what happens after each Commit call) the nodes will hold only the collapsed version of the nodes on the last stored level. Then, at each traversing, the trie might try to fetch a collapsed node that will trigger the DB loading and altering of its containing pointers. So, because of the usage of the `resolveCollapsed` function, we can not use RLock/RUnlock functions. The concurrency test might not fail because is not big enough, but otherwise, it will fail on concurrent operations. Changed the test to have only one level in memory so it will fail if we change the Lock/Unlock to RLock/RUnlock.

AdoAdoAdo (approved):

BeniaminDrasovean (approved):

ssd04 (approved):

gabi-vuls (approved):
##                            trie-mutex-refactoring                       ##
- block hash does not match                  10283
- receipts hash missmatch                    2
- wrong nonce in block                       3761

PR #4998, opened 2023-02-17
Bugfixing and stabilization for sovereign chain

SebastianMarian:

- Fixed signing and checking aggregate signature on consensus flow for sovereign chain
 
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/consensus/constants.go
+++ b/consensus/constants.go
@@ -1,11 +1,11 @@
 package consensus
 
-// SubroundType defines the types of subround type
-type SubroundType string
+// ConsensusModel defines the consensus models
+type ConsensusModel string
 
 const (
-	// SubroundTypeV1 defines the first variant of the subround type
-	SubroundTypeV1 SubroundType = "subround v1"
-	// SubroundTypeV2 defines the second variant of the subround type
-	SubroundTypeV2 SubroundType = "subround v2"
+	// ConsensusModelV1 defines the first variant of the consensus model
+	ConsensusModelV1 ConsensusModel = "consensus model v1"
+	// ConsensusModelV2 defines the second variant of the consensus model
+	ConsensusModelV2 ConsensusModel = "consensus model v2"
 )
--- a/consensus/constants.go
+++ b/consensus/constants.go
@@ -1,11 +1,11 @@
 package consensus
 
-// SubroundBlockType defines the types of subround block type
-type SubroundBlockType string
+// SubroundType defines the types of subround type
+type SubroundType string
 
 const (
-	// SubroundBlockTypeV1 defines the first variant of the subround block type
-	SubroundBlockTypeV1 SubroundBlockType = "subround v1"
-	// SubroundBlockTypeV2 defines the second variant of the subround block type
-	SubroundBlockTypeV2 SubroundBlockType = "subround v2"
+	// SubroundTypeV1 defines the first variant of the subround type
+	SubroundTypeV1 SubroundType = "subround v1"
+	// SubroundTypeV2 defines the second variant of the subround type
+	SubroundTypeV2 SubroundType = "subround v2"
 )
--- a/consensus/spos/bls/blsSubroundsFactory.go
+++ b/consensus/spos/bls/blsSubroundsFactory.go
@@ -23,7 +23,7 @@ type factory struct {
 	outportHandler   outport.OutportHandler
 	chainID          []byte
 	currentPid       core.PeerID
-	subroundType     consensus.SubroundType
+	consensusModel   consensus.ConsensusModel
 }
 
 // NewSubroundsFactory creates a new factory object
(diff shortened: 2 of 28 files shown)


sasurobert (approved):

consensus/constants.go
```diff
@@ -1,11 +1,11 @@
 package consensus
 
-// SubroundBlockType defines the types of subround block type
-type SubroundBlockType string
+// ConsensusModel defines the consensus models
+type ConsensusModel string
```
iulianpascalau, line 4: :+1:

consensus/spos/bls/blsSubroundsFactory.go
```diff
@@ -246,25 +273,53 @@ func (fct *factory) generateSignatureSubround() error {
...
+
+func (fct *factory) generateEndRoundSubround() error {
+	subroundEndRoundInstance, err := fct.generateEndRoundSubroundV1()
 	if err != nil {
 		return err
 	}
 
-	fct.worker.AddReceivedMessageCall(MtSignature, subroundSignatureObject.receivedSignature)
-	fct.consensusCore.Chronology().AddSubround(subroundSignatureObject)
+	switch fct.consensusModel {
```
iulianpascalau, line 297: Can be made in another PR: instead of having multiple `switch` operations on each subround, we can have only one switch to create all the subrounds following the model 1 or model 2.

testscommon/components/components.go
```diff
@@ -173,7 +173,7 @@ func GetConsensusArgs(shardCoordinator sharding.Coordinator) consensusComp.Conse
 		StatusComponents:     statusComponents,
 		StatusCoreComponents: GetStatusCoreComponents(),
 		ScheduledProcessor:   scheduledProcessor,
-		SubroundBlockType:    consensus.SubroundBlockTypeV1,
+		ConsensusModel:       consensus.ConsensusModelV1,
```
iulianpascalau, line 176: For the next PR: can we have the consensus integration tests running on both variants? It will be way easier to debug if we need to in the future

iulianpascalau (dismissed):

process/block/baseProcess.go
```diff
@@ -341,6 +341,11 @@ func displayHeader(headerHandler data.HeaderHandler) []*display.LineData {
 	} else {
 		shardHeader, isShardHeader := headerHandler.(data.ShardHeaderHandler)
 		if isShardHeader {
+			_, isHeaderWithValidatorStats := headerHandler.(*block.HeaderWithValidatorStats)
```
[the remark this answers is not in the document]
sasurobert: +1

process/common.go
```diff
@@ -26,6 +26,8 @@ import (
 
 var log = logger.GetOrCreate("process")
 
+var SovereignHeaderVersion = []byte("S1")
```
sasurobert, line 29: missing comment

process/common.go
```diff
@@ -867,6 +875,11 @@ func UnmarshalShardHeader(marshalizer marshal.Marshalizer, hdrBuff []byte) (data
 		return hdr, nil
 	}
 
+	hdr, err = UnmarshalHeaderWithValidatorStats(marshalizer, hdrBuff)
```
sasurobert, line 878: why do we need a code like this ? can't we have a factory/interface for this unmarshal and 3 implementations. Why should we try unmarshalling for header v2 and only after doing for the one with validators stats ? It is double work on CPU. In case of importDB, it is triple work.

process/common.go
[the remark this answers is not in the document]
SebastianMarian: Yep, added a TODO here about the approach that should be used whenever a new version of sovereign header will be created

SebastianMarian: Added

process/common.go
SebastianMarian: Unfortunately could not be resolved for all the cases, as on header interceptor you will receive some bytes. All you know is that this should be a header. Then you should try to unmarshall in all these variants. To optimize this stuff at least we put the most probable implementation header, like v2 for now, as the first try.

process/block/baseProcess.go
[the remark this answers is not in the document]

process/common.go
sasurobert: can't we have sovereignHeaderInterceptor and hearderV2Interceptor ? I think this part has to be refactored.

iulianpascalau: I can get the task of injecting an abstract factory that will get only the marshaled bytes and try to output a headerHandler. We will have 2 versions: one for normal chain operation, and one for the sovereign. Then we can try to optimize it further if we want. This will be a whole new PR.

sasurobert (approved):

iulianpascalau (approved):

PR #5002, opened 2023-02-19
Sharded setup per db

ssd04:
- Set sharded db setup per each storer separately
  
## Proposed changes
- Separate config fields in DBConfig
- Create db toml config file in Storer FilePath

## Testing procedure
- TDB

## Pre-requisites


storage/factory/persisterFactory.go
```diff
@@ -1,59 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return persister, nil
+}
+
+func (pf *PersisterFactory) getDBConfig(path string) (*config.DBConfig, error) {
+	dbConfigFromFile := &config.DBConfig{}
+	err := core.LoadTomlFile(dbConfigFromFile, getPersisterConfigFilePath(path))
```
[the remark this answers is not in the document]
ssd04: the idea was to load the new configuration only after next epoch change, this way we would end up with only one config per a particular epoch, but indeed, i was considering this especially when thinking on moving to different db type, but there are also other options that we might want to consider immediately, will think on this more

storage/factory/persisterFactory_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		pf, _ := factory.NewPersisterFactory(createDefaultDBConfig())
+
+		dirPath := t.TempDir()
+
+		conf, err := pf.GetDBConfig(dirPath)
+		require.Nil(t, err)
+		require.Equal(t, &expectedDBConfig, conf)
+	})
+}
```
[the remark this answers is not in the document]
ssd04: added test here, based on current implementation it will load from config

storage/factory/persisterFactory.go
```diff
@@ -1,59 +1,186 @@
...
+			return nil
+		}
+
+		return err
+	}
+
+	configFilePath := getPersisterConfigFilePath(path)
+	f, err := core.OpenFile(configFilePath)
+	if err == nil {
+		// config file already exists, no need to create config
```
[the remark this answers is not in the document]
ssd04: moved defer close above, if it failed opening the config file it will write the new one

storage/factory/persisterFactory.go
```diff
@@ -1,59 +1,186 @@
...
 }
 
 // Create will return a new instance of a DB with a given path
 func (pf *PersisterFactory) Create(path string) (storage.Persister, error) {
 	if len(path) == 0 {
-		return nil, errors.New("invalid file path")
+		return nil, storage.ErrInvalidFilePath
+	}
+
+	dbConfig, err := pf.getDBConfig(path)
```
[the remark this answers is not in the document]
ssd04: added separate component for db config handler

storage/factory/persisterFactory.go
```diff
@@ -1,59 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	persister, err := pf.createDB(path, dbConfig)
```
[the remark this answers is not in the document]
ssd04: :+1:

storage/interface.go
```diff
@@ -47,25 +47,7 @@ type Cacher interface {
...
-	Remove(key []byte) error
-	// Destroy removes the persistence medium stored data
-	Destroy() error
-	// DestroyClosed removes the already closed persistence medium stored data
-	DestroyClosed() error
-	RangeKeys(handler func(key []byte, val []byte) bool)
-	// IsInterfaceNil returns true if there is no value under the interface
-	IsInterfaceNil() bool
-}
+type Persister = types.Persister
```
iulianpascalau, line 50: :+1:

storage/factory/dbConfigHandler.go
```diff
@@ -0,0 +1,131 @@
...
+			// in memory db, no files available
+			log.Debug("SaveDBConfigToFilePath: provided path not available, config file will not be created")
+			return nil
+		}
+
+		return err
+	}
+
+	configFilePath := getPersisterConfigFilePath(path)
+	f, err := core.OpenFile(configFilePath)
```
iulianpascalau, line 90: something tells me that we should not re-open the file here

storage/factory/persisterCreator.go
```diff
@@ -0,0 +1,75 @@
...
+type persisterCreator struct {
+	dbType              string
+	batchDelaySeconds   int
+	maxBatchSize        int
+	maxOpenFiles        int
+	shardIDProviderType string
+	numShards           int32
+}
+
+func newPersisterCreator(config config.DBConfig) (*persisterCreator, error) {
```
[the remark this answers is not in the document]
ssd04: added unit tests for this file as well

storage/factory/dbConfigHandler.go
```diff
@@ -0,0 +1,123 @@
...
+func (dh *dbConfigHandler) GetDBConfig(path string) (*config.DBConfig, error) {
+	dbConfigFromFile := &config.DBConfig{}
+	err := core.LoadTomlFile(dbConfigFromFile, getPersisterConfigFilePath(path))
+	if err == nil {
+		log.Debug("GetDBConfig: loaded db config from toml config file", "path", dbConfigFromFile)
+		return dbConfigFromFile, nil
+	}
+
+	empty := checkIfDirIsEmpty(path)
+	if !empty {
```
BeniaminDrasovean, line 51: Is this ok? If not empty load the default config? In which case do you load config from file and when do you use the one from main config file and why?

storage/interface.go
```diff
@@ -209,3 +191,12 @@ type ShardIDProvider interface {
 	GetShardIDs() []uint32
 	IsInterfaceNil() bool
 }
+
+// PersisterCreator defines the behavour of a component which is able to create a persister
+type PersisterCreator = types.PersisterCreator
+
+type DBConfigHandler interface {
```
BeniaminDrasovean, line 198: Missing comment.

storage/factory/dbConfigHandler.go
ssd04: if not empty it will load the default, that means it already running with the current (default) setup, if it is a new epoch, it will be an empty dir so it will load the one from main config file; it will load from toml config file if the node will be started until next epoch change

storage/interface.go
ssd04: added

BeniaminDrasovean (dismissed):

storage/factory/dbConfigHandler.go
```diff
@@ -75,29 +75,21 @@ func (dh *dbConfigHandler) GetDBConfig(path string) (*config.DBConfig, error) {
 
 // SaveDBConfigToFilePath will save the provided db config to specified path
 func (dh *dbConfigHandler) SaveDBConfigToFilePath(path string, dbConfig *config.DBConfig) error {
-	_, err := os.Stat(path)
```
iulianpascalau: please put back this code. You can end up writing directories for the memory DB as you rely only on the implementation of SaveTomlFile which can be changed

iulianpascalau (changes requested):

storage/factory/dbConfigHandler.go
ssd04: reverted to add check for directory

iulianpascalau (approved):

BeniaminDrasovean (approved):

iulianpascalau (approved):

PR #5006, opened 2023-02-20
Fixed self-blacklisted event in consensus due to edge case when setting the epoch as proposer

iulianpascalau:
- when the epoch is set on the shard block processor in `CreateNewHeader` the leader might get the new epoch start metablock in an async matter, thus incorrectly setting the `EpochStartMetaHash` value in `CreateBlock` function but on the previous epoch
  
## Proposed changes
- if the metablock to be referenced is of type start of epoch block, rewrite the epoch on the shard block, if required, preventing the block to be discarded

## Testing procedure
- normal system testnet, grepping the logs for `shardProcessor.CreateBlock: epoch from the header is not the same as epoch from epoch start trigger, overwriting`

## Pre-requisites


process/block/shardblock.go
```diff
@@ -807,6 +807,16 @@ func (sp *shardProcessor) CreateBlock(
 		if err != nil {
 			return nil, nil, err
 		}
+
+		epoch := sp.epochStartTrigger.MetaEpoch()
+		if initialHdr.GetEpoch() != epoch {
+			log.Debug("shardProcessor.CreateBlock: epoch from header is not the same as epoch from epoch start trigger, overwriting",
+				"epoch from header", initialHdr.GetEpoch(), "epoch from epoch start trigger", epoch)
+			err = shardHdr.SetEpoch(epoch)
```
SebastianMarian, line 815: :+1:

SebastianMarian (approved):

raduchis (approved):

PR #5010, opened 2023-02-21
Feat/sharded persister

ssd04:
- Added option for sharded persister
  
## Proposed changes
- Added a shardID provider component per each storer, currently the only available option is "BinarySplit" shard id provider type
- Added a general sharded persister implementation that can be used with every type of persister
- The persister config is saved on disk into a toml config file and the latest config options are loaded only at next epoch start

## Testing procedure
- Standard testing scenario with sharded persister type after upgrade and also along the way:
    - at upgrade, sharded persister can be set, for example, to `MiniBlocksStrorage` by adding these 2 options to `[MiniBlocksStorage.DB]` section in node's main config file `config.toml`:
        > ShardIDProviderType = "BinarySplit"
        > NumShards = 4
    - after several epocs, sharded storage can be added also to other storers (it doesn't matter which) by adding these lines (num of shards can vary):
         > ShardIDProviderType = "BinarySplit"
         > NumShards = 8
    - we can have multiple storers with sharded persister type, at least at some points in time
    - sharded persister can be enabled to multiple nodes, observers and validators; after the config is modified and the node is restarted, the new storage config changes will be applied at next epoch change
- we need to make sure the nodes are starting correctly, the sync works properly

## Pre-requisites


ssd04:
Note to add proper releases

gabi-vuls (dismissed):
System test passed. 
##                            feat/sharded-persister                             ##
- Known Warnings              7
- block hash does not match                  1545
- wrong nonce in block                       500

gabi-vuls (approved):

BeniaminDrasovean (approved):

iulianpascalau (approved):

PR #5013, opened 2023-02-21
original caller

matei-p:
We need to know the at any point in a call chain, the original address that initiated the first SC call.
  
## Proposed changes
A new OriginalCallerAddr will be added to VMInput and will be filled by SCRs OriginalSender.
Corresponding changes will be done on VM side. 

## Testing procedure
Some of the existing tests will be updated to store the original caller and to check for it in the end.


process/smartContract/processorV2/vmInputV2.go
```diff
@@ -35,7 +35,12 @@ func (sc *scProcessor) createVMDeployInput(tx data.TransactionHandler) (*vmcommo
 
 func (sc *scProcessor) initializeVMInputFromTx(vmInput *vmcommon.VMInput, tx data.TransactionHandler) error {
 	var err error
-
+	origScr, ok := (tx).(*smartContractResult.SmartContractResult)
+	if ok {
```
sasurobert, line 39: make a small separate function - getOriginalCaller - and do the same if else there with return.

matei-p: done

go.mod
```diff
@@ -180,3 +181,5 @@ require (
 )
 
 replace github.com/gogo/protobuf => github.com/multiversx/protobuf v1.3.2
+
+// replace github.com/multiversx/mx-chain-vm-go v1.5.3-0.20230221130555-7aad95ed6cf1 => /home/bogdan/go2/src/github.com/multiversx/mx-chain-vm-go
```
sasurobert, line 185: delete this line

process/smartContract/processorV2/processV2.go
```diff
@@ -2325,12 +2329,19 @@ func (sc *scProcessor) preprocessOutTransferToSCR(
...
 	result.CallType = outputTransfer.CallType
 	setOriginalTxHash(result, txHash, tx)
-	if result.Value.Cmp(zero) > 0 {
-		result.OriginalSender = tx.GetSndAddr()
-	}
+	result.OriginalSender = GetOriginalSenderForTx(tx)
 	return result
 }
 
+func GetOriginalSenderForTx(tx data.TransactionHandler) []byte {
```
sasurobert, line 2336: add comment to public function.

sasurobert: It would be harder but we could have implemented this function inside data.TransactionHandler - but that is a bigger job. can leave as it is. only add the comment

camilbancioiu (approved):
Robert's comments

sasurobert (approved):

PR #5018, opened 2023-02-22
Fix import db on rc v1.6.0

iulianpascalau:
- import DB stopped abnormally on epoch 3
  
## Proposed changes
- due to the refactoring of the resolvers/requesters components, we need to call the SetEpochHandler on the storage header requester in order to make the import-db process run as designed. This call is necessary so the manual epochs handler component can open one epoch in advance as of the current processing epoch so the start of epochs events will work

## Testing procedure
- standard system test
- after the standard system test, execute an import-db process on all shards. Import-db should work.

## Pre-requisites


dataRetriever/mock/requestersContainerStub.go
```diff
@@ -0,0 +1,77 @@
+package mock
+
+import (
+	"github.com/multiversx/mx-chain-go/dataRetriever"
+)
+
+// RequestersContainerStub -
+type RequestersContainerStub struct {
```
sstanculeanu, line 8: why not use the one from `testscommon/dataRetriever` ? `common_test.go` is part of `dataRetriever_test` package so there should be no cyclic dependency

dataRetriever/common_test.go
```diff
@@ -13,63 +13,127 @@ import (
...
-		},
-	}
-	epochHandler := &mock.EpochHandlerStub{}
-
-	err := dataRetriever.SetEpochHandlerToHdrResolver(resolverContainer, epochHandler)
-	require.Equal(t, localErr, err)
+	t.Run("get function errors should return error", func(t *testing.T) {
+		t.Parallel()
+
+		localErr := errors.New("err")
```
sstanculeanu, line 19: can be extracted as global

dataRetriever/common_test.go
```diff
@@ -13,63 +13,127 @@ import (
...
+		}
+		epochHandler := &mock.EpochHandlerStub{}
+
+		err := dataRetriever.SetEpochHandlerToHdrResolver(resolverContainer, epochHandler)
+		require.Nil(t, err)
+	})
 }
 
-func TestSetEpochHandlerToHdrResolver_CannotSetEpoch(t *testing.T) {
+func TestSetEpochHandlerToHdrRequester_GetErr(t *testing.T) {
```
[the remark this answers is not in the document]
iulianpascalau: fixed

dataRetriever/mock/requestersContainerStub.go
iulianpascalau: done

dataRetriever/common_test.go
iulianpascalau: done

sstanculeanu (approved):

alyn509 (approved):

gabi-vuls (approved):
System test passed. 

##                            fix-import-db-on-rc-v1.6.0                   ##
- Known Warnings              8
- block hash does not match                  1951
- wrong nonce in block                       698

PR #5031, opened 2023-02-24
Propagate sync missing snapshot nodes error

BeniaminDrasovean:
- When a missing trie node is found during snapshot, the sync process is triggered for that node. If the sync fails, the trie db for that epoch should not be marked as complete.
  
## Proposed changes
- If the sync process that was triggered by a missing snapshot node fails, propagate the sync err. This will not mark the db as complete, because the sync failed.

## Testing procedure
- Normal testing procedure. 
- If everything is ok, take the dbs, remove a trie node from the db, and start a testnet with the faulty dbs. When the first snapshot occurs,  the sync process should be triggered, but the node is missing from all the nodes. This will result in the failing of the sync. Check that the db is not marked as active.

## Pre-requisites


state/accountsDB.go
```diff
@@ -1263,6 +1263,7 @@ func (adb *AccountsDB) processSnapshotCompletion(
 	defer func() {
 		adb.isSnapshotInProgress.Reset()
 		adb.updateMetricsOnSnapshotCompletion(metrics, stats)
+		close(errChan)
```
iulianpascalau, line 1266: :+1: this ensures the correct closing of the error chan. Good catch.

state/accountsDB_test.go
```diff
@@ -2790,6 +2795,127 @@ func TestAccountsDB_PrintStatsForRootHash(t *testing.T) {
...
+		for tr.GetStorageManager().IsPruningBlocked() {
+			time.Sleep(time.Millisecond * 100)
+		}
+
+		assert.True(t, isMissingNodeCalled)
+	})
+}
+
+func prepareTrie(tr common.Trie) {
+	key1 := []byte("key1")
```
iulianpascalau, line 2906: could have been written as:
```GO
func prepareTrie(tr common.Trie, numKeys int) {  
    for i := 0; i < numKeys; i++{
       key := fmt.Sprintf("key%d", i)
       val := fmt.Sprintf("val%d", i)
        _ = tr.Update([]byte(key), []byte(val))
    }

    _ = tr.Commit()
}
```

in this way we can have very large tries (like 10000 or more keys)

raduchis (commented):
small question

state/accountsDB.go
BeniaminDrasovean: It is ok. This will be called only after all other launched goroutines will finish the execution. (See `adb.finishSnapshotOperation`)

state/accountsDB_test.go
BeniaminDrasovean: Refactored.

iulianpascalau (approved):

gabi-vuls (approved):
System test passed 

##                            propagate-sync-missing-snapshot-nodes-error  ##
- Known Warnings              11
- block hash does not match                  17362
- receipts hash missmatch                    1
- wrong nonce in block                       5080

raduchis (approved):

PR #5059, opened 2023-03-06
Bug-fix `gasUsed` and `fee` transasctionsFeeProcessor

miiu96:
- Bug-fix the case when a smart contract result with a refund is generated but the receiver is not the sender of the initial transaction.  In this case, the `fee` and `gasUsed` was calculated based on a refund that never arrived at the sender of the transaction, which is wrong.
  
## Proposed changes
- Added an extra check that verifies if the receiver of the smart contract result with refund is the sender of the initial transaction. 

## Testing procedure
- Verify that for the case from above the `gasused` and `fee` fields are calculated correctly

## Pre-requisites

### DIFF
--- a/node/external/transactionAPI/gasUsedAndFeeProcessor.go
+++ b/node/external/transactionAPI/gasUsedAndFeeProcessor.go
@@ -36,6 +36,9 @@ func (gfp *gasUsedAndFeeProcessor) computeAndAttachGasUsedAndFee(tx *transaction
 		if !scr.IsRefund {
 			continue
 		}
+		if scr.RcvAddr != tx.Sender {
+			continue
+		}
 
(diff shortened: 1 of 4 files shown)


bogdan-rosianu (dismissed):

andreibancioiu (dismissed):

iulianpascalau (dismissed):

iulianpascalau (approved):

andreibancioiu (approved):

bogdan-rosianu (approved):

gabi-vuls (approved):
System test passed 

##                            bug-fix-gas-used-and-fee-scr-with-refund     ##
- Known Warnings              12
- Known Errors                1
- block hash does not match                  8714
- receipts hash missmatch                    1
- wrong nonce in block                       3030

PR #5062, opened 2023-03-07
Snapshot get test point

BeniaminDrasovean:
- If a trie node can not be retrieved during a snapshot, that node is synced from the network. We need a way to easily test this.
  
## Proposed changes
- Add a test point so that during snapshot some nodes are not retrieved from storage, triggering the sync process.

## Testing procedure
- Test with the activated sed, and check that all nodes that are not retrieved will be synced.

## Pre-requisites


trie/node_extension.go
```diff
@@ -22,3 +24,14 @@ func shouldTestNode(n node, key []byte) bool {
 
 	return false
 }
+
+func snapshotGetTestPoint(key []byte, faultyChance int) error {
```
bogdan-rosianu, line 28: we have the `unused function` linter issue. add a dummy test that calls this function so we can get rid of the linter issue

BeniaminDrasovean: Added.

bogdan-rosianu (dismissed):

iulianpascalau (approved):

bogdan-rosianu (approved):

iulianpascalau (approved):

gabi-vuls (approved):

PR #5071, opened 2023-03-09
extracted configuration value for vm-queries delay

bogdan-rosianu:
- the delay before the node allows external (API) vm queries was fixed to 2 minutes
  
## Proposed changes
- extracted configuration value

## Testing procedure
- play around with the `VmQueryDelayAfterStartInSec` value and (re)start a node. after the number of seconds specified in the config, it should accept API VM Queries. Query example:
```
curl --request POST \
  --url http://127.0.0.1:8080/vm-values/query \
  --header 'Content-Type: application/json' \
  --data '{
	"scAddress": "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u",
	"funcName": "getContractConfig"
}'
```

## Pre-requisites

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -589,6 +589,8 @@
     TrieOperationsDeadlineMilliseconds = 10000
     # GetAddressesBulkMaxSize represents the maximum number of addresses to be fetched in a bulk per API request. 0 means unlimited
     GetAddressesBulkMaxSize = 100
+    # VmQueryDelayAfterStartInSec represents the number of seconds to wait when starting node before accepting vm query requests
+    VmQueryDelayAfterStartInSec = 5
     # EndpointsThrottlers represents a map for maximum simultaneous go routines for an endpoint
     EndpointsThrottlers = [{ Endpoint = "/transaction/:hash", MaxNumGoRoutines = 10 },
                            { Endpoint = "/transaction/send", MaxNumGoRoutines = 2 },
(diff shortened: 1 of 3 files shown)


node/nodeRunner.go
```diff
@@ -516,9 +516,13 @@ func (nr *nodeRunner) executeOneComponentCreationCycle(
 
 	log.Info("application is now running")
 
+	delayInSecBeforeAllowingVmQueries := configs.GeneralConfig.WebServerAntiflood.VmQueryDelayAfterStartInSec
+	if delayInSecBeforeAllowingVmQueries == 0 {
```
mariusmihaic, line 520: Shall we add a warn here to specify that we override the config value?

cmd/node/config/config.toml
```diff
@@ -589,6 +589,8 @@
     TrieOperationsDeadlineMilliseconds = 10000
     # GetAddressesBulkMaxSize represents the maximum number of addresses to be fetched in a bulk per API request. 0 means unlimited
     GetAddressesBulkMaxSize = 100
+    # VmQueryDelayAfterStartInSec represents the number of seconds to wait when starting node before accepting vm query requests
+    VmQueryDelayAfterStartInSec = 5
```
mariusmihaic, line 593: Isn't 5 sec even for testnets too low for a value?

node/nodeRunner.go
bogdan-rosianu: done

cmd/node/config/config.toml
bogdan-rosianu: used this for testing, forgot to change back. done

mariusmihaic (approved):

gabi-vuls (approved):
System test passed

PR #5090, opened 2023-03-13
always return empty responses when accounts are not found

bogdan-rosianu:
- there were some inconsistencies between how several endpoints returned data when the account was not found 
  
## Proposed changes
- return empty responses instead of errors each time an account that does not exist is passed as input

**This PR brings breaking changes on the node's API level, so they should be communicated properly before reaching a release**

## Testing procedure
Call the following endpoints 2 ways: for a node running this branch and for a node running `feat/polaris-fixes`. Node with this branch should return empty responses, while the node with feat/polaris-fixes _could_ return errors (not in each case).
_these examples require that the used address does not exist or is in another shard_
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/balance
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/esdt
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/esdt/new-esdt
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/code-hash
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/username
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/keys
- 127.0.0.1:8080/address/erd12fzhhrv4pnjflvsy0d7zpq54lz4fjuafqnarlwvspsz569zj4dfqlquphc/key/0a0a0a

## Pre-requisites

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -157,8 +157,8 @@ func (n *Node) GetConsensusGroupSize() int {
 func (n *Node) GetBalance(address string, options api.AccountQueryOptions) (*big.Int, api.BlockInfo, error) {
 	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
 	if err != nil {
-		adaptedBlockInfo, adaptErr := adaptBlockInfoIfError(err)
-		if adaptErr == nil {
+		adaptedBlockInfo, isEmptyAccount := extractBlockInfoIfNewAccount(err)
+		if isEmptyAccount {
 			return big.NewInt(0), adaptedBlockInfo, nil
 		}
 
@@ -172,8 +172,8 @@ func (n *Node) GetBalance(address string, options api.AccountQueryOptions) (*big
 func (n *Node) GetUsername(address string, options api.AccountQueryOptions) (string, api.BlockInfo, error) {
 	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
 	if err != nil {
-		adaptedBlockInfo, adaptErr := adaptBlockInfoIfError(err)
-		if adaptErr == nil {
+		adaptedBlockInfo, isEmptyAccount := extractBlockInfoIfNewAccount(err)
+		if isEmptyAccount {
 			return "", adaptedBlockInfo, nil
 		}
 
@@ -188,8 +188,8 @@ func (n *Node) GetUsername(address string, options api.AccountQueryOptions) (str
 func (n *Node) GetCodeHash(address string, options api.AccountQueryOptions) ([]byte, api.BlockInfo, error) {
 	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
 	if err != nil {
-		adaptedBlockInfo, adaptErr := adaptBlockInfoIfError(err)
-		if adaptErr == nil {
+		adaptedBlockInfo, isEmptyAccount := extractBlockInfoIfNewAccount(err)
+		if isEmptyAccount {
 			return make([]byte, 0), adaptedBlockInfo, nil
 		}
 
@@ -286,8 +286,8 @@ func (n *Node) getEsdtDataFromLeaf(leaf core.KeyValueHolder, userAccount state.U
 func (n *Node) GetKeyValuePairs(address string, options api.AccountQueryOptions, ctx context.Context) (map[string]string, api.BlockInfo, error) {
 	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
 	if err != nil {
-		adaptedBlockInfo, adaptErr := adaptBlockInfoIfError(err)
-		if adaptErr == nil {
+		adaptedBlockInfo, isEmptyAccount := extractBlockInfoIfNewAccount(err)
+		if isEmptyAccount {
 			return make(map[string]string), adaptedBlockInfo, nil
 		}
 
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -870,6 +901,21 @@ func (n *Node) GetAccount(address string, options api.AccountQueryOptions) (api.
...
+func adaptBlockInfoIfError(err error) (api.BlockInfo, error) {
+	if err == nil {
+		return api.BlockInfo{}, nil
+	}
+
+	apiBlockInfo, ok := extractApiBlockInfoIfErrAccountNotFoundAtBlock(err)
+	if ok {
+		return apiBlockInfo, nil
+	}
+	if err == ErrCannotCastAccountHandlerToUserAccountHandler {
```
andreibancioiu, line 913: A comment explaining why this isn't an actual error would be helpful for posterity.

node/node.go
```diff
@@ -870,6 +901,21 @@ func (n *Node) GetAccount(address string, options api.AccountQueryOptions) (api.
 	}, blockInfo, nil
 }
 
+func adaptBlockInfoIfError(err error) (api.BlockInfo, error) {
```
andreibancioiu, line 904: The name is a bit imprecise (since the "thing to adapt" isn't received as a parameter). Could be `extendErrorWithBlockInfo` (just an option, perhaps not a very good one).

Could also be `detectEmptyAccount() -> (blockInfo, bool)` (as far as I see, the returned error type is never used - the error is simply compared to nil).

node/node.go
```diff
@@ -174,6 +172,11 @@ func (n *Node) GetBalance(address string, options api.AccountQueryOptions) (*big
 func (n *Node) GetUsername(address string, options api.AccountQueryOptions) (string, api.BlockInfo, error) {
 	userAccount, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
 	if err != nil {
+		adaptedBlockInfo, adaptErr := adaptBlockInfoIfError(err)
```
andreibancioiu, line 175: `adaptErr` vs. `adaptedErr`.

If the suggestion below would be applied, then we can have `blockInfoOfEmptyAccount, isEmptyAccount := detectEmptyAccount(...)`.

andreibancioiu (approved):

node/node.go
iulianpascalau, line 913: Is this ok? What happens if you provide an encoded code-hash instead of an address?
The `ErrCannotCastAccountHandlerToUserAccountHandler` is returned both if the account handler is nil or it is not of type userAccountHandler. Don't we want to signal an error in the case the loaded account is of bad type?

node/node_test.go
```diff
@@ -62,6 +62,25 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type testBlockInfo struct {
+}
+
+func (t testBlockInfo) asApiResult() api.BlockInfo {
```
iulianpascalau, line 68: missing pointers?

node/node_test.go
```diff
@@ -62,6 +62,25 @@ import (
...
+	return api.BlockInfo{
+		Nonce:    37,
+		Hash:     hex.EncodeToString([]byte("hash")),
+		RootHash: hex.EncodeToString([]byte("root")),
+	}
+}
+
+func (t testBlockInfo) asForProcessing() common.BlockInfo {
```
iulianpascalau, line 76: this sounds bad. Remove `as` ?

node/node.go
bogdan-rosianu: If we provide a non-address an error will be returned because of the bech32 validation.
@andreibancioiu do you think about other situations where this error might be thrown and it really indicates an issue?

node/node.go
bogdan-rosianu: renamed to `extractBlockInfoIfNewAccount`

node/node.go
bogdan-rosianu: renamed / refactored

node/node.go
bogdan-rosianu: added

node/node_test.go
bogdan-rosianu: no need, since the structure doesn't have any member

iulianpascalau: then why do we use a struct?

node/node_test.go
bogdan-rosianu: renamed

andreibancioiu (approved):

iulianpascalau (approved):

node/node_test.go
bogdan-rosianu: just to organize the code, to have a single instance that I'll use in all the tests

gabi-vuls (approved):
System test passed

PR #5106, opened 2023-03-16
Block tracker implementation for extended shard headers

SebastianMarian:
- 
- Added support for extended shard headers in block tracker component
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


iulianpascalau (dismissed):

process/track/sovereignChainShardBlockTrack.go
```diff
@@ -50,3 +57,168 @@ func (scsbt *sovereignChainShardBlockTrack) ComputeLongestSelfChain() (data.Head
...
+	if check.IfNil(extendedShardHeaderHandler) {
+		return
+	}
+	if scsbt.isExtendedShardHeaderOutOfRange(extendedShardHeaderHandler) {
+		return
+	}
+
+	miniBlockHandlers := extendedShardHeaderHandler.GetIncomingMiniBlockHandlers()
+
+	keys := make([][]byte, 0)
```
iulianpascalau, line 129: Does L129-L140 is common code? I think all block trackers instances do the whitelisting in the same way. Can we extract it in base?

sasurobert: +1

SebastianMarian: Not really the same. In shard header and metablock instances we are whitelisting the miniblocks hashes and not the tx hashes. But because for the simplicity, in sovereign, we already have incoming miniblocks included in extended shard headers received, we need to whitelist only the txs now. For whitelisting the txs, block tracker uses another component implemented in miniBlockTrack.go but this component is useless for sovereign, as the miniblocks wil not be sent and received in the "traditional" way. So, we will not have any received miniblock that will notify this mini block tracker component which is hooked to the miniblock's pool.

dataRetriever/requestHandlers/requestHandler.go
```diff
@@ -414,6 +414,11 @@ func (rrh *resolverRequestHandler) RequestShardHeaderByNonce(shardID uint32, non
 	rrh.addRequestedItems([][]byte{key}, suffix)
 }
 
+// RequestExtendedShardHeaderByNonce method does nothing in main chain
+func (rrh *resolverRequestHandler) RequestExtendedShardHeaderByNonce(_ uint64) {
```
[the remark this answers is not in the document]
SebastianMarian: We can, but we will have more code to add. Also here we have useless methods implemented, where theirs usage depends of the node instance, if it is meta or shard node.

process/track/sovereignChainShardBlockTrack.go
iulianpascalau: ok

iulianpascalau (approved):

sasurobert (approved):

PR #5128, opened 2023-03-28
Fix rewards broadcast

iulianpascalau:
- the metachain nodes (proposer + validator) were unable to broadcast the rewards txs and the shard nodes relied only on requests which are known as not optimized. 
- on testnets we found prints like these on the proposer machine:
```
TRACE[2023-03-28 07:47:46.586]   p2p validator                            error = intercepted data not for current shard topic = rewardsTransactions_0_META originator = ...mZ7CtMrPMQ7X from connected peer = ...DXAkkEDxQuTx seq no = 1750857df01fa9f2 topic identifier = default interceptor 
TRACE[2023-03-28 07:47:46.586]   error sending data                       error = validation failed 
```
  
## Proposed changes
- removed exception for the metachain in interceptedRewardTransaction.go
- fixed a print in miniblock tracker

## Testing procedure
- standard system testnet with the following log level `*:DEBUG,p2p:TRACE,process:TRACE,broadcast:TRACE,debug:DEBUG` and search for the following prints:
* we should find prints like `received miniblock from network in block tracker` on all nodes 
* search for a proposer on the metachain that produced an epoch start metablock. Validate on the logs that it did not produced p2p-validation error logs
* for the same epoch, search through logs on the shard nodes and validate that they do received the rewards without requesting them at ~1 second when they received the rewards miniblock. Bonus: the meta-cross-notarization on shards should not have gaps.
* confirm that a cleanup log appears on meta nodes. Something like: 
```
TRACE[2023-03-28 10:18:05.217]   transaction has been cleaned             hash = 201925a901fdccc8758d1b7837bd69681c188f9d7c795fa00f809a0023af656c round = 1292278 sender = 4294967295 receiver = 0 type = rewardTx 
```

## Pre-requisites


process/rewardTransaction/interceptedRewardTransaction_test.go
```diff
@@ -317,6 +318,88 @@ func TestNewInterceptedRewardTransaction_CheckValidityShouldWork(t *testing.T) {
...
+		irt, err := rewardTransaction.NewInterceptedRewardTransaction(
+			txBuff,
+			marshalizer,
+			&hashingMocks.HasherMock{},
+			createMockPubkeyConverter(),
+			mockShardCoordinator)
+		assert.Nil(t, err)
+		assert.True(t, irt.IsForCurrentShard())
+	})
+	t.Run("different shard should return true", func(t *testing.T) {
```
iulianpascalau, line 380: should return false *

raduchis: still true here? The check is with false: assert.False(t, irt.IsForCurrentShard())

iulianpascalau: done, thanks

raduchis: fixed

raduchis (approved):

sstanculeanu (approved):

AdoAdoAdo (approved):

SebastianMarian (approved):

PR #5135, opened 2023-03-29
Fixes outport refactor

miiu96:
- Fixed the part where the transactions were fetched from the pool in the execution order component.
  
## Pre-requisites


outport/process/executionOrder/transactionsExecutionOrder.go
```diff
@@ -173,7 +173,7 @@ func (s *sorter) getInvalidTxsExecutedInCurrentBlock(scheduledMbsFromPreviousBlo
 	for _, hash := range mb.TxHashes {
 		_, found := allScheduledTxs[string(hash)]
 		if found {
-			scheduledExecutedInvalidTxsHashesPrevBlock = append(scheduledExecutedInvalidTxsHashesPrevBlock, string(hash))
+			scheduledExecutedInvalidTxsHashesPrevBlock = append(scheduledExecutedInvalidTxsHashesPrevBlock, hex.EncodeToString(hash))
```
iulianpascalau, line 176: is this still backward compatible?

outport/process/executionOrder/transactionsExecutionOrder.go
```diff
@@ -275,9 +277,10 @@ func extractSCRsFromMap(txsHashes [][]byte, scrs map[string]*outport.SCRInfo) ([
 func extractRewardsFromMap(txsHashes [][]byte, rewards map[string]*outport.RewardInfo) ([]data.TxWithExecutionOrderHandler, error) {
 	result := make([]data.TxWithExecutionOrderHandler, 0, len(txsHashes))
 	for _, txHash := range txsHashes {
-		reward, found := rewards[string(txHash)]
+		txHashHex := hex.EncodeToString(txHash)
```
iulianpascalau, line 280: why do we keep the hash in hex format? This degrades the performance of the node as the hex.EncodeToString is more CPU intensive than the regular cast to string.

outport/process/executionOrder/transactionsExecutionOrder.go
miiu96: yes is not a problem. because this part is used only in indexer

outport/process/executionOrder/transactionsExecutionOrder.go
miiu96: we need the transaction hash hex-encoded in order can send it over to the WS. 
If I will let it non hex I will have problems with the marshaller because the hash is used as a string key in map structures.

mariusmihaic (dismissed):

go.mod
```diff
@@ -13,9 +13,9 @@ require (
 	github.com/google/gops v0.3.18
 	github.com/gorilla/websocket v1.5.0
 	github.com/mitchellh/mapstructure v1.5.0
-	github.com/multiversx/mx-chain-core-go v1.2.1-0.20230322093158-35195fa155c0
+	github.com/multiversx/mx-chain-core-go v1.2.1-0.20230329082847-b78e96c3ad5a
```
mariusmihaic, line 16: This is not the latest commit hash
I think this should be `932a718276f65ff8f820720adf5e98c4cfdd3074`.

Check here: https://github.com/multiversx/mx-chain-core-go/pull/171

outport/notifier/eventNotifier.go
```diff
@@ -46,10 +48,15 @@ func NewEventNotifier(args ArgsEventNotifier) (*eventNotifier, error) {
 	if err != nil {
 		return nil, err
 	}
+	blockContainer, err := createBlockCreatorsContainer()
```
mariusmihaic, line 51: Please move it at the same level as factory: `CreateEventNotifier`, just like you did on indexer

go.mod
miiu96: fixed

outport/notifier/eventNotifier.go
miiu96: done

mariusmihaic (approved):

outport/process/executionOrder/transactionsExecutionOrder.go
iulianpascalau: agree, did not noticed that the package was actually `outport...`

outport/process/executionOrder/transactionsExecutionOrder.go
iulianpascalau: agree, this does not degrade the performance of the validators as the code impacted is in `outport...` package

iulianpascalau (approved):

PR #5172, opened 2023-04-13
Return err every time system vm

sasurobert:
To increase consistency in system smart contract we should return error on case of all user errors. Otherwise it is hard to make integrations of these functions into other SCs.
  
## Proposed changes
Return error when needed.

## Testing procedure
Test withdraw from a delegation contract when there is nothing to unBond. After flag is activated, it should return error.

## Pre-requisites


SebastianMarian (approved):

iulianpascalau (approved):

gabi-vuls (dismissed):
System test passed
##                            returnErr-every-time-systemVM                ##
- Known Warnings              10
- block hash does not match                  3554
- wrong nonce in block                       1101

iulianpascalau (approved):

SebastianMarian (approved):

PR #5176, opened 2023-04-19
Improve requesting of missing mini blocks dest me

SebastianMarian:
- 
- Improved requesting for missing mini blocks with destination in self shard
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- All In and after several epochs we should check if we will find these messages multiple times, having printed different hashes which have been requested, all done in the same millisecond(s): **_transactionCoordinator.requestMissingMiniBlocks: mini block not found and was requested_** or **_transactionCoordinator.RequestMiniBlocks: mini block not found and was requested_**
- 

## Pre-requisites


process/coordinator/process.go
```diff
@@ -748,6 +749,29 @@ func (tc *transactionCoordinator) CreateMbsAndProcessCrossShardTransactionsDstMe
 	return createMBDestMeExecutionInfo.miniBlocks, createMBDestMeExecutionInfo.numTxAdded, allMBsProcessed, nil
 }
 
+func (tc *transactionCoordinator) requestMissingMiniBlocks(mbsInfo []*data.MiniBlockInfo) {
```
raduchis, line 752: functionality of _requestMissingMiniBlocks_  and _RequestMiniBlocks_ is very similar, could the refactored with a method that gets a list of miniblockHashes and does the request for those and be called from both other methods.

raduchis (changes requested):

process/coordinator/process.go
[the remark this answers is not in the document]
SebastianMarian: Yes, but this is solved with this new approach by calling requestMissingMiniBlocks above. For all these mbs requested in this method, once they will be received in L1138 -> method func (tc *transactionCoordinator) receivedMiniBlock(key []byte, value interface{}) we will finally know all the tx hashes and request them if missing in L1162 -> numTxsRequested := preproc.RequestTransactionsForMiniBlock(miniBlock). Actually you cannot request txs for the missing mbs as you don't know their hashes. But I can add the txs request for all the miniblocks which are not missing in the same method -> requestMissingMiniBlocks.

SebastianMarian: Right, will do this refactor

process/coordinator/process.go
```diff
@@ -748,6 +749,48 @@ func (tc *transactionCoordinator) CreateMbsAndProcessCrossShardTransactionsDstMe
...
+			continue
+		}
+
+		preproc := tc.getPreProcessor(miniBlock.Type)
+		if check.IfNil(preproc) {
+			log.Warn("transactionCoordinator.requestMissingMiniBlocksAndTransactions: getPreProcessor", "mb type", miniBlock.Type, "error", process.ErrNilPreProcessor)
+			continue
+		}
+
+		numTxsRequested := preproc.RequestTransactionsForMiniBlock(miniBlock)
```
iulianpascalau, line 782: this will cause the requesting of the missing transactions to be done before the miniblocks are actually requested. This will work as the request handler will whitelist the required data and the interceptors will then accept the missing txs. We should prevent the future disruption of the whitelist mechanisms between the request handler and the interceptors.

iulianpascalau (approved):

process/coordinator/process.go
SebastianMarian: Actually this is the case when some of the mini blocks, from the same header, are already in pool and it requests txs from them, if they are missing.

iulianpascalau: right

gabi-vuls (approved):

PR #5188, opened 2023-04-24
pass storageMarker as argument when SyncAccounts() is called

BeniaminDrasovean:
- When a missing trie node is found, the subtrie starting from that hash will be synced. This is triggered on `SyncBlock()`. However, the storage should not be marked as synced and active - not the whole trie was synced, but just a missing subtrie.
  
## Proposed changes
- Pass `storageMarker` to `SyncAccounts()` func. This way, a disabled storage marker can be given in case the db should not be marked as synced.

## Testing procedure
- Normal testing procedure with missing trie nodes.

## Pre-requisites

### DIFF
--- a/state/syncer/userAccountsSyncer.go
+++ b/state/syncer/userAccountsSyncer.go
@@ -12,7 +12,6 @@ import (
 	"github.com/multiversx/mx-chain-core-go/core/check"
 	"github.com/multiversx/mx-chain-core-go/data"
 	"github.com/multiversx/mx-chain-go/common"
-	"github.com/multiversx/mx-chain-go/epochStart"
 	"github.com/multiversx/mx-chain-go/process/factory"
 	"github.com/multiversx/mx-chain-go/state"
 	"github.com/multiversx/mx-chain-go/trie"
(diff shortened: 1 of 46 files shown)


state/syncer/userAccountsSyncer.go
```diff
@@ -104,7 +102,7 @@ func NewUserAccountsSyncer(args ArgsNewUserAccountsSyncer) (*userAccountsSyncer,
 }
 
 // SyncAccounts will launch the syncing method to gather all the data needed for userAccounts - it is a blocking method
-func (u *userAccountsSyncer) SyncAccounts(rootHash []byte) error {
+func (u *userAccountsSyncer) SyncAccounts(rootHash []byte, storageMarker common.StorageMarker) error {
 	u.mutex.Lock()
```
iulianpascalau, line 106: Please add here a test for nil storageMarker argument. The code can panic on L135. You need to add CheckIfNil on the StorageMarker interface. Valid for validatorAccountsSyncer.go L82. 
+ unit tests

BeniaminDrasovean: Done.

iulianpascalau (approved):

PR #5219, opened 2023-05-05
do not add intra shard scr in case of failed txs.

sasurobert:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/smartContract/processorV2/process_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
+	"github.com/multiversx/mx-chain-go/process/transactionLog"
```
iulianpascalau, line 7: go imports

process/smartContract/processorV2/processV2.go
```diff
@@ -1474,7 +1474,8 @@ func (sc *scProcessor) processIfErrorWithAddedLogs(acntSnd state.UserAccountHand
 
 	userErrorLog := createNewLogFromSCRIfError(scrIfError)
 
-	if !sc.isInformativeTxHandler(scrIfError) {
+	isRecvSelfShard := sc.shardCoordinator.SelfId() == sc.shardCoordinator.ComputeId(scrIfError.RcvAddr)
```
iulianpascalau, line 1477: is this condition sufficient? I remember the case was invalid with the receiver in another shard.

process/smartContract/processorV2/process_test.go
sasurobert: done

process/smartContract/processorV2/processV2.go
AdoAdoAdo: I think the condition below will be easier to read if we check with a isReceiverOtherShard instead of !isReceiverSelfShard

AdoAdoAdo (approved):

iulianpascalau (approved):

PR #5331, opened 2023-06-09
fix snapshot after restart

BeniaminDrasovean:
- When a snapshot starts, the root hash of that snapshot is saved in storage at a special key. After that snapshot is finished, that key is removed from storage. If a node restarts before that snapshot is finished, when the node restarts that snapshot will be also restarted.
- If a snapshot takes more than one epoch and the node restarts before the snapshot is finished, the snapshot will restart in the new epoch but with the old root hash (that was saved at the special key in storage) 

  
## Proposed changes
- Save the root hash of the snapshot in storage even if the snapshot is skipped. This way, when a restart occurs, the snapshot that will start after the restart will be of the latest snapshot root hash
- After a snapshot is finished, remove that special key (which holds the last snapshot root hash) from all the epochs. This way, when a node restarts, it will not restart an old snapshot

## Testing procedure
- Normal testing procedure
- Add a delay during snapshots that will force the snapshots to last more than one epoch, and then restart. Check that the correct snapshot is started after the node restarts.

## Pre-requisites

### DIFF
--- a/state/accountsDB.go
+++ b/state/accountsDB.go
@@ -28,7 +28,7 @@ import (
 const (
 	leavesChannelSize       = 100
 	missingNodesChannelSize = 100
-	lastSnapshotStarted     = "lastSnapshot"
+	lastSnapshot            = "lastSnapshot"
 	userTrieSnapshotMsg     = "snapshotState user trie"
 	peerTrieSnapshotMsg     = "snapshotState peer trie"
 )
@@ -202,7 +202,7 @@ func startSnapshotAfterRestart(adb AccountsAdapter, tsm common.StorageManager, p
 		return
 	}
 
-	rootHash, err := tsm.Get([]byte(lastSnapshotStarted))
+	rootHash, err := tsm.GetFromCurrentEpoch([]byte(lastSnapshot))
 	if err != nil {
 		log.Debug("startSnapshotAfterRestart root hash", "error", err)
 		return
@@ -1183,6 +1183,11 @@ func (adb *AccountsDB) prepareSnapshot(rootHash []byte) (common.StorageManager, 
 		return nil, 0, false
 	}
 
+	defer func() {
+		err = trieStorageManager.PutInEpoch([]byte(lastSnapshot), rootHash, epoch)
+		handleLoggingWhenError("could not set lastSnapshot", err, "rootHash", rootHash)
+	}()
+
 	if !adb.shouldTakeSnapshot(trieStorageManager, rootHash, epoch) {
 		log.Debug("skipping snapshot",
 			"last snapshot rootHash", adb.lastSnapshot.rootHash,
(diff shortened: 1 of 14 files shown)


state/accountsDB.go
```diff
@@ -1282,8 +1285,8 @@ func (adb *AccountsDB) processSnapshotCompletion(
 		return
 	}
 
-	err := trieStorageManager.Remove([]byte(lastSnapshotStarted))
-	handleLoggingWhenError("could not remove lastSnapshotStarted", err, "rootHash", rootHash)
+	err := trieStorageManager.RemoveFromAllEpochs([]byte(lastSnapshot))
```
iulianpascalau, line 1288: I think this works, as seen in the following scenario:

Epoch 100 -> snapshot started (takes a long time to finalize it) `lastSnapshot` written in epoch 100
Epoch 101 -> snapshot skipped `lastSnapshot` written in epoch 101
Epoch 102 -> snapshot skipped `lastSnapshot` written in epoch 102
Epoch 100 snapshot finished, removed all `lastSnapshot` entries from all opened epochs, marks the Epoch 100 as active.
*** restarts the node ***
The node searches the ActiveDB marker, opens Epoch 100 & one ore more previous epochs and all the DBs up until Epoch 102. Then, it will start the snapshot from Epoch 102.

Alternatively, if the node would have been restarted before the removal of the `lastSnapshot` marker, it would have opened more DBs (until it could have found 2 ActiveDB markers) & start the snapshot from the most recent Epoch (102) not the uncompleted Epoch 100

ssd04 (commented):
no other things from me

storage/pruning/triePruningStorer.go
```diff
@@ -164,6 +164,23 @@ func (ps *triePruningStorer) GetLatestStorageEpoch() (uint32, error) {
 	return ps.activePersisters[currentEpochIndex].epoch, nil
 }
 
+// RemoveFromAllEpochs removes the data associated to the given key from both cache and epochs storers
+func (ps *triePruningStorer) RemoveFromAllEpochs(key []byte) error {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Changed.

iulianpascalau (approved):

state/export_test.go
```diff
@@ -17,6 +18,11 @@ func NewEmptyBaseAccount(address []byte, tracker DataTrieTracker) *baseAccount {
 	}
 }
 
+// LoadCode -
+func (adb *AccountsDB) IsSnapshotInProgress() *atomic.Flag {
```
ssd04, line 22: update mock comment

BeniaminDrasovean: Done

ssd04 (approved):

ssd04 (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed. 
##                            fix-snapshot-after-restart                   ##
- block hash does not match                  10495
- wrong nonce in block                       4009

PR #5340, opened 2023-06-13
handle nodes config from storage

ssd04:
- When loading from storage, get from most recent loaded epoch nodes config

## Testing procedure
- N/A

## Pre-requisites

### DIFF
--- a/sharding/nodesCoordinator/indexHashedNodesCoordinator.go
+++ b/sharding/nodesCoordinator/indexHashedNodesCoordinator.go
@@ -251,8 +251,7 @@ func (ihnc *indexHashedNodesCoordinator) getNodesConfig(epoch uint32) (*epochNod
 		}
 	}
 
-	ncInternalkey := append([]byte(common.NodesCoordinatorRegistryKeyPrefix), []byte(fmt.Sprint(epoch))...)
-	epochConfigBytes, err := ihnc.bootStorer.GetFromEpoch(ncInternalkey, epoch)
+	epochConfigBytes, err := ihnc.getNodesConfigFromStorer(epoch)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sharding/nodesCoordinator/indexHashedNodesCoordinator.go
```diff
@@ -277,6 +276,38 @@ func (ihnc *indexHashedNodesCoordinator) getNodesConfig(epoch uint32) (*epochNod
 	return nodesConfigEpoch, ok
 }
 
+func (ihnc *indexHashedNodesCoordinator) getNodesConfigFromStorer(epoch uint32) ([]byte, error) {
+	var err error
+
+	mostRecentEpoch := ihnc.getMostRecentLoadedEpoch()
```
AdoAdoAdo, line 282: This should be returning the most recent epoch that would have (according to the config) the requested epoch config (parameter)

e.g 
a. we are in epoch 10
b. config is to keep latest 4 epochs in memory
c. request is to get the nodes config for epoch 2

according to (b.) epoch 2 config should be available (in memory and also stored on disk) in epochs 2, 3, 4, 5.
the most recent epoch from those would be epoch 5, (you wouldn't be able to find the config stored in persisters for epochs 6, 7, 8, 9, 10).

So in this case you should try to get it from epoch 5 persister `ihnc.bootStorer.GetFromEpoch(ncInternalkey, 5)`
maybe also do a retrial if not found with previous epoch `ihnc.bootStorer.GetFromEpoch(ncInternalkey, 4)`

AdoAdoAdo (changes requested):

BeniaminDrasovean (approved):

PR #5362, opened 2023-06-20
Feat/multikey metrics

sstanculeanu:
- no info about the managed keys at API level
  
## Proposed changes
- added multiple new endpoints:
   - `node/managed-keys/count` for the total number of managed keys
   - `node/managed-keys` for all keys managed by the node
   - `node/managed-keys/eligible` for the eligible keys managed, in the current epoch, from the same shard as the node
   - `node/managed-keys/waiting` for the waiting keys managed, in a the current epoch, from the same shard as the node
- also updated leader metrics to keep track of managed keys leaders

## Testing procedure
- standard system test
- to be checked the new metrics both on multikey and regular(should see empty/0 on regular mode)
- to be checked that the following metrics are updated when a managed key is leader: `erd_count_consensus`, `erd_consensus_state`, `erd_count_consensus_accepted_blocks`, `erd_count_leader`, `erd_consensus_round_state`, 

## Pre-requisites


iulianpascalau (approved):

process/block/shardblock.go
```diff
@@ -1043,6 +1043,7 @@ func (sp *shardProcessor) CommitBlock(
 		highestFinalBlockNonce,
 		lastCrossNotarizedHeader,
 		header,
+		sp.managedPeersHolder,
```
iulianpascalau, line 1046: you forgot to set the managedPeersHolder value and is always nil. Failed at system testnet.
Please add this line on L121:
```GO
managedPeersHolder: arguments.ManagedPeersHolder
```
something similar to what we have on metablock.go L137

iulianpascalau (changes requested):

process/block/shardblock.go
sstanculeanu: :man_facepalming: fixed in #5371

iulianpascalau (dismissed):

iulianpascalau (dismissed):

ssd04 (approved):

PR #5372, opened 2023-06-22
Merge rc 1 6 into sovereign 22 jun 2023

mariusmihaic:
Merge rc 1 6 into sovereign 22 jun 2023


genesis/process/disabled/disabled_test.go
```diff
@@ -271,12 +272,12 @@ func TestScheduledTxsExecutionHandler(t *testing.T) {
 	require.False(t, handler.IsInterfaceNil())
 
 	hash, err := handler.GetScheduledRootHashForHeader([]byte{})
-	require.Equal(t, make([]byte, 0), hash)
-	require.Nil(t, err)
```
iulianpascalau: why this change? also on L280

mariusmihaic: Because the stub was changed to return `nil, err` instead of `make(), nil`

iulianpascalau (approved):

raduchis (approved):

PR #5384, opened 2023-06-28
Multiple instances HostDriver

miiu96:
-  The configuration for the HostDriver has to support multiple instances.
  
## Proposed changes
- Extended the `external.toml` config file in order to have multiple instances of `HostDriversConfig`. 
- This change is needed because we want to have multiple host drivers on the same node. ( we need an instance of the host driver for the Elastic indexer and one instance for the Event notifier)

## Testing procedure
- Do a system test with multiple instances of the host driver on the same node. 

## Pre-requisites

### DIFF
--- a/factory/status/statusComponents.go
+++ b/factory/status/statusComponents.go
@@ -255,7 +255,7 @@ func (scf *statusComponentsFactory) makeHostDriversArgs() ([]outportDriverFactor
 	for idx := 0; idx < len(scf.externalConfig.HostDriversConfig); idx++ {
 		hostConfig := scf.externalConfig.HostDriversConfig[idx]
 		if !hostConfig.Enabled {
-			return argsHostDriverFactorySlice, nil
+			continue
 		}
 
 		marshaller, err := factoryMarshalizer.NewMarshalizer(hostConfig.MarshallerType)
--- a/factory/status/statusComponents.go
+++ b/factory/status/statusComponents.go
@@ -190,7 +190,7 @@ func (pc *statusComponents) Close() error {
 // createOutportDriver creates a new outport.OutportHandler which is used to register outport drivers
 // once a driver is subscribed it will receive data through the implemented outport.Driver methods
 func (scf *statusComponentsFactory) createOutportDriver() (outport.OutportHandler, error) {
-	hostDriverArgs, err := scf.makeHostDriverArgs()
+	hostDriversArgs, err := scf.makeHostDriversArgs()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


factory/status/statusComponents.go
```diff
@@ -250,18 +250,24 @@ func (scf *statusComponentsFactory) makeEventNotifierArgs() (*outportDriverFacto
...
-func (scf *statusComponentsFactory) makeHostDriverArgs() (outportDriverFactory.ArgsHostDriverFactory, error) {
-	if !scf.externalConfig.HostDriverConfig.Enabled {
-		return outportDriverFactory.ArgsHostDriverFactory{}, nil
-	}
+func (scf *statusComponentsFactory) makeHostDriversArgs() ([]outportDriverFactory.ArgsHostDriverFactory, error) {
+	argsHostDriverFactorySlice := make([]outportDriverFactory.ArgsHostDriverFactory, 0, len(scf.externalConfig.HostDriversConfig))
+	for idx := 0; idx < len(scf.externalConfig.HostDriversConfig); idx++ {
+		hostConfig := scf.externalConfig.HostDriversConfig[idx]
+		if !hostConfig.Enabled {
+			return argsHostDriverFactorySlice, nil
```
iulianpascalau, line 258: we need a continue instead of this return. Reason: we might define an array of configs and we decide, temporarily, to switch one host off

outport/factory/outportFactory_test.go
```diff
@@ -104,3 +107,46 @@ func TestCreateOutport_SubscribeNotifierDriver(t *testing.T) {
 
 	require.True(t, outPort.HasDrivers())
 }
+
+func TestCreateOutport_SubscribeMultipleHostDrivers(t *testing.T) {
```
iulianpascalau, line 111: we need to add a test with 3 hosts having one inactive. Can be the one with index 1.

outport/factory/outportFactory.go
```diff
@@ -52,7 +52,14 @@ func createAndSubscribeDrivers(outport outport.OutportHandler, args *OutportFact
 		return err
 	}
 
-	return createAndSubscribeHostDriverIfNeeded(outport, args.HostDriverArgs)
+	for idx := 0; idx < len(args.HostDriversArgs); idx++ {
+		err = createAndSubscribeHostDriverIfNeeded(outport, args.HostDriversArgs[idx])
+		if err != nil {
+			return err
```
bogdan-rosianu, line 58: might also log the index of the driver that failed during the creation

outport/factory/outportFactory_test.go
bogdan-rosianu: :+1:

bogdan-rosianu (approved):

outport/factory/outportFactory.go
```diff
@@ -55,6 +55,7 @@ func createAndSubscribeDrivers(outport outport.OutportHandler, args *OutportFact
 	for idx := 0; idx < len(args.HostDriversArgs); idx++ {
 		err = createAndSubscribeHostDriverIfNeeded(outport, args.HostDriversArgs[idx])
 		if err != nil {
+			log.Error("createAndSubscribeHostDriverIfNeeded", "host index", idx, "error", err)
```
iulianpascalau, line 58: instead of logging this, we might wrap the returned `err`. This will prevent logging 2 log.Error events

bogdan-rosianu (approved):

iulianpascalau (approved):

gabi-vuls (approved):
System test passed.
##                            multiple-instances-host-driver               ##
- Known Warnings              10
- New Warnings                4
- block hash does not match                  2318
- wrong nonce in block                       744

PR #5392, opened 2023-07-04
Fix the logviewer CLI commands

iulianpascalau:
- the logviewer's CLI commands were outdated
  
## Proposed changes
- fixed the logviewer CLI commands

## Testing procedure
- not required, no production code affected

## Pre-requisites


miiu96 (approved):

bogdan-rosianu (changes requested):
let's do this the proper way so we'll have this updated each time: 
1. create a github workflow that builds all the binaries (node, seednode, loviewer, and so on)
2. run `cmd/CLI.md.sh` and check if there are any changes (could use `git status` or something else)

This way, we'll know if we didn't update the MD files while adding a new flag to one of the binaries

bogdan-rosianu (approved):

miiu96 (approved):

raduchis (approved):

PR #5397, opened 2023-07-05
Fix missing log event. 

miiu96:
- Fixed the problem when the log event with `CompletedTxEvent` identifier is missing in case of a smart contract call that transfer value cross-shard.
  
## Proposed changes
- Changed the condition when the log event with the identifier `CompletedTxEvent` is generated for a value transfer through a smart contract result.

## Testing procedure
- Deploy on a testnet a smart contract that transfers EGLD value to a given address. Check that the smart contract result with the value transfer has a log with the identifier `CompletedTxEvent`.

## Pre-requisites


sasurobert (dismissed):

integrationTests/vm/txsFee/multiShard/scCallWithValueTransfer_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func testDeployContractAndTransferValue(t *testing.T, scProcessorV2Enabled bool) {
+	testContextSource, err := vm.CreatePreparedTxProcessorWithVMsMultiShard(0, config.EnableEpochs{})
+	require.Nil(t, err)
+	defer testContextSource.Close()
+
+	configEnabledEpochs := config.EnableEpochs{}
+	if !scProcessorV2Enabled {
+		configEnabledEpochs.SCDeployEnableEpoch = 1000
```
iulianpascalau, line 34: is this ok? Should have been the other flag. Also, instead of providing the bool, you can provide the epoch directly. One test will be with epoch 0, the other, with a random value (1000 is good enough)
Reason: I've commented out the fix for the processor v1 and the tests still pass

integrationTests/vm/txsFee/utils/utils.go
```diff
@@ -88,6 +89,28 @@ func doDeployInternal(
 	return scAddr, owner
 }
 
+func generateRandomArray(length int) []byte {
+	rand.Seed(time.Now().UnixNano())
```
iulianpascalau, line 93: this can elegantly be refactored as:
```GO
buff := make([]byte, length)
_, _ = rand.Read(buff)
```

`rand` package from the crypto lib should be used

integrationTests/vm/txsFee/utils/utils.go
```diff
@@ -97,13 +120,14 @@ func DoDeployWithCustomParams(
 	gasLimit uint64,
 	contractHexParams []string,
 ) (scAddr []byte, owner []byte) {
-	owner = []byte("12345678901234567890123456789011")
+	owner = generateAddressForContextShardID(testContext)
```
iulianpascalau, line 123: :+1:

iulianpascalau (changes requested):

integrationTests/vm/txsFee/utils/utils.go
miiu96: done

integrationTests/vm/txsFee/multiShard/scCallWithValueTransfer_test.go
miiu96: fixed

iulianpascalau (approved):

sasurobert (approved):

gabi-vuls (approved):
##                            fix-missing-complete-tx-event                ##
- block hash does not match                  10698
- wrong nonce in block                       3956

PR #5447, opened 2023-07-21
Block coordinates on vm query

sstanculeanu:
- add block coordinates on vm query
  
## Proposed changes
- added block coordinates option on vm query

## Testing procedure
- will be tested on feat branch

## Pre-requisites


dataRetriever/blockchain/apiBlockchain.go
```diff
@@ -0,0 +1,95 @@
...
+	return abc.mainBlockchain.GetCurrentBlockHeaderHash()
+}
+
+// SetCurrentBlockHeaderHash does nothing
+func (abc *apiBlockchain) SetCurrentBlockHeaderHash(_ []byte) {
+}
+
+// GetCurrentBlockRootHash returns the current block root hash from the main chain handler, if no local root hash is set
+// if there is a local root hash, it will be returned until its reset
+func (abc *apiBlockchain) GetCurrentBlockRootHash() []byte {
```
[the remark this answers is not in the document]
sstanculeanu: tried to cover all the cases when `executeScCall` would early exit, in order to properly reset the root hash and avoid this situation

factory/processing/processComponents.go
```diff
@@ -340,8 +340,25 @@ func (pcf *processComponentsFactory) Create() (*processComponents, error) {
...
+		pcf.coreData.Hasher(),
+		pcf.bootstrapComponents.ShardCoordinator(),
+	)
+	if err != nil {
+		return nil, err
+	}
+
 	pcf.txLogsProcessor = txLogsProcessor
-	genesisBlocks, initialTxs, err := pcf.generateGenesisHeadersAndApplyInitialBalances()
+	genesisBlocks, initialTxs, err := pcf.generateGenesisHeadersAndApplyInitialBalances(scheduledTxsExecutionHandler)
```
iulianpascalau, line 361: why this change? Why do we need to inject the scheduledTxsExecutionHandler as a parameter?

genesis/process/genesisBlockCreator.go
```diff
@@ -193,6 +193,12 @@ func checkArgumentsForBlockCreator(arg ArgsGenesisBlockCreator) error {
 	if arg.RoundConfig == nil {
 		return genesis.ErrNilRoundConfig
 	}
+	if check.IfNil(arg.HistoryRepository) {
+		return process.ErrNilHistoryRepository
+	}
+	if check.IfNil(arg.ScheduledTxsExecutionHandler) {
+		return process.ErrNilScheduledTxsExecutionHandler
```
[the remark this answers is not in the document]
iulianpascalau: exactly, I do not see the reason for injecting this on the genesis processor

factory/api/apiResolverFactory.go
```diff
@@ -469,19 +495,123 @@ func createScQueryElement(
...
+		StorageService:               args.dataComponents.StorageService(),
+		Marshaller:                   args.coreComponents.InternalMarshalizer(),
+		ScheduledTxsExecutionHandler: args.processComponents.ScheduledTxsExecutionHandler(),
+		Uint64ByteSliceConverter:     args.coreComponents.Uint64ByteSliceConverter(),
 	}
 
 	return smartContract.NewSCQueryService(argsNewSCQueryService)
 }
 
+func createNewAccountsAdapterApi(args *scQueryElementArgs, chainHandler data.ChainHandler) (state.AccountsAdapterAPI, error) {
```
iulianpascalau, line 517: so we will create a new trie for each SC query service instance.
I think we have this, in one form or another, when we create the AccountsDBApi instance. We just need to create more of those instances and pass them to each SC query service element.

process/smartContract/scQueryService.go
```diff
@@ -141,11 +192,28 @@ func (service *SCQueryService) executeScCall(query *process.SCQuery, gasPrice ui
...
+		_ = service.blockChain.SetCurrentBlockHeaderAndRootHash(nil, blockRootHash)
+	}
+
 	query = prepareScQuery(query)
 	vmInput := service.createVMCallInput(query, gasPrice)
 	vmOutput, err := vm.RunSmartContractCall(vmInput)
 	service.wasmVMChangeLocker.RUnlock()
 	if err != nil {
+		// Cleaning the current root hash so the real one would be returned further
+		_ = service.blockChain.SetCurrentBlockHeaderAndRootHash(nil, nil)
```
iulianpascalau, line 216: can call this on a defer function after L207

factory/processing/processComponents.go
AdoAdoAdo: agree, I don't see where this will be needed for the VM query

process/smartContract/scQueryService.go
```diff
@@ -168,6 +241,124 @@ func (service *SCQueryService) executeScCall(query *process.SCQuery, gasPrice ui
...
+	header, err := process.UnmarshalHeader(shardId, service.marshaller, headerBuffer)
+	if err != nil {
+		return nil, err
+	}
+
+	return header, nil
+}
+
+func (service *SCQueryService) getBlockRootHash(headerHash []byte, header data.HeaderHandler) []byte {
+	blockRootHash, err := service.scheduledTxsExecutionHandler.GetScheduledRootHashForHeaderWithEpoch(
```
AdoAdoAdo, line 325: OK I see why you added the scheduled execution handler, but you can use instead getBlockHeaderByNonce and request the header.Nonce+1 and check on that the scheduledRootHash.

AdoAdoAdo: We can discuss as well, I see that this is similar to how it is done for on nodeBlocks.go
This code appears duplicated, maybe it would be better to extract a component and reuse it in both places instead.

AdoAdoAdo (changes requested):

process/smartContract/scQueryService.go
sstanculeanu: added a TODO for the duplicated code

factory/processing/processComponents.go
sstanculeanu: removed it

dataRetriever/blockchain/apiBlockchain.go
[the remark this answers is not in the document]
sstanculeanu: implemented the suggestion

api/groups/vmValuesGroup.go
```diff
@@ -152,6 +158,25 @@ func (vvg *vmValuesGroup) doExecuteQuery(context *gin.Context) (*vm.VMOutputApi,
...
+		return core.OptionalUint64{}, nil, nil, err
+	}
+
+	blockRootHash, err := parseHexBytesUrlParam(context, urlParamBlockRootHash)
+	if err != nil {
+		return core.OptionalUint64{}, nil, nil, err
+	}
+
+	return blockNonce, blockHash, blockRootHash, nil
+}
```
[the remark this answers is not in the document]
sstanculeanu: updated

genesis/process/genesisBlockCreator.go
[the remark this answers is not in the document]
sstanculeanu: removed

factory/api/apiResolverFactory.go
```diff
@@ -286,17 +300,19 @@ func createScQueryService(
...
 		gasScheduleNotifier:   args.gasScheduleNotifier,
 		messageSigVerifier:    args.messageSigVerifier,
 		systemSCConfig:        args.systemSCConfig,
-		workingDir:            args.workingDir,
 		bootstrapper:          args.bootstrapper,
 		guardedAccountHandler: args.guardedAccountHandler,
 		allowVMQueriesChan:    args.allowVMQueriesChan,
+		workingDir:            args.workingDir,
 		index:                 0,
+		processingMode:        args.processingMode,
```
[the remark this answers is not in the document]
sstanculeanu: it is used for ArgsAccountsDB

api/groups/vmValuesGroup_test.go
```diff
@@ -480,11 +465,11 @@ func TestVMValuesGroup_UpdateFacade(t *testing.T) {
 		require.Equal(t, hex.EncodeToString(valueBuff), response.Data)
 
 		newFacade := &mock.FacadeStub{
-			ExecuteSCQueryHandler: func(query *process.SCQuery) (vmOutput *vm.VMOutputApi, e error) {
+			ExecuteSCQueryHandler: func(query *process.SCQuery) (*vm.VMOutputApi, api.BlockInfo, error) {
 
 				return &vm.VMOutputApi{
 					ReturnData: nil,
-				}, expectedErr
+				}, api.BlockInfo{}, expectedErr
```
bogdan-rosianu, line 472: please add a test (or update an existing one) where the blockinfo isn't empty

factory/api/apiResolverFactory.go
```diff
@@ -413,8 +402,22 @@ func createScQueryElement(
 		MissingTrieNodesNotifier: syncer.NewMissingTrieNodesNotifier(),
 	}
 
+	var apiBlockchain data.ChainHandler
 	maxGasForVmQueries := args.generalConfig.VirtualMachine.GasConfig.ShardMaxGasPerVmQuery
 	if args.processComponents.ShardCoordinator().SelfId() == core.MetachainShardId {
+		apiBlockchain, err = blockchain.NewMetaChain(disabled.NewAppStatusHandler())
```
bogdan-rosianu, line 408: extract in 2 functions, so the code would be:
```
if shard == metachain:
    createMetaVmFactory(...)
else
    createShardVmFactory(...)
```

process/smartContract/scQueryService.go
```diff
@@ -167,104 +169,111 @@ func (service *SCQueryService) shouldAllowQueriesExecution() bool {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var blockHash []byte
+	var blockNonce uint64
+
+	if !check.IfNil(blockHeader) {
+		blockNonce = blockHeader.GetNonce()
+		blockHash, err = service.getBlockHashByNonce(blockNonce)
```
bogdan-rosianu, line 191: you don't need storage access just for fetching the hash. it's better to simply compute the hash of the already existing header.
also, these lines can be moved before the end of function, since blockHash and blockNonce are only used for populating the `blockInfo`

api/groups/vmValuesGroup_test.go
sstanculeanu: added

factory/api/apiResolverFactory.go
sstanculeanu: done

process/smartContract/scQueryService.go
sstanculeanu: updated as suggested

process/smartContract/scQueryService.go
```diff
@@ -116,60 +176,204 @@ func (service *SCQueryService) shouldAllowQueriesExecution() bool {
...
+func (service *SCQueryService) extractBlockHeaderAndRootHash(query *process.SCQuery) (data.HeaderHandler, []byte, error) {
+
+	if len(query.BlockHash) > 0 {
+		blockHeader, err := service.getBlockHeaderByHash(query.BlockHash)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		header := service.getBlockHeader(blockHeader)
+		return header, header.GetRootHash(), nil
```
AdoAdoAdo, line 269: I see that you extract the roothash from the following block after the 'block by hash'

I think the header to return is exactly the block by hash, but the rootHash to return here should be the ScheduledRootHash from the following block after the `block by hash`

``` go
return blockHeader, header.GetScheduledRootHash, nil
```

process/smartContract/scQueryService.go
```diff
@@ -116,60 +176,204 @@ func (service *SCQueryService) shouldAllowQueriesExecution() bool {
...
+	header, err := process.UnmarshalHeader(shardId, service.marshaller, headerBuffer)
+	if err != nil {
+		return nil, err
+	}
+
+	return header, nil
+}
+
+func (service *SCQueryService) getBlockHeader(currentHeader data.HeaderHandler) data.HeaderHandler {
+	header, _, err := service.getBlockHeaderByNonce(currentHeader.GetNonce() + 1)
```
AdoAdoAdo, line 340: why nonce +1 here?
Maybe rename the caller method to getNextBlockHeader?

also I don't think we need this method, as we already have getBlockByNonce (current header nonce +1), and an error on this get should not use the current header, but propagate the error.

process/smartContract/scQueryService.go
```diff
@@ -116,60 +176,204 @@ func (service *SCQueryService) shouldAllowQueriesExecution() bool {
...
+		service.storageService,
+		service.uint64ByteSliceConverter,
+		service.marshaller,
+		hashByNonceUnit,
+	)
 }
 
 func (service *SCQueryService) checkForRootHashChanges(rootHashBefore []byte) error {
-	rootHashAfter := service.blockChain.GetCurrentBlockRootHash()
+	rootHashAfter := service.mainBlockChain.GetCurrentBlockRootHash()
```
AdoAdoAdo, line 376: shouldn't it verify this against the api blockchain block roothash?

I would only use the mainBlockChain to set the apiBlockChain data when the API is called without block parameters (so for current)

state/accountsDBApi.go
```diff
@@ -164,9 +164,10 @@ func (accountsDB *accountsDBApi) RootHash() ([]byte, error) {
 	return blockInfo.GetRootHash(), nil
 }
 
-// RecreateTrie is a not permitted operation in this implementation and thus, will return an error
-func (accountsDB *accountsDBApi) RecreateTrie(_ []byte) error {
-	return ErrOperationNotPermitted
+// RecreateTrie is used to reload the trie based on an existing rootHash
+func (accountsDB *accountsDBApi) RecreateTrie(rootHash []byte) error {
+	_, err := accountsDB.doRecreateTrieWithBlockInfo(holders.NewBlockInfo([]byte{}, 0, rootHash))
```
AdoAdoAdo, line 169: do you afterwards use the blockInfo set on the accountsDBApi?
if yes, the maybe use instead 
```go
accountsDB.innerAccountsAdapter.RecreateTrie(rootHash)
```

AdoAdoAdo (changes requested):

factory/api/apiResolverFactory.go
iulianpascalau: We need more AccountsDBApi instances as it is easier to manage the inner trie instances that way. AccountsDB is just a wrapper over the trie.

genesis/process/shardGenesisBlockCreator.go
```diff
@@ -692,14 +694,26 @@ func createProcessorsForShardGenesisBlock(arg ArgsGenesisBlockCreator, enableEpo
...
 		Bootstrapper:             syncDisabled.NewDisabledBootstrapper(),
 		AllowExternalQueriesChan: common.GetClosedUnbufferedChannel(),
+		HistoryRepository:        arg.HistoryRepository,
+		ShardCoordinator:         arg.ShardCoordinator,
+		StorageService:           arg.Data.StorageService(),
+		Marshaller:               arg.Core.InternalMarshalizer(),
+		Hasher:                   arg.Core.Hasher(),
+		Uint64ByteSliceConverter: arg.Core.Uint64ByteSliceConverter(),
 	}
 	queryService, err := smartContract.NewSCQueryService(argsNewSCQueryService)
```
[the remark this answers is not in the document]
iulianpascalau: At the time it was the easiest way to verify that the genesis was constructed correctly. As seen, during testing, without these checks we can have a malformed genesis state if we got some genesis parameters wrong (like the initial number of staked nodes). I strongly encourage to keep them as they are now in order to safeguard the genesis state creation.

process/interface.go
```diff
@@ -538,6 +538,7 @@ type BlockChainHookHandler interface {
 	GetCounterValues() map[string]uint64
 	IsInterfaceNil() bool
 	IsBuiltinFunctionName(functionName string) bool
+	GetAccountsAdapter() state.AccountsAdapter
```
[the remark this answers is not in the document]
sstanculeanu: updated

process/smartContract/scQueryService.go
sstanculeanu: updated

process/smartContract/scQueryService.go
sstanculeanu: removed the method and updated as suggested

state/accountsDBApi.go
sstanculeanu: reused this method as the blockInfo is the returned on `GetAccountWithBlockInfo` and `RootHash`

bogdan-rosianu (dismissed):

process/smartContract/scQueryService.go
```diff
@@ -260,23 +260,41 @@ func (service *SCQueryService) executeScCall(query *process.SCQuery, gasPrice ui
...
 	if query.BlockNonce.HasValue {
-		blockHeader, _, err := service.getBlockHeaderByNonce(query.BlockNonce.Value)
+		currentHeader, _, err := service.getBlockHeaderByNonce(query.BlockNonce.Value)
 		if err != nil {
 			return nil, nil, err
 		}
 
-		header := service.getBlockHeader(blockHeader)
-		return header, header.GetRootHash(), nil
+		blockHeader, _, err := service.getBlockHeaderByNonce(currentHeader.GetNonce() + 1)
```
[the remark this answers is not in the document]
sstanculeanu: pushed

AdoAdoAdo (approved):

iulianpascalau (approved):

PR #5544, opened 2023-08-30
Extended the search for the full history resolver

iulianpascalau:
- in the case of full history observers, if the epoch is wrongly provided, the node does not search the current active storers like the normal nodes. Default the search first call if the epoch search fails.
  
## Proposed changes
- extended the search for the full history resolver in case the epoch is wrongly provided

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/dataRetriever/resolvers/baseFullHistoryResolver_test.go
+++ b/dataRetriever/resolvers/baseFullHistoryResolver_test.go
@@ -1,0 +1,101 @@
+package resolvers
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/multiversx/mx-chain-go/testscommon/storage"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBaseFullHistoryResolver_SearchFirst(t *testing.T) {
+	t.Parallel()
+
+	testKey := []byte("key")
+	testValue := []byte("value")
+	resolver := &baseFullHistoryResolver{
+		storer: &storage.StorerStub{
+			SearchFirstCalled: func(key []byte) ([]byte, error) {
+				assert.Equal(t, testKey, key)
+
+				return testValue, nil
+			},
+		},
+	}
+
+	val, err := resolver.searchFirst(testKey)
+	assert.Nil(t, err)
+	assert.Equal(t, testValue, val)
+}
+
+func TestBaseFullHistoryResolver_GetFromStorage(t *testing.T) {
+	t.Parallel()
+
+	testKey := []byte("key")
+	testValue := []byte("value")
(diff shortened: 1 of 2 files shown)


dataRetriever/resolvers/baseFullHistoryResolver_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+					return testValue, nil
+				},
+			},
+		}
+
+		val, err := resolver.getFromStorage(testKey, testEpoch)
+		assert.Nil(t, err)
+		assert.Equal(t, testValue, val)
+	})
```
andreibancioiu, line 58: Formatting / newline needed (and below).

sstanculeanu (dismissed):

dataRetriever/resolvers/baseFullHistoryResolver_test.go
```diff
@@ -0,0 +1,101 @@
...
+					return nil, nil
+				},
+			},
+		}
+
+		val, err := resolver.getFromStorage(testKey, testEpoch)
+		assert.Nil(t, err)
+		assert.Equal(t, 0, len(val))
+	})
+	t.Run("get from epoch returned error and will default to search first", func(t *testing.T) {
```
andreibancioiu, line 80: `"will fallback to search first"`.

andreibancioiu (dismissed):

dataRetriever/resolvers/baseFullHistoryResolver_test.go
iulianpascalau: changed

dataRetriever/resolvers/baseFullHistoryResolver_test.go
iulianpascalau: I generally prefer inner tests not separated by a new line. We should agree in a future clean code meeting if we will add empty lines or not in this particular case as to slowly align all existing code

sstanculeanu (approved):

andreibancioiu (approved):

gabi-vuls (approved):
Normal allin test: v1.5.13-dev-config-8c685c8e00 -> extend-full-history-resolv-0739badb6e

--- Specific errors ---

block hash does not match 7756
wrong nonce in block 2769
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  247
Nr. of new ERRORS:  0
Nr. of new WARNS:  2
Nr. of PANICS:  0

/------/

PR #5545, opened 2023-08-30
added additionalStorageService

raduchis:
- this was the last place where the chainRunType was actively used
  
## Proposed changes
- creation of two different versions of StorageServiceCreators for normal and sovereign mode

## Testing procedure
- #allIn

## Pre-requisites

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -664,3 +664,6 @@ var ErrNilTransactionCoordinatorCreator = errors.New("nil transaction coordinato
 
 // ErrNilValidatorStatisticsProcessorCreator signals that a nil validator statistics processor creator has been provided
 var ErrNilValidatorStatisticsProcessorCreator = errors.New("nil validator statistics processor creator")
+
+// ErrNilFunction signals that a nil function has been provided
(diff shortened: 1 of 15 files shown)


sasurobert (dismissed):

mariusmihaic (dismissed):
This approach is still a hack.

We should have a `StorageServiceFactoryCreator` interface that has only one method: **Create**(instead of **CreateForShard** +  **CreateForMeta**).

We should split this file in multiple files like` shardStorageServiceFactory.go, sovereignStorageServiceFactory.go, metaStorageServiceFactory.go`

We need to create a `SovereignStorageServiceFactory` ,which holds a pointer internally to `StorageServiceFactory`. On `CreateForSovereign`  (which shall be `Create` (meta/shard/sovereign) we call `CreateForShard` -> which returns a `dataRetriever.StorageService`. Then we add additional sovereign storage units to that one

Also, `AdditionalStorageServiceCreator` shall be deleted and replaced with  a `StorageServiceFactoryCreator`

sasurobert (approved):

PR #5572, opened 2023-09-08
Relayed v3

sstanculeanu:
- New approach for relayed txs, v3, where the user tx is set as a new field of the parent tx, which would simplify and lower the processing cost to the cost of 2 different txs, as there will be no more extra data field.
  
## Proposed changes
- implement the changes needed

## Testing procedure
- standard system test
- relayed v3 specific tests

## Pre-requisites


process/coordinator/transactionType.go
```diff
@@ -185,6 +194,15 @@ func (tth *txTypeHandler) isRelayedTransactionV2(functionName string) bool {
 	return functionName == core.RelayedTransactionV2
 }
 
+func (tth *txTypeHandler) isRelayedTransactionV3(tx data.TransactionHandler) bool {
+	rtx, ok := tx.(relayedV3TransactionHandler)
+	if !ok {
+		return false
+	}
+
+	return rtx.GetInnerTransaction() != nil
```
sasurobert, line 203: check.Ifnil

process/coordinator/transactionType.go
```diff
@@ -16,6 +17,10 @@ import (
 
 var _ process.TxTypeHandler = (*txTypeHandler)(nil)
 
+type relayedV3TransactionHandler interface {
+	GetInnerTransaction() *transaction.Transaction
```
sasurobert, line 21: can't you add this function to the data.TransactionHandler interface ? and have GetInnerTransaction() data.TransactionHandler as the actual function ? It is better if we do not have concrete structure here.

process/dataValidators/txValidator.go
```diff
@@ -115,6 +121,12 @@ func (txv *txValidator) getSenderUserAccount(
 }
 
 func (txv *txValidator) checkBalance(interceptedTx process.InterceptedTransactionHandler, account state.UserAccountHandler) error {
+	rTx, ok := interceptedTx.Transaction().(relayedV3TransactionHandler)
+	if ok && len(rTx.GetRelayerAddr()) > 0 {
+		// early return if this is a user tx of relayed v3, no need to check balance
```
sasurobert, line 126: check balance of relayer is a must.

Like the transaction sender is the relayer, if it has InnerTransaction - the inner TX should be for the user. First we execute the balance transfer from the relayer to the user, and execute the inner TX after. 

GetRelayerAddr should be a field in the InnerTransaction and in the SmartContractResults.

process/transaction/interceptedTransaction.go
```diff
@@ -199,51 +199,64 @@ func (inTx *InterceptedTransaction) CheckValidity() error {
...
 	}
 
-	userTx, err := createRelayedV2(tx, userTxArgs)
-	if err != nil {
-		return err
+	innerTx := tx.InnerTransaction
+	if !bytes.Equal(innerTx.SndAddr, tx.RcvAddr) {
+		return process.ErrRelayedTxV3BeneficiaryDoesNotMatchReceiver
+	}
+	if len(innerTx.RelayerAddr) == 0 {
```
sasurobert, line 228: why do you need a relayerAddress in the innerTX ? it is not needed. Duplicate information. You have the relayerAddress from the tx.SndAddr

process/transactionEvaluator/transactionEvaluator.go
```diff
@@ -103,7 +103,7 @@ func (ate *apiTransactionEvaluator) ComputeTransactionGasLimit(tx *transaction.T
 	switch txTypeOnSender {
 	case process.SCDeployment, process.SCInvoking, process.BuiltInFunctionCall, process.MoveBalance:
 		return ate.simulateTransactionCost(tx, txTypeOnSender)
-	case process.RelayedTx, process.RelayedTxV2:
+	case process.RelayedTx, process.RelayedTxV2, process.RelayedTxV3:
 		// TODO implement in the next PR
```
sasurobert, line 107: let's add a task for this on Jira - a long forgotten TODO.

process/transaction/shardProcess.go
```diff
@@ -612,6 +614,35 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
 	return txProc.accounts.SaveAccount(acntDst)
 }
 
+func (txProc *txProcessor) processRelayedTxV3(
+	tx *transaction.Transaction,
+	relayerAcnt, acntDst state.UserAccountHandler,
+) (vmcommon.ReturnCode, error) {
+	if !txProc.enableEpochsHandler.IsRelayedTransactionsV3FlagEnabled() {
+		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3Disabled)
```
sasurobert, line 622: this has to be backward compatible. So let's make sure that TX is not accepted in the pool if it is of RelayedV3 and the flag is not active. This check should be done at transaction interceptor level.

The problem is nodes with different software will run at the same time in the first few days of the announcement of the new release.

process/transaction/shardProcess.go
```diff
@@ -612,6 +614,35 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+	}
+	if tx.GetValue().Cmp(big.NewInt(0)) != 0 {
+		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3ZeroVal)
+	}
+
+	userTx := tx.GetInnerTransaction()
+	if !bytes.Equal(tx.RcvAddr, userTx.SndAddr) {
+		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3BeneficiaryDoesNotMatchReceiver)
+	}
+	if len(userTx.RelayerAddr) == 0 {
```
sasurobert, line 632: I do not see the need for the user to put a relayerAddress in his own TX. This was used internally, in case of relayedV1 and v2. It was not the user who field this value, but the protocol in case of processing.

sasurobert (changes requested):

integrationTests/multiShard/relayedTx/relayedTxV2_test.go
```diff
@@ -1,104 +0,0 @@
...
-	"time"
-
-	"github.com/multiversx/mx-chain-core-go/data/transaction"
-	"github.com/multiversx/mx-chain-go/integrationTests"
-	"github.com/multiversx/mx-chain-go/integrationTests/vm/wasm"
-	vmFactory "github.com/multiversx/mx-chain-go/process/factory"
-	"github.com/stretchr/testify/assert"
-)
-
-func TestRelayedTransactionV2InMultiShardEnvironmentWithSmartContractTX(t *testing.T) {
```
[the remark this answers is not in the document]
sstanculeanu: the code is not lost, but moved [here](https://github.com/multiversx/mx-chain-go/blob/relayedV3/integrationTests/multiShard/relayedTx/relayedTx_test.go#L26), in order to reuse it for different types of relayed

node/node.go
```diff
@@ -54,7 +54,8 @@ var log = logger.GetOrCreate("node")
 var _ facade.NodeHandler = (*Node)(nil)
 
 // Option represents a functional configuration parameter that can operate
-//  over the None struct.
+//
```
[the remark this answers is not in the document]
sstanculeanu: fixed

process/coordinator/transactionType.go
sstanculeanu: updated the TransactionHandler interface

sstanculeanu: updated TransactionHandler interface

process/dataValidators/txValidator.go
sstanculeanu: it won't apply for the relayed transaction

this check is only valid for the inner transaction. Reason: only transaction is castable to this structure, then only an inner transaction would have the relayer address set

the reason why this bypass is needed is because now, at api level, we also [create the inner transaction](https://github.com/multiversx/mx-chain-go/blob/relayedV3/api/groups/transactionGroup.go#L310), which would fail in the following scenario:
- User A with 0 egld signs a tx to move some ESDT
- Relayer B will pay the fee for it
- Since A does not have EGLD, the inner tx creation at API level would have failed

process/transaction/interceptedTransaction.go
sstanculeanu: added it for an extra check, to be the same with (parent tx).SndAddr, as discussed with @AdoAdoAdo in order to prevent sandwich attacks

process/transactionEvaluator/transactionEvaluator.go
sstanculeanu: [added](https://multiversxlabs.atlassian.net/browse/MX-14626)

process/transaction/shardProcess.go
sstanculeanu: added on inner txs only, for extra check on relayed v3, as mentioned above

process/transaction/shardProcess.go
sstanculeanu: added

process/transaction/shardProcess_test.go
```diff
@@ -2019,6 +2019,208 @@ func TestTxProcessor_ProcessRelayedTransactionV2(t *testing.T) {
...
+		args.PubkeyConv = pubKeyConverter
+		args.ArgsParser = smartContract.NewArgumentParser()
+		args.EnableEpochsHandler = &enableEpochsHandlerMock.EnableEpochsHandlerStub{}
+		execTx, _ := txproc.NewTxProcessor(args)
+
+		returnCode, err := execTx.ProcessTransaction(tx)
+		assert.Equal(t, process.ErrFailedTransaction, err)
+		assert.Equal(t, vmcommon.UserError, returnCode)
+	})
+	t.Run("value on relayed tx should error", func(t *testing.T) {
```
AdoAdoAdo, line 2104: "value on relay ...tx"?
the relayed tx is the user tx

process/transaction/interceptedTransaction_test.go
```diff
@@ -1497,6 +1573,111 @@ func TestInterceptedTransaction_CheckValidityOfRelayedTxV2(t *testing.T) {
...
+		GasLimit:         10,
+		GasPrice:         4,
+		RcvAddr:          recvAddress,
+		SndAddr:          senderAddress,
+		Signature:        sigOk,
+		ChainID:          chainID,
+		Version:          minTxVersion,
+		InnerTransaction: innerTx,
+	}
+	txi, _ := createInterceptedTxFromPlainTxWithArgParser(tx)
```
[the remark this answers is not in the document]
sstanculeanu: done

AdoAdoAdo (approved):

process/transaction/interceptedTransaction.go
sasurobert: that's good. As the user can choose the relayer.

sasurobert: But add the check

process/transaction/shardProcess.go
sasurobert: add the full check here as well userTx.RelayerAddr = tx.SenderAddr

process/dataValidators/txValidator.go
sasurobert: Ok. I get that this is for user tx only. so it is the check for inner tx

sstanculeanu: exactly, but does not work as expected.. on it

process/transaction/interceptedTransaction.go
sstanculeanu: added

process/transaction/shardProcess.go
sstanculeanu: added

process/dataValidators/txValidator.go
sstanculeanu: completely removed it, not needed

AdoAdoAdo (approved):

sasurobert (approved):

PR #5595, opened 2023-09-20
Fix snapshot in the wrong epoch

BeniaminDrasovean:
- There were some edge cases in which a snapshot would start right before the node would reach the next epoch, so there were 2 snapshots in the same epoch.
  
## Proposed changes
- Get the right epoch when starting a snapshot from the epochStartTrigger

## Testing procedure
- Normal testing procedure

## Pre-requisites

### DIFF
--- a/process/block/shardblock.go
+++ b/process/block/shardblock.go
@@ -1258,7 +1258,7 @@ func (sp *shardProcessor) snapShotEpochStartFromMeta(header data.ShardHeaderHand
 				log.Debug("using scheduled root hash for snapshotting", "schRootHash", schRootHash)
 				rootHash = schRootHash
 			}
-			epoch := header.GetEpoch()
+			epoch := sp.epochStartTrigger.MetaEpoch()
 			log.Debug("shard trie snapshot from epoch start shard data", "rootHash", rootHash, "epoch", epoch)
 			accounts.SnapshotState(rootHash, epoch)
 			sp.markSnapshotDoneInPeerAccounts()


process/block/shardblock.go
```diff
@@ -1258,7 +1258,7 @@ func (sp *shardProcessor) snapShotEpochStartFromMeta(header data.ShardHeaderHand
 				log.Debug("using scheduled root hash for snapshotting", "schRootHash", schRootHash)
 				rootHash = schRootHash
 			}
-			epoch := header.GetEpoch()
+			epoch := sp.epochStartTrigger.MetaEpoch()
```
iulianpascalau, line 1261: I think this is pretty solid as the rollback of the epoch start metablock is not possible in this case. I remember that the epoch start trigger triggers the new epoch after it has the confirmation metablock

iulianpascalau (approved):

gabi-vuls (approved):
Normal allin test: v1.5.13-dev-config-8c685c8e00 -> fix-snapshot-in-the-wrong--7eed456aa5

--- Specific errors ---

block hash does not match 5150
wrong nonce in block 1963
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  235
Nr. of new ERRORS:  0
Nr. of new WARNS:  0
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

--- WARNINGS ---

/------/

AdoAdoAdo (approved):

PR #5601, opened 2023-09-25
Node redundancy system fixes

iulianpascalau:
- the redundancy system did not perform as expected
  
## Proposed changes
1. added & integrated a new component called `sentSignatureTracker` in order to avoid false positives for a redundancy multikey node when the main multikey node sent the signatures directly as part of the final info. This situation happens when the multikey node has one key as a leader and one or more keys as validators in the same round
2. `subroundSignature` component needs to take into account that the provided key should be active to send signature messages (situation appeared when the single-key backup node kept sending signatures despite the fact that the main machine recovered)
3. removed an early exit in `subroundStartRound` component that made the backup node process one block behind the current state of the chain
4. made the call to `sr.IncrementRoundsWithoutReceivedMessages(pkBytes)` in `subroundStartRound` unconditional to the managed keys info as otherwise the redundancy mechanism won't work
5. refactored the old method `UpdatePublicKeyLiveness` as the peer ID condition was not ok, moved the condition inside the `managedPeersHolder` component where we have the correct peer ID.
6. `IsMultiKeyMode` function from the `managedPeersHolder` should return true if we start the node in multikey mode regardless the redundancy level so the factories can create the correct components (because of this bug we had created `heartbeatSender` component instead of `multikeyHeartbeatSender` component)
7. `peerInfo` component had a bug in the `isNodeActiveOnMainMachine` function as the comparison was done strictly on the provided `maxRoundsWithoutReceivedMessages`. If that value was 1, it would have false-positively triggered the backup node because the normal working interval for the `pInfo.roundsWithoutReceivedMessages` variable is [0; 1].
8. fixed behavior when no validatorKey.pem was provided but the allValidatorsKeys.pem contained keys. Deprecated `-no-key` option.
9. removed duplicated code from the single-key & multi-key handler that kept the number of rounds missed by the main machine. Uniformized the behavior between the 2 components and extracted one config value. 
10. when the node is running in multi-key mode, for the node's heartbeat message, send the provided key, not the one handled in the redundancyHandler used in single-key mode as to bring clarity whenever the node needs to be identified in the explorer

## Testing procedure
- redundancy tests with single key mode & multikey mode

## Pre-requisites


keysManagement/keysHandler.go
```diff
@@ -115,13 +115,9 @@ func (handler *keysHandler) IsOriginalPublicKeyOfTheNode(pkBytes []byte) bool {
...
-// UpdatePublicKeyLiveness update the provided public key liveness if the provided pid is not managed by the current node
-func (handler *keysHandler) UpdatePublicKeyLiveness(pkBytes []byte, pid core.PeerID) {
-	if bytes.Equal(handler.pid.Bytes(), pid.Bytes()) {
-		return
-	}
-
-	handler.managedPeersHolder.ResetRoundsWithoutReceivedMessages(pkBytes)
+// ResetRoundsWithoutReceivedMessages calls the ResetRoundsWithoutReceivedMessages on the managed peers holder
+func (handler *keysHandler) ResetRoundsWithoutReceivedMessages(pkBytes []byte, pid core.PeerID) {
+	handler.managedPeersHolder.ResetRoundsWithoutReceivedMessages(pkBytes, pid)
```
iulianpascalau, line 120: the condition was moved in the `managedPeersHolder.ResetRoundsWithoutReceivedMessages` function L270

consensus/spos/bls/export_test.go
```diff
@@ -332,3 +332,11 @@ func (sr *subroundEndRound) CreateAndBroadcastInvalidSigners(invalidSigners []by
 func (sr *subroundEndRound) GetFullMessagesForInvalidSigners(invalidPubKeys []string) ([]byte, error) {
 	return sr.getFullMessagesForInvalidSigners(invalidPubKeys)
 }
+
+func (sr *subroundEndRound) GetSentSignatureTracker() spos.SentSignaturesTracker {
```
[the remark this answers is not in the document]
iulianpascalau: export_test.go file. Will add mock comments on all methods

consensus/spos/bls/subroundEndRound.go
```diff
@@ -838,23 +853,34 @@ func (sr *subroundEndRound) doEndRoundConsensusCheck() bool {
 	return false
 }
 
-func (sr *subroundEndRound) checkSignaturesValidity(bitmap []byte) error {
+// ComputeSignersPublicKeys will extract from the provided consensus group slice only the strings that matched with the bitmap
+func ComputeSignersPublicKeys(consensusGroup []string, bitmap []byte) []string {
```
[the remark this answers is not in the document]
iulianpascalau: done

sstanculeanu (approved):

consensus/spos/bls/export_test.go
```diff
@@ -332,3 +332,11 @@ func (sr *subroundEndRound) CreateAndBroadcastInvalidSigners(invalidSigners []by
...
+	return sr.sentSignatureTracker
+}
+
+func (sr *subroundStartRound) GetSentSignatureTracker() spos.SentSignaturesTracker {
```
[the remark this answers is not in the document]
iulianpascalau: reorganized

consensus/spos/bls/subroundStartRound_test.go
```diff
@@ -275,9 +365,16 @@ func TestSubroundStartRound_DoStartRoundConsensusCheckShouldReturnTrueWhenInitCu
 	container.SetBootStrapper(bootstrapperMock)
 
 	sr := *initSubroundStartRoundWithContainer(container)
+	sentTrackerInterface := sr.GetSentSignatureTracker()
+	sentTracker := sentTrackerInterface.(*mock.SentSignatureTrackerStub)
+	startRoundCalled := true
```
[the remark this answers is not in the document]
iulianpascalau: right, thanks

consensus/spos/sentSignatureTracker_test.go
```diff
@@ -0,0 +1,94 @@
+package spos
```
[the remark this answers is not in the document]
iulianpascalau: renamed

sstanculeanu (dismissed):

SebastianMarian (approved):

factory/crypto/cryptoComponents.go
```diff
@@ -334,14 +329,16 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
...
 		return ccf.generateCryptoParams(keygen, "in import mode", handledPrivateKeys)
 	}
-	if ccf.noKeyProvided {
-		return ccf.generateCryptoParams(keygen, "with no-key flag enabled", make([][]byte, 0))
-	}
-	if len(handledPrivateKeys) > 0 {
-		return ccf.generateCryptoParams(keygen, "running with a provided allValidatorsKeys.pem", handledPrivateKeys)
+	cp, err := ccf.readCryptoParams(keygen)
+	if err == nil {
+		cp.handledPrivateKeys = handledPrivateKeys
```
sstanculeanu, line 334: with this new approach, if `validatorKey.pem` is present, we'll use it as the **observer's** key that manages all keys

may be confusing, perhaps we should update the `Usage` field of the `validator-key-pem-file` flag?

factory/crypto/cryptoComponents.go
```diff
@@ -334,14 +329,16 @@ func (ccf *cryptoComponentsFactory) createCryptoParams(
...
+
+		return cp, nil
 	}
 
-	return ccf.readCryptoParams(keygen)
+	log.Debug("failure while reading the BLS key, will autogenerate one", "error", err)
+
+	return ccf.generateCryptoParams(keygen, "running without a provided BLS key", handledPrivateKeys)
```
sstanculeanu, line 341: perhaps improve this message to mention if node is multikey or not? (and add one more similar warning on L335)

cmd/node/config/config.toml
```diff
@@ -955,3 +955,8 @@
         Capacity = 50000
         Type = "SizeLRU"
         SizeInBytes = 314572800 #300MB
+
+[Redundancy]
+    # MaxRoundsOfInactivityAccepted defines the number of rounds missed by a main or higher level backup machine before
+    # the current machine will take over and propose/sign blocks. Used in both single-key and multi-key modes.
+    MaxRoundsOfInactivityAccepted = 3
```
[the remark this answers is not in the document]
iulianpascalau: done

factory/crypto/cryptoComponents.go
iulianpascalau: updated the Usage section.

factory/crypto/cryptoComponents.go
iulianpascalau: done, the messages can be in this form (switched to INFO, btw)
```
the node loaded the validatorKey.pem file and is running in multi-key mode, managing 4 keys
the node loaded the validatorKey.pem file and is running in single-key mode 
the node is running in single-key mode! Will generate a fresh new BLS key 
the node is running in multi-key mode, managing 4 keys! Will generate a fresh new BLS key 
the node is in import-db mode! Will generate a fresh new BLS key 
```

sstanculeanu (dismissed):

redundancy/redundancy.go
```diff
@@ -39,19 +44,24 @@ func NewNodeRedundancy(arg ArgNodeRedundancy) (*nodeRedundancy, error) {
...
+		observerPrivateKey:    arg.ObserverPrivateKey,
 	}
 
 	return nr, nil
 }
 
 // IsRedundancyNode returns true if the current instance is used as a redundancy node
 func (nr *nodeRedundancy) IsRedundancyNode() bool {
-	return nr.redundancyLevel != 0
+	return nr.handler.IsRedundancyNode(nr.maxRoundsOfInactivity)
```
SebastianMarian, line 64: nr.mutNodeRedundancy.RLock()
	defer nr.mutNodeRedundancy.RUnlock()
?

redundancy/redundancy.go
```diff
@@ -100,16 +111,16 @@ func (nr *nodeRedundancy) ResetInactivityIfNeeded(selfPubKey string, consensusMs
...
 
 	nr.mutNodeRedundancy.Lock()
-	nr.roundsOfInactivity = 0
+	nr.handler.ResetRoundsOfInactivity()
 	nr.mutNodeRedundancy.Unlock()
 }
 
 func (nr *nodeRedundancy) isMainMachineActive() bool {
-	if nr.redundancyLevel < 0 {
+	if !nr.handler.IsRedundancyNode(nr.maxRoundsOfInactivity) {
```
SebastianMarian, line 119: I think that this condition could be deleted if the change suggested above in redundancyHandler.go, regarding the condition **<=** instead **<** will be done! We also can keep it like this, if we want that this method always returns true for the main machine, regardless the number of missed rounds by itself

redundancy/common/redundancyHandler.go
```diff
@@ -0,0 +1,60 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+)
+
+const minRoundsOfInactivity = 2 // the system does not work as expected with the value of 1
```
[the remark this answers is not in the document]
iulianpascalau: Technically, it can work with the value of 2. Maybe someone wants to set it to 2 as the backups will kick in more faster. I would keep it set to 2 but the config value from the .toml file set to 3.

redundancy/common/redundancyHandler.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// ResetRoundsOfInactivity will reset the rounds of inactivity
+func (handler *redundancyHandler) ResetRoundsOfInactivity() {
+	handler.roundsOfInactivity = 0
+}
+
+// IsMainMachineActive returns true if the main machine is still active
+func (handler *redundancyHandler) IsMainMachineActive(maxRoundsOfInactivity int) bool {
+	return handler.roundsOfInactivity < maxRoundsOfInactivity
```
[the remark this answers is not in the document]
iulianpascalau: right, will change. Studied the code yesterday and the naming is a bit confusing, indeed. All calls of this method first check whether the node is main or not. Will think of a cleaner solution to this.

redundancy/redundancy.go
iulianpascalau: can add but is not needed due to the fact that the handler's pointer won't change and the function IsRedundancyNode does not need a mutex. Will refactor

redundancy/redundancy.go
iulianpascalau: will refactor

redundancy/common/redundancyHandler.go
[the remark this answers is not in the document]
SebastianMarian: ok

redundancy/common/redundancyHandler.go
```diff
@@ -33,10 +33,10 @@ func CheckMaxRoundsOfInactivity(maxRoundsOfInactivity int) error {
 	return nil
 }
 
-// IsRedundancyNode returns true if the provided maxRoundsOfInactivity value is higher than the
+// IsMainNode returns true if the provided maxRoundsOfInactivity value is equal than the
```
SebastianMarian, line 36: equal with ?

SebastianMarian (approved):

redundancy/common/redundancyHandler.go
iulianpascalau: changed

sstanculeanu (approved):

SebastianMarian (approved):

gabi-vuls (approved):

PR #5605, opened 2023-09-27
Fixed tx removal from inner tx pool

sstanculeanu:
- after latest fixes, the structure was not castable anymore
  
## Proposed changes
- updated the registration process

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/dataRetriever/txpool/shardedTxPool.go
+++ b/dataRetriever/txpool/shardedTxPool.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/core/check"
 	"github.com/multiversx/mx-chain-core-go/core/counting"
 	"github.com/multiversx/mx-chain-core-go/data"
 	"github.com/multiversx/mx-chain-go/dataRetriever"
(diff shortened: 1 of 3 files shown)


dataRetriever/txpool/shardedTxPool.go
```diff
@@ -387,6 +387,11 @@ func (txPool *shardedTxPool) Diagnose(deep bool) {
 	}
 }
 
+// NotifyEviction will be called when a transaction is removed from the main tx pool
+func (txPool *shardedTxPool) NotifyEviction(txHash []byte) {
```
andreibancioiu, line 391: Who calls `NotifyEviction`? If called on a large collection, might cause performance issues (since `removeTxFromAllShards` isn't extremely fast (if I remember well).

andreibancioiu: Also, does not seem to be covered by tests.

sstanculeanu: `NotifyEviction` is called on a new go routine, from mx-chain-storage-go, and the calls are protected through a worker pool, allowing max 5 go routines at a time. 

Added tests

AdoAdoAdo (approved):

dataRetriever/txpool/shardedTxPool.go
AdoAdoAdo: maybe we can do some benchmarking to have a  clearer view on the performance

andreibancioiu (approved):

PR #5611, opened 2023-09-29
Revert tx that migrates data bugfix

BeniaminDrasovean:
- There is an edgecase relating to trie pruning + migrate data trie + revert. If a transaction that triggers the migration of a certain data trie value is reverted, the old version of the trie leaf that was migrated is added back to the trie, but the new migrated leaf is not removed. This means that the root hash of the data trie after the revert will not be the same as before.
- This bug can be noticed only if pruning is enabled because even though the data trie will reload the old rootHash correctly, it will also mark it for eviction. 
- Another edgecase which is fixed in this PR: If a trie branch node migrates only one of it's children and then that child is removed/the migration is reverted, the branch node will have an initialized `ChildrenVersion` slice instead of a nil byte slice. This will result in a different root hash. 
  
## Proposed changes
- To prevent the eviction of the needed root hash, the data trie needs to have the same root hash after the revert as before. This will mean that the root hash will be present in both `oldHashes` and `newHashes`, thus the removal will be canceled.

## Testing procedure
- Start the chain with snapshots disabled and pruning enabled, and after the `autoBalanceDataTriesFlag` is enabled, send some txs (which will change the existing data trie leaves) that will fail. 
- Normal testing procedure

## Pre-requisites


state/accountsDB_test.go
```diff
@@ -3065,6 +3065,82 @@ func TestAccountsDB_SaveKeyValAfterAccountIsReverted(t *testing.T) {
...
+	spm.MarkForEvictionCalled = func(_ []byte, _ []byte, oldHashes common.ModifiedHashes, newHashes common.ModifiedHashes) error {
+		_, ok := oldHashes[string(userAccRootHash)]
+		require.True(t, ok)
+		_, ok = newHashes[string(userAccRootHash)]
+		require.True(t, ok)
+		markForEvictionCalled = true
+
+		return nil
+	}
+	_, err = adb.Commit()
```
iulianpascalau, line 3140: from linter: `ineffectual assignment to err (ineffassign)`

state/trackableDataTrie/trackableDataTrie.go
```diff
@@ -234,12 +234,20 @@ func (tdt *trackableDataTrie) updateTrie(dtr state.DataTrie) ([]core.TrieData, e
...
 
-		err = tdt.modifyTrie([]byte(key), dataEntry, oldVal, dtr)
+		newKey, err := tdt.modifyTrie([]byte(key), dataEntry, oldVal, dtr)
 		if err != nil {
 			return nil, err
 		}
 
 		index++
+
+		isMigration := oldVal.Version == core.NotSpecified && dataEntry.newVersion == core.AutoBalanceEnabled
```
iulianpascalau, line 244: isFirstMigration? As we might have multiple ones

iulianpascalau (commented):
:+1:

state/accountsDB_test.go
BeniaminDrasovean: Fixed.

state/trackableDataTrie/trackableDataTrie.go
BeniaminDrasovean: Renamed.

trie/branchNode.go
```diff
@@ -647,10 +649,23 @@ func (bn *branchNode) setNewChild(childPos byte, newNode node) error {
...
+	if childVersion == core.NotSpecified {
+		bn.revertChildrenVersionSlice()
+	}
 
 	return nil
 }
 
+func (bn *branchNode) revertChildrenVersionSlice() {
+	for i := range bn.ChildrenVersion {
+		if bn.ChildrenVersion[i] == 1 {
```
iulianpascalau, line 661: hardcoded version?

BeniaminDrasovean: Fixed.

iulianpascalau (dismissed):

trie/branchNode.go
```diff
@@ -647,10 +649,23 @@ func (bn *branchNode) setNewChild(childPos byte, newNode node) error {
...
 	}
 	bn.setVersionForChild(childVersion, childPos)
+	if childVersion == core.NotSpecified {
+		bn.revertChildrenVersionSlice()
+	}
 
 	return nil
 }
 
+func (bn *branchNode) revertChildrenVersionSlice() {
```
AdoAdoAdo, line 659: ca you add unit tests?

state/trackableDataTrie/trackableDataTrie.go
```diff
@@ -234,12 +234,20 @@ func (tdt *trackableDataTrie) updateTrie(dtr state.DataTrie) ([]core.TrieData, e
...
-		err = tdt.modifyTrie([]byte(key), dataEntry, oldVal, dtr)
+		newKey, err := tdt.modifyTrie([]byte(key), dataEntry, oldVal, dtr)
 		if err != nil {
 			return nil, err
 		}
 
 		index++
+
+		isFirstMigration := oldVal.Version == core.NotSpecified && dataEntry.newVersion == core.AutoBalanceEnabled
+		if isFirstMigration && len(newKey) != 0 {
```
AdoAdoAdo, line 245: we only have len(newKey) == 0 here in case of deletes right?

BeniaminDrasovean: Yes.

iulianpascalau (dismissed):

AdoAdoAdo (approved):

trie/branchNode.go
BeniaminDrasovean: Added.

iulianpascalau (dismissed):

AdoAdoAdo (approved):

trie/branchNode_test.go
```diff
@@ -1436,7 +1436,7 @@ func TestBranchNode_VerifyChildrenVersionIsSetCorrectlyAfterInsertAndDelete(t *t
 		}
 		newBn, _, err := bn.insert(data, &testscommon.MemDbMock{})
 		assert.Nil(t, err)
-		assert.Equal(t, 0, len(newBn.(*branchNode).ChildrenVersion))
+		assert.Equal(t, []byte(nil), newBn.(*branchNode).ChildrenVersion)
```
iulianpascalau, line 1439: assert.Nil here and on L1452 & L1492

BeniaminDrasovean: Done.

iulianpascalau (approved):

gabi-vuls (approved):
Normal allin test: v1.5.13-dev-config-8c685c8e00 -> reverted-migration-tx-fix-a3cfd08194

--- Specific errors ---

block hash does not match 7771
wrong nonce in block 2861
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  1
Nr. of all WARNS:  216
Nr. of new ERRORS:  1
Nr. of new WARNS:  3
Nr. of PANICS:  0

/------/

AdoAdoAdo (approved):

PR #5618, opened 2023-10-02
adjust `account not found` error when sending a transaction

bogdan-rosianu:
- previously, when sending a transaction from a 'new' account ('new' means that it didn't have any interaction with the blockchain, that is, neither sending nor receiving a transaction) the returned error would have been 'account not found for address X and shard Y"
- it was not consistent with other situations (an address that exists in the trie but does not have any balance - would have returned other errors)
  
## Proposed changes
- return `insufficient balance` instead of `account not found`

## Testing procedure
- generate a valid transaction from an address that doesn't have any interaction with the chain. When sending it, you should receive an `insufficient balance` instead of `account not found`

## Pre-requisites


api/groups/transactionGroup.go
```diff
@@ -465,6 +468,20 @@ func (tg *transactionGroup) sendMultipleTransactions(c *gin.Context) {
...
+		return nil
+	}
+
+	// TODO: might move these errors from the process package so they can be used here as well
+
+	if !strings.Contains(err.Error(), "account not found") {
+		return err
+	}
+
+	return fmt.Errorf("insufficient balance for address %s", address)
```
iulianpascalau, line 482: this is a bad assumption that any other error other than account not found is caused by insufficient balance. I would add something like:
```
return fmt.Errorf("%w that might be caused by insufficient balance for address %s", err, address)
```

api/groups/transactionGroup.go
```diff
@@ -465,6 +468,20 @@ func (tg *transactionGroup) sendMultipleTransactions(c *gin.Context) {
 	)
 }
 
+func adjustTxValidatorErrorIfNeeded(err error, address string) error {
```
iulianpascalau, line 471: wrapSendError?

iulianpascalau (changes requested):

api/groups/transactionGroup.go
bogdan-rosianu: refactored

api/groups/transactionGroup.go
bogdan-rosianu: refactored

node/node.go
```diff
@@ -54,7 +54,8 @@ var log = logger.GetOrCreate("node")
 var _ facade.NodeHandler = (*Node)(nil)
 
 // Option represents a functional configuration parameter that can operate
-//  over the None struct.
+//
+//	over the None struct.
```
iulianpascalau, line 58: Node struct :)

node/node.go
```diff
@@ -705,7 +706,15 @@ func (n *Node) ValidateTransaction(tx *transaction.Transaction) error {
 		return err
 	}
 
-	return txValidator.CheckTxValidity(intTx)
+	err = txValidator.CheckTxValidity(intTx)
+	if errors.Is(err, process.ErrAccountNotFound) {
```
iulianpascalau, line 710: :+1:

iulianpascalau (dismissed):

miiu96 (dismissed):

gabi-vuls (approved):
Normal allin test: v1.5.13-dev-config-8c685c8e00 -> adjust-account-not-found-e-e890eaa707

--- Specific errors ---

block hash does not match 9265
wrong nonce in block 3496
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  225
Nr. of new ERRORS:  0
Nr. of new WARNS:  1
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

iulianpascalau (approved):

gabi-vuls (approved):

miiu96 (approved):

PR #5642, opened 2023-10-06
lastSnapshot marker fix

BeniaminDrasovean:
- There was an edgecase in which the `lastSnapshot` marker was put in the new epoch before the storers could actually open the new epoch storer. This triggered an `WARN` print, and the marker was not saved to storage.
  
## Proposed changes
- Before calling `PutInEpoch` for the `lastSnapshot` marker, wait for the storers to change epochs.
- Created a new component which handles the `lastSnapshot` marker operations

## Testing procedure
- Normal testing procedure

## Pre-requisites


state/accountsDB_test.go
```diff
@@ -1205,34 +1208,38 @@ func TestAccountsDB_SnapshotStateSkipSnapshotIfSnapshotInProgress(t *testing.T)
...
+					}
 					snapshotMutex.Unlock()
 				},
 				PutInEpochCalled: func(key []byte, val []byte, epoch uint32) error {
 					assert.Equal(t, []byte(state.LastSnapshotStarted), key)
-					assert.Equal(t, rootHashes[epoch], val)
+					assert.Equal(t, rootHashes[epoch-1], val)
 
-					numPutInEpochCalled++
+					numPutInEpochCalled.Set(numPutInEpochCalled.Get() + 1)
```
iulianpascalau, line 1233: careful with these, they won't yield expected results if run in a concurrent manner. Suggestion to use the `atomic` package.

state/export_test.go
```diff
@@ -1,16 +1,13 @@
...
-
 	"github.com/multiversx/mx-chain-core-go/marshal"
 	"github.com/multiversx/mx-chain-go/common"
-	"github.com/multiversx/mx-chain-go/testscommon/storageManager"
 	vmcommon "github.com/multiversx/mx-chain-vm-common-go"
 )
 
 // LastSnapshotStarted -
-const LastSnapshotStarted = lastSnapshot
+const LastSnapshotStarted = "lastSnapshot"
```
iulianpascalau, line 10: export const from `state/lastSnapshotMarker` ?

state/interface.go
```diff
@@ -183,7 +183,8 @@ type DataTrie interface {
 }
 
 // PeerAccountHandler models a peer state account, which can journalize a normal account's data
-//  with some extra features like signing statistics or rating information
+//
```
iulianpascalau, line 186: why this?

state/lastSnapshotMarker/lastSnapshotMarker.go
```diff
@@ -0,0 +1,82 @@
...
+const (
+	lastSnapshot = "lastSnapshot"
+)
+
+type lastSnapshotMarker struct {
+	mutex                       sync.RWMutex
+	latestFinishedSnapshotEpoch uint32
+}
+
+func NewLastSnapshotMarker() *lastSnapshotMarker {
```
iulianpascalau, line 22: missing comments on all exported items

state/lastSnapshotMarker/lastSnapshotMarker.go
```diff
@@ -0,0 +1,82 @@
...
+	err := trieStorageManager.RemoveFromAllActiveEpochs([]byte(lastSnapshot))
+	if err != nil {
+		log.Warn("could not remove lastSnapshot", err, "rootHash", rootHash, "epoch", epoch)
+	}
+
+	lsm.latestFinishedSnapshotEpoch = epoch
+}
+
+func (lsm *lastSnapshotMarker) GetMarkerInfo(trieStorageManager common.StorageManager) ([]byte, error) {
+	lsm.mutex.RLock()
```
iulianpascalau, line 68: useless RLock ?

state/lastSnapshotMarker/lastSnapshotMarker.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	err = trieStorageManager.PutInEpoch([]byte(lastSnapshot), rootHash, epoch)
+	if err != nil {
+		log.Warn("could not set lastSnapshot", err, "rootHash", rootHash, "epoch", epoch, "rootHash", rootHash)
+	}
+}
+
+func (lsm *lastSnapshotMarker) RemoveMarker(trieStorageManager common.StorageManager, epoch uint32, rootHash []byte) {
+	lsm.mutex.RLock()
```
iulianpascalau, line 56: Because of L64 I think this should have been a Lock. Also. The mutex is used here as a critical section or just to protect the `latestFinishedSnapshotEpoch` value?

state/lastSnapshotMarker/lastSnapshotMarker.go
```diff
@@ -0,0 +1,82 @@
...
+			"epoch", epoch,
+			"latestFinishedSnapshotEpoch", lsm.latestFinishedSnapshotEpoch,
+		)
+		return
+	}
+
+	err = trieStorageManager.PutInEpoch([]byte(lastSnapshot), rootHash, epoch)
+	if err != nil {
+		log.Warn("could not set lastSnapshot", err, "rootHash", rootHash, "epoch", epoch, "rootHash", rootHash)
+	}
```
iulianpascalau, line 52: forgotten write to lsm.latestFinishedSnapshotEpoch ? How is this gonna work? :D

iulianpascalau (changes requested):

state/snapshotsManager.go
```diff
@@ -215,10 +211,7 @@ func (sm *snapshotsManager) prepareSnapshot(rootHash []byte, epoch uint32, trieS
 		return nil, true
 	}
 
-	defer func() {
-		err := trieStorageManager.PutInEpoch([]byte(lastSnapshot), rootHash, epoch)
-		handleLoggingWhenError("could not set lastSnapshot", err, "rootHash", rootHash)
-	}()
+	go sm.lastSnapshotMarker.AddMarker(trieStorageManager, epoch, rootHash)
```
ssd04, line 214: AddMarker will check internally for `WaitForStorageEpochChange` , should we avoid that in ImportDB processing mode?

state/snapshotsManager.go
```diff
@@ -71,6 +63,9 @@ func NewSnapshotsManager(args ArgsNewSnapshotsManager) (*snapshotsManager, error
 	if check.IfNil(args.AccountFactory) {
 		return nil, ErrNilAccountFactory
 	}
+	if check.IfNil(args.LastSnapshotMarker) {
+		return nil, ErrNilLastSnapshotMarker
```
ssd04, line 67: unit test for nil check LastSnapshotMarker?

state/accountsDB_test.go
AdoAdoAdo: can you use numPutInEpochCalled.Add(1) instead?

common/storage/storage.go
```diff
@@ -0,0 +1,67 @@
...
+// WaitForStorageEpochChange waits for the storage epoch to change to the given epoch
+func WaitForStorageEpochChange(args StorageEpochChangeWaitArgs) error {
+	log.Debug("waiting for storage epoch change", "epoch", args.Epoch, "wait timeout", args.SnapshotWaitTimeout)
+
+	if args.SnapshotWaitTimeout < args.WaitTimeForSnapshotEpochCheck {
+		return fmt.Errorf("timeout (%s) must be greater than wait time between snapshot epoch check (%s)", args.SnapshotWaitTimeout, args.WaitTimeForSnapshotEpochCheck)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), args.SnapshotWaitTimeout)
+	defer cancel()
```
AdoAdoAdo, line 40: this is not needed as there is no separate goroutine where you treat the ctx.Done, that one is in the same routine, so it will never have any effect.

iulianpascalau: as far as I know, it is go idiomatic to call `defer cancel()` (or to call the `cancel()` function based on a logic)  whenever you work with contexts even if the code is written in such a way that it is impossible to cause memory leaks. If Beni removes that cancel call the linter will complain.

iulianpascalau: In this case, the `cancel()` function needs to be called as we might exit the function through L49, L54, or L58 so we can free up the context resources. 
The same thing is valid for the timer used in the function which is done on L43.

common/storage/storage.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-go/common"
+	logger "github.com/multiversx/mx-chain-logger-go"
+)
+
+var log = logger.GetOrCreate("state")
```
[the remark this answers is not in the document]
BeniaminDrasovean: Changed.

common/storage/storage.go
```diff
@@ -0,0 +1,67 @@
...
+// StorageEpochChangeWaitArgs are the args needed for calling the WaitForStorageEpochChange function
+type StorageEpochChangeWaitArgs struct {
+	TrieStorageManager            common.StorageManager
+	Epoch                         uint32
+	WaitTimeForSnapshotEpochCheck time.Duration
+	SnapshotWaitTimeout           time.Duration
+}
+
+// WaitForStorageEpochChange waits for the storage epoch to change to the given epoch
+func WaitForStorageEpochChange(args StorageEpochChangeWaitArgs) error {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Moved.

state/accountsDB_test.go
BeniaminDrasovean: Done.

state/export_test.go
BeniaminDrasovean: Done.

state/lastSnapshotMarker/lastSnapshotMarker.go
BeniaminDrasovean: Added.

state/lastSnapshotMarker/lastSnapshotMarker.go
BeniaminDrasovean: Yes, removed.

state/lastSnapshotMarker/lastSnapshotMarker.go
BeniaminDrasovean: Indeed, changed to `Lock()` and `Unlock()` . Critical section needed here.

state/lastSnapshotMarker/lastSnapshotMarker.go
BeniaminDrasovean: The marker is added at the start of the snapshot, and is removed at the end of the snapshot. `latestFinishedSnapshotEpoch` is used in case `RemoveMarker` is called before `AddMarker` (possible because of goroutines). So if `Remove` is called first, `Put` will not do anything because `latestFinishedSnapshotEpoch` is set

BeniaminDrasovean: So there is no need to write in `latestFinishedSnapshotEpoch`

state/snapshotsManager.go
BeniaminDrasovean: Indeed, there is no need for this if we run in importDB mode.

storage/storageEpochChange/storageEpochChange.go
```diff
@@ -10,7 +10,7 @@ import (
 	logger "github.com/multiversx/mx-chain-logger-go"
 )
 
-var log = logger.GetOrCreate("state")
+var log = logger.GetOrCreate("storage")
```
iulianpascalau, line 13: `storage/storageEpochChange` to comply with the rest of the project

iulianpascalau (approved):

iulianpascalau (approved):

ssd04 (dismissed):

gabi-vuls (approved):
Normal allin test: v1.5.13-dev-config-8c685c8e00 -> test-last-snapshot-marker--53d20a600c

--- Specific errors ---

block hash does not match 8716
wrong nonce in block 2958
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  385
Nr. of new ERRORS:  0
Nr. of new WARNS:  20
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

--- WARNINGS ---
/------/

AdoAdoAdo (approved):

ssd04 (approved):

iulianpascalau (approved):

gabi-vuls (approved):

PR #5665, opened 2023-10-23
Implemented part 4 of the fix for db lookup extension and partial miniblocks

iulianpascalau:
- integrated miniblock handler in history repository
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


dblookupext/export_test.go
```diff
@@ -0,0 +1,13 @@
+package dblookupext
+
+import (
+	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/data/block"
+)
+
+func (hr *historyRepository) computeMiniblockHash(miniblock *block.MiniBlock) []byte {
```
bogdan-rosianu, line 8: this could stay in a test file, since we don't export anything here

dblookupext/historyRepository.go
```diff
@@ -383,112 +240,51 @@ func (hr *historyRepository) onNotarizedMiniblock(metaBlockNonce uint64, metaBlo
...
-			continue
+	if isNotarizedAtDestination {
+		updateHandler := func(mbMetadataOnBlock *MiniblockMetadataOnBlock) {
+			mbMetadataOnBlock.NotarizedAtDestinationInMetaHash = metaBlockHash
+			mbMetadataOnBlock.NotarizedAtDestinationInMetaNonce = metaBlockNonce
 		}
 
-		patchMetadataFunc(metadata, notificationTyped)
-		err = hr.putMiniblockMetadata(miniblockHash, metadata)
+		err := hr.miniblocksHandler.updateMiniblockMetadataOnBlock(miniblockHash, headerHash, updateHandler)
```
bogdan-rosianu, line 279: I think that this code can be extracted outside the `if` clauses and called only once with the obtained `updateHandler`:
```
err := hr.miniblocksHandler.updateMiniblockMetadataOnBlock(miniblockHash, headerHash, updateHandler)
if err != nil {
    ...
}
```

dblookupext/miniblocksHandler.go
```diff
@@ -366,3 +366,53 @@ func hasTxHashInMiniblockMetadataOnBlock(mbInfo *MiniblockMetadataOnBlock, txHas
...
+	}
+	if !shouldContinue {
+		return nil
+	}
+
+	shouldContinue, err = mh.updateMiniblockMetadataOnBlockInEpoch(epoch+1, miniblockHash, headerHash, updateHandler)
+	if err != nil {
+		return err
+	}
+	if shouldContinue {
```
bogdan-rosianu, line 392: `if !shouldContinue`? 
if not, there is a slight inconsistency here ("if should continue, then return err not found in storage")

dblookupext/export_test.go
iulianpascalau: refactored

dblookupext/historyRepository.go
iulianpascalau: refactored code

integrationTests/node/dblookupext/partialMiniblockSave_test.go
```diff
@@ -129,5 +126,5 @@ func TestSavePartialMiniblocksShouldKeepAllData(t *testing.T) {
 	txHash5 := []byte(fmt.Sprintf(txHashPattern, 5)) // hash 5 should not be linked, it was not processed yet
 	mbData, err = historyRepository.GetMiniblockMetadataByTxHash(txHash5)
 	require.Nil(t, mbData)
-	assert.Equal(t, dblookupext.ErrNotFoundInStorage, err)
+	assert.NotNil(t, err)
```
andreibancioiu, line 129: Is it possible to assert for a more precise error here?

dblookupext/miniblocksHandler.go
iulianpascalau: Actually, this is correct. 
Reason: the `updateMiniblockMetadataOnBlockInEpoch` returns (false, nil) if we found a miniblock metadata that contained the blockhash. (true, nil) signals that on this epoch we did not find the required block hash and also we did not encounter any error. So the algorithm will try to check the next epoch. On that epoch, if we still do not find the block hash (the function still returns (true, nil)) we need to force an error as that part of the miniblock wasn't notarized yet.
The inner test `partial miniblock hash with invalid header hash should not call update` test this exact scenario.

dblookupext/miniblocksHandler_test.go
```diff
@@ -792,3 +792,100 @@ func TestMiniblocksHandler_getMiniblockMetadataByTxHashAndMbHash(t *testing.T) {
...
+	err = mbHandler.commitMiniblock(header1, headerHash1, partialMiniblock)
+	assert.Nil(t, err)
+
+	t.Run("completed hash with invalid header hash should not call update", func(t *testing.T) {
+		err = mbHandler.updateMiniblockMetadataOnBlock(completeMiniblockHash, []byte("missing header hash"), func(mbMetadataOnBlock *MiniblockMetadataOnBlock) {
+			assert.Fail(t, "should have not called update handler")
+		})
+		assert.Equal(t, ErrNotFoundInStorage, err)
+	})
+	t.Run("completed hash with valid header hash should call update", func(t *testing.T) {
```
andreibancioiu, line 824: `completed miniblock hash` instead of `completed hash` (and above).

dblookupext/miniblocksHandler.go
```diff
@@ -366,3 +366,53 @@ func hasTxHashInMiniblockMetadataOnBlock(mbInfo *MiniblockMetadataOnBlock, txHas
...
+) (bool, error) {
+	miniblockMetadata, err := mh.loadExistingMiniblocksMetadata(miniblockHash, epoch)
+	if err != nil {
+		return false, err
+	}
+
+	for _, mbInfo := range miniblockMetadata.MiniblocksInfo {
+		if bytes.Equal(mbInfo.HeaderHash, headerHash) {
+			updateHandler(mbInfo)
+			return false, mh.saveMiniblocksMetadata(miniblockHash, miniblockMetadata, epoch)
```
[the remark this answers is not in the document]
iulianpascalau: already done like this after the feedback received from @bogdan-rosianu

integrationTests/node/dblookupext/partialMiniblockSave_test.go
iulianpascalau: done & simplified the error definition & usage

dblookupext/miniblocksHandler_test.go
iulianpascalau: done

dblookupext/historyRepository.go
```diff
@@ -240,51 +235,61 @@ func (hr *historyRepository) onNotarizedMiniblock(
 		return
 	}
 
+	updateHandler := hr.createUpdateHandler(metaBlockHash, metaBlockNonce, miniblockHeader, shardOfContainingBlock)
```
andreibancioiu, line 238: :+1: 

(alternative name: `createMiniblockMetadataUpdateHandler` - since here we are in the history repository)

andreibancioiu (approved):

andreibancioiu (approved):

PR #5666, opened 2023-10-24
Equivalent messages filter on consensus topic

sstanculeanu:
- implement equivalent messages filter on consensus topic, in order to stop further broadcasts of messages containing same block final info
  
## Proposed changes
- added filter for equivalent messages
- added specific debugger

## Testing procedure
- with feat branch

## Pre-requisites


consensus/spos/interface.go
```diff
@@ -170,3 +172,9 @@ type PeerBlackListCacher interface {
 	Sweep()
 	IsInterfaceNil() bool
 }
+
+// EquivalentMessagesDebugger defines the specific debugger for equivalent messages
+type EquivalentMessagesDebugger interface {
+	DisplayEquivalentMessagesStatistics(getDataHandler func() map[string]uint64)
```
iulianpascalau, line 178: let's make the signature more straightforward. Why not declare the function like
```GO
DisplayEquivalentMessagesStatistics(messages map[string]uint64)
```

consensus/spos/worker.go
```diff
@@ -711,6 +728,48 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+	if !wrk.consensusService.IsMessageWithFinalInfo(msgType) {
+		return nil
+	}
+
+	hdrHash := string(blockHeaderHash)
+
+	wrk.mutEquivalentMessages.Lock()
+	defer wrk.mutEquivalentMessages.Unlock()
+
+	// if an equivalent message was seen before, return error to stop further broadcasts
```
iulianpascalau, line 742: slightly optimization for L743 - L49
```GO
numMessages := wrk.equivalentMessages[hdrHash]
wrk.equivalentMessages[hdrHash] = numMessages+1
if numMessages > 0 {
    return ErrEquivalentMessageAlreadyReceived
}
```
as this one does one read & one write on the map. The original code does 2 reads & one write

consensus/spos/worker.go
```diff
@@ -711,6 +728,48 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+
+	return nil
+}
+
+// getEquivalentMessages returns a copy of the equivalent messages
+func (wrk *Worker) getEquivalentMessages() map[string]uint64 {
+	wrk.mutEquivalentMessages.RLock()
+	defer wrk.mutEquivalentMessages.RUnlock()
+
+	equivalentMessagesCopy := make(map[string]uint64, len(wrk.equivalentMessages))
```
iulianpascalau, line 758: :+1:

consensus/spos/worker_test.go
```diff
@@ -582,6 +584,80 @@ func TestWorker_ProcessReceivedMessageRedundancyNodeShouldResetInactivityIfNeede
...
+	err = wrk.ProcessReceivedMessage(
+		&p2pmocks.P2PMessageMock{
+			DataField:      buffEquiv,
+			PeerField:      currentPid,
+			SignatureField: []byte("signatureEquiv"),
+		},
+		equivMsgFrom,
+		&p2pmocks.MessengerStub{},
+	)
+	assert.Equal(t, spos.ErrEquivalentMessageAlreadyReceived, err)
```
iulianpascalau, line 658: test also the inner counter? Or a special unit test for the reset & read of the counters?

consensus/spos/worker.go
sstanculeanu: applied

consensus/spos/worker_test.go
sstanculeanu: improved the test

consensus/spos/worker.go
```diff
@@ -711,6 +728,47 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+		return nil
+	}
+
+	hdrHash := string(blockHeaderHash)
+
+	wrk.mutEquivalentMessages.Lock()
+	defer wrk.mutEquivalentMessages.Unlock()
+
+	// if an equivalent message was seen before, return error to stop further broadcasts
+	numMessages := wrk.equivalentMessages[hdrHash]
```
AdoAdoAdo, line 743: Is the validation of the equivalent message in a different PR?
Here if for example the first message received is invalid, you still count it.

sstanculeanu: right, handled invalid message received in case of validation error

consensus/spos/worker.go
```diff
@@ -749,6 +750,19 @@ func (wrk *Worker) processEquivalentMessage(msgType consensus.MessageType, block
...
+	if !wrk.consensusService.IsMessageWithFinalInfo(msgType) {
+		return
+	}
+
+	hdrHash := string(blockHeaderHash)
+
+	wrk.mutEquivalentMessages.Lock()
+	defer wrk.mutEquivalentMessages.Unlock()
+
+	delete(wrk.equivalentMessages, hdrHash)
```
iulianpascalau, line 763: I do not think this is the proper way to do this.
Reason: what happens in this scenario?
1. receive the correct message
2. receive the same block final info but for some reason `checkConsensusMessageValidity` returns an error. We delete the equivalent message for the hdrHash
3. we receive the correct message again and we re-broadcast it

sstanculeanu: pushed

consensus/spos/worker.go
```diff
@@ -398,8 +398,10 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedP
 
 	msgType := consensus.MessageType(cnsMsg.MsgType)
 
-	err = wrk.processEquivalentMessage(msgType, cnsMsg.BlockHeaderHash)
+	wrk.mutEquivalentMessages.Lock()
```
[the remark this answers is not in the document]
sstanculeanu: updated as suggested

iulianpascalau (dismissed):

consensus/spos/worker.go
```diff
@@ -386,23 +396,14 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedP
...
-	log.Trace("received message from consensus topic",
-		"msg type", wrk.consensusService.GetStringValue(msgType),
-		"from", cnsMsg.PubKey,
-		"header hash", cnsMsg.BlockHeaderHash,
-		"round", cnsMsg.RoundIndex,
-		"size", len(message.Data()),
-	)
-
-	err = wrk.consensusMessageValidator.checkConsensusMessageValidity(cnsMsg, message.Peer())
+	err = wrk.checkValidityAndProcessEquivalentMessages(cnsMsg, message)
```
AdoAdoAdo, line 399: I think we need to add an enable flag for equivalent messages activation, as the process will be different.
We also need to keep the previous behavior for before activation.

AdoAdoAdo (changes requested):

consensus/spos/worker.go
sstanculeanu: pushed

iulianpascalau (dismissed):

consensus/spos/worker.go
```diff
@@ -711,6 +720,89 @@ func (wrk *Worker) ResetConsensusMessages() {
 	wrk.consensusMessageValidator.resetConsensusMessages()
 }
 
+func (wrk *Worker) checkValidityAndProcessEquivalentMessages(cnsMsg *consensus.Message, p2pMessage p2p.MessageP2P) error {
+	wrk.mutEquivalentMessages.Lock()
```
AdoAdoAdo, line 724: can we separate the validation of the final info which becomes equivalent message validation into a separate function?

also would be good if we only add the mutex on the equivalent message validation (after we check the enable flag so we don't affect the previous behavior)

for the equivalent message validation we need to do also aggregated signature verification before we start rejecting other equivalent messages.

AdoAdoAdo (changes requested):

consensus/spos/worker.go
sstanculeanu: pushed as discussed:
- now the mutex should only be locked for final info messages, when the flag is active
- added signature verification

consensus/spos/worker.go
```diff
@@ -711,6 +720,112 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+}
+
+func (wrk *Worker) verifyEquivalentMessageSignature(cnsMsg *consensus.Message) error {
+	if check.IfNil(wrk.consensusState.Header) {
+		return ErrNilHeader
+	}
+
+	header := wrk.consensusState.Header.ShallowClone()
+
+	err := header.SetSignature(cnsMsg.Signature)
```
AdoAdoAdo, line 785: did we agree to only temporarily set this on the header and remove it afterwards?
otherwise the header sig verifier should accept separately the header and the proof (bitmap + aggregated signature) and then verify

consensus/spos/bls/subroundStartRound.go
```diff
@@ -56,12 +58,16 @@ func NewSubroundStartRound(
 	if check.IfNil(sentSignatureTracker) {
 		return nil, spos.ErrNilSentSignatureTracker
 	}
+	if removeAllEquivalentMessages == nil {
```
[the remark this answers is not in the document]
sstanculeanu: indeed, removed the new function

consensus/spos/worker.go
```diff
@@ -386,23 +402,14 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedP
...
-	log.Trace("received message from consensus topic",
-		"msg type", wrk.consensusService.GetStringValue(msgType),
-		"from", cnsMsg.PubKey,
-		"header hash", cnsMsg.BlockHeaderHash,
-		"round", cnsMsg.RoundIndex,
-		"size", len(message.Data()),
-	)
-
-	err = wrk.consensusMessageValidator.checkConsensusMessageValidity(cnsMsg, message.Peer())
+	err = wrk.checkValidityAndProcessEquivalentMessages(cnsMsg, message)
```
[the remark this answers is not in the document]
sstanculeanu: added extra check on shouldBlacklistPeer method

consensus/spos/worker.go
```diff
@@ -711,6 +720,112 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+
+	log.Trace("received message from consensus topic",
+		"msg type", wrk.consensusService.GetStringValue(msgType),
+		"from", cnsMsg.PubKey,
+		"header hash", cnsMsg.BlockHeaderHash,
+		"round", cnsMsg.RoundIndex,
+		"size", len(p2pMessage.Data()),
+	)
+
+	if !wrk.enableEpochsHandler.IsFlagEnabled(common.EquivalentMessagesFlag) {
```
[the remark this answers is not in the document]
sstanculeanu: applied

consensus/spos/worker.go
```diff
@@ -711,6 +720,112 @@ func (wrk *Worker) ResetConsensusMessages() {
...
+	if err != nil {
+		return err
+	}
+
+	hdrHash := string(cnsMsg.BlockHeaderHash)
+
+	// if a valid equivalent message was seen before, return error to stop further broadcasts
+	numMessages := wrk.equivalentMessages[hdrHash]
+	wrk.equivalentMessages[hdrHash] = numMessages + 1
+	if numMessages > 0 {
```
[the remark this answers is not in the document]
sstanculeanu: updated

consensus/spos/worker.go
sstanculeanu: removed and added todo for next PR, where the header structure is updated

consensus/spos/worker.go
```diff
@@ -757,48 +757,46 @@ func (wrk *Worker) checkValidityAndProcessEquivalentMessages(cnsMsg *consensus.M
...
-	}
-
-	err = wrk.headerSigVerifier.VerifySignature(header)
-	if err != nil {
-		log.Debug("verifyEquivalentMessageSignature", "error", err.Error())
-		return err
-	}
-
-	return nil
+	// TODO[Sorin]: after flag enabled, VerifySignature on previous hash, with the signature and bitmap from the proof on cnsMsg
```
AdoAdoAdo, line 800: verify on current header with the signature and bitmap from cnsMsg, but anyway would be removed in next PR.

AdoAdoAdo (approved):

iulianpascalau (approved):

PR #5705, opened 2023-11-13
Trie sync progress status

ssd04:
- Add trie sync progress status percentage
  
## Proposed changes
- Added a new check into gateway in order to fetch estimated number of trie nodes from last snapshot
- export gateway url from config on bootstrap metrics
- Added additional config section in `config.toml` to specify gateway url

## Testing procedure
- Standard system test
- Check in termui that trie sync progress percentage works, on testnet, devnet or mainnet (it can be tested also on internal testnet if that's feasable)
     - set gateway URL from newly added section in `config.toml`
          - start termui while node is syncing and check trie sync area
     - if gateway URL is not set, it should not crash, it should show N/A on trie sync progress section

## Pre-requisites


cmd/termui/provider/metricsProvider.go
```diff
@@ -27,17 +33,37 @@ type responseFromApi struct {
 	Code  string                    `json:"code"`
 }
 
+type trirStatisticsResponseData struct {
```
BeniaminDrasovean, line 36: typo (trir)

cmd/termui/presenter/chainInfoGetters.go
```diff
@@ -185,6 +186,23 @@ func (psh *PresenterStatusHandler) GetTrieSyncNumProcessedNodes() uint64 {
 	return psh.getFromCacheAsUint64(common.MetricTrieSyncNumProcessedNodes)
 }
 
+// GetTrieSyncNumProcessedNodes will return the number of processed nodes during trie sync
```
BeniaminDrasovean, line 189: GetTrieSyncProcessedPercentage

cmd/termui/provider/metricsProvider.go
ssd04: done

cmd/termui/presenter/chainInfoGetters.go
ssd04: done

cmd/termui/main.go
```diff
@@ -46,6 +47,15 @@ VERSION:
 		Value:       "127.0.0.1:8080",
 		Destination: &argsConfig.address,
 	}
+
+	// gatewayAddress defines a flag for setting the address and port for gateway interactions
+	gatewayAddress = cli.StringFlag{
+		Name:        "gateway-address",
+		Usage:       "Address and port number on which the application will try to connect to gateway",
+		Value:       "https://testnet-gateway.multiversx.com",
```
iulianpascalau, line 55: no one will pass this parameter and the termui will end up showing misleading information. I guess we need to find another approach here.

What I can figure out now, is to have a new component *in the node* that will execute the code written in the `loadMetricsFromGatewayApi` function periodically. And the gateway API string will be provided in the config.toml file so we can set it at config time.

cmd/termui/view/termuic/termuiRenders/widgetsRender.go
```diff
@@ -194,7 +194,16 @@ func (wr *WidgetsRender) prepareChainInfo(numMillisecondsRefreshTime int) {
 	case isNodeSyncingTrie:
 		syncingStr = statusSyncing
 		bytesReceived := wr.presenter.GetTrieSyncNumBytesReceived()
-		statusMessage = fmt.Sprintf("Trie sync: %d nodes, %s state size", nodesProcessed, core.ConvertBytes(bytesReceived))
+		syncPercentage := wr.presenter.GetTrieSyncProcessedPercentage()
+
+		syncPercentageOut := ""
+		if syncPercentage == 0 {
+			syncPercentageOut = statusNotApplicable
```
iulianpascalau, line 201: The status not applicable will be shown shortly before the node reaches 1%. Please refactor this check.

iulianpascalau (changes requested):

cmd/termui/main.go
ssd04: remove this cli parameter and added a separate metric on node bootstrap metrics

cmd/termui/view/termuic/termuiRenders/widgetsRender.go
ssd04: refactored here

cmd/termui/presenter/chainInfoGetters.go
```diff
@@ -187,20 +187,20 @@ func (psh *PresenterStatusHandler) GetTrieSyncNumProcessedNodes() uint64 {
 }
 
 // GetTrieSyncProcessedPercentage will return the number of processed nodes during trie sync
-func (psh *PresenterStatusHandler) GetTrieSyncProcessedPercentage() uint64 {
+func (psh *PresenterStatusHandler) GetTrieSyncProcessedPercentage() (uint64, bool) {
```
iulianpascalau, line 190: ... or could have used an optional uint64 struct that already has the IsSetValue field

cmd/node/config/config.toml
```diff
@@ -805,6 +805,9 @@
     StableTagLocation = "https://api.github.com/repos/multiversx/mx-chain-go/releases/latest"
     PollingIntervalInMinutes = 65
 
+[GatewayMetricsConfig]
+    URL = "https://testnet-gateway.multiversx.com"
```
iulianpascalau, line 809: I would leave this blank and add a TODO to change this on testnet/devnet/mainnet
In this way it won't show misleading info whenever we start a local testnet

cmd/termui/provider/metricsProvider.go
```diff
@@ -94,24 +89,24 @@ func (smp *StatusMetricsProvider) updateMetrics() {
 	smp.fetchAndApplyMetrics(statusMetricsUrlSuffix)
 	smp.fetchAndApplyBootstrapMetrics(bootstrapStatusMetricsUrlSuffix)
 
-	if smp.shardID != "" {
+	if smp.shardID != "" && smp.gatewayAddress != "" {
 		metricsURLSuffix := trieStatisticsMetricsUrlSuffix + smp.shardID
-		smp.fetchAndApplyGatewayStatusMetrics(metricsURLSuffix)
+		statusMetricsURL := smp.gatewayAddress + metricsURLSuffix
+		smp.fetchAndApplyGatewayStatusMetrics(statusMetricsURL)
```
iulianpascalau, line 95: careful, this will be called each ~500 milliseconds. The `fetchAndApplyGatewayStatusMetrics` calls the `loadMetricsFromGatewayApi` function which does an API call, which is not free :)
Suggestion: I would call `loadMetricsFromGatewayApi` only if the gateway address changed (we might encounter this situation when the termui window was opened before the node, or there were some connection issues that were resolved). The result of this load would then be stored inside the presenter and used to actually output the trie sync string.

cmd/termui/presenter/chainInfoGetters.go
ssd04: changed to use OptionalUint64

cmd/termui/provider/metricsProvider.go
ssd04: added a separate bool check for this

cmd/termui/provider/metricsProvider.go
```diff
@@ -92,22 +93,32 @@ func (smp *StatusMetricsProvider) updateMetrics() {
...
 }
 
 func (smp *StatusMetricsProvider) fetchAndApplyGatewayStatusMetrics(statusMetricsURL string) {
+	noErrors := true
 	numTrieNodes, err := smp.loadMetricsFromGatewayApi(statusMetricsURL)
 	if err != nil {
 		log.Info("fetch from Gateway API",
 			"path", statusMetricsURL,
 			"error", err.Error())
+		noErrors = false
```
iulianpascalau, line 110: noErrors = false :D 
Could have been foundErrors = true to avoid double negation

iulianpascalau (approved):

cmd/termui/provider/metricsProvider.go
ssd04: changed

iulianpascalau (dismissed):

BeniaminDrasovean (dismissed):

iulianpascalau (approved):

miiu96 (approved):

gabi-vuls (approved):

PR #5741, opened 2023-11-27
Feat/RelayedV3

sstanculeanu:
- Relayed tx v3
- Fix base cost of user tx for all relayed transactions
- Only consume move balance cost of user tx in case of relayed move balance
 
## Proposed changes
- Implemented relayed tx v3. Added a new field on the tx, InnerTransactions that can hold a configurable max number of relayed user transactions (currently set to 50 max).
- Fixed base cost of relayed tx for all types. Ex: let's consider a relayed v1 tx, with a move balance as user tx, having len of data 200. Considering we would have passed the fixed gas limit, it should be: 50.000 for user move balance + 50.000 for relayed tx + 300.000 for len of data. So total of 400.000. But, due to the way it was computed, it would have cost less than that (350.000 for relayed tx + 500 considered the processing cost). After this fix, this tx should consume exactly 400.000. Also applies for sc calls, where the base cost was omitted.
- Now when a move balance is detected as a user tx, only the cost of move balance is removed from relayer, no matter how much extra gas is provided.

## Testing procedure
- standard system test
- test relayed v1 and relayed v2 fees for all kind of transactions(failure/refund scenarios to be verified as well). Make sure the fees consumed by the same transaction are different before and after the fix. After fix, all of them should consume more gas as described above.
- test relayed v3 transaction execution + fees consumed + events generated for all user txs, in respect with the [confluence specs](https://multiversxlabs.atlassian.net/wiki/spaces/EN/pages/2207285249/WIP-ERDIP-113+Relayed+Transaction+V3)

## Pre-requisites


AdoAdoAdo (dismissed):

miiu96 (approved):

elvisboghiu (approved):
System Test is finished.
The nodes were closed when they gave each of the errors. 

`Normal allin test: v1.7.13-dev-config-aebb55e8e0 -> feat\relayedv3

--- Specific errors --- 

block hash does not match 2095
wrong nonce in block 975
miniblocks does not match 0
block bodies does not match 1
receipts hash missmatch 1

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  4
Nr. of all WARNS:  1038
Nr. of new ERRORS:  3
Nr. of new WARNS:  6
Nr. of PANICS:  0

/------/ 

--- ERRORS ---  

ovh-p04-observer-3 : 
Error: error calling SaveAccounts, will retry  driver = *host.hostDriver retrial in = error = connection not open while sending data on route for topic SaveAccounts  3

/------/ 

--- WARNINGS ---  

ovh-p04-observer-3 : 
Warn: wt.Listen()-> connection problem     error = connection not open  5
Warn: outport.monitorCompletionOnDriver took too long function = saveAccountsBlocking driver = *host.hostDriver message counter = time =  1`

AdoAdoAdo (approved):

miiu96 (approved):

PR #5760, opened 2023-12-08
SubroundSignature changes

sstanculeanu:
- continue consensus finality improvements on subroundSignature
  
## Proposed changes
- all validators wait for signatures

## Testing procedure
- with feat branch

## Pre-requisites


consensus/spos/bls/subroundSignature_test.go
```diff
@@ -663,49 +724,106 @@ func TestSubroundSignature_DoSignatureConsensusCheckShouldReturnFalseWhenSignatu
...
 
-	container := mock.InitConsensusCore()
-	sr := *initSubroundSignatureWithContainer(container)
-	sr.WaitingAllSignaturesTimeOut = false
+	t.Run("with flag active", testSubroundSignatureDoSignatureConsensusCheckShouldReturnFalseWhenNotAllSignaturesCollectedAndTimeIsNotOut(true))
+	t.Run("with flag inactive", testSubroundSignatureDoSignatureConsensusCheckShouldReturnFalseWhenNotAllSignaturesCollectedAndTimeIsNotOut(false))
+}
 
-	sr.SetSelfPubKey(sr.ConsensusGroup()[0])
+func testSubroundSignatureDoSignatureConsensusCheckShouldReturnFalseWhenNotAllSignaturesCollectedAndTimeIsNotOut(flagActive bool) func(t *testing.T) {
```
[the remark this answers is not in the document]
sstanculeanu: updated to use only one test func

consensus/spos/bls/subroundSignature.go
```diff
@@ -100,6 +100,7 @@ func (sr *subroundSignature) doSignatureJob(_ context.Context) bool {
 			return false
 		}
 
+		// leader already sent his signature on subround block
```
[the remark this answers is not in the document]
sstanculeanu: I've only added the comment, the isSelfLeader check did not change

consensus/spos/bls/subroundSignature.go
```diff
@@ -378,6 +389,12 @@ func (sr *subroundSignature) doSignatureJobForManagedKeys() bool {
 			return false
 		}
 
+		isKeyLeader := idx == spos.IndexOfLeaderInConsensusGroup
+		// TODO[cleanup cns finality]: update the check
+		// with the equivalent messages feature on, signatures from all managed keys must be broadcast, as the aggregation is done by any participant
+		if isFlagActive {
+			isMultiKeyLeader = isKeyLeader
+		}
 		if !isMultiKeyLeader {
```
ssd04, line 398: i think here it would help to have a separate variable like `shouldBroadcastSignature`, since if `isFlagActive` = true `isMultiKeyLeader` variable name is confusing

ssd04: if we can refactor a bit here, it's a bit difficult to track `isFlagActive` `isMultiKeyLeader` `isKeyLeader` conditions in this func, at least at first

consensus/spos/bls/subroundSignature.go
```diff
@@ -386,10 +403,11 @@ func (sr *subroundSignature) doSignatureJobForManagedKeys() bool {
 
 			numMultiKeysSignaturesSent++
 		}
+		// with the equivalent messages feature on, the leader signature is sent on subroundBlock, thus we should update its status here as well
 		sr.sentSignatureTracker.SignatureSent(pkBytes)
 
-		isLeader := idx == spos.IndexOfLeaderInConsensusGroup
-		ok := sr.completeSignatureSubRound(pk, isLeader)
+		shouldWaitForAllSigsAsync := isKeyLeader && !isFlagActive
+		ok := sr.completeSignatureSubRound(pk, shouldWaitForAllSigsAsync)
```
ssd04, line 410: if `isFlagActive` if false shouldn't we wait for all sigs also if `isMultiKeyLeader` is true?

consensus/spos/bls/subroundSignature.go
```diff
@@ -79,45 +78,49 @@ func (sr *subroundSignature) doSignatureJob(_ context.Context) bool {
...
 
-	isSelfLeader := sr.IsSelfLeaderInCurrentRound() && sr.ShouldConsiderSelfKeyInConsensus()
-	isSelfInConsensusGroup := sr.IsNodeInConsensusGroup(sr.SelfPubKey()) && sr.ShouldConsiderSelfKeyInConsensus()
+	isSelfSingleKeyLeader := sr.IsSelfLeaderInCurrentRound() && sr.ShouldConsiderSelfKeyInConsensus()
+	isSelfSingleKeyInConsensusGroup := sr.IsNodeInConsensusGroup(sr.SelfPubKey()) && sr.ShouldConsiderSelfKeyInConsensus()
+	isFlagActive := sr.EnableEpochsHandler().IsFlagEnabledInEpoch(common.EquivalentMessagesFlag, sr.Header.GetEpoch())
+	// if single key leader, the signature has been sent on subroundBlock, thus the current round can be marked as finished
+	if isSelfSingleKeyLeader && isFlagActive {
+		log.Debug("step 2: subround has been finished for leader",
+			"subround", sr.Name())
```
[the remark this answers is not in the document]
sstanculeanu: right, moved at the end

consensus/spos/bls/subroundSignature.go
sstanculeanu: renamed a bit for clarity

consensus/spos/bls/subroundSignature.go
sstanculeanu: when the flag is not active(as before), we only want to start one waiting go routine, thus we start it only for the specific managed key that is waiting leader

after flag activation, we don't want to wait here.. this will happen on subroundEndRound

consensus/spos/bls/subroundSignature.go
```diff
@@ -258,13 +258,21 @@ func (sr *subroundSignature) doSignatureConsensusCheck() bool {
 		return false
 	}
 
+	isSelfInConsensusGroup := sr.IsNodeInConsensusGroup(sr.SelfPubKey()) || sr.IsMultiKeyInConsensusGroup()
+	if !isSelfInConsensusGroup {
+		log.Debug("step 2: subround has been finished",
```
ssd04, line 263: if we have this early check here, i think that the reference check to `!isSelfInConsensusGroup` on L302 is not really needed anymore

consensus/spos/bls/subroundSignature.go
```diff
@@ -378,18 +399,24 @@ func (sr *subroundSignature) doSignatureJobForManagedKeys() bool {
 			return false
 		}
 
-		if !isMultiKeyLeader {
+		isCurrentManagedKeyLeader := idx == spos.IndexOfLeaderInConsensusGroup
+		// TODO[cleanup cns finality]: update the check
+		// with the equivalent messages feature on, signatures from all managed keys must be broadcast, as the aggregation is done by any participant
+		shouldBroadcastSignatureShare := (!isCurrentNodeMultiKeyLeader && !isFlagActive) ||
```
AdoAdoAdo, line 405: what is the difference between isCurrentNodeMultiKeyLeader and isCurrentManagedKeyLeader ?

to me it seems that they should represent the same thing

and since there is an && with !isFlagActive and isFlagActive, that could be simplified/removed if isCurrentNodeMultiKeyLeader and isCurrentManagedKeyLeader indeed represent the same thing, and only go forward if current index is not leader.

consensus/spos/bls/subroundSignature_test.go
```diff
@@ -650,62 +862,74 @@ func TestSubroundSignature_DoSignatureConsensusCheckShouldReturnTrueWhenSignatur
...
-	sr.WaitingAllSignaturesTimeOut = false
-
-	sr.SetSelfPubKey(sr.ConsensusGroup()[0])
-
-	for i := 0; i < sr.Threshold(bls.SrSignature); i++ {
-		_ = sr.SetJobDone(sr.ConsensusGroup()[i], bls.SrSignature, true)
-	}
-
-	assert.False(t, sr.DoSignatureConsensusCheck())
+	t.Run("with flag active, should return false - will be done on subroundEndRound", testSubroundSignatureDoSignatureConsensusCheck(true, setThresholdJobsDone, false, false))
```
AdoAdoAdo, line 880: can you have an arg struct for the argument, then it will be clearer which fields are false and which are true.

consensus/spos/bls/subroundSignature.go
sstanculeanu: `isCurrentNodeMultiKeyLeader` = the leader is managed by this node
`isCurrentManagedKeyLeader` = that specific index is the leader, which is managed by the node

this is already removed and simplified a lot in next PRs, matching your suggestion :pray:

consensus/spos/bls/subroundSignature_test.go
sstanculeanu: pushed, thanks

AdoAdoAdo (approved):

ssd04 (approved):

PR #5827, opened 2024-01-11
scripting files updates for sovereign

axenteoctavian:
- MX-14982 task
- 
- 
  
## Proposed changes
- merge xDay sovereign script files updates into main files
- 
- 

## Testing procedure
- running sovereign node
- 
- 

## Pre-requisites


scripts/testnet/deploySovereignScreen.sh
```diff
@@ -0,0 +1,36 @@
...
+echo -e "${GREEN}Starting sending the basic scenario transactions...${NC}"
+screen -L -Logfile txgenBasicLog.txt -d -m -S txgenBasicScreen ./sovereign-txgen-basic.sh
+
+echo -e "${GREEN}Starting sending the erc20 scenario transactions...${NC}"
+screen -L -Logfile txgenErc20Log.txt -d -m -S txgenErc20Screen ./sovereign-txgen-erc20.sh
+
+echo -e "${GREEN}Starting sending the esdt scenario transactions...${NC}"
+screen -L -Logfile txgenEsdtLog.txt -d -m -S txgenEsdtScreen ./sovereign-txgen-esdt.sh
+
+echo -e "${GREEN}Finished the sovereign chain deployment. Don't forget to stop it with ./stop.sh at the end.${NC}"
```
bogdan-rosianu, line 36: add empty line at the end of the file

bogdan-rosianu (approved):

sasurobert (dismissed):

scripts/testnet/variables.sh
```diff
@@ -60,14 +60,16 @@ export GENESIS_STAKE_TYPE="direct" #'delegated' or 'direct' as in direct stake
...
 export OBSERVERS_ANTIFLOOD_DISABLE=0
 
-# If set to true, this will deploy nodes in a sovereign shard.
+# If set to 1, this will deploy nodes in a sovereign shard.
 # All variables from metashard structure(validators, observers, consensus) should be set to zero and SHARDCOUNT to 1
 # For now, make sure that you checkout feat/sovereign branch from mx-chain-deploy repo when using these scripts
-export SOVEREIGN_DEPLOY=true
+export SOVEREIGN_DEPLOY=1
+
+export ROUND_DURATION=4000
```
mariusmihaic, line 68: I don't think this is ok.
We are overwriting the RoundDuration from `nodesSetup.json` with this.
Is this really needed/expected?
cc @bogdan-rosianu @raduchis

scripts/testnet/include/config.sh
```diff
@@ -221,6 +246,13 @@ updateTxGenConfig() {
   popd
 }
 
+updateSovereignTxGenConfig() {
+  updateTxGenConfig
+  pushd $TESTNETDIR/txgen/config/nodeConfig/config
+  updateTOMLValue enableEpochs.toml "SCDeployEnableEpoch" 0
+  updateTOMLValue enableEpochs.toml "BuiltInFunctionsEnableEpoch" 0
+  updateTOMLValue enableEpochs.toml "ESDTEnableEpoch" 0
```
mariusmihaic, line 254: I think this func could be removed.
We can set these flags to 0 from `enableEpochs.toml`

scripts/testnet/include/config.sh
```diff
@@ -8,6 +8,11 @@ generateConfig() {
     TMP_META_OBSERVERCOUNT=0
   fi
 
+  SOVEREIGN_BOOL="false"
+  if [ $USE_HARDFORK -eq 1 ]; then
+    SOVEREIGN_BOOL="true"
```
mariusmihaic, line 13: Why set this to true if `USE_HARDFORK`?
This is wrong.
We need to use  `SOVEREIGN_DEPLOY`

scripts/testnet/include/config.sh
```diff
@@ -146,6 +152,8 @@ updateNodeConfig() {
 	  sed -i '/^\[ElasticSearchConnector\]/,/^\[/ s/Enabled *= *false/Enabled = true/' external_observer.toml
 	fi
 
+  sed -i '/^\[DbLookupExtensions\]/,/^\[/ s/Enabled *= *false/Enabled = true/' config_observer.toml
```
mariusmihaic, line 155: I don't think this should be by default set to true.
I think we should use an exported variable for this...

Opinions?
@bogdan-rosianu @raduchis

scripts/testnet/deploySovereignScreen.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
mariusmihaic, line 1: I think we can rename this file to smth like "deployFullStackSovereignChainWithScreen.sh"

scripts/testnet/include/config.sh
bogdan-rosianu: right

scripts/testnet/include/config.sh
bogdan-rosianu: in my opinion, it's ok to be default to true. we can have an exported var for this, but I would like to make it default to true

scripts/testnet/variables.sh
bogdan-rosianu: yes, we can revert. we needed a shorter round on the demo for a more responsive feedback

raduchis: yes, can be reverted. As @bogdan-rosianu mentioned it was updated for the xDay demo

scripts/testnet/include/config.sh
raduchis: if this will be connected to a proxy and an api, it should be default true.

scripts/testnet/include/config.sh
raduchis: yes, but it should be set to 0 there... some of the flags could not be set to 0 and this is why an update at start time was needed

scripts/testnet/include/config.sh
```diff
@@ -8,9 +8,9 @@ generateConfig() {
     TMP_META_OBSERVERCOUNT=0
   fi
 
-  SOVEREIGN_BOOL="false"
+  SOVEREIGN_DEPLOY=0
   if [ $USE_HARDFORK -eq 1 ]; then
-    SOVEREIGN_BOOL="true"
+    SOVEREIGN_DEPLOY=1
```
mariusmihaic, line 13: Wrong

```
  SOVEREIGN_BOOL="false"
  if [ $SOVEREIGN_DEPLOY -eq 1 ]; then
    SOVEREIGN_BOOL="true"
```

SOVEREIGN_DEPLOY -> search name of this variable exists  and is 1/0 to make sure

scripts/testnet/include/config.sh
```diff
@@ -26,7 +26,7 @@ generateConfig() {
     -stake-type $GENESIS_STAKE_TYPE \
     -hysteresis $HYSTERESIS \
     -round-duration $ROUND_DURATION \
-    -sovereign=$SOVEREIGN_BOOL
+    -sovereign=SOVEREIGN_DEPLOY
```
mariusmihaic, line 29: Wrong

```suggestion
    -sovereign=SOVEREIGN_BOOL
```

scripts/testnet/variables.sh
```diff
@@ -65,8 +65,6 @@ export OBSERVERS_ANTIFLOOD_DISABLE=0
 # For now, make sure that you checkout feat/sovereign branch from mx-chain-deploy repo when using these scripts
 export SOVEREIGN_DEPLOY=1
 
-export ROUND_DURATION=4000
```
mariusmihaic: If you removed this, you should remove its usage from entire project. Remove if as well from `scripts/testnet/include/config.sh`

mariusmihaic (changes requested):

scripts/testnet/variables.sh
```diff
@@ -60,14 +60,14 @@ export GENESIS_STAKE_TYPE="direct" #'delegated' or 'direct' as in direct stake
...
+# If set to 1, this will deploy nodes in a sovereign shard.
 # All variables from metashard structure(validators, observers, consensus) should be set to zero and SHARDCOUNT to 1
 # For now, make sure that you checkout feat/sovereign branch from mx-chain-deploy repo when using these scripts
-export SOVEREIGN_DEPLOY=true
+export SOVEREIGN_DEPLOY=1
 
 # Shard structure
 export SHARDCOUNT=1
-export SHARD_VALIDATORCOUNT=2
+export SHARD_VALIDATORCOUNT=3
```
mariusmihaic, line 70: We should not change this. the script to deploy sovereign full stack works with a min of 3 here.
Inside `scripts/testnet/deployFullStackSovereignChainWithScreen.sh`, when updating variables we need to check if `SHARD_VALIDATORCOUNT` is less than 3, and if so, set to it to 3, otherwise leave it unchanged

bogdan-rosianu (approved):

PR #5855, opened 2024-01-19
Fixes after tests

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/headerCheck/headerSignatureVerify.go
```diff
@@ -171,7 +178,16 @@ func (hsv *HeaderSigVerifier) VerifySignature(header data.HeaderHandler) error {
 		return err
 	}
 
-	return hsv.VerifySignatureForHash(header, hash, header.GetPubKeysBitmap(), header.GetSignature())
+	bitmap := header.GetPubKeysBitmap()
+	sig := header.GetSignature()
+	if hsv.enableEpochsHandler.IsFlagEnabledInEpoch(common.EquivalentMessagesFlag, headerCopy.GetEpoch()) {
```
ssd04, line 183: i think that header signature verifier component should not know if we want to handle current or previous header info; can we use only `VerifySingatureForHash` and decide from the caller which header info to use?

ssd04: how about other components where `VerifySignature` is being used? does this also fit with the check from process interceptors?

process/headerCheck/headerSignatureVerify.go
```diff
@@ -171,7 +178,16 @@ func (hsv *HeaderSigVerifier) VerifySignature(header data.HeaderHandler) error {
...
+		headerCopy, hash, sig, bitmap, err = hsv.getPrevHeaderInfo(headerCopy)
```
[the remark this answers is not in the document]
AdoAdoAdo: I will start working on the change regarding option b) so we can simplify here and for cross chain proofs as well.

process/headerCheck/headerSignatureVerify.go
sstanculeanu: yes, it should be compatible in all places. We can keep it as it is for compatibility. After this release, a refactor will come in the next one probably and completely remove EquivalentMessagesFlag and only keep the new version.

consensus/spos/bls/subroundSignature.go
```diff
@@ -299,7 +299,7 @@ func (sr *subroundSignature) doSignatureConsensusCheck() bool {
 	}
 	isJobDoneByConsensusNode := !isSelfLeader && isSelfInConsensusGroup && selfJobDone && multiKeyJobDone
 
-	isSubroundFinished := !isSelfInConsensusGroup || isJobDoneByConsensusNode || isJobDoneByLeader
+	isSubroundFinished := isJobDoneByConsensusNode || isJobDoneByLeader
```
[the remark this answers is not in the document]
sstanculeanu: this is already done on L261 as an early exit. It was part of the subroundSignature changes PR: https://github.com/multiversx/mx-chain-go/pull/5760/files#diff-4f4adc2b25d7edb3422708759a12cbcc7be62a34a1c0970a8339a4a531ddf385R261

AdoAdoAdo (approved):

ssd04 (approved):

PR #5858, opened 2024-01-22
do not activate more nodes on stake if too many nodes

sasurobert:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


vm/systemSmartContracts/delegation.go
```diff
@@ -1226,6 +1232,19 @@ func (d *delegation) stakeNodes(args *vmcommon.ContractCallInput) vmcommon.Retur
 	return vmcommon.Ok
 }
 
+func tooManyNodesLogs(logEntries []*vmcommon.LogEntry) bool {
+	for _, logEntry := range logEntries {
+		if len(logEntry.Topics) > 1 {
+			continue
+		}
+		if !bytes.Equal(logEntry.Topics[0], []byte(numberOfNodesTooHigh)) {
```
mariusmihaic, line 1240: Can't this panic? : `logEntry.Topics[0]`
You check above `if len(logEntry.Topics) > 1`, so this can come down here with `len == 0`

Also, just a question for my own clarification: Is it ok that you check in logs this when calling `stakeNodes` ? Is it 100% sure that this logs will be put here when calling `stakeNodes`?

vm/systemSmartContracts/validator_test.go
```diff
@@ -472,7 +469,7 @@ func TestStakingValidatorSC_ExecuteStakeTooManyNodes(t *testing.T) {
 	arguments.Arguments = [][]byte{big.NewInt(3).Bytes(), key1, []byte("msg1"), key2, []byte("msg2"), key3, []byte("msg3")}
 
 	errCode := stakingValidatorSc.Execute(arguments)
-	assert.Equal(t, vmcommon.UserError, errCode)
+	assert.Equal(t, vmcommon.Ok, errCode)
```
mariusmihaic, line 472: I'd suggest we also check that the log entry is present or test `stakeNodes` returns error when the log is present.

vm/systemSmartContracts/delegation.go
sasurobert: changed to != 1. 

Yes, it is 100% sure, as stakeNodes calls "stake" endpoint from ValidatorSC, which creates this log in those cases.

mariusmihaic (approved):

vm/systemSmartContracts/delegation.go
```diff
@@ -1226,6 +1232,19 @@ func (d *delegation) stakeNodes(args *vmcommon.ContractCallInput) vmcommon.Retur
 	return vmcommon.Ok
 }
 
+func tooManyNodesLogs(logEntries []*vmcommon.LogEntry) bool {
+	for _, logEntry := range logEntries {
+		if len(logEntry.Topics) != 1 {
+			continue
+		}
+		if !bytes.Equal(logEntry.Topics[0], []byte(numberOfNodesTooHigh)) {
```
AdoAdoAdo, line 1240: shouldn't this be `if bytes.Equal(logEntry.Topics[0], []byte(numberOfNodesTooHigh))`

vm/systemSmartContracts/delegation.go
```diff
@@ -1215,6 +1215,12 @@ func (d *delegation) stakeNodes(args *vmcommon.ContractCallInput) vmcommon.Retur
 		return vmOutput.ReturnCode
 	}
 
+	allLogs := d.eei.GetLogs()
+	if tooManyNodesLogs(allLogs) {
```
raduchis, line 1219: is there no way to see that the number of nodes it too high from the vmOutput or the error? Looking in the logs seems more like a "hack".

Also I think an unit tests is missing.

vm/systemSmartContracts/validator.go
```diff
@@ -1064,17 +1065,21 @@ func (v *validatorSC) stake(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
-		registrationData,
-		validatorConfig.NodePrice,
-		registrationData.RewardAddress,
-		args.CallerAddr,
-	)
-
-	if v.isNumberOfNodesTooHigh(registrationData) {
-		v.eei.AddReturnMessage("number of nodes is too high")
-		return vmcommon.UserError
+	if !v.isNumberOfNodesTooHigh(registrationData) {
```
raduchis, line 1068: if I have 100 keys registered, but only 10 staked, would it activate the next key?

vm/systemSmartContracts/delegation.go
mariusmihaic: Right, fixed it

vm/systemSmartContracts/validator.go
mariusmihaic: Added unit test for this scenario

raduchis (dismissed):

AdoAdoAdo (approved):

vm/systemSmartContracts/validator.go
```diff
@@ -935,8 +936,12 @@ func (v *validatorSC) isNumberOfNodesTooHigh(registrationData *ValidatorDataV2)
 		return false
 	}
 
+	return len(registrationData.BlsPubKeys) > v.calcNodeLimit()
+}
+
+func (v *validatorSC) calcNodeLimit() int {
```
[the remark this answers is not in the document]
mariusmihaic: Renamed it

PR #5923, opened 2024-02-07
No registration for too many key

sasurobert:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


vm/systemSmartContracts/validator.go
```diff
@@ -649,6 +649,10 @@ func (v *validatorSC) registerBLSKeys(
 	}
 
 	for _, blsKey := range newKeys {
+		if v.isNumberOfNodesTooHigh(registrationData) {
```
iulianpascalau, line 652: why do we have to put the check in the for loop? Can it be taken outside?

vm/systemSmartContracts/validator_test.go
```diff
@@ -466,6 +466,15 @@ func TestStakingValidatorSC_ExecuteStakeTooManyNodes(t *testing.T) {
 		assert.Equal(t, entry.Topics[0], []byte(numberOfNodesTooHigh))
 	}
 
+	stakeCalledInStakingSC := false
+	eei.ExecuteOnDestContextCalled = func(destination, sender []byte, value *big.Int, input []byte) (*vmcommon.VMOutput, error) {
+		if strings.Contains(string(input), "stake") {
+			stakeCalledInStakingSC = true
+			assert.False(t, stakeCalledInStakingSC)
```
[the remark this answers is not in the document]
sasurobert: the idea is that this code should be never reached.

vm/systemSmartContracts/validator.go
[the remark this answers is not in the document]
mariusmihaic: In this case, it means that those 50 will also be staked, not only registered, right?

raduchis: registrationData.BLSKeys is changed inside the for on line 675

vm/systemSmartContracts/validator_test.go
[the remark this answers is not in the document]
mariusmihaic: Ok, understood, although it looks confusing.
Perhaps make it more clear by removing the bool variable and replace those lines with smth like:

```go
require.Fail(t, "should not stake nodes")
```

vm/systemSmartContracts/validator.go
raduchis, line 652: if you have 60 nodes registered, but only 40 active, there will be no activateStakingFor for 10 nodes, even if you have stake for them. 
so it will behave different than having 40 nodes, and registering and staking another 20.

iulianpascalau: :man_facepalming:

vm/systemSmartContracts/validator_test.go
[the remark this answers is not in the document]
iulianpascalau: Yup, assert.Fail directly

vm/systemSmartContracts/validator.go
```diff
@@ -1069,46 +1075,73 @@ func (v *validatorSC) stake(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+
+func (v *validatorSC) activateNewBLSKeys(
+	registrationData *ValidatorDataV2,
+	blsKeys [][]byte,
+	newKeys [][]byte,
+	validatorConfig *ValidatorConfig,
+	args *vmcommon.ContractCallInput,
+) {
+	numRegisteredBlsKeys := len(registrationData.BlsPubKeys)
+	numNodesTooHigh := v.activateStakingFor(
```
mariusmihaic, line 1097: It is a bit confusing having a function called `activateStakingFor` -> which **should do** something, but returns true when it **doesn't do** it (activate), e.g.: number of nodes is too high). Perhaps we can either change return logic of variable/func names

vm/systemSmartContracts/validator.go
```diff
@@ -1069,46 +1075,73 @@ func (v *validatorSC) stake(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+		blsKeys,
+		newKeys,
+		registrationData,
+		validatorConfig.NodePrice,
+		registrationData.RewardAddress,
+		args.CallerAddr,
+	)
+
+	if numNodesTooHigh && len(blsKeys) > 0 {
```
mariusmihaic, line 1106: Can it ever happen that len(blsKeys) is 0 here?

vm/systemSmartContracts/validator.go
```diff
@@ -1069,46 +1075,73 @@ func (v *validatorSC) stake(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
+	if v.isNumberOfNodesTooHigh(numRegisteredKeys) {
+		return true
+	}
+
+	maxNumNodesToActivate := len(blsKeys)
+	if v.enableEpochsHandler.IsFlagEnabled(common.StakeLimitsFlag) {
+		maxNumNodesToActivate = v.computeNodeLimit() - numRegisteredKeys + len(newKeys)
+	}
+	nodesActivated := 0
+	if nodesActivated >= maxNumNodesToActivate && len(blsKeys) >= maxNumNodesToActivate {
```
mariusmihaic, line 1142: Is it ok that you
1. `nodesActivated := 0`
2. check `if nodesActivated >= maxNumNodesToActivate`
?

vm/systemSmartContracts/validator.go
```diff
@@ -1069,46 +1075,73 @@ func (v *validatorSC) stake(args *vmcommon.ContractCallInput) vmcommon.ReturnCod
...
 	registrationData *ValidatorDataV2,
 	fixedStakeValue *big.Int,
 	rewardAddress []byte,
 	ownerAddress []byte,
-) {
-	numRegistered := uint64(registrationData.NumRegistered)
+) bool {
+	numActivatedKey := uint64(registrationData.NumRegistered)
+
+	numRegisteredKeys := len(registrationData.BlsPubKeys)
```
mariusmihaic, line 1132: Again the naming here is a bit unfortunate. Having a variable called `numRegistered`, which is not `registrationData.NumRegistered`

vm/systemSmartContracts/validator.go
sasurobert: that is true. but you can always unstake and unbond those nodes.

vm/systemSmartContracts/validator.go
```diff
@@ -1127,12 +1160,19 @@ func (v *validatorSC) activateStakingFor(
 		}
 
 		if stakedData.UnStakedNonce == 0 {
-			numRegistered++
+			numActivatedKey++
+		}
+
+		nodesActivated++
```
mariusmihaic, line 1166: Is it ok that you can increase two times the num of activated nodes here?

vm/systemSmartContracts/validator.go
```diff
@@ -1127,12 +1160,19 @@ func (v *validatorSC) activateStakingFor(
...
+		if nodesActivated >= maxNumNodesToActivate {
+			break
 		}
 	}
 
-	registrationData.NumRegistered = uint32(numRegistered)
-	registrationData.LockedStake.Mul(fixedStakeValue, big.NewInt(0).SetUint64(numRegistered))
+	registrationData.NumRegistered = uint32(numActivatedKey)
+	registrationData.LockedStake.Mul(fixedStakeValue, big.NewInt(0).SetUint64(numActivatedKey))
```
mariusmihaic, line 1173: If you increase `nodesActivated` two times above, is this still backwards compatbile?

vm/systemSmartContracts/validator.go
raduchis: ok

vm/systemSmartContracts/validator_test.go
```diff
@@ -466,6 +466,66 @@ func TestStakingValidatorSC_ExecuteStakeTooManyNodes(t *testing.T) {
...
+		}
+		return &vmcommon.VMOutput{}, nil
+	}
+
+	key1 := []byte("Key1")
+	key2 := []byte("Key2")
+	key3 := []byte("Key3")
+	arguments.Function = "stake"
+	arguments.CallValue = big.NewInt(0).Mul(big.NewInt(3), big.NewInt(10000000))
+	arguments.Arguments = [][]byte{big.NewInt(3).Bytes(), key1, []byte("msg1"), key2, []byte("msg2"), key3, []byte("msg3")}
```
mariusmihaic, line 481: Test is called TestStakingValidatorSC_ExecuteStakeTooManyNodes, but here we don't check that the number of nodes is too high. Can we change the limit nodes to 3 in this case and check that this still works?

vm/systemSmartContracts/validator.go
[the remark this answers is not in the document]
sasurobert: exactly.

vm/systemSmartContracts/validator.go
sasurobert: yes. in 99% of cases. blsKeys are there when automatic activation is on for delegation contract. Or the validator explicitely wants to stake more nodes.

vm/systemSmartContracts/validator.go
sasurobert: yes. this is the case when you cannot activate any new node.

vm/systemSmartContracts/validator.go
sasurobert: those are 2 different variables.

vm/systemSmartContracts/validator.go
sasurobert: those are 2 different variables.

vm/systemSmartContracts/validator_test.go
sasurobert: It checks. as AddLogEntry must be called. Line 465 and 485

mariusmihaic: Right, missed it

vm/systemSmartContracts/validator.go
sasurobert: renamed to numAllBlsKeys

vm/systemSmartContracts/validator.go
sasurobert: pushed

raduchis (approved):

mariusmihaic (approved):

iulianpascalau (approved):

PR #5928, opened 2024-02-08
new tests and fixes

axenteoctavian:
- Unit tests for code coverage
- Updates and fixes
- Deploy scripts without `txgen`
  
## Proposed changes
- 
- 
- 

## Testing procedure
- running sovereign node
- 
- 

## Pre-requisites


scripts/testnet/prerequisites.sh
```diff
@@ -75,7 +75,7 @@ if [ "$SOVEREIGN_DEPLOY" -eq 1 ]; then
     popd
 fi
 
-if [[ $PRIVATE_REPOS -eq 1 ]]; then
+if [[ $USE_TXGEN -eq 1 ]]; then
```
mariusmihaic, line 78: Can we better add an or here and keep also PRIVATE_REPOS check?

scripts/testnet/deploySovereignWithIndexer.sh
```diff
@@ -0,0 +1,31 @@
...
+source variables.sh
+if [ "$SHARD_VALIDATORCOUNT" -lt 3 ]; then
+  sed -i 's/export SHARD_VALIDATORCOUNT=.*/export SHARD_VALIDATORCOUNT=3/' variables.sh
+fi
+
+echo -e "${GREEN}Generating the configuration files...${NC}"
+./config.sh
+
+echo -e "${GREEN}Starting the sovereign chain (in a screen)...${NC}"
+screen -L -Logfile sovereignStartLog.txt -d -m -S sovereignStartScreen ./sovereignStart.sh debug
```
mariusmihaic, line 26: Not sure if we need to keep starting sovereign with screen.
I think we can remove the screen in this case

scripts/testnet/deploySovereignWithIndexer.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo -e "${GREEN}Sleeping few minutes so the sovereign chain will begin...${NC}"
+sleep 120
```
mariusmihaic, line 29: This sleep I think it can be reduced to 60 seconds

node/sovereignNodeFactory_test.go
```diff
@@ -25,3 +25,17 @@ func TestSovereignNodeFactory_CreateNewNode(t *testing.T) {
...
+	t.Parallel()
+
+	sovereignNodeFactory := node.NewSovereignNodeFactory()
+
+	options := []node.Option{
+		node.WithStatusCoreComponents(nil),
+	}
+
+	sn, err := sovereignNodeFactory.CreateNewNode(options...)
+	require.Equal(t, "error applying option: nil status core components", err.Error())
```
mariusmihaic, line 39: We should better check that the error contains the `ErrNilStatusCoreComp...` here instead of checking for string.
Reason being that if somebody changes the error message your test would fail

node/node_test.go
```diff
@@ -1344,6 +1344,14 @@ func testNodeGetESDTsWithRole(t *testing.T, nodeFactory node.NodeFactory, shardI
 	tokenResult, _, err = n.GetESDTsWithRole(testscommon.TestAddressAlice, core.ESDTRoleNFTCreate, api.AccountQueryOptions{}, context.Background())
 	require.NoError(t, err)
 	require.Len(t, tokenResult, 0)
+
+	tokenResult, _, err = n.GetESDTsWithRole(testscommon.TestAddressAlice, "invalid role", api.AccountQueryOptions{}, context.Background())
```
mariusmihaic, line 1348: The naming for all these tests is a bit unfortunate.
All tests names contain "shouldWork", but they test invalid cases. We should have perhaps separate tests for invalid cases or put inside a `t.Run`

mariusmihaic (approved):

sasurobert (approved):

PR #5943, opened 2024-02-09
New endpoint /node/loaded-keys

sstanculeanu:
- new enpoint needed for all the keys loaded by a node
  
## Proposed changes
- added new endpoint
image

## Testing procedure
- standard system test + multi key test and check `/node/loaded-keys` on both single key(validator and observer) and multi key 

## Pre-requisites


facade/nodeFacade.go
```diff
@@ -36,7 +36,8 @@ import (
 const DefaultRestInterface = "localhost:8080"
 
 // DefaultRestPortOff is the default value that should be passed if it is desired
-//  to start the node without a REST endpoint available
+//
```
iulianpascalau, line 39: indenting went wrong here and on L167

node/external/nodeApiResolver.go
```diff
@@ -339,12 +339,18 @@ func (nar *nodeApiResolver) GetManagedKeysCount() int {
...
-// GetManagedKeys returns all keys managed by the current node when running in multikey mode
+// GetManagedKeys returns all keys that should act as validator(main or backup that took over) and will be managed by this node
 func (nar *nodeApiResolver) GetManagedKeys() []string {
 	managedKeys := nar.managedPeersMonitor.GetManagedKeys()
 	return nar.parseKeys(managedKeys)
 }
 
+// GetLoadedKeys returns all keys that were loaded and will be managed by this node
+func (nar *nodeApiResolver) GetLoadedKeys() []string {
+	loadedKeys := nar.managedPeersMonitor.GetLoadedKeys()
```
iulianpascalau, line 350: Question: should we append the main BLS key from crypto components to this list? Reason: maybe we are using the node in single key mode, not multikey and this route should return the main BLS key. 
Or return that single key if the list returned by the managedPeersMonitor is empty. I guess this will be better as the multikey list won't be polluted by an observer BLS key.

keysManagement/managedPeersHolder_test.go
```diff
@@ -751,6 +751,24 @@ func TestManagedPeersHolder_GetManagedKeysByCurrentNode(t *testing.T) {
 	})
 }
 
+func TestManagedPeersHolder_GetLoadedKeysByCurrentNode(t *testing.T) {
+	t.Parallel()
+
+	args := createMockArgsManagedPeersHolder()
+	holder, _ := keysManagement.NewManagedPeersHolder(args)
+	_ = holder.AddManagedPeer(skBytes1)
```
iulianpascalau, line 759: :+1: testing out the sort call

node/external/nodeApiResolver.go
sstanculeanu: pushed

it will also return the public key for single key observers

iulianpascalau (approved):

keysManagement/managedPeersHolder.go
```diff
@@ -299,6 +300,23 @@ func (holder *managedPeersHolder) GetManagedKeysByCurrentNode() map[string]crypt
...
+	defer holder.mut.RUnlock()
+
+	allLoadedKeys := make([][]byte, 0, len(holder.data))
+	for pk := range holder.data {
+		allLoadedKeys = append(allLoadedKeys, []byte(pk))
+	}
+
+	sort.Slice(allLoadedKeys, func(i, j int) bool {
+		return string(allLoadedKeys[i]) < string(allLoadedKeys[j])
+	})
```
ssd04, line 315: extract this logic into a separate func and used in ManagedKeysByCurrentNode and remove sorting logic from managePeersMonitor?

sstanculeanu: ManagedKeysByCurrentNode has an extra check than GetLoadedKeysByCurrentNode, the one for validators.. thus the common code is just the sorting which can be done separate

ssd04 (approved):

gabi-vuls (approved):
- allin ✅ 
- scenario ✅ 
- on multikey nodes the route prins all keys , and on single key nodes the route prints that one specific key ✅ 


Normal allin test: v1.6.15-dev-config-eb2e06c06d -> loaded-keys-ebeaab4359

--- Specific errors ---

block hash does not match 399
wrong nonce in block 295
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  1
Nr. of all WARNS:  258
Nr. of new ERRORS:  1
Nr. of new WARNS:  23
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

PR #5944, opened 2024-02-11
Configurable epoch change delay

miiu96:
- Moved the hard-coded wait duration from the epoch start trigger in `config.toml` file

## Pre-requisites

### DIFF
--- a/epochStart/shardchain/trigger.go
+++ b/epochStart/shardchain/trigger.go
@@ -46,14 +46,15 @@ type ArgsShardEpochStartTrigger struct {
 	HeaderValidator epochStart.HeaderValidator
 	Uint64Converter typeConverters.Uint64ByteSliceConverter
 
-	DataPool             dataRetriever.PoolsHolder
-	Storage              dataRetriever.StorageService
-	RequestHandler       epochStart.RequestHandler
-	EpochStartNotifier   epochStart.Notifier
-	PeerMiniBlocksSyncer process.ValidatorInfoSyncer
-	RoundHandler         process.RoundHandler
-	AppStatusHandler     core.AppStatusHandler
-	EnableEpochsHandler  common.EnableEpochsHandler
+	DataPool                      dataRetriever.PoolsHolder
+	Storage                       dataRetriever.StorageService
+	RequestHandler                epochStart.RequestHandler
+	EpochStartNotifier            epochStart.Notifier
+	PeerMiniBlocksSyncer          process.ValidatorInfoSyncer
+	RoundHandler                  process.RoundHandler
+	AppStatusHandler              core.AppStatusHandler
+	EnableEpochsHandler           common.EnableEpochsHandler
+	ExtraDelayForRequestBlockInfo time.Duration
 
 	Epoch    uint32
 	Validity uint64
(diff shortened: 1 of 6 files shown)


epochStart/shardchain/trigger.go
```diff
@@ -260,6 +263,7 @@ func NewEpochStartTrigger(args *ArgsShardEpochStartTrigger) (*trigger, error) {
 		appStatusHandler:              args.AppStatusHandler,
 		roundHandler:                  args.RoundHandler,
 		enableEpochsHandler:           args.EnableEpochsHandler,
+		extraDelayForRequestBlockInfo: args.ExtraDelayForRequestBlockInfo,
```
iulianpascalau, line 266: I think we should add a warning in the logs if we provide a different value of 3 seconds for this (as seen in the chain simulator)
Reason: If we change this in the future, at least we should have a warning that the epoch change trigger might not behave normally.

epochStart/shardchain/trigger.go
```diff
@@ -586,7 +590,7 @@ func (t *trigger) receivedMetaBlock(headerHandler data.HeaderHandler, metaBlockH
 		t.newEpochHdrReceived = true
 		t.mapEpochStartHdrs[string(metaBlockHash)] = metaHdr
 		// waiting for late broadcast of mini blocks and transactions to be done and received
-		wait := common.ExtraDelayForRequestBlockInfo
```
iulianpascalau, line 589: remove the constant?

iulianpascalau (commented):
:+1:

sstanculeanu (commented):
no other comments

epochStart/shardchain/trigger.go
miiu96: done

epochStart/shardchain/trigger.go
miiu96: this const is used also in `baseProcessor.go` file

iulianpascalau (approved):

sstanculeanu (approved):

node/chainSimulator/chainSimulator.go
```diff
@@ -163,6 +163,7 @@ func (s *simulator) incrementRoundOnAllValidators() {
 
 func (s *simulator) allNodesCreateBlocks() error {
 	for _, node := range s.handlers {
+		time.Sleep(2 * time.Millisecond)
```
iulianpascalau, line 166: you will have a conflict here with what we have on staking-v4 branch:
```
// TODO MX-15150 remove this when we remove all goroutines
time.Sleep(2 * time.Millisecond)

err := node.CreateNewBlock()
if err != nil {
	return err
}
```

iulianpascalau (approved):

sstanculeanu (approved):

gabi-vuls (approved):
Normal allin test: v1.6.13-dev-config-06f147d698 -> move-in-config.toml-epoch--9b5f0ab6cf

--- Specific errors ---

block hash does not match 912
wrong nonce in block 301
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  216
Nr. of new ERRORS:  0
Nr. of new WARNS:  0
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

--- WARNINGS ---

/------/

PR #5969, opened 2024-02-16
Staking v4 nodes config provider for api calls

mariusmihaic:
- Created nodes config provider for api calls.
- Added  `GetCurrentNodesConfig` function which retrieves the current configuration of nodes. However, when invoked during epoch stakingV4 step 2 through API calls, it will provide the nodes configuration as it will appear in epoch stakingV4 step 3. This adjustment is made because, with the transition to step 3 at the epoch change, the maximum number of nodes will be reduced. Therefore, calling this API during step 2 aims to offer a preview of the upcoming epoch, accurately reflecting the adjusted number of nodes that will qualify from the auction.

## Testing procedure
We need a system test which has num nodes > max num nodes from configs, meaning that probably before stakingV4 we need to stake extra nodes that will be placed in queue, then afterwards in auction.

Call to api route `/auction` in stakingV4 step2 and check that the correct number of qualified nodes is displayed.

Taking this config as example:
```
    StakingV4Step1EnableEpoch = 4
    StakingV4Step2EnableEpoch = 5
    StakingV4Step3EnableEpoch = 6

        { EpochEnable = 0, MaxNumNodes = 48, NodesToShufflePerShard = 4 },
        { EpochEnable = 1, MaxNumNodes = 64, NodesToShufflePerShard = 2 },
(numOfShards+1)*NodesToShufflePerShard)
        { EpochEnable = 6, MaxNumNodes = 56, NodesToShufflePerShard = 2 },
```

**!!And suppose we have 66 staked nodes.!!**

Upon calling `/auction` route we should get:
- in **step1** (epoch 4) : all nodes from queue qualified (2 nodes should be in auction)
- in **step2** (epoch 5):  **8 nodes qualified + 2 nodes unqualified**
- in **step** 3 and forward on: **8 nodes qualified + 2 nodes unqualified**


iulianpascalau (approved):

raduchis (approved):

gabi-vuls (approved):
Normal allin test: v1.6.15-dev-config-eb2e06c06d -> MX-15156-api-nodes-config--162c4a77ce

--- Specific errors ---

block hash does not match 124
wrong nonce in block 149
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  227
Nr. of new ERRORS:  0
Nr. of new WARNS:  2
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

--- WARNINGS ---

up-mad-validator-11 :
Warn: trie recreate error:           error = getNodeFromDB error: key not found for key root =  1

up-mad-validator-7 :
Warn: trie recreate error:           error = getNodeFromDB error: key not found for key root =  1

/------/

PR #5987, opened 2024-02-21
Sketch example for codecs

andreibancioiu:
Sandbox for codec examples.

Also see: https://github.com/multiversx/mx-sdk-abi-incubator/tree/init/golang/abi


andreibancioiu:
We should move these examples to docs.

PR #5998, opened 2024-02-27
Docker local-testnet scripts

cristure:
- Scipt to setup a local testnet with docker.
  
## Proposed changes
- Added scripts to setup local testnet.

## Testing procedure
- Run scripts/docker-testnet/start.sh
- Inspect the created containers.

## Pre-requisites
- Docker installed.


scripts/docker-testnet/helpers.sh
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env bash
```
iulianpascalau, line 1: why not rename this to "functions.sh" ?

scripts/docker-testnet/helpers.sh
```diff
@@ -0,0 +1,153 @@
...
+    # Truncate config.toml before the [[Observers]] list
+    sed -i -n '/\[\[Observers\]\]/q;p' config_edit.toml
+
+    if [ "$SHARD_OBSERVERCOUNT" -le 0 ]; then
+        generateProxyValidatorListDocker config_edit.toml
+    else
+        generateProxyObserverListDocker config_edit.toml
+    fi
+
+    cp config_edit.toml config.toml
```
iulianpascalau, line 85: `mv` directly and remove next line?

scripts/docker-testnet/start.sh
```diff
@@ -0,0 +1,38 @@
...
+
+prepareFolders
+
+buildConfigGenerator
+
+generateConfig
+
+copyConfig
+
+copySeednodeConfig
```
iulianpascalau, line 22: interesting, so you actually use a lot of the existing testnet scripts configuration functions :thinking:

cristure: yep, it's mostly the same configuration. The only difference in configuration are the ports and the ip addresses.

scripts/docker-testnet/helpers.sh
cristure: done.

scripts/docker-testnet/start.sh
iulianpascalau: right, I've noticed that: the first question was: how was possible to function if all nodes run on the same ports. But then I've saw the IPs.

iulianpascalau (approved):

iulianpascalau (approved):

AdoAdoAdo (dismissed):

iulianpascalau (approved):
will test this ASAP

scripts/docker-testnet/build.sh
```diff
@@ -0,0 +1,6 @@
+pushd ../..
```
ssd04, line 1: add `#!/usr/bin/env bash` here also? i think pushd is bash specific

ssd04: is this script intended to be run separately?

scripts/docker-testnet/functions.sh
```diff
@@ -0,0 +1,196 @@
...
+  # this variable is used to keep track of the allocated IP addresses in the network, by removing the last byte
+  # of the DOCKER_NETWORK_SUBNET. One can consider this the host network address without the last byte at the end.
+  export NETWORK_ADDRESS=$(echo "$DOCKER_NETWORK_SUBNET" | rev | cut -d. -f2- | rev)
+}
+
+startSeedNode() {
+  docker run -d --name seednode -v ${TESTNETDIR}/seednode/config:/go/mx-chain-go/cmd/seednode/config \
+  --network ${DOCKER_NETWORK_NAME} \
+  seednode:dev \
+  --rest-api-interface=0.0.0.0:10000
```
ssd04, line 31: indentation here?

scripts/docker-testnet/functions.sh
```diff
@@ -0,0 +1,196 @@
...
+   done
+ done
+
+ for ((i = 0; i < META_OBSERVERCOUNT; i++)); do
+    KEY_INDEX=$((TOTAL_NODECOUNT - observerIdx - 1))
+
+    docker run -d --name "observer${observerIdx}-${NETWORK_ADDRESS}.${IP_HOST_BYTE}-10200-metachain" \
+        -v $TESTNETDIR/node/config:/go/mx-chain-go/cmd/node/config \
+        --network ${DOCKER_NETWORK_NAME} \
+        node:dev \
```
ssd04, line 64: please use same indentation pattern for the entire script and remove unnecessary empty lines

scripts/docker-testnet/functions.sh
```diff
@@ -0,0 +1,196 @@
...
+  else
+    cd $(dirname $MULTIVERSXDIR)
+
+    git clone git@github.com:multiversx/mx-chain-deploy-go.git || true
+    git clone git@github.com:multiversx/mx-chain-proxy-go.git || true
+  fi
+}
+
+createDockerNetwork() {
+  docker network create -d bridge --subnet=${DOCKER_NETWORK_SUBNET} ${DOCKER_NETWORK_NAME}
```
ssd04, line 20: maybe check if `DOCKER_NETWOR_NAME` already exists

scripts/docker-testnet/clean.sh
```diff
@@ -0,0 +1,19 @@
...
+
+export MULTIVERSXTESTNETSCRIPTSDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+source "$MULTIVERSXTESTNETSCRIPTSDIR/variables.sh"
+
+echo "Stopping all containers..."
+docker stop $(docker ps -a -q)
+
+echo "Removing all containers..."
+docker container prune -f
```
ssd04, line 13: the scripts will most probably be used in development environment where there might be also other local containers running, i don't think it's fine to stop and prune all containers, we should stop only the intended containers

ssd04 (changes requested):

scripts/docker-testnet/build.sh
cristure: yes, it is. I could have added the build step inside the `./start.sh` script, but it was easier for me to debug, since it didn't build the images at every run. Don't have a strong opinion on this, I can add it to the wrapper script instead. I was planning on adding a `README.md` nevertheless.

scripts/docker-testnet/functions.sh
cristure: fixed.

scripts/docker-testnet/functions.sh
cristure: I hope it's fixed.

scripts/docker-testnet/functions.sh
cristure: added `set -u`. Unbound variables will throw an error now.

scripts/docker-testnet/build.sh
iulianpascalau: I think it is better that we keep the build separate from the run script. Not a strong opinion, thus.

scripts/docker-testnet/variables.sh
```diff
@@ -114,32 +108,28 @@ export PORT_ORIGIN_VALIDATOR_REST="9500"
...
-
-
-if [ "$TESTNETMODE" == "debug" ]; then
-  LOGLEVEL="*:DEBUG,api:INFO"
-fi
-
-if [ "$TESTNETMODE" == "trace" ]; then
-  LOGLEVEL="*:TRACE"
-fi
+## Use tmux or not. If set to 1, only 2 terminal windows will be opened, and
```
iulianpascalau, line 111: commented code. Can use a local.sh file instead. We have exceptions for it in the .gitignore

scripts/docker-testnet/functions.sh
```diff
@@ -6,14 +6,18 @@
...
 cloneRepositories() {
-  if [[ -n $CI_RUN ]]; then
-    echo "Repositories have been cloned in the CI"
-  else
-    cd $(dirname $MULTIVERSXDIR)
-
-    git clone git@github.com:multiversx/mx-chain-deploy-go.git || true
-    git clone git@github.com:multiversx/mx-chain-proxy-go.git || true
-  fi
+  cd $(dirname $MULTIVERSXDIR)
```
iulianpascalau, line 9: Why this change? Does this function will **always** clone the proxy and deploy repos?

scripts/docker-testnet/functions.sh
ssd04: here I was referring to the fact that if the network is already created, it will fail because it is not able to create docker network again

scripts/docker-testnet/README.md
```diff
@@ -0,0 +1,10 @@
+# Setting up a local-testnet with Docker
+
+First and foremost, one needs to build the **seednode** & **node** images. Hence, the **_build.sh_**
+script is provided. This can be done, by invoking the script or building the images manually.
+
+```
+./build.sh  # (Optional) Can be ignored if you already have the images stored in the local registry.
+./start.sh  # Will start the local-testnet.
+./clean.sh  # Will stop and remove the containers related to the local-testnet.
```
ssd04, line 9: i think that a good flow here is to have also a `stop` step, in order to be able to have the following development flow:
- initially: `build`
- multiple times: `start` , `stop`
- and finally: `clean`

i think is not very good to have to clean everything on each run, something we just wan't to stop the squad and start it again, with the same/setup configuration, later on

ssd04: this is also related to the docker create comment

scripts/docker-testnet/variables.sh
cristure: removed altogether. not needed in our context.

scripts/docker-testnet/functions.sh
cristure: yes, it will. But if the repository already exists, it won't clone them again.

scripts/docker-testnet/README.md
cristure: This intended workflow adds some complexity. The `start.sh` uses the `docker run` command. If you stop all containers, they will only start again with`docker start` instead, otherwise it will throw an error. I don't see much of a benefit to this behaviour, as I need to persists all the `containerIds/containerNames` between stopping and restarting. One can achieve this manually by leveraging already available commands in the `clean.sh` script.

scripts/docker-testnet/functions.sh
iulianpascalau: confirmed that is working properly :+1: 
```
git clone git@github.com:multiversx/mx-chain-deploy-go.git || true
fatal: destination path 'mx-chain-deploy-go' already exists and is not an empty directory.
```

iulianpascalau (dismissed):
tested locally with system-tests-go and worked perfectly

scripts/docker-testnet/start.sh
```diff
@@ -2,42 +2,17 @@
...
+    echo "File $file_path not found."
+    exit 1
 fi
 
+# Read the file line by line
+while IFS= read -r line; do
+    docker start $line
+done < "$file_path"
+
+rmdir ./tmp
```
iulianpascalau, line 18: add empty line?

scripts/docker-testnet/stop.sh
```diff
@@ -0,0 +1,21 @@
...
+export CONTAINER_IDS=$(docker network inspect -f '{{range $k, $v := .Containers}}{{printf "%s\n" $k}}{{end}}' "$DOCKER_NETWORK_NAME")
+
+mkdir -p ./tmp
+
+# Stop each container
+echo "Stopping containers..."
+for CONTAINER_ID in $CONTAINER_IDS; do
+    docker stop "$CONTAINER_ID"
+    echo "$CONTAINER_ID" >> ./tmp/stopped_containers
+done
```
iulianpascalau, line 21: add empty line?

iulianpascalau (dismissed):

ssd04 (approved):

iulianpascalau (approved):

AdoAdoAdo (approved):

PR #6038, opened 2024-03-11
Mx 15248 outgoing operations hasher

axenteoctavian:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/cmd/sovereignnode/sovereignNodeRunner.go
+++ b/cmd/sovereignnode/sovereignNodeRunner.go
@@ -1333,6 +1333,11 @@ func (snr *sovereignNodeRunner) CreateManagedProcessComponents(
 		return nil, err
 	}
 
+	outgoingOperationsHasher, err := hasherFactory.NewHasher("sha256")
+	if err != nil {
+		return nil, err
+	}
+
 	processArgs := processComp.ProcessComponentsFactoryArgs{
 		Config:                                *configs.GeneralConfig,
 		EpochConfig:                           *configs.EpochConfig,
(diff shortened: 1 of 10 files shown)


cmd/sovereignnode/sovereignNodeRunner.go
```diff
@@ -1337,6 +1337,11 @@ func (snr *sovereignNodeRunner) CreateManagedProcessComponents(
 		return nil, err
 	}
 
+	outgoingOperationsHasher, err := hasherFactory.NewHasher("sha256")
```
mariusmihaic, line 1340: Hasher type should be a configurable paramter found and loaded from `sovereignConfig.toml`

process/block/argProcessor.go
```diff
@@ -99,6 +99,7 @@ type ArgBaseProcessor struct {
 	OutGoingOperationsPool         OutGoingOperationsPool
 	DataCodec                      sovereign.DataCodecProcessor
 	TopicsChecker                  sovereign.TopicsChecker
+	OperationsHasher               hashing.Hasher
```
mariusmihaic, line 102: This approach is not correct.
You should not pass this interface everywhere from factories up to baseProcess and metaBlock process which will never use this.

This is the reason why you have `sovereignBlockProcessorFactory`. That factory should create the hasher based on a config string. Max you can do is to add `OperationHasherType` as `string` in `BaseArgs` and only make use of it inside sovereign factory.

Also, you will have to delete the disabled operation hasher

mariusmihaic (changes requested):
This approach is wrong

cmd/sovereignnode/sovereignNodeRunner.go
```diff
@@ -1376,6 +1381,7 @@ func (snr *sovereignNodeRunner) CreateManagedProcessComponents(
 		OutGoingOperationsPool:                outGoingOperationsPool,
 		DataCodec:                             dataCodec,
 		TopicsChecker:                         topicsChecker,
+		OperationsHasher:                      outgoingOperationsHasher,
```
raduchis, line 1384: the args below RunTypeComponents should be maybe moved to runTypeComponents in a future cleanup&refactor PR.

process/block/argProcessor.go
raduchis: :+1: also the other members of ArgBaseProcessor should be used just where they are needed

axenteoctavian:
cancelled, changed with https://github.com/multiversx/mx-chain-go/pull/6051

PR #6040, opened 2024-03-11
Added stake-unstake-unbond scenario + EEI fix

iulianpascalau:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


vm/systemSmartContracts/eei.go
```diff
@@ -339,8 +340,11 @@ func (host *vmContext) properMergeContexts(parentContext *vmContext, returnCode
...
 	host.AddReturnMessage(parentContext.returnMessage)
-	if returnCode != vmcommon.Ok {
-		// no need to merge - revert was done - transaction will fail
+
+	// merge contexts if the return code is OK or the fix flag is activated because it was wrong not to merge them if the call failed
+	shouldMergeContexts := returnCode == vmcommon.Ok || host.enableEpochsHandler.IsFlagEnabled(common.AlwaysMergeContextsInEEIFlag)
+	if !shouldMergeContexts {
+		// backwards compatibility
 		return
 	}
```
AdoAdoAdo, line 350: Is what happens in L360 correct?
is it possible for example for rightAccount to exist outside of outputAccount and to be set to a clean state (leftAccount on line 355)?

This should be taken also in the context of sovereign chains where the systemscs are in the sovereign chain, not in a dedicated meta chain.

sasurobert: Yes. This is correct.

Merging of context happen roughly the same way on WasmVM. In case of SovereignShards, wasm SC will can executeOnDest which executes here, and will interpret and merge the vmOutput created by the execution.

Contexts, merging, popping is independent of VM.

sasurobert (dismissed):

vm/systemSmartContracts/eei.go
sasurobert: I introduced that bug, the if should have not existed. Merging, popping and pushing is formally verified in WasmVM.

iulianpascalau: After a talk with @sasurobert , I can conclude that L360 is executed if I have a change in the parent context (let's say the parent context executed something and for account A set the BalanceDelta to 300) and that account is not found in the current context. Example here is that the child context did not touch account A. Well, the `!exist` is satisfied and the left account is created as new, with 0s on it. This newly created left account pointer is assigned to the host.outputAccounts map and then provided to the `addOutputAccounts` as **destination** parameter. Then, we can further see that on L380 we actually do an Add operation on which the provided leftAccount (which in this function is declared as destination) will have its BalanceDelta field added with the provided rightAccount, yielding to a value of 300 (0+300).
 Add is also done for the following fields: `GasUsed`,  `BytesAddedToStorage`, `BytesDeletedFromStorage` and there is an append for the output transfers (the order of appending does not matter as we have a field called Index on the OutputTransfer struct that will decide the execution order). The Nonce field will be set to the higher value of both right & destination. The `Code`, `CodeMetadata`, `CodeDeployerAddress` will be set to the values of the **parent** context, if the parent context set them.

sasurobert (approved):

AdoAdoAdo (approved):

liviuf-ancas (approved):
:green_circle:  on system testnet

PR #6041, opened 2024-03-12
FIX: Remove createSovereignGenesisESDTTransfers to avoid double spending

mariusmihaic:
- Removed createSovereignGenesisESDTTransfers to avoid double spending.

## Proposed changes
- Since we introduced sovereign accounts creator, initial esdt balances are directly set to trie into genesis processor, so there is no need to create initial cross shard esdt transfers, since that would double the balance from initial config. 

## Testing procedure
- Local testnet

## Pre-requisites

### DIFF
--- a/genesis/process/sovereignGenesisBlockCreator.go
+++ b/genesis/process/sovereignGenesisBlockCreator.go
@@ -9,7 +9,6 @@ import (
 	"github.com/multiversx/mx-chain-core-go/core"
 	"github.com/multiversx/mx-chain-core-go/data"
 	"github.com/multiversx/mx-chain-core-go/data/block"
-	"github.com/multiversx/mx-chain-core-go/data/smartContractResult"
 	"github.com/multiversx/mx-chain-core-go/data/transaction"
 	"github.com/multiversx/mx-chain-go/config"
 	"github.com/multiversx/mx-chain-go/factory"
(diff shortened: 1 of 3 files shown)


genesis/process/sovereignGenesisBlockCreator.go
```diff
@@ -193,12 +192,6 @@ func createSovereignShardGenesisBlock(
 	indexingData.StakingTxs = stakingTxs
 
 	metaScrsTxs := metaProcessor.txCoordinator.GetAllCurrentUsedTxs(block.SmartContractResultBlock)
-	genesisESDTTransfers, err := createSovereignGenesisESDTTransfers(arg, shardProcessors.scrProcessor)
```
raduchis, line 196: Will you be able to see these transactions in the api/explorer? 
For the mainnet the genesis transactions had to be indexed after the beginning of the mainnet. 
They should be visible right from the start

mariusmihaic: Transfers from `createSovereignGenesisESDTTransfers` are deleted.
The other ones, remain the same as in mainnet.

sasurobert: At least generate LogEvents for the initial balance setups - than it should be good.

mariusmihaic: Will add a task for it to be done on a separate PR

sasurobert (approved):

raduchis (approved):

PR #6111, opened 2024-04-08
Fix gas for view "getAllNodeStates"

gfusee:
- Delegator SC's "getAllNodeState" view uses DelegationOps gas
- There is a gas amount in the config called GetAllNodeStates. So I guess "getAllNodeState" should use it instead of DelegationOps
- GetAllNodeStates = 20 * DelegationOps
  
## Proposed changes
- Added a flag "FixDelegationGetAllNodeStatesViewGasFlag"
- Used the gas value "GetAllNodeStates" when the above flag is activated
- Added and changed tests accordingly

## Testing procedure
- go test ./... on a Ubuntu AMD64 server

## Notes
- I'm not used to the mx-chain-go codebase therefore I might have missed something, I mainly checked at the pre-existing tests to learn what I had to do in order to fix this issue
- I see y'all are pushing fixes into the rc/v1.7.0 branch so I did the same, maybe I had to create the PR to target master or another branch?

## Pre-requisites


sasurobert (changes requested):
checking this. will come with a feedback later.

but this has to target rc/v1.7.0

PR #6114, opened 2024-04-11
POC: Distribute to waiting from auction based on leaving nodes

mariusmihaic:
- In case of leaving active nodes in epoch (waiting/eligible), as long as there leaving nodes per shard < number of active nodes && leaving nodes per shard  < number nodes to shuffle per shard, those will not be shuffled out, they can leave. If there are more nodes per shard that are leaving than the number of nodes to shuffle per shard, the remaining will be forced to stay
  
## Proposed changes
- Fix auction list selection algorithm to consider leaving active nodes and forced nodes to stay when computing number of avaiable slots to select qualified nodes.

## Testing procedure
- One easy starting point would be to do something similar to this test: `TestChainSimulator_UnStakeOneActiveNodeAndCheckAPIAuctionList`

Scenario 1: 
- Have every epoch auction list with enough nodes (let's say 8 qualified, 2 unqualified)
- Unstake 1 or more eligible/waiting nodes (no more than all nodes per shard or no more than all eligible) and call auction list api; we should see 8 qualified nodes. Next epoch we need to check that exactly 8 nodes were qualified. Afterwards(some epochs), those leaving nodes will be leaving and replaced by the auction nodes

Scenario 2:
- Have every epoch auction list with MORE nodes (let's say 8 qualified, 20 unqualified)
- Let's say we have number of nodes to shuffle per shard = 2 and 8 eligible nodes + 4 waiting nodes
- Unstake from one or more shards >=2 nodes. We should see in api auction list that we have number of qualified nodes = maxNumNodesFromConfig -  (numRemainingActiveNodes + number of forced nodes to stay); _forced node = every other unstaked node over 2 node unstaked per shard_

**At all times for all these tests, we should always have in system eligible + waiting nodes <= maxNumNodes from config.
Please check this at all times**

### DIFF
--- a/epochStart/metachain/auctionListSelector.go
+++ b/epochStart/metachain/auctionListSelector.go
@@ -198,7 +198,7 @@ func (als *auctionListSelector) SelectNodesFromAuctionList(
 
 	currNodesConfig := als.nodesConfigProvider.GetCurrentNodesConfig()
 	currNumOfValidators := als.stakingDataProvider.GetNumOfValidatorsInCurrentEpoch()
-	numOfShuffledNodes := als.computeNumShuffledNodes(currNodesConfig)
+	numOfShuffledNodes, numForcedToStay := als.computeNumShuffledNodes(currNodesConfig)
 	numOfValidatorsAfterShuffling, err := safeSub(currNumOfValidators, numOfShuffledNodes)
 	if err != nil {
 		log.Warn(fmt.Sprintf("auctionListSelector.SelectNodesFromAuctionList: %v when trying to compute numOfValidatorsAfterShuffling = %v - %v (currNumOfValidators - numOfShuffledNodes)",
@@ -210,12 +210,12 @@ func (als *auctionListSelector) SelectNodesFromAuctionList(
 	}
 
 	maxNumNodes := currNodesConfig.MaxNumNodes
-	availableSlots, err := safeSub(maxNumNodes, numOfValidatorsAfterShuffling)
+	availableSlots, err := safeSub(maxNumNodes, numOfValidatorsAfterShuffling+numForcedToStay)
 	if availableSlots == 0 || err != nil {
-		log.Info(fmt.Sprintf("auctionListSelector.SelectNodesFromAuctionList: %v or zero value when trying to compute availableSlots = %v - %v (maxNodes - numOfValidatorsAfterShuffling); skip selecting nodes from auction list",
+		log.Info(fmt.Sprintf("auctionListSelector.SelectNodesFromAuctionList: %v or zero value when trying to compute availableSlots = %v - %v (maxNodes - numOfValidatorsAfterShuffling+numForcedToStay); skip selecting nodes from auction list",
 			err,
 			maxNumNodes,
-			numOfValidatorsAfterShuffling,
+			numOfValidatorsAfterShuffling+numForcedToStay,
 		))
 		return nil
 	}
@@ -224,9 +224,10 @@ func (als *auctionListSelector) SelectNodesFromAuctionList(
 		"max nodes", maxNumNodes,
 		"current number of validators", currNumOfValidators,
 		"num of nodes which will be shuffled out", numOfShuffledNodes,
+		"num forced to stay", numForcedToStay,
 		"num of validators after shuffling", numOfValidatorsAfterShuffling,
 		"auction list size", auctionListSize,
-		fmt.Sprintf("available slots (%v - %v)", maxNumNodes, numOfValidatorsAfterShuffling), availableSlots,
+		fmt.Sprintf("available slots (%v - %v)", maxNumNodes, numOfValidatorsAfterShuffling+numForcedToStay), availableSlots,
 	)
 
 	als.auctionListDisplayer.DisplayOwnersData(ownersData)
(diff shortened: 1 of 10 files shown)


sasurobert (changes requested):

epochStart/metachain/auctionListSelector.go
```diff
@@ -198,7 +198,7 @@ func (als *auctionListSelector) SelectNodesFromAuctionList(
 
 	currNodesConfig := als.nodesConfigProvider.GetCurrentNodesConfig()
 	currNumOfValidators := als.stakingDataProvider.GetNumOfValidatorsInCurrentEpoch()
-	numOfShuffledNodes := currNodesConfig.NodesToShufflePerShard * (als.shardCoordinator.NumberOfShards() + 1)
+	numOfShuffledNodes, numForcedToStay := als.computeNumShuffledNodes(currNodesConfig)
```
raduchis, line 201: could you add a test with 100 leaving on each shard? so that numOfShuffledNodes = 0?

epochStart/metachain/stakingDataProvider.go
```diff
@@ -532,6 +544,48 @@ func (sdp *stakingDataProvider) GetNumOfValidatorsInCurrentEpoch() uint32 {
...
+	}
+
+	validatorPreviousList := common.PeerType(validator.GetPreviousList())
+	if sdp.isValidatorLeaving(validatorCurrentList, validatorPreviousList) {
+		sdp.validatorStatsInEpoch.Leaving[shardID]++
+	}
+}
+
+func (sdp *stakingDataProvider) isValidatorLeaving(validatorCurrentList, validatorPreviousList common.PeerType) bool {
+	if validatorCurrentList != common.LeavingList {
```
raduchis, line 568: the forcedToStay validators in the current epoch for the next epoch their list is updated to Eligible or Waiting right?

mariusmihaic: I checked an integration tests I did superficially for forced to stay nodes and checked with debug what happens in that case.

For every forced to stay epoch, those nodes will allways have:
- Current list = Leaving
- Previous List = Eligible/Waiting

So I guess it is correct, right?

testscommon/stakingcommon/stakingDataProviderStub.go
```diff
@@ -88,6 +88,14 @@ func (sdps *StakingDataProviderStub) GetNumOfValidatorsInCurrentEpoch() uint32 {
 	return 0
 }
 
+func (sdps *StakingDataProviderStub) GetCurrentEpochValidatorStats() epochStart.ValidatorStatsInEpoch {
```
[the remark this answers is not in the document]
mariusmihaic: Added comm

epochStart/metachain/stakingDataProvider.go
```diff
@@ -532,6 +544,48 @@ func (sdp *stakingDataProvider) GetNumOfValidatorsInCurrentEpoch() uint32 {
...
+		return false
+	}
+
+	// If no previous list is set, means that staking v4 is not activated or node is leaving right before activation
+	// and this node will be considered as eligible by the nodes coordinator with legacy bug.
```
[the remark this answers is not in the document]
mariusmihaic: Refactored

epochStart/metachain/stakingDataProvider.go
```diff
@@ -532,6 +544,48 @@ func (sdp *stakingDataProvider) GetNumOfValidatorsInCurrentEpoch() uint32 {
...
+	// Otherwise, it will have it set, and we should check its previous list in the current epoch
+	if len(validatorPreviousList) == 0 || validatorPreviousList == common.EligibleList || validatorPreviousList == common.WaitingList {
```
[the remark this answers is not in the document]
mariusmihaic: Right, good suggestion

epochStart/metachain/stakingDataProvider.go
raduchis: in the next epoch, do they have the CurrentList=Leaving? or the owner has to "unstake" them again because they are eligible/waiting?

epochStart/metachain/auctionListSelector.go
mariusmihaic: Added test for that + also unstake 80 nodes

sasurobert (approved):

epochStart/metachain/stakingDataProvider.go
mariusmihaic: Nodes that are forced to stay will remain in system no matter what the owner does.
As I've said: 
>For every forced to stay epoch, those nodes will allways have:
> - Current list = Leaving
> - Previous List = Eligible/Waiting

sasurobert (approved):

epochStart/metachain/auctionListSelector.go
raduchis: ok

raduchis (approved):

epochStart/metachain/stakingDataProvider.go
raduchis: ok

AdoAdoAdo (approved):

gabi-vuls (approved):

PR #6118, opened 2024-04-15
Relayed v3 with multiple transactions

sstanculeanu:
- Relayed v3 should be able to support multiple inner transactions
  
## Proposed changes
- Added support for multiple inner transactions for relayed v3(senders should be in the same shard with the relayer)
- Added proper fix for the relayed move balance fee
- Added integration test with chain simulator for relayed v3

## Testing procedure
- Will be tested with feat branch

## Pre-requisites


process/transaction/baseProcess.go
```diff
@@ -146,7 +146,11 @@ func (txProc *baseTxProcessor) checkTxValues(
 			return process.ErrNotEnoughGasInUserTx
 		}
 		if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedMoveBalanceFlag) {
-			txFee = txProc.economicsFee.ComputeTxFee(tx)
+			moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(tx)
+			gasToUse := tx.GetGasLimit() - moveBalanceGasLimit
+			moveBalanceUserFee := txProc.economicsFee.ComputeMoveBalanceFee(tx)
```
sasurobert, line 151: this will further increase the cost of relayed txs as they are now. 

It can stay, but for relayed v1 and v2 the txData is computed twice. Once in the relayer and once when processing the underlying tx.

process/transaction/shardProcess.go
```diff
@@ -298,7 +304,14 @@ func (txProc *txProcessor) executingFailedTransaction(
 		return nil
 	}
 
-	txFee := txProc.economicsFee.ComputeTxFee(tx)
+	txFee := txProc.economicsFee.ComputeFeeForProcessing(tx, tx.GasLimit)
+	if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedMoveBalanceFlag) {
```
sasurobert, line 308: move the common code and use a common function. This is the exactly same lines of codes as in baseProcess.

process/transaction/shardProcess.go
```diff
@@ -391,7 +404,11 @@ func (txProc *txProcessor) processTxFee(
 	if isUserTxOfRelayed {
 		totalCost := txProc.economicsFee.ComputeFeeForProcessing(tx, tx.GasLimit)
 		if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedMoveBalanceFlag) {
-			totalCost = txProc.economicsFee.ComputeTxFee(tx)
+			moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(tx)
+			gasToUse := tx.GetGasLimit() - moveBalanceGasLimit
```
sasurobert, line 408: duplicated code again.

process/transaction/shardProcess.go
```diff
@@ -580,41 +597,41 @@ func (txProc *txProcessor) finishExecutionOfRelayedTx(
...
+		txHash, err := core.CalculateHash(txProc.marshalizer, txProc.hasher, tx)
+		if err != nil {
+			return err
 		}
 
 		txProc.txFeeHandler.ProcessTransactionFee(relayerFee, big.NewInt(0), txHash)
 	}
 
-	return txHash, nil
+	return nil
```
sasurobert, line 634: txHash was returned in order to not calculate that too many times. Actually we should create a task where we calculate txHash only once in the lifetime of a tx.

sstanculeanu: created new task [MX-15423](https://multiversxlabs.atlassian.net/browse/MX-15423)

api/groups/transactionGroup.go
```diff
@@ -387,23 +401,28 @@ func (tg *transactionGroup) sendMultipleTransactions(c *gin.Context) {
...
+		if len(receivedTx.InnerTransactions) != 0 {
+			for _, innerTx := range receivedTx.InnerTransactions {
+				if len(innerTx.InnerTransactions) != 0 {
+					// if one of the inner txs is invalid, break the loop and move to the next transaction received
+					break
+				}
+
+				newInnerTx, _, err := tg.createTransaction(innerTx, nil)
+				if err != nil {
+					// if one of the inner txs is invalid, break the loop and move to the next transaction received
```
sasurobert, line 414: why would you send the tx further if the inner tx is wrong? Why not error?

process/transaction/shardProcess.go
```diff
@@ -633,34 +642,62 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+		}
+
+		executedUserTxs = append(executedUserTxs, innerTx)
 	}
-	if !bytes.Equal(userTx.RelayerAddr, tx.SndAddr) {
-		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3RelayerMismatch)
+
+	allUserTxsSucceeded := len(executedUserTxs) == len(innerTxs) && innerTxErr == nil && innerTxRetCode == vmcommon.Ok
+	if !allUserTxsSucceeded {
+		log.Debug("failed to execute all inner transactions", "total", len(innerTxs), "executed transactions", len(executedUserTxs))
```
sasurobert, line 681: we do not need information like this in log.Debug - maybe in trace.

sasurobert (changes requested):

api/groups/transactionGroup.go
sstanculeanu: right, updated

process/transaction/baseProcess.go
```diff
@@ -176,6 +176,16 @@ func (txProc *baseTxProcessor) checkTxValues(
 	return nil
 }
 
+func (txProc *baseTxProcessor) computeTxFeeAfterMoveBalanceFix(tx *transaction.Transaction) *big.Int {
+	moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(tx)
```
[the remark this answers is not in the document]
sstanculeanu: updated

process/transaction/shardProcess.go
```diff
@@ -963,10 +1000,15 @@ func (txProc *txProcessor) makeSCRFromUserTx(
 	relayerAdr []byte,
 	relayedTxValue *big.Int,
 	txHash []byte,
+	isRevertSCR bool,
 ) (*smartContractResult.SmartContractResult, error) {
+	scrValue := tx.Value
+	if isRevertSCR {
+		scrValue = big.NewInt(0).Neg(tx.Value)
```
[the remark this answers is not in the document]
sstanculeanu: indeed not needed, I forgot to remove it after the last discussion

process/transaction/relayedTxV3Processor.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return nil
+}
+
+// ComputeRelayedTxFees returns the both the total fee for the entire relayed tx and the relayed only fee
+func (proc *relayedTxV3Processor) ComputeRelayedTxFees(tx *transaction.Transaction) (*big.Int, *big.Int) {
+	relayerMoveBalanceFee := proc.economicsFee.ComputeMoveBalanceFee(tx)
+	uniqueSenders := proc.GetUniqueSendersRequiredFeesMap(tx.InnerTransactions)
+
+	relayerFee := big.NewInt(0).Mul(relayerMoveBalanceFee, big.NewInt(int64(len(uniqueSenders))))
```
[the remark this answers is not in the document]
sstanculeanu: considered this a cost optimization, although it makes more sense to keep each tx independent including the move balance fee.. will update as suggested

process/transaction/shardProcess.go
```diff
@@ -1071,6 +1120,76 @@ func isNonExecutableError(executionErr error) bool {
...
+		return make(map[state.UserAccountHandler]*big.Int), err
+	}
+
+	uniqueSendersMap := txProc.relayedTxV3Processor.GetUniqueSendersRequiredFeesMap(tx.InnerTransactions)
+	uniqueSendersSlice := mapToSlice(uniqueSendersMap)
+	sendersBalancesSnapshot := make(map[state.UserAccountHandler]*big.Int, len(uniqueSendersMap))
+	var lastTransferErr error
+	for _, uniqueSender := range uniqueSendersSlice {
+		totalFeesForSender := uniqueSendersMap[uniqueSender]
+		senderAcnt, prevBalanceForSender, err := txProc.addFeesToDest([]byte(uniqueSender), totalFeesForSender)
```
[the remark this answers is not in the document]
sstanculeanu: considered this a cost optimization.. will update in order to treat each tx independent, thus no need to add the fees to maps anymore

process/transaction/shardProcess.go
```diff
@@ -633,34 +642,62 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+) (vmcommon.ReturnCode, error) {
+	acntSnd, err := txProc.getAccountFromAddress(innerTx.SndAddr)
+	if err != nil {
+		return vmcommon.UserError, err
 	}
-	remainingGasLimit := tx.GasLimit - txProc.economicsFee.ComputeGasLimit(tx)
-	if userTx.GasLimit != remainingGasLimit {
-		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3GasLimitMismatch)
+
+	if check.IfNil(acntSnd) {
```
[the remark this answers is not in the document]
sstanculeanu: added return error as suggested

this check already done at interceptor level, so this check here should never ever be reached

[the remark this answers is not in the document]
sasurobert: that is true. But it is ok to check here as well and return error.

process/transaction/shardProcess.go
```diff
@@ -633,34 +637,68 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
-	if len(userTx.RelayerAddr) == 0 {
-		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3EmptyRelayer)
+
+	innerTxs := tx.GetInnerTransactions()
+
+	var innerTxRetCode vmcommon.ReturnCode
+	var innerTxErr error
+	executedUserTxs := make([]*transaction.Transaction, 0)
+	for _, innerTx := range innerTxs {
+		innerTxRetCode, innerTxErr = txProc.finishExecutionOfInnerTx(tx, innerTx)
```
sasurobert, line 666: maybe a better name is processInnerTx

process/transaction/shardProcess.go
```diff
@@ -633,34 +637,68 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
-	if userTx.GasLimit != remainingGasLimit {
-		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3GasLimitMismatch)
+
+	if check.IfNil(acntSnd) {
+		return vmcommon.UserError, process.ErrRelayedTxV3SenderShardMismatch
 	}
 
-	return txProc.finishExecutionOfRelayedTx(relayerAcnt, acntDst, tx, userTx)
+	txFee := txProc.computeTxFee(innerTx)
+	err = txProc.addFeeAndValueToDest(acntSnd, tx, txFee)
```
sasurobert, line 696: is tx.value == big.NewInt(0) ? I think it would be better to have a localTX here, where we have value at 0 clearly.

sasurobert, line 696: you can change the addFeeAndValueToDest function and have 2 bigInts to be received there. it is much more clear. and here you could have big.newint(0), txFee.

process/transaction/shardProcess.go
```diff
@@ -633,34 +637,68 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
-	remainingGasLimit := tx.GasLimit - txProc.economicsFee.ComputeGasLimit(tx)
-	if userTx.GasLimit != remainingGasLimit {
-		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrRelayedTxV3GasLimitMismatch)
+
+	if check.IfNil(acntSnd) {
+		return vmcommon.UserError, process.ErrRelayedTxV3SenderShardMismatch
 	}
 
-	return txProc.finishExecutionOfRelayedTx(relayerAcnt, acntDst, tx, userTx)
+	txFee := txProc.computeTxFee(innerTx)
```
sasurobert, line 695: we need another check - inveriant check Sum of TXFee == ComputeRelayedFee. you can sum all the txFees you have here, and at the end of processing to check it.

process/transaction/shardProcess.go
```diff
@@ -633,34 +637,68 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+
+	innerTxs := tx.GetInnerTransactions()
+
+	var innerTxRetCode vmcommon.ReturnCode
+	var innerTxErr error
+	executedUserTxs := make([]*transaction.Transaction, 0)
+	for _, innerTx := range innerTxs {
+		innerTxRetCode, innerTxErr = txProc.finishExecutionOfInnerTx(tx, innerTx)
+		if innerTxErr != nil || innerTxRetCode != vmcommon.Ok {
+			continue
```
sasurobert, line 668: we should create a logEvent for failed innerTX, for those cases which is not caught by processUserTX - I think when the code reaches processUserTX a logEvent is done at least for that.

process/transaction/shardProcess.go
```diff
@@ -633,34 +637,68 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+		}
+
+		executedUserTxs = append(executedUserTxs, innerTx)
```
sasurobert, line 671: we have to check whether an additional logEvent would be needed here - I think it is better to have that innerTX was executed.

process/transaction/shardProcess.go
sstanculeanu: we do have an early check for the tx value to be 0, but the suggestion from the next comment with the specific parameter should clarify.. applied it

process/transaction/shardProcess.go
sstanculeanu: that's true, when the code reaches processUserTX it generates a logEvent.. will add one in case the processing fails before that point

process/transaction/shardProcess.go
sstanculeanu: we do have a SCR. do we also a logEvent?

process/transaction/shardProcess.go
sstanculeanu: added check as discussed :pray:

process/transaction/shardProcess.go
sstanculeanu: now I understand.. it would make sense to also have events for the succeeded inner txs

sstanculeanu: added in the next pr :pray:

process/transaction/shardProcess.go
```diff
@@ -1018,15 +1115,22 @@ func (txProc *txProcessor) executeFailedRelayedUserTx(
 	}
 
 	totalFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, userTx.GasLimit)
+	moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(userTx)
+	gasToUse := userTx.GetGasLimit() - moveBalanceGasLimit
+	processingUserFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, gasToUse)
 	if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedMoveBalanceFlag) {
-		totalFee = txProc.economicsFee.ComputeTxFee(userTx)
+		moveBalanceUserFee := txProc.economicsFee.ComputeMoveBalanceFee(userTx)
```
sasurobert, line 1122: can't you reuse the common code for this?

sstanculeanu: some of the above variables are used in further ifs

sasurobert (dismissed):

AdoAdoAdo (changes requested):

api/groups/transactionGroup.go
```diff
@@ -387,23 +401,36 @@ func (tg *transactionGroup) sendMultipleTransactions(c *gin.Context) {
...
-				c.JSON(
-					http.StatusBadRequest,
-					shared.GenericAPIResponse{
-						Data:  nil,
-						Error: fmt.Sprintf("%s: %s", errors.ErrTxGenerationFailed.Error(), err.Error()),
-						Code:  shared.ReturnCodeRequestError,
-					},
-				)
-				return
+		innerTxs := make([]*transaction.Transaction, 0, len(receivedTx.InnerTransactions))
```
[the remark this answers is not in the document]
sstanculeanu: updated as suggested and reused the code

factory/processing/processComponents.go
```diff
@@ -377,8 +379,18 @@ func (pcf *processComponentsFactory) Create() (*processComponents, error) {
...
+		MaxTransactionsAllowed: pcf.config.RelayedTransactionConfig.MaxTransactionsAllowed,
+	}
+	relayedTxV3Processor, err := transaction.NewRelayedTxV3Processor(argsRelayedTxV3Processor)
+	if err != nil {
+		return nil, err
+	}
+
 	pcf.txLogsProcessor = txLogsProcessor
-	genesisBlocks, initialTxs, err := pcf.generateGenesisHeadersAndApplyInitialBalances()
+	genesisBlocks, initialTxs, err := pcf.generateGenesisHeadersAndApplyInitialBalances(relayedTxV3Processor)
```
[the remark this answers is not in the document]
sstanculeanu: right, removed it and passed a disabled component where needed

integrationTests/multiShard/relayedTx/common.go
```diff
@@ -15,6 +15,25 @@ import (
...
+	initialVal := big.NewInt(10000000000)
+	nodes, idxProposers := createAndMintNodes(initialVal)
+
+	players, relayerAccount := createAndMintPlayers(false, nodes, initialVal)
+
+	return nodes, idxProposers, players, relayerAccount
+}
+
+// CreateGeneralSetupForRelayedV3TxTest will create the general setup for relayed transactions v3
+func CreateGeneralSetupForRelayedV3TxTest() ([]*integrationTests.TestProcessorNode, []int, []*integrationTests.TestWalletAccount, *integrationTests.TestWalletAccount) {
```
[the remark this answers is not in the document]
sstanculeanu: updated

process/transaction/relayedTxV3Processor.go
```diff
@@ -0,0 +1,133 @@
...
+	if !bytes.Equal(tx.RcvAddr, tx.SndAddr) {
+		return process.ErrRelayedTxV3SenderDoesNotMatchReceiver
+	}
+	if tx.GasLimit < proc.computeRelayedTxMinGasLimit(tx) {
+		return process.ErrRelayedTxV3GasLimitMismatch
+	}
+
+	innerTxs := tx.InnerTransactions
+	for _, innerTx := range innerTxs {
+		if len(innerTx.RelayerAddr) == 0 {
```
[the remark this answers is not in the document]
sstanculeanu: true, should be safe to only keep the one on L76.. updated

process/transaction/relayedTxV3Processor.go
```diff
@@ -0,0 +1,133 @@
...
+			return process.ErrRelayedTxV3EmptyRelayer
+		}
+		if !bytes.Equal(innerTx.RelayerAddr, tx.SndAddr) {
+			return process.ErrRelayedTxV3RelayerMismatch
+		}
+		if tx.GasPrice != innerTx.GasPrice {
+			return process.ErrRelayedV3GasPriceMismatch
+		}
+
```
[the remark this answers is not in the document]
sstanculeanu: added check

process/transaction/relayedTxV3Processor.go
```diff
@@ -0,0 +1,133 @@
...
+		totalFees.Add(totalFees, innerTxFee)
+	}
+
+	return totalFees
+}
+
+func (proc *relayedTxV3Processor) computeRelayedTxMinGasLimit(tx *transaction.Transaction) uint64 {
+	relayedTxGasLimit := proc.economicsFee.ComputeGasLimit(tx)
+
+	totalGasLimit := relayedTxGasLimit * uint64(len(tx.InnerTransactions))
```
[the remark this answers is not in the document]
sstanculeanu: added check for data field to be empty. this way, we also avoid the case when a transaction could be both v3 and v1/v2 in the same time(special checks were added in the next PR, but with data field empty requirement, checks are not needed for v3)

very good point with the guarded relayed tx, implemented the suggestion with the min gas limit

process/transaction/relayedTxV3Processor.go
```diff
@@ -0,0 +1,133 @@
...
+		if senderShard != relayerShard {
+			return process.ErrRelayedTxV3SenderShardMismatch
+		}
+	}
+
+	return nil
+}
+
+// ComputeRelayedTxFees returns the both the total fee for the entire relayed tx and the relayed only fee
+func (proc *relayedTxV3Processor) ComputeRelayedTxFees(tx *transaction.Transaction) (*big.Int, *big.Int) {
```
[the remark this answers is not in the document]
sstanculeanu: updated + added unittest

process/transaction/shardProcess.go
```diff
@@ -298,7 +304,7 @@ func (txProc *txProcessor) executingFailedTransaction(
 		return nil
 	}
 
-	txFee := txProc.economicsFee.ComputeTxFee(tx)
+	txFee := txProc.computeTxFee(tx)
```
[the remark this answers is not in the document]
sstanculeanu: true, this should have not been modified.. reverted it

integrationTests/vm/txsFee/relayedScDeploy_test.go
```diff
@@ -70,7 +70,7 @@ func TestRelayedScDeployInvalidCodeShouldConsumeGas(t *testing.T) {
 		t.Skip("this is not a short test")
 	}
 
-	t.Run("before relayed fix", testRelayedScDeployInvalidCodeShouldConsumeGas(integrationTests.UnreachableEpoch, big.NewInt(17030), big.NewInt(32970)))
+	t.Run("before relayed fix", testRelayedScDeployInvalidCodeShouldConsumeGas(integrationTests.UnreachableEpoch, big.NewInt(8890), big.NewInt(41110)))
```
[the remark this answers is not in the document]
sstanculeanu: the fix should only have effects on txs with move balance move balance.

for this specific test, both before and after should have the same results, as the code of the contract is invalid. As part of the fix, I've also fixed these tests which were wrong. On current rc branch, [here](https://github.com/multiversx/mx-chain-go/blob/rc/v1.7.next1/integrationTests/vm/txsFee/relayedScDeploy_test.go#L59), we can see that the gas limit of the relayed tx was set to 500, [here](https://github.com/multiversx/mx-chain-go/blob/rc/v1.7.next1/integrationTests/vm/txsFee/relayedScDeploy_test.go#L73), which was too little thus the gas consumed was completely different. The actual error was not the one for invalid contract code, but for insufficient gas.

This PR is targeted towards the feat/relayedV3 branch, which already has the first version of the fix for the move balance cost, which was not the good one, thus the values changed.

Will add an integration test with chain simulator and price modifier on relayed v1 with move balance, in order to confirm the gas consumed fix :pray:

process/transaction/shardProcess.go
```diff
@@ -832,20 +935,14 @@ func (txProc *txProcessor) processUserTx(
 	userTx *transaction.Transaction,
 	relayedTxValue *big.Int,
 	relayedNonce uint64,
-	txHash []byte,
+	originalTxHash []byte,
 ) (vmcommon.ReturnCode, error) {
 
 	acntSnd, acntDst, err := txProc.getAccounts(userTx.SndAddr, userTx.RcvAddr)
```
AdoAdoAdo, line 941: is it possible that during the execution of relayed tx we reach this point and we can get an error here?
if so we need extra cleanup.

process/transaction/shardProcess.go
```diff
@@ -633,34 +654,117 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+			innerTx.RelayerAddr,
+			big.NewInt(0),
+			tx.Nonce,
+			tx,
+			originalTxHash,
+			process.ErrRelayedTxV3SenderShardMismatch.Error())
 	}
 
-	return txProc.finishExecutionOfRelayedTx(relayerAcnt, acntDst, tx, userTx)
+	err = txProc.addFeeAndValueToDest(acntSnd, big.NewInt(0), txFee)
```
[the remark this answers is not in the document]
sstanculeanu: added TODO

process/transaction/shardProcess.go
sstanculeanu: handled the situation. should not happen though

sasurobert (approved):

AdoAdoAdo (approved):

PR #6121, opened 2024-04-18
[sovereign] Fix non payable by sc cross shard

mariusmihaic:
- When checking the payable property for a "cross" shard (meta chain) system account(esdt, delegation, etc.)  for calls between scs -> sys accounts, the cross check in `BlockChainHook` would fail to detect it as "cross shard". This caused reading meta data properties for system accounts for payable byte, which is by default false for all of them.
## Proposed changes
- "Faked" cross shard(to meta sys accounts) checks in blockchain hook.
- Newly created func for sovereign blockchain hook `isCrossShardForPayableCheck` should be equivalent to `isNotSystemAccountAndCrossShard` from normal blockchain hook

## Testing procedure
- Tested on a separate branch with chain simulator.

## Pre-requisites

### DIFF
--- a/process/smartContract/hooks/blockChainHook.go
+++ b/process/smartContract/hooks/blockChainHook.go
@@ -556,7 +556,11 @@ func (bh *BlockChainHookImpl) IsPayable(sndAddress []byte, recvAddress []byte) (
 		return true, nil
 	}
 
-	userAcc, err := bh.GetUserAccount(recvAddress)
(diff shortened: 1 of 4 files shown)


sasurobert (changes requested):
Reject. That is not the clear solution.

mariusmihaic:
This is the correct solution: https://github.com/multiversx/mx-chain-go/pull/6129

PR #6123, opened 2024-04-19
add withKeys option on account

raduchis:
- do one single request on /address/:address?withKeys=true, so that also keys are returned together with the account information.
  
## Proposed changes
- add withKeys option

## Testing procedure
- #allIn

## Pre-requisites


facade/nodeFacade.go
```diff
@@ -358,9 +364,11 @@ func (nf *nodeFacade) GetAccounts(addresses []string, options apiData.AccountQue
 
 	response := make(map[string]*apiData.AccountResponse)
 	var blockInfo apiData.BlockInfo
+	ctx, cancel := nf.getContextForApiTrieRangeOperations()
+	defer cancel()
 
 	for _, address := range addresses {
-		accountResponse, blockInfoForAccount, err := nf.node.GetAccount(address, options)
+		accountResponse, blockInfoForAccount, err := nf.node.GetAccount(address, options, ctx)
```
AdoAdoAdo, line 371: I think we have an old issue here

seems that here we could return accounts with different block info, but in the end we reply with the latest block info, which might not be correct.

Could you request with the same roothash for all accounts?

facade/nodeFacade.go
```diff
@@ -336,7 +339,10 @@ func (nf *nodeFacade) ComputeTransactionGasLimit(tx *transaction.Transaction) (*
 
 // GetAccount returns a response containing information about the account correlated with provided address
 func (nf *nodeFacade) GetAccount(address string, options apiData.AccountQueryOptions) (apiData.AccountResponse, apiData.BlockInfo, error) {
-	accountResponse, blockInfo, err := nf.node.GetAccount(address, options)
+	ctx, cancel := nf.getContextForApiTrieRangeOperations()
```
AdoAdoAdo, line 342: could we do the getContextForApiTrieRangeOperations only if withKeys option is set?

facade/nodeFacade.go
```diff
@@ -321,7 +321,10 @@ func (nf *nodeFacade) GetLastPoolNonceForSender(sender string) (uint64, error) {
 
 // GetTransactionsPoolNonceGapsForSender will return the nonce gaps from pool for sender, if exists, that is to be returned on API calls
 func (nf *nodeFacade) GetTransactionsPoolNonceGapsForSender(sender string) (*common.TransactionsPoolNonceGapsForSenderApiResponse, error) {
-	accountResponse, _, err := nf.node.GetAccount(sender, apiData.AccountQueryOptions{})
+	ctx, cancel := nf.getContextForApiTrieRangeOperations()
```
AdoAdoAdo, line 324: I think here we can remove the getContextForApiTrieRangeOperations as this call does not support options.

facade/nodeFacade.go
raduchis: yes, changed it

node/node.go
```diff
@@ -291,13 +291,26 @@ func (n *Node) GetKeyValuePairs(address string, options api.AccountQueryOptions,
 		return map[string]string{}, api.BlockInfo{}, nil
 	}
 
+	mapToReturn, err := n.getKeys(userAccount, ctx)
+	if err != nil {
+		return nil, api.BlockInfo{}, err
+	}
+
+	if common.IsContextDone(ctx) {
```
mariusmihaic, line 299: Can this ever happen with `error == nil` from above when iterating? Or just in an edge case of micro seconds?

node/node.go
```diff
@@ -954,6 +962,11 @@ func (n *Node) GetAccount(address string, options api.AccountQueryOptions) (api.
 		}
 	}
 
+	var keys map[string]string
+	if options.WithKeys {
+		keys, _ = n.getKeys(account, ctx)
```
mariusmihaic, line 967: Why no error check?

node/node_test.go
```diff
@@ -3447,6 +3447,75 @@ func TestNode_GetAccountAccountExistsShouldReturn(t *testing.T) {
...
+		HistoricalStateAccountsWrapper: accDB,
+	}
+	stateComponents.AccountsRepo, _ = state.NewAccountsRepository(args)
+	n, _ := node.NewNode(
+		node.WithCoreComponents(coreComponents),
+		node.WithDataComponents(dataComponents),
+		node.WithStateComponents(stateComponents),
+	)
+
+	recovAccnt, _, err := n.GetAccount(testscommon.TestAddressBob, api.AccountQueryOptions{WithKeys: false}, context.Background())
```
mariusmihaic, line 3506: Would it make sens to test the Getters in parallel here on different go routines?

facade/nodeFacade.go
raduchis: refactored to have a method GetAccountWithKeys

facade/nodeFacade.go
raduchis: good observation: refactored to have GetAccountWithKeys

node/node.go
raduchis: added err check.

node/node_test.go
raduchis: do you mean to try to have multiple requests and check the result from different blocks?

facade/interface.go
```diff
@@ -74,7 +74,11 @@ type NodeHandler interface {
 
 	// GetAccount returns an accountResponse containing information
 	//  about the account correlated with provided address
-	GetAccount(address string, options api.AccountQueryOptions, ctx context.Context) (api.AccountResponse, api.BlockInfo, error)
+	GetAccount(address string, options api.AccountQueryOptions) (api.AccountResponse, api.BlockInfo, error)
+
+	// GetAccountWithKeys returns an accountResponse containing information
+	//  about the account correlated with provided address and all keys
```
mariusmihaic, line 80:
```suggestion
	// about the account correlated with provided address and all keys
```

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.2.5
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/multiversx/mx-chain-communication-go v1.0.14
-	github.com/multiversx/mx-chain-core-go v1.2.20-0.20240418121049-6970013b49d9
+	github.com/multiversx/mx-chain-core-go v1.2.20
```
mariusmihaic, line 18: Is it ok that you changed this?

node/node.go
```diff
@@ -938,7 +938,45 @@ func (n *Node) setTxGuardianData(guardian string, guardianSigHex string, tx *tra
 }
 
 // GetAccount will return account details for a given address
-func (n *Node) GetAccount(address string, options api.AccountQueryOptions, ctx context.Context) (api.AccountResponse, api.BlockInfo, error) {
+func (n *Node) GetAccount(address string, options api.AccountQueryOptions) (api.AccountResponse, api.BlockInfo, error) {
+	account, blockInfo, err := n.loadUserAccountHandlerByAddress(address, options)
```
mariusmihaic, line 942: This first part of the code is duplicated from `GetAccount` -> `GetAccountWithKeys`.
I think we should extract it to a common base function

node/node_test.go
```diff
@@ -3447,6 +3447,51 @@ func TestNode_GetAccountAccountExistsShouldReturn(t *testing.T) {
 	require.Equal(t, testscommon.TestAddressAlice, recovAccnt.OwnerAddress)
 }
 
+func TestNode_GetAccountAccountWithKeysErrorShouldErr(t *testing.T) {
```
mariusmihaic, line 3450: Is this test name ok?

node/node_test.go
```diff
@@ -3447,6 +3447,121 @@ func TestNode_GetAccountAccountExistsShouldReturn(t *testing.T) {
...
+	)
+
+	recovAccnt, blockInfo, err := n.GetAccountWithKeys(testscommon.TestAddressBob, api.AccountQueryOptions{WithKeys: true}, context.Background())
+
+	require.Equal(t, expectedErr, err)
+	require.Equal(t, api.AccountResponse{}, recovAccnt)
+	require.Equal(t, api.BlockInfo{}, blockInfo)
+}
+
+func TestNode_GetAccountAccountWithKeysShouldReturn(t *testing.T) {
```
mariusmihaic, line 3495: I think test name is incomplete: Should return what?

node/node_test.go
```diff
@@ -3447,6 +3447,121 @@ func TestNode_GetAccountAccountExistsShouldReturn(t *testing.T) {
...
+	accDB := &stateMock.AccountsStub{
+		GetAccountWithBlockInfoCalled: func(address []byte, options common.RootHashHolder) (vmcommon.AccountHandler, common.BlockInfo, error) {
+			return accnt, nil, nil
+		},
+		RecreateTrieCalled: func(rootHash []byte) error {
+			return nil
+		},
+	}
+
+	coreComponents := getDefaultCoreComponents()
```
mariusmihaic, line 3473: This code part is duplicated with the one from `TestNode_GetAccountAccountWithKeysShouldReturn`. 
Perhaps we can extract it to a common func

mariusmihaic (changes requested):

node/node_test.go
mariusmihaic: Yes, or from the same block, just making parallel requests.
If it doesn't make sense or it's already being tested somewhere else, we can skip it

go.mod
raduchis: yes, it is the actual release

facade/interface.go
raduchis: changed

node/node.go
raduchis: fixed

node/node_test.go
raduchis: rename to TestNode_GetAccountAccountWithKeysErrorShouldFail

node/node_test.go
raduchis: renamed to TestNode_GetAccountAccountWithKeysShouldWork
Copy-Pasted from above tests

node/node_test.go
raduchis: fixed

mariusmihaic (approved):

AdoAdoAdo (approved):

gabi-vuls (approved):
Normal allin test: v1.6.15-dev-config-eb2e06c06d -> MX-15381-add-withKeys-on-a-74c90c5230

--- Specific errors ---

block hash does not match 1725
wrong nonce in block 679
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  2
Nr. of all WARNS:  498
Nr. of new ERRORS:  1
Nr. of new WARNS:  5
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

PR #6129, opened 2024-04-22
[sovereign] Add system sc code data

mariusmihaic:
- Fix unknown async call error when a sc tries to call a meta chain sc address(like esdt issue)
## Proposed changes
- Reused init func for esdt sys accounts from epoch change flow
- Fixed sovereign blockchain pointers in run type components 

## Testing procedure
- 

## Pre-requisites


sasurobert (approved):
this is not a non payable fix - but add systemSC code data on genesis on Sovereign.

axenteoctavian (approved):

PR #6143, opened 2024-04-26
[sovereign] Add factory for NodesSetupChecker to override it easily.

raress96:

We need to override the NodesSetupChecker easily for a Sovereign Chain from the `sovereignNodeRunner.go` file. This way we can have `mx-chain-go` as a dependency of a Sovereign Chain GO project and not have to fork the code.

Currently this can be overwritten by implementing a custom `managedProcessComponents` and `processComponentsFactory`, but the `processComponentsFactory` `Create` method has a lot of logic that we would not want to maintain, hence we need an easier way to override the NodesSetupChecker
  
## Proposed changes
Added a NodesSetupCheckerFactory and a NodeSetupChecker interface to allow easier customization.

## Testing procedure
- 

## Pre-requisites


raress96:
Closing this since it is outdated.

PR #6176, opened 2024-05-16
Append log events for all inner transactions failure

sstanculeanu:
- only the last failed transaction was seen in the log events of a relayed v3 with multiple failed inner txs, because the original tx has was used
  
## Proposed changes
- now logs processor appends new logs to the old ones

## Testing procedure

## Pre-requisites


process/transaction/shardProcess.go
```diff
@@ -662,7 +662,12 @@ func (txProc *txProcessor) processRelayedTxV3(
 	if check.IfNil(relayerAcnt) {
 		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAcnt, process.ErrNilRelayerAccount)
 	}
-	err := txProc.relayedTxV3Processor.CheckRelayedTx(tx)
+	funcName, _, err := txProc.argsParser.ParseCallData(string(tx.Data))
```
sasurobert, line 665: can't you add this check inside the relayedTxV3Process.CheckRelayedTX function?

process/transaction/shardProcess.go
```diff
@@ -970,6 +982,10 @@ func (txProc *txProcessor) processUserTx(
 	switch txType {
 	case process.MoveBalance:
 		err = txProc.processMoveBalance(userTx, acntSnd, acntDst, dstShardTxType, originalTxHash, true)
+		isUserTxOfRelayedV3 := len(originalTx.InnerTransactions) > 0
```
sasurobert, line 985: this if can be deleted and add a complete event log for each such processing.

sasurobert: But actually the completeEventLog should be done only if destination shard is the current shard.

otherwise you may have execution on the destination shard and create another log event.

so add log event is this is intra shard only. otherwise, completeLogEvent is added by scProcessor. plus add an integration test and verify how many completeLogEvents are generated.

process/transactionLog/process.go
```diff
@@ -161,25 +162,65 @@ func (tlp *txLogProcessor) SaveLog(txHash []byte, tx data.TransactionHandler, lo
...
+	err := tlp.marshalizer.Unmarshal(oldLogs, oldLogsBuff)
 	if err != nil {
 		return err
 	}
 
-	return tlp.storer.Put(txHash, buff)
+	if oldLogs.Address == nil {
+		oldLogs.Address = newLog.Address
+	}
+	oldLogs.Events = append(oldLogs.Events, newLog.Events...)
```
sasurobert, line 196: this is super inefficient. until now for a hash we added the logEvent only once when all the execution was completed. Please do that here as well. 

In the current format this PR is not ok.

sasurobert (changes requested):

process/transactionLog/process.go
sstanculeanu: pushed

updated the flow a bit so this only happens for failed inner intra shard transactions. The append functionality is still needed, as for failed intra shard  inner transactions we don't have scrs, but only events on the big relayed tx(key = relayed tx hash).

sasurobert: but you could accumulate the logs at the transaction processor level and push here once.

sstanculeanu: pushed.. added a new component able to accumulate all logs during execution, then at the end only one SaveLog is done

also removed AppendLogs functionality

sasurobert (approved):

process/transaction/interceptedTransaction.go
```diff
@@ -314,6 +323,10 @@ func (inTx *InterceptedTransaction) verifyIfRelayedTx(tx *transaction.Transactio
 		return process.ErrInvalidArguments
 	}
 
+	if len(tx.InnerTransactions) > 0 {
```
ssd04, line 326: should we check here also for relayed v2 not allowed? or is this already checked?

process/transaction/shardProcess.go
```diff
@@ -1187,6 +1208,36 @@ func isNonExecutableError(executionErr error) bool {
 		errors.Is(executionErr, process.ErrTransactionNotExecutable)
 }
 
+func (txProc *txProcessor) createCompleteEventLog(scr data.TransactionHandler, originalTxHash []byte) {
+	scrHash, err := core.CalculateHash(txProc.marshalizer, txProc.hasher, scr)
+	if err != nil {
+		scrHash = originalTxHash
```
ssd04, line 1214: do we need a log message for this error?

process/transactionLog/failedTxLogsAccumulator.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+type logData struct {
+	tx   data.TransactionHandler
+	logs []*vmcommon.LogEntry
+}
+
+type failedTxLogsAccumulator struct {
+	mut     sync.RWMutex
+	logsMap map[string]*logData
```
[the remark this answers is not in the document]
ssd04: maybe evaluating using LRU mechanism to cleanup old entries

process/transactionLog/failedTxLogsAccumulator.go
```diff
@@ -0,0 +1,109 @@
...
+		tx: logsData.tx,
+	}
+
+	logsDataCopy.logs = append(logsDataCopy.logs, logsData.logs...)
+
+	return logsDataCopy, found
+}
+
+// SaveLogs saves the logs into the internal map
+func (accumulator *failedTxLogsAccumulator) SaveLogs(txHash []byte, tx data.TransactionHandler, logs []*vmcommon.LogEntry) error {
```
ssd04, line 63: since we are not saving the logs to persistent storage, is it possible that we lose unintended failed tx logs in case of a restart? or the logs will be saved later via tx logs processor if needed?

process/transactionLog/failedTxLogsAccumulator.go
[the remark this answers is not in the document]
sstanculeanu: it should not use a lot of memory. This is how it works: during the execution of a relayed tx v3, all inner tx scrs will be pushed here, then at the end of the execution, they all acumulated scrs will be forwarded to txLogsProcessor, in order to write into that component only once. Once they are forwarded, the entry for the current tx will be deleted from the map.

process/smartContract/processorV2/processV2.go
```diff
@@ -1507,10 +1513,15 @@ func (sc *scProcessor) processIfErrorWithAddedLogs(acntSnd state.UserAccountHand
...
 	}
 
-	logsTxHash := sc.getOriginalTxHashIfIntraShardRelayedSCR(tx, failureContext.txHash)
-	ignorableError := sc.txLogsProcessor.SaveLog(logsTxHash, tx, processIfErrorLogs)
+	logsTxHash, isRelayed := sc.getOriginalTxHashIfIntraShardRelayedSCR(tx, failureContext.txHash)
+	var ignorableError error
+	if isRelayed {
+		ignorableError = sc.failedTxLogsAccumulator.SaveLogs(logsTxHash, tx, processIfErrorLogs)
+	} else {
+		ignorableError = sc.txLogsProcessor.SaveLog(logsTxHash, tx, processIfErrorLogs)
```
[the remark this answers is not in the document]
sstanculeanu: failedTxLogsAccumulator only accumulates them during the execution of the tx. At the end, the accumulated logs are forwarded to the txLogsProcessor. Should be no difference for relayedV1 and relayedV2

process/transaction/interceptedTransaction.go
sstanculeanu: should be imposible to have v1 and v2 on the same tx, as they both are part of the data field. Thus no need for check

process/transaction/shardProcess.go
sstanculeanu: not needed

process/transactionLog/failedTxLogsAccumulator.go
sstanculeanu: I don’t think we can lose logs. If something fails during tx execution and we somehow lose this, the tx won’t be seen as executed too.

ssd04 (dismissed):

sasurobert (approved):

ssd04 (approved):

PR #6205, opened 2024-05-24
FIX: First version bugfix consensus received header

mariusmihaic:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/consensus/spos/bls/export_test.go
+++ b/consensus/spos/bls/export_test.go
@@ -328,6 +328,11 @@ func (sr *subroundEndRound) ReceivedBlockHeaderFinalInfo(cnsDta *consensus.Messa
 	return sr.receivedBlockHeaderFinalInfo(context.Background(), cnsDta)
 }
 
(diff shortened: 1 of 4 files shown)


scripts/testnet/variables.sh
```diff
@@ -122,7 +122,7 @@ export PORT_ORIGIN_VALIDATOR_REST="9500"
 export USETMUX=1
 
 # Log level for the logger in the Node.
-export LOGLEVEL="*:INFO"
+export LOGLEVEL="*:DEBUG"
```
axenteoctavian, line 125: Revert to `INFO`?

mariusmihaic: We always need it on debug for this feat branch, instead of always modifying it locally, will set it like this.

sasurobert (approved):

axenteoctavian (approved):

PR #6243, opened 2024-06-06
Mx 15529 esdt prefix new tests

axenteoctavian:
- tests for brige oprations, execute (mint) on esdt-safe then deposit (burn) on esdt-safe
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


integrationTests/chainSimulator/esdt/bridge_test.go
```diff
@@ -1,190 +0,0 @@
-package esdt
```
mariusmihaic, line 1: Why did you delete it to move it in another package `bridge`?
Please do a `git checkout feat/esdt-prefix integrationTests/chainSimulator/esdt/bridge_test.go` and revert it.
These are esdt tests for prefixed tokens(the fact that we use bridge contracts it's not in the scope, that's just the tool we used for testing the  `feat/esdt-prefix` ), Should've left it here, along with other tests

integrationTests/chainSimulator/esdt/issue_test.go
```diff
@@ -36,21 +30,18 @@ func TestChainSimulator_IssueESDTWithPrefix(t *testing.T) {
...
 
-	numOfShards := uint32(1)
 	cs, err := chainSimulator.NewChainSimulator(chainSimulator.ArgsChainSimulator{
 		BypassTxSignatureCheck:      false,
 		TempDir:                     t.TempDir(),
 		PathToInitialConfig:         defaultPathToInitialConfig,
-		NumOfShards:                 numOfShards,
-		GenesisTimestamp:            startTime,
-		RoundDurationInMillis:       roundDurationInMillis,
+		NumOfShards:                 1,
```
mariusmihaic, line 42: We should put 3 here. 
With your current implementation, this test might fail. You should revert the common funcs back to their original implementation.

integrationTests/chainSimulator/common.go
```diff
@@ -43,3 +98,103 @@ func GenerateTransaction(sender []byte, nonce uint64, receiver []byte, value *bi
...
+
+// RequireAccountHasToken checks if the account has the amount of tokens (can also be zero)
+func RequireAccountHasToken(
+	t *testing.T,
+	cs ChainSimulator,
+	token string,
+	address string,
+	value *big.Int,
+) {
+	tokens, _, err := cs.GetNodeHandler(0).GetFacadeHandler().GetAllESDTTokens(address, dataApi.AccountQueryOptions{})
```
mariusmihaic, line 134: This is wrong, it will fail for multiple shard tests.
Revert back to original code.
Also, delete that `if value.Cmp(big.NewInt(0)) == 0`, it is redundant

integrationTests/chainSimulator/common.go
```diff
@@ -43,3 +98,103 @@ func GenerateTransaction(sender []byte, nonce uint64, receiver []byte, value *bi
...
+	data string,
+	gasLimit uint64,
+) *transaction.ApiTransactionResult {
+	tx := GenerateTransaction(sender, *nonce, receiver, value, data, gasLimit)
+	txResult, err := cs.SendTxAndGenerateBlockTilTxIsExecuted(tx, maxNumOfBlocksToGenerateWhenExecutingTx)
+	*nonce++
+	require.Nil(t, err)
+	require.NotNil(t, txResult)
+	require.Equal(t, transaction.TxStatusSuccess, txResult.Status)
+	if txResult.Logs != nil && txResult.Logs.Events != nil && len(txResult.Logs.Events) > 0 {
```
mariusmihaic, line 119: Second check is redundant.
```suggestion
	if txResult.Logs != nil && len(txResult.Logs.Events) > 0 {
```

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
```diff
@@ -0,0 +1,308 @@
...
+		Identifier: token,
+		Nonce:      tokenNonce,
+		Amount:     big.NewInt(10),
+		Type:       core.Fungible,
+	})
+
+	simulateExecutionAndDeposit(t, mintTokens, depositTokens)
+}
+
+func TestChainSimulator_ExecuteWithMintAndBurnNftWithDeposit(t *testing.T) {
```
mariusmihaic, line 65: Perhaps rename a bit more suggestive, since you "execute" nft + esdt, but only deposit nft

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
```diff
@@ -0,0 +1,308 @@
...
+		Type:       core.SemiFungible,
+	})
+
+	simulateExecutionAndDeposit(t, mintTokens, depositTokens)
+}
+
+func simulateExecutionAndDeposit(
+	t *testing.T,
+	mintTokens []chainSim.ArgsDepositToken,
+	depositTokens []chainSim.ArgsDepositToken,
```
mariusmihaic, line 129: I think we should find a better naming for these variables, here and everywhere in tests(also rename the tests-that execute is the endpoint of the contract, we are not testing that actually).

Something perhaps like bridgedIn/bridgedOut tokens?

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// We will deposit an array of prefixed tokens from a sovereign chain to the main chain,
+	// expecting these tokens to be minted by the whitelisted ESDT safe sc and transferred to our wallet address.
+	executeBridgeOperation(t, cs, wallet, &nonce, bridgeData.ESDTSafeAddress, mintTokens)
+
+	// Deposit an array of tokens from main chain to sovereign chain,
+	// expecting these tokens to be burned by the whitelisted ESDT safe sc
+	Deposit(t, cs, wallet.Bytes, &nonce, bridgeData.ESDTSafeAddress, depositTokens, wallet.Bytes)
+	mintedTokens := groupTokens(mintTokens)
+	for _, token := range groupTokens(depositTokens) {
```
mariusmihaic, line 197:
```suggestion
	for _, bridgedOutToken := range groupTokens(depositTokens) {
```

Perhaps after refactor? or burned token?

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
```diff
@@ -0,0 +1,308 @@
...
+		Identifier: sft,
+		Nonce:      sftNonce,
+		Amount:     big.NewInt(20),
+		Type:       core.SemiFungible,
+	})
+
+	simulateExecutionAndDeposit(t, mintTokens, depositTokens)
+}
+
+func simulateExecutionAndDeposit(
```
mariusmihaic, line 126: :+1: 
Nice and clean tests, using reusable code.

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
```diff
@@ -0,0 +1,308 @@
...
+	nonce := uint64(0)
+
+	bridgeData := DeployBridgeSetup(t, cs, wallet.Bytes, &nonce, esdtSafeWasmPath, feeMarketWasmPath)
+
+	esdtSafeEncoded, _ := nodeHandler.GetCoreComponents().AddressPubKeyConverter().Encode(bridgeData.ESDTSafeAddress)
+	require.Equal(t, esdtSafeEncoded, whiteListedAddress)
+
+	// We will deposit an array of prefixed tokens from a sovereign chain to the main chain,
+	// expecting these tokens to be minted by the whitelisted ESDT safe sc and transferred to our wallet address.
+	executeBridgeOperation(t, cs, wallet, &nonce, bridgeData.ESDTSafeAddress, mintTokens)
```
mariusmihaic, line 191: We should also rename this func

mariusmihaic (changes requested):

mariusmihaic (changes requested):
We also need to use this vm common commit hash for cross chain nonce fix: https://github.com/multiversx/mx-chain-vm-common-go/pull/287

integrationTests/chainSimulator/common.go
axenteoctavian: I reverted the code to work with multiple shards, it was a mistake.
That condition is to check the account doesn't have that token (value i zero), it is needed for other tests (e.g. NFT/SFT execute then burn)

integrationTests/chainSimulator/bridge/executeAndBurn_test.go
axenteoctavian: changed to `executeMintOperation`

integrationTests/chainSimulator/common.go
mariusmihaic: Then shouldn't we actually check something like `GetNFTTokenIDsRegisteredByAddress` or anything similar to check the token actually exists? Otherwise, you check nothing for those cases

mariusmihaic (commented):
Just the one unresolved comment from previous review

mariusmihaic (dismissed):

sasurobert (approved):

mariusmihaic (approved):

sasurobert (approved):

PR #6245, opened 2024-06-07
FIX: Outgoing op tx hash

mariusmihaic:
- When setting outgoing mb after whole processing is done (which contain outgoing operations),  there are some extra checks that can cause the chain to halt. Mainly, we were missing some saved txs for the outgoing tx hashes, which were not found and also failed gas limit checks
  
## Proposed changes
-  Since we weren't saving outgoing operations anyway (especially the data inside time), I think it's better to save it in the storage as a "dummy" tx with the outGoingTxData that is used by the bridge service to format txs for main chain.

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
+++ b/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
@@ -40,6 +40,8 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 	if err != nil {
 		return nil, err
 	}
+
+	alteredSovConfigs := config.SovereignConfig{}
 	args.AlterConfigsFunction = func(cfg *config.Configs) {
 		cfg.EconomicsConfig = configs.EconomicsConfig
 		cfg.EpochConfig = configs.EpochConfig
@@ -50,6 +52,8 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		if alterConfigs != nil {
 			alterConfigs(cfg)
 		}
+
+		alteredSovConfigs = cfg.GeneralConfig.SovereignConfig
 	}
 
 	args.CreateGenesisNodesSetup = func(nodesFilePath string, addressPubkeyConverter core.PubkeyConverter, validatorPubkeyConverter core.PubkeyConverter, _ uint32) (sharding.GenesisNodesSetupHandler, error) {
@@ -63,9 +67,11 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		return rating.NewSovereignRatingsData(arg)
 	}
 	args.CreateIncomingHeaderSubscriber = func(config *config.NotifierConfig, dataPool dataRetriever.PoolsHolder, mainChainNotarizationStartRound uint64, runTypeComponents factory.RunTypeComponentsHolder) (process.IncomingHeaderSubscriber, error) {
+		config = &alteredSovConfigs.NotifierConfig
 		return incomingHeader.CreateIncomingHeaderProcessor(config, dataPool, mainChainNotarizationStartRound, runTypeComponents)
 	}
 	args.CreateRunTypeComponents = func(argsRunType runType.ArgsRunTypeComponents) (factory.RunTypeComponentsHolder, error) {
+		*configs.SovereignExtraConfig = alteredSovConfigs
 		return createSovereignRunTypeComponents(argsRunType, *configs.SovereignExtraConfig)
 	}
 	args.NodeFactory = node.NewSovereignNodeFactory()
--- a/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
+++ b/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
@@ -41,7 +41,7 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		return nil, err
 	}
 
-	alteredSovConfigs := config.SovereignConfig{}
+	alteredSovConfigs := *configs.SovereignExtraConfig
 	args.AlterConfigsFunction = func(cfg *config.Configs) {
 		cfg.EconomicsConfig = configs.EconomicsConfig
 		cfg.EpochConfig = configs.EpochConfig
@@ -51,9 +51,8 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 
 		if alterConfigs != nil {
 			alterConfigs(cfg)
+			alteredSovConfigs = cfg.GeneralConfig.SovereignConfig
 		}
-
-		alteredSovConfigs = cfg.GeneralConfig.SovereignConfig
 	}
 
 	args.CreateGenesisNodesSetup = func(nodesFilePath string, addressPubkeyConverter core.PubkeyConverter, validatorPubkeyConverter core.PubkeyConverter, _ uint32) (sharding.GenesisNodesSetupHandler, error) {
@@ -67,12 +66,10 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		return rating.NewSovereignRatingsData(arg)
 	}
 	args.CreateIncomingHeaderSubscriber = func(config *config.NotifierConfig, dataPool dataRetriever.PoolsHolder, mainChainNotarizationStartRound uint64, runTypeComponents factory.RunTypeComponentsHolder) (process.IncomingHeaderSubscriber, error) {
-		config = &alteredSovConfigs.NotifierConfig
-		return incomingHeader.CreateIncomingHeaderProcessor(config, dataPool, mainChainNotarizationStartRound, runTypeComponents)
+		return incomingHeader.CreateIncomingHeaderProcessor(&alteredSovConfigs.NotifierConfig, dataPool, mainChainNotarizationStartRound, runTypeComponents)
 	}
 	args.CreateRunTypeComponents = func(argsRunType runType.ArgsRunTypeComponents) (factory.RunTypeComponentsHolder, error) {
-		*configs.SovereignExtraConfig = alteredSovConfigs
-		return createSovereignRunTypeComponents(argsRunType, *configs.SovereignExtraConfig)
+		return createSovereignRunTypeComponents(argsRunType, alteredSovConfigs)
 	}
 	args.NodeFactory = node.NewSovereignNodeFactory()
 	args.ChainProcessorFactory = NewSovereignChainHandlerFactory()
--- a/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
+++ b/cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
@@ -41,7 +41,6 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		return nil, err
 	}
 
-	alteredSovConfigs := *configs.SovereignExtraConfig
 	args.AlterConfigsFunction = func(cfg *config.Configs) {
 		cfg.EconomicsConfig = configs.EconomicsConfig
 		cfg.EpochConfig = configs.EpochConfig
@@ -51,7 +50,7 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 
 		if alterConfigs != nil {
 			alterConfigs(cfg)
-			alteredSovConfigs = cfg.GeneralConfig.SovereignConfig
+			configs.SovereignExtraConfig = &cfg.GeneralConfig.SovereignConfig
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
```diff
@@ -50,6 +52,8 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		if alterConfigs != nil {
 			alterConfigs(cfg)
 		}
+
+		alteredSovConfigs = cfg.GeneralConfig.SovereignConfig
```
axenteoctavian, line 56: You can put this line inside the `if alterConfigs != nil`

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
```diff
@@ -63,9 +67,11 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
 		return rating.NewSovereignRatingsData(arg)
 	}
 	args.CreateIncomingHeaderSubscriber = func(config *config.NotifierConfig, dataPool dataRetriever.PoolsHolder, mainChainNotarizationStartRound uint64, runTypeComponents factory.RunTypeComponentsHolder) (process.IncomingHeaderSubscriber, error) {
+		config = &alteredSovConfigs.NotifierConfig
```
axenteoctavian, line 70: This is not needed because the config comes from func args and is already altered.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
```diff
@@ -63,9 +67,11 @@ func NewSovereignChainSimulator(args ArgsSovereignChainSimulator) (chainSimulato
...
 		return incomingHeader.CreateIncomingHeaderProcessor(config, dataPool, mainChainNotarizationStartRound, runTypeComponents)
 	}
 	args.CreateRunTypeComponents = func(argsRunType runType.ArgsRunTypeComponents) (factory.RunTypeComponentsHolder, error) {
+		*configs.SovereignExtraConfig = alteredSovConfigs
```
axenteoctavian, line 74: I don't think that is really necessary to change the pointer here, you can just put `alteredSovConfigs` below in the return.
Or, If you need to have the `configs.SovereignExtraConfig` changed, just updated it at line 54 with `configs.SovereignExtraConfig = &cfg.GeneralConfig.SovereignConfig` and deleted the alteredSovConfigs var

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
```diff
@@ -5,9 +5,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/multiversx/mx-chain-core-go/data/transaction"
```
axenteoctavian, line 8: Please fix this, move this line at line 18

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
```diff
@@ -55,12 +75,27 @@ func TestSovereignChainSimulator_DeployBridgeContractsThenIssueAndDeposit(t *tes
...
+	require.Nil(t, err)
+
+	wallet := dtos.WalletAddress{Bech32: initialAddress, Bytes: initialAddrBytes}
+
+	expectedESDTSafeAddressBytes, err := nodeHandler.GetCoreComponents().AddressPubKeyConverter().Decode(outGoingSubscribedAddress)
 	require.Nil(t, err)
-	nonce := uint64(0)
 
-	bridgeData := DeploySovereignBridgeSetup(t, cs, esdtSafeWasmPath, feeMarketWasmPath)
+	bridgeData := DeploySovereignBridgeSetup(t, cs, wallet, esdtSafeWasmPath, feeMarketWasmPath)
```
axenteoctavian, line 95: I would just send a bech32 address as parameter (maybe a common constant in common.go) and do the SetStateMultiple inside the deploy function, and return OwnerWallet: dtos.WalletAddress as another parameter in case we want to make other transactions on the bridge contracts.

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
```diff
@@ -55,12 +75,27 @@ func TestSovereignChainSimulator_DeployBridgeContractsThenIssueAndDeposit(t *tes
...
+	require.Equal(t, expectedESDTSafeAddressBytes, bridgeData.ESDTSafeAddress)
 
+	nonce := uint64(5)
```
axenteoctavian, line 98: This is not really ok because the number of deploy transactions will be changed in the future. Just create another wallet and start with nonce 0 (do all the operations as a new user, we don't need to do them from bridge owner wallet)

process/block/preprocess/basePreProcess.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/multiversx/mx-chain-core-go/data/block"
 	"github.com/multiversx/mx-chain-core-go/hashing"
 	"github.com/multiversx/mx-chain-core-go/marshal"
+
```
axenteoctavian, line 15: Please revert this, or move the mx-chain-go imports above

process/block/sovereignChainBlock.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/multiversx/mx-chain-core-go/data/transaction"
```
axenteoctavian, line 9: Please move this at line 22

process/block/sovereignChainBlock.go
```diff
@@ -916,17 +921,31 @@ func (scbp *sovereignChainBlockProcessor) createOutGoingMiniBlockData(outGoingOp
...
+			}),
+		GasPrice: scbp.economicsData.MinGasPrice(),
+		Data:     outGoingOp.Data,
+	}
+
+	scbp.dataPool.Transactions().AddData(
+		outGoingOp.Hash,
+		tx,
+		tx.Size(),
+		fmt.Sprintf("%d_%d", core.SovereignChainShardId, core.MainChainShardId),
```
axenteoctavian, line 945: You can use here `cacheId := process.ShardCacherIdentifier(core.SovereignChainShardId,  core.MainChainShardId)`

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
```diff
@@ -87,4 +123,34 @@ func TestSovereignChainSimulator_DeployBridgeContractsThenIssueAndDeposit(t *tes
...
+			&transaction.Transaction{
+				Data: outGoingOp.Data,
+			}),
+		Data: outGoingOp.Data,
+	}
+	require.Equal(t, expectedSavedTx, savedTx)
+
+	// Generate extra blocks after outgoing operations are created
+	err = cs.GenerateBlocks(10)
+	require.Nil(t, err)
```
axenteoctavian, line 155: Maybe a TODO for later to add IncomingHeader with confirmation and then check UnconfirmedOperations == 0. This way we'll have a full cross-chain transaction test.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
sasurobert: +1

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
sasurobert: +1

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
sasurobert: +1

process/block/sovereignChainBlock.go
sasurobert: +1

process/block/sovereignChainBlock.go
```diff
@@ -916,17 +921,31 @@ func (scbp *sovereignChainBlockProcessor) createOutGoingMiniBlockData(outGoingOp
...
 	return &block.MiniBlock{
 		TxHashes:        outGoingOpHashes,
 		ReceiverShardID: core.MainChainShardId,
 		SenderShardID:   scbp.shardCoordinator.SelfId(),
 	}, outGoingOperationsHash
 }
 
+func (scbp *sovereignChainBlockProcessor) addOutGoingTxToPool(outGoingOp *sovCore.OutGoingOperation) {
+	tx := &transaction.Transaction{
+		GasLimit: scbp.economicsData.ComputeGasLimit(
```
sasurobert, line 933: I think gasLimit should be much higher. not only outGoingOpData - but it needs to be computed from somewhere.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
mariusmihaic: Moved it there, but had to initialize it above as well.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
mariusmihaic: That's not the case here. These functions are called before `AlterConfig` is called further down in base chain simulator, so I need the altered configs beforehand. Test would not pass anyway otherwise.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
mariusmihaic: See above comment, as for extra variable, deleted it and passed directly

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
mariusmihaic: 1. It needs now bech32
2. Wallet should be input paramter, not randomly generated there and returned.

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
mariusmihaic: I was aware of this limitation, but didn't want to do any more refactor already in this test.
Made a common func to return the acc nonce

process/block/sovereignChainBlock.go
mariusmihaic: I'm very much afraid this will cause a go linter issue which will say to put it directly without variable, but changed as request

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
mariusmihaic: Yes, could be done in another PR

process/block/sovereignChainBlock.go
mariusmihaic: As discussed, this is just a dummy stored tx, so actually no gas is used; this value is used just to pass some basic gas checks futher down in code

sasurobert (approved):

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
axenteoctavian: No, this `CreateIncomingHeaderSubscriber` is called after `AlterConfig`.
AlterConfig is the first thing, then is Create Core comps, Crypto comps, ..., IncomingHeader, and so on

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
axenteoctavian: Yes, you could send `initialAddress` as parameter and do the set state and stuff inside the `DeploySovereignBridgeSetup` and return the wallet+nonce in `bridgeData: ArgsBridgeSetup`. Then the rest of the transactions could be done from bridge owner wallet or a new random wallet.

Anyways, it's fine as it is, maybe we'll do this later.

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
axenteoctavian: The comment above is not valid. The `Or` is still an alternative if you want.

cmd/sovereignnode/chainSimulator/tests/bridge/bridge_test.go
mariusmihaic: 1. Why would you generate a random wallet, since you want to set in genesis the exact sc address that will be deployed? no need to use extra wallets.
2. No need to return the nonce, added func to retrieve the nonce

cmd/sovereignnode/chainSimulator/sovereignChainSimulator.go
mariusmihaic: Not the case for runType comps, those are called before, please check for yourself, I need that extra variable and to change it. Maybe not also for incoming, but definetely for runType

axenteoctavian (approved):

sasurobert (approved):

PR #6247, opened 2024-06-07
fixed error on withKeys

raduchis:
- account with Keys does not work for empty accounts
  
## Proposed changes
- take into consideration empty accounts and check for nil

## Testing procedure
- run testnet and call address/:address?withKeys=true for multiple accounts, some with DataTries, some without and some not created yet

## Pre-requisites


node/node.go
```diff
@@ -960,6 +960,10 @@ func (n *Node) GetAccountWithKeys(address string, options api.AccountQueryOption
 		return api.AccountResponse{}, api.BlockInfo{}, err
 	}
 
+	if accInfo.account == nil || accInfo.account.DataTrie() == nil {
```
AdoAdoAdo, line 963: is this method only used when options.WithKeys is set? 

if yes then all good, otherwise maybe you need to move the check for DataTrie in the code block starting at line 968

gabi-vuls (approved):
Normal allin test: 1.7.12-2a6916fbce -> MX-15556-Bugfix-Account-Wi-3dc4427bd2

--- Specific errors ---

block hash does not match 2133
wrong nonce in block 878
miniblocks does not match 0
block bodies does not match 1
receipts hash missmatch 0

/------/


--- Statistics ---

Nr. of all ERRORS:  0
Nr. of all WARNS:  599
Nr. of new ERRORS:  0
Nr. of new WARNS:  17
Nr. of PANICS:  0

/------/

--- ERRORS ---

/------/

--- WARNINGS ---

/------/

node/node.go
```diff
@@ -962,6 +962,10 @@ func (n *Node) GetAccountWithKeys(address string, options api.AccountQueryOption
 
 	var keys map[string]string
 	if options.WithKeys {
+		if accInfo.account == nil || accInfo.account.DataTrie() == nil {
+			return accInfo.accountResponse, accInfo.block, nil
```
mariusmihaic, line 966: Shouldn't we return the default values like in other cases?
``` go
return api.AccountResponse{}, api.BlockInfo{}, err
```

raduchis: the accountResponse and BlockInfo come populated even if the accounts are empty or have nil datatries. So you would get a valid accountResponse with nonce, address etc and a valid blockInfo even in the case of nil account or datatrie.

mariusmihaic (approved):

node/node.go
raduchis: moved inside the if on line 968

AdoAdoAdo (dismissed):

node/node.go
```diff
@@ -290,20 +290,20 @@ func (n *Node) GetKeyValuePairs(address string, options api.AccountQueryOptions,
 			return make(map[string]string), adaptedBlockInfo, nil
 		}
 
-		return nil, api.BlockInfo{}, err
+		return nil, blockInfo, err
 	}
 
 	if check.IfNil(userAccount.DataTrie()) {
-		return map[string]string{}, api.BlockInfo{}, nil
```
AdoAdoAdo, line 297: maybe just here return you need to return the blockinfo, the others with error should still return zero values

mariusmihaic (approved):

node/node.go
raduchis: fixed

mariusmihaic (approved):

AdoAdoAdo (approved):

lcswillems:
Thank you very much for fixing this! Do you confirm that it would solve those two errors:

https://gateway.multiversx.com/address/erd174wnc5jlss2uxesrdn03dllegvpyd5cs476qewqym8c6eeqwu5qqr7pna8?withKeys=true

https://gateway.multiversx.com/address/erd1pgeh3jva2pdaz65ak262s68z9fpq2sqwq36v3kr59aa2vvetl4fsrjxy4j?withKeys=true

danidrasovean (approved):
Normal allin test: v1.7.13-dev-config-aebb55e8e0 -> MX-15556-Bugfix-Account-Wi-e520f75f28

--- Specific errors --- 

block hash does not match 1165
wrong nonce in block 477
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  1
Nr. of all WARNS:  675
Nr. of new ERRORS:  1
Nr. of new WARNS:  3
Nr. of PANICS:  0

/------/

PR #6252, opened 2024-06-11
RelayedV3 fixes

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/transaction/shardProcess.go
```diff
@@ -467,6 +467,7 @@ func (txProc *txProcessor) processMoveBalance(
 	destShardTxType process.TransactionType,
 	originalTxHash []byte,
 	isUserTxOfRelayed bool,
+	isUserTxOfRelayedV3 bool,
```
sasurobert, line 470: can you make something better than 2 bools ? this is not really good.

process/transaction/shardProcess.go
```diff
@@ -1216,6 +1231,31 @@ func (txProc *txProcessor) createCompleteEventLog(scr data.TransactionHandler, o
...
+		SndAddr:        tx.SndAddr,
+		PrevTxHash:     txHash,
+		OriginalTxHash: originalTxHash,
+		GasPrice:       tx.GetGasPrice(),
+		CallType:       vm.DirectCall,
+		ReturnMessage:  []byte(core.GasRefundForRelayerMessage),
+		OriginalSender: tx.RelayerAddr,
+	}
+
+	return txProc.scrForwarder.AddIntermediateTransactions([]data.TransactionHandler{refundGasToRelayerSCR})
```
sasurobert, line 1256: err = txProc.createReceiptWithReturnedGas(txHash, tx, acntSrc, moveBalanceCost, totalCost, destShardTxType, isUserTxOfRelayed)
	if err != nil {
		return err
	} 

this is called before. Why would you want to create refundSCRForMoveBalance? A receipt is enough and better.

sasurobert (changes requested):

process/transaction/shardProcess.go
sstanculeanu: added a new scr because the receipt is not created for user tx of relayed: https://github.com/multiversx/mx-chain-go/blob/75ef2631e72298207500b02251518f3943a32702/process/transaction/shardProcess.go#L356

second reason was to make it compatible with the method used while computing the gasUsed field on api response, that only checks scrs: https://github.com/multiversx/mx-chain-go/blob/75ef2631e72298207500b02251518f3943a32702/outport/process/transactionsfee/transactionChecker.go#L49

integrationTests/multiShard/relayedTx/relayedTx_test.go
```diff
@@ -447,8 +447,11 @@ func checkPlayerBalances(
 	t *testing.T,
 	nodes []*integrationTests.TestProcessorNode,
 	players []*integrationTests.TestWalletAccount) {
-	for _, player := range players {
+	for idx, player := range players {
 		userAcc := GetUserAccount(nodes, player.Address)
+		if idx == 5 {
+			print("x")
+		}
```
ssd04, line 454: do not forget to remove debugging code

integrationTests/chainSimulator/relayedTx/relayedTx_test.go
```diff
@@ -255,7 +263,199 @@ func TestRelayedTransactionInMultiShardEnvironmentWithChainSimulatorScCalls(t *t
...
 }
 
-func startChainSimulator(t *testing.T) testsChainSimulator.ChainSimulator {
+func TestFixRelayedMoveBalanceWithChainSimulator(t *testing.T) {
+	if testing.Short() {
+		t.Skip("this is not a short test")
+	}
+
+	expectedFeeScCall := "815285920000000"
+	t.Run("sc call", testFixRelayedMoveBalanceWithChainSimulatorScCall(expectedFeeScCall, expectedFeeScCall))
```
ssd04, line 272: so for this test the expected fee, before and after fix, is the same?

sstanculeanu: yes, the fix should not influence sc calls, but only move balance

sasurobert (approved):

ssd04 (approved):

PR #6257, opened 2024-06-13
Improved and updated sovereign deploy scrips

axenteoctavian:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


scripts/testnet/prerequisites.sh
```diff
@@ -8,15 +8,15 @@ export DISTRIBUTION=$(cat /etc/os-release | grep "^ID=" | sed 's/ID=//')
 
 if [[ "$DISTRIBUTION" =~ ^(fedora|centos|rhel)$ ]]; then
   export PACKAGE_MANAGER="dnf"
-  export REQUIRED_PACKAGES="git golang gcc lsof jq curl"
+  export REQUIRED_PACKAGES="git golang gcc lsof jq curl tmux"
```
mariusmihaic, line 11: Why did you change these?
You are affecting all other scripts/user scripts of normal chain with this.

How safe are these changes?
How compatible is it for other users using it?
Why are these required?

scripts/testnet/sovereignBridge/config/common.snippets.sh
```diff
@@ -1,55 +1,59 @@
...
+}
 
-    setFeeMarketAddressSovereign
+gitPullAllChanges()
+{
+    pushd .
 
-    disableFeeMarketContractSovereign
+    # Traverse up to the parent directory of "mx-chain-go"
+    while [[ ! -d "mx-chain-go" ]]; do
```
mariusmihaic, line 34: We should Have I think an extra exit condition here,
If you call it from another directory, this might act weird.

scripts/testnet/sovereignBridge/config/configs.cfg
```diff
@@ -1,54 +1,49 @@
...
 
-#============ WASM FILES ==============
-
-ESDT_SAFE_WASM="~/contracts/esdt-safe.wasm"
-FEE_MARKET_WASM="~/contracts/fee-market.wasm"
-MULTISIG_VERIFIER_WASM="~/contracts/multisigverifier.wasm"
+#=========== NETWORK CONFIGURATION ===========
+PROXY=https://testnet-gateway.multiversx.com
+CHAIN_ID=T
+PROXY_SOVEREIGN=http://localhost:7950
```
mariusmihaic, line 12: We should use the proxy port and chain id from variables.sh here I guess

scripts/testnet/sovereignBridge/config/configs.cfg
```diff
@@ -1,54 +1,49 @@
...
+
+#============ FEE MARKET DEPLOY ARGUMENTS ==============
 PRICE_AGGREGATOR_ADDRESS=erd1qqqqqqqqqqqqqpgqfsfcrxnlym0fjlnuvus2dmlz6ju9cpsfexks8wq4d4
 WEGLD_TOKEN_ID=WEGLD-a28c59
 USDC_TOKEN_ID=USDC-350c4e
 
-PROXY_SOVEREIGN=http://localhost:7950
-CHAIN_ID_SOVEREIGN=local-testnet
-
-#============ ESDT SAFE DEPLOY ARGUMENTS ==============
```
mariusmihaic, line 24: So these are not needed anymore?

scripts/testnet/sovereignBridge/config/deploy.snippets.sh
```diff
@@ -0,0 +1,125 @@
...
+    stopObserver
+}
+
+# This function will:
+# - stop sovereign nodes and services
+# - clean the sovereign configuration
+stopAndCleanSovereign() {
+    stopSovereign
+
+    ../clean.sh
```
mariusmihaic, line 124: Why two dots here and above?

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
```diff
@@ -2,167 +2,225 @@ ESDT_SAFE_ADDRESS=$(mxpy data load --partition=${CHAIN_ID} --key=address-esdt-sa
...
+        exit 1
+    fi
+
+    local ADDRESS=$1
+    local URL=$2
+    local CHAIN=$3
+    local OUTFILE=$4
+
+    mxpy contract upgrade ${ADDRESS} \
         --bytecode=$(eval echo ${ESDT_SAFE_WASM}) \
```
mariusmihaic, line 62: Is this wasm file same for both main/sovereign chain?

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
```diff
@@ -2,167 +2,225 @@ ESDT_SAFE_ADDRESS=$(mxpy data load --partition=${CHAIN_ID} --key=address-esdt-sa
...
 }
 
-setMultisigAddress() {
-    checkVariables ESDT_SAFE_ADDRESS MULTISIG_VERIFIER_ADDRESS || return
+setHeaderVerifierAddressInEsdtSafe() {
+    echo "Setting Header Verifier address in ESDT Safe contract on main chain..."
+    checkVariables ESDT_SAFE_ADDRESS HEADER_VERIFIER_ADDRESS || return
 
-    mxpy --verbose contract call ${ESDT_SAFE_ADDRESS} \
+    local OUTFILE="${OUTFILE_PATH}/set-headerverifier-address.interaction.json"
```
mariusmihaic, line 192:
```suggestion
    local OUTFILE="${OUTFILE_PATH}/set-header-verifier-address.interaction.json"
```

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
```diff
@@ -2,167 +2,225 @@ ESDT_SAFE_ADDRESS=$(mxpy data load --partition=${CHAIN_ID} --key=address-esdt-sa
...
+    mxpy contract call ${ESDT_SAFE_ADDRESS} \
         --pem=${WALLET} \
         --proxy=${PROXY} \
         --chain=${CHAIN_ID} \
         --gas-limit=10000000 \
-        --function="setMultisigAddress" \
-        --arguments ${MULTISIG_VERIFIER_ADDRESS} \
+        --function="setHeaderVerifierAddress" \
```
mariusmihaic, line 198: Why header verifier?
Since when are we not verifying multisigs anymore and only headers?

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
```diff
@@ -2,167 +2,225 @@ ESDT_SAFE_ADDRESS=$(mxpy data load --partition=${CHAIN_ID} --key=address-esdt-sa
...
+        --outfile=${OUTFILE} \
         --recall-nonce \
         --wait-result \
         --send || return
+
+    printTxStatus ${OUTFILE}
 }
 
-setSovereignBridgeAddress() {
+setSovereignBridgeAddressInEsdtSafe() {
```
mariusmihaic, line 208: What was this address used for?

scripts/testnet/sovereignBridge/config/header-verifier.snippets.sh
```diff
@@ -0,0 +1,69 @@
...
+    echo -e "Header Verifier contract: ${ADDRESS}\n"
+}
+
+upgradeHeaderVerifierContract() {
+    manualUpdateConfigFile #update config file
+
+    echo "Upgrading Header Verifier contract on main chain..."
+
+    local OUTFILE="${OUTFILE_PATH}/upgrade-header-verifier.interaction.json"
+    mxpy contract upgrade ${HEADER_VERIFIER_ADDRESS} \
```
mariusmihaic, line 37: I feel like we re-write this function of upgrading contract in many other places;
Is there any way to make a base func out of it?

scripts/testnet/sovereignBridge/config/deploy.snippets.sh
```diff
@@ -0,0 +1,125 @@
...
+
+    sovereignDeploy
+}
+
+# This function will:
+# - deploy all main chain contracts
+# - update sovereign configs
+# - prepare a main chain observer for sovereign nodes
+deployMainChainContractsAndSetupObserver() {
+    checkWalletBalanceOnMainChain || return
```
mariusmihaic, line 15: Shouldn't we do a` || return` for each of the functions below?

scripts/testnet/sovereignBridge/config/token.snippets.sh
```diff
@@ -1,9 +1,11 @@
 issueToken() {
     manualUpdateConfigFile #update config file
 
-    local TOKENS_TO_MINT=$(echo "scale=0; $INITIAL_SUPPLY*10^$NR_DECIMALS/1" | bc)
+    echo "Issuing fungible token on main chain..."
 
-    mxpy --verbose contract call ${ESDT_SYSTEM_SC_ADDRESS} \
+    local TOKENS_TO_MINT=$(echo "scale=0; $INITIAL_SUPPLY*10^$NR_DECIMALS/1" | bc)
```
mariusmihaic, line 6: Is this multiplication safe for very large numbers?

scripts/testnet/sovereignBridge/config/token.snippets.sh
```diff
@@ -47,17 +53,22 @@ depositTokenInSC() {
...
         --send || return
+
+    printTxStatus ${OUTFILE}
 }
 
 issueTokenSovereign() {
     manualUpdateConfigFile #update config file
 
-    local TOKENS_TO_MINT=$(echo "scale=0; $INITIAL_SUPPLY_SOVEREIGN*10^$NR_DECIMALS_SOVEREIGN/1" | bc)
+    echo "Issuing fungible token on sovereign chain..."
```
mariusmihaic, line 67: Can't we use a base func to issue token as you did before?
This code seems duplicated with the one from normal chain

scripts/testnet/sovereignBridge/prerequisites.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+runBridgePrerequisites() {
```
mariusmihaic, line 3: Where is this called and used?

scripts/testnet/prerequisites.sh
axenteoctavian: 1/ `tmux` is what we use, right?
2/ from what I know, we need golang 1.20 to be able to run the nodes, so we don't want to install the latest version
3/ clone that repo with SSH to be conistent

scripts/testnet/sovereignBridge/config/common.snippets.sh
axenteoctavian: fixed

scripts/testnet/sovereignBridge/config/configs.cfg
axenteoctavian: Correct. We created another version of the contract without those arguments.

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
axenteoctavian: Currently, yes. Same wasm is for both chains.

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
axenteoctavian: I check again and this is not used anymore, deleted this function.
I'm not sure why it was there in the first place, probably introduced with some bad logic that was deleted.

scripts/testnet/sovereignBridge/config/esdt-safe.snippets.sh
axenteoctavian: Andrei renamed the contract with request from Robert.

scripts/testnet/sovereignBridge/config/header-verifier.snippets.sh
axenteoctavian: I would have to send so many parameters to a base function (message, outfile, address, wasm, proxy, chain, gas). Also arguments can be sent if we need in the future and that would be variable, etc... 
Also, probably, these upgrade funcs will be deleted in the future from deploy scrips when we'll have factory/config SC.

scripts/testnet/sovereignBridge/prerequisites.sh
axenteoctavian: It's in the docs.
Step 2 is clone `mx-chain-go` and run `./prerequisites.sh` to install node dependencies.
Step 3 is `runBridgePrerequisites` to install all software needed for scripts and download the contracts automatically.

scripts/testnet/sovereignBridge/config/token.snippets.sh
axenteoctavian: ChatGPT: `By default, bc supports up to approximately 9223372036854775807 (2^63 - 1) for integers on many systems`.

scripts/testnet/sovereignBridge/config/token.snippets.sh
axenteoctavian: It's just an mxpy command with arguments from config. Will live it as it is for now. Maybe in the future we'll even delete these issue examples..

scripts/testnet/sovereignBridge/config/deploy.snippets.sh
axenteoctavian: The `.sh` script is in the parent directory. But I changed it so we can call the functions from anywhere not a specific path.

scripts/testnet/sovereignBridge/config/deploy.snippets.sh
axenteoctavian: I added more `|| return` for some more functions that can actually fail because of missing stuff.

scripts/testnet/prerequisites.sh
mariusmihaic: tmux is optional actually, not required, we should be able to run nodes without it, not enforce it;
I think the best way to do it is to add that extra options with `tmux` only if the flag to `USETMUX` is enabled

scripts/testnet/include/config.sh
```diff
@@ -89,7 +89,12 @@ updateSeednodeConfig() {
 prepareElasticsearch() {
   echo "Starting Elasticsearch Docker container..."
   pwd
-  export ES_CONTAINER_ID=$(docker run -d -p 9200:9200 -p 9301:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.10.2)
+  ES_CONTAINER="sov-elastic"
```
mariusmihaic, line 92: Is it mandatory to add `sov` prefix? 
I think this should work for normal chain as well

scripts/testnet/include/config.sh
```diff
@@ -89,7 +89,12 @@ updateSeednodeConfig() {
...
+  if [ -n "$(docker ps -aqf name=$ES_CONTAINER)" ]; then
```
mariusmihaic, line 93: Nice one :+1:

scripts/testnet/sovereignBridge/config/contracts.snippets.sh
```diff
@@ -0,0 +1,9 @@
+downloadCrossChainContracts() {
+    echo "Downloading cross-chain contracts..."
+
+    mkdir -p $(eval echo "${CONTRACTS_DIRECTORY}")
+    version=$(basename `curl -s https://github.com/multiversx/mx-sovereign-sc/releases/latest -I | grep location | awk -F"https:/" '{print $2}' | tr -d "\r"`)
+    wget -O $(eval echo ${ESDT_SAFE_WASM}) https://github.com/multiversx/mx-sovereign-sc/releases/download/${version}/esdt-safe.wasm
+    wget -O $(eval echo ${FEE_MARKET_WASM}) https://github.com/multiversx/mx-sovereign-sc/releases/download/${version}/fee-market.wasm
+    wget -O $(eval echo ${HEADER_VERIFIER_WASM}) https://github.com/multiversx/mx-sovereign-sc/releases/download/${version}/header-verifier.wasm
+}
```
mariusmihaic, line 9: Missing EOF empty line

scripts/testnet/sovereignBridge/prerequisites.sh
```diff
@@ -1,31 +1,33 @@
...
+sudo apt install -y python3-pip pipx screen ca-certificates curl wget
+pip install multiversx-sdk
+pipx install multiversx-sdk-cli --force
+pipx ensurepath
 
-    downloadCrossChainContracts
-}
+source config/configs.cfg
+source config/contracts.snippets.sh
+downloadCrossChainContracts
```
mariusmihaic, line 12: These 3 extra functions should not be done by default.
Please put them inside an `if SOVEREIGN_CHAIN`

scripts/testnet/include/config.sh
axenteoctavian: renamed to `mvx-elastic` to be more unique

scripts/testnet/sovereignBridge/prerequisites.sh
axenteoctavian: This bridge software prerequisites and we need this by default.

mariusmihaic (approved):

sasurobert (approved):

PR #6274, opened 2024-06-21
Rename FixRelayedMoveBalanceFlag to FixRelayedBaseCostFlag + proper fix

sstanculeanu:
- the fix for move balance should actually be fix for base cost
  
## Proposed changes
- updated the fix to also cover other inner txs than move balance

## Testing procedure
- will be verified on feat branch

## Pre-requisites

### DIFF
--- a/process/transaction/shardProcess.go
+++ b/process/transaction/shardProcess.go
@@ -1154,20 +1154,18 @@ func (txProc *txProcessor) executeFailedRelayedUserTx(
 		return err
 	}
 
-	_, dstShardTxType := txProc.txTypeHandler.ComputeTransactionType(userTx)
-	isMoveBalance := dstShardTxType == process.MoveBalance
 	totalFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, userTx.GasLimit)
 	moveBalanceGasLimit := txProc.economicsFee.ComputeGasLimit(userTx)
 	gasToUse := userTx.GetGasLimit() - moveBalanceGasLimit
 	processingUserFee := txProc.economicsFee.ComputeFeeForProcessing(userTx, gasToUse)
-	if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedBaseCostFlag) && isMoveBalance {
+	if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedBaseCostFlag) {
 		moveBalanceUserFee := txProc.economicsFee.ComputeMoveBalanceFee(userTx)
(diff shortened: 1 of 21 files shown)


process/transaction/shardProcess.go
```diff
@@ -744,9 +743,7 @@ func (txProc *txProcessor) processInnerTx(
 	originalTxHash []byte,
 ) (*big.Int, vmcommon.ReturnCode, error) {
 
-	_, dstShardTxType := txProc.txTypeHandler.ComputeTransactionType(innerTx)
-	isMoveBalance := dstShardTxType == process.MoveBalance
-	txFee := txProc.computeTxFee(innerTx, isMoveBalance)
+	txFee := txProc.computeTxFee(innerTx)
```
sasurobert, line 746: rename to computeTxFeeForRelayedTx?

sstanculeanu: pushed

sasurobert (approved):

process/transaction/shardProcess.go
```diff
@@ -865,10 +862,8 @@ func (txProc *txProcessor) processRelayedTx(
 func (txProc *txProcessor) computeRelayedTxFees(tx, userTx *transaction.Transaction) relayedFees {
 	relayerFee := txProc.economicsFee.ComputeMoveBalanceFee(tx)
 	totalFee := txProc.economicsFee.ComputeTxFee(tx)
-	_, dstShardTxType := txProc.txTypeHandler.ComputeTransactionType(userTx)
-	isMoveBalance := dstShardTxType == process.MoveBalance
-	if txProc.enableEpochsHandler.IsFlagEnabled(common.FixRelayedMoveBalanceFlag) && isMoveBalance {
```
raduchis, line 870: there was a check for && isMoveBalance. Is it ok that it's not used any more? 

New lines 1163 and 1170 on executeFailedRelayedUserTx still check for isMoveBalance.

sasurobert: Good catch. I think it needs to be deleted from there as well.

We should remove that isMoveBalance check from everywhere when we speak about relayed TX.  As you want to process the baseFee (gasPrice*(50K + len(data)*1500)) for every inner operation, as it happens on any tx.

sstanculeanu: right, pushed

raduchis: ok!

raduchis (approved):

sasurobert (approved):

PR #6280, opened 2024-06-25
Relayedv3 further fixes

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/economics/economicsData.go
```diff
@@ -348,6 +361,23 @@ func (ed *economicsData) getTotalFeesRequiredForInnerTxs(innerTxs []data.Transac
 	return totalFees
 }
 
+func (ed *economicsData) isMoveBalance(tx data.TransactionHandler) bool {
```
sasurobert, line 364: this function is not good. 

you would also need to check on line 373 function to be not a built in function. and you can have execution of contracts without arguments. with only function call.

sasurobert: we need the txTypeHandler here - or to have the txType added as input parameter in the function.

process/economics/economicsData.go
```diff
@@ -337,6 +343,13 @@ func (ed *economicsData) ComputeRelayedTxFees(tx data.TransactionWithFeeHandler)
 func (ed *economicsData) getTotalFeesRequiredForInnerTxs(innerTxs []data.TransactionHandler) *big.Int {
 	totalFees := big.NewInt(0)
 	for _, innerTx := range innerTxs {
+		if ed.isMoveBalance(innerTx) {
+			innerTxFee := ed.ComputeMoveBalanceFee(innerTx)
```
sasurobert, line 347: is this function used in other places than transactionProcessor ?

sstanculeanu: it is also called from `ComputeTxFee`

process/economics/economicsData.go
sstanculeanu: I was afraid this won't be enough.. will think about another solution.. having txTypeHandler here won't work, as it is created later on, during process components creation

sstanculeanu: added a setter for the real tx type handler and used it

sasurobert (dismissed):

process/transaction/shardProcess.go
```diff
@@ -539,10 +539,6 @@ func (txProc *txProcessor) processMoveBalance(
 		txProc.txFeeHandler.ProcessTransactionFee(moveBalanceCost, big.NewInt(0), txHash)
 	}
 
-	if len(tx.RelayerAddr) > 0 {
```
AdoAdoAdo, line 542: Can you remind me why is this no longer needed?
I remember discussing about this, but don't remember exactly the reason.

factory/api/apiResolverFactory.go
```diff
@@ -185,6 +185,11 @@ func CreateApiResolver(args *ApiResolverArgs) (facade.ApiResolver, error) {
 		return nil, err
 	}
 
+	err = args.CoreComponents.EconomicsData().SetTxTypeHandler(txTypeHandler)
```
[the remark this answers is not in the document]
sstanculeanu: true, removed some of the overwrites that were not needed

integrationTests/vm/txsFee/multiShard/relayedMoveBalance_test.go
```diff
@@ -267,14 +264,21 @@ func testRelayedMoveBalanceExecuteOnSourceAndDestinationRelayerAndInnerTxSenderS
 		require.Nil(t, err)
 
 		// check relayed balance
-		// 100000 - rTxFee(163)*gasPrice(10) - innerTxFee(1000) = 97370
-		utils.TestAccount(t, testContextSource.Accounts, relayerAddr, 1, big.NewInt(97370))
+		// before base cost fix: 100000 - rTxFee(163)*gasPrice(10) - innerTxFee(1000) = 97370
+		//  after base cost fix: 100000 - rTxFee(163)*gasPrice(10) - innerTxFee(10)   = 98360
+		expectedConsumedFee := big.NewInt(97370)
```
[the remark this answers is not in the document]
sstanculeanu: renamed

integrationTests/vm/txsFee/multiShard/relayedMoveBalance_test.go
```diff
@@ -267,14 +264,21 @@ func testRelayedMoveBalanceExecuteOnSourceAndDestinationRelayerAndInnerTxSenderS
...
+		expectedAccumulatedFees := big.NewInt(2630)
+		if relayedFixActivationEpoch != integrationTests.UnreachableEpoch {
+			expectedConsumedFee = big.NewInt(98360)
+			expectedAccumulatedFees = big.NewInt(1640)
```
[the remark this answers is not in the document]
sstanculeanu: same reason, due to price modifier = 1, the computation returns a higher number before fix

process/transaction/shardProcess.go
sstanculeanu: this is not needed as now we'll only use the exact gas needed for move balance, thus no refund will happen. Also, this was wrong here, it would apply for all inner txs, not move balance only as it was supposed to be

AdoAdoAdo (approved):

sasurobert (approved):

PR #6282, opened 2024-06-25
native esdt not printed as issued esdt

axenteoctavian:
- NativeESDT is printed as Issued ESDT
- 
- 
  
## Proposed changes
- Don't print tokens from ESDT SC Address if they contain prefix `ELRONDesdt`
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -23,6 +23,9 @@ import (
 	"github.com/multiversx/mx-chain-core-go/data/transaction"
 	"github.com/multiversx/mx-chain-core-go/data/validator"
 	disabledSig "github.com/multiversx/mx-chain-crypto-go/signing/disabled/singlesig"
+	logger "github.com/multiversx/mx-chain-logger-go"
+	vmcommon "github.com/multiversx/mx-chain-vm-common-go"
+
 	"github.com/multiversx/mx-chain-go/common"
 	"github.com/multiversx/mx-chain-go/common/errChan"
 	"github.com/multiversx/mx-chain-go/dataRetriever"
@@ -41,13 +44,12 @@ import (
 	"github.com/multiversx/mx-chain-go/trie"
 	"github.com/multiversx/mx-chain-go/vm"
 	"github.com/multiversx/mx-chain-go/vm/systemSmartContracts"
-	logger "github.com/multiversx/mx-chain-logger-go"
-	vmcommon "github.com/multiversx/mx-chain-vm-common-go"
 )
 
 const (
 	// esdtTickerNumChars represents the number of hex-encoded characters of a ticker
 	esdtTickerNumChars = 6
+	baseESDTKeyPrefix  = core.ProtectedKeyPrefix + core.ESDTKeyIdentifier
 )
 
 var log = logger.GetOrCreate("node")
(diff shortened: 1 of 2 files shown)


raduchis (approved):

node/node.go
```diff
@@ -247,6 +249,10 @@ func (n *Node) baseGetAllIssuedESDTs(tokenType string, ctx context.Context) ([]s
 			continue
 		}
 
+		if strings.HasPrefix(tokenName, baseESDTKeyPrefix) {
```
mariusmihaic, line 252: This is wrong.

The correct approach would be to filter out the native token id from config only inside `sovereignNode.go`, not touching base code.

mariusmihaic (changes requested):

node/node.go
axenteoctavian: Ok. Changed the implementation.

node/chainSimulator/components/nodeFacade.go
```diff
@@ -102,6 +102,7 @@ func (node *testOnlyProcessingNode) createFacade(configs config.Configs, apiInte
 		nodePack.WithNodeStopChannel(node.CoreComponentsHolder.ChanStopNodeProcess()),
 		nodePack.WithImportMode(configs.ImportDbConfig.IsImportDBMode),
 		nodePack.WithESDTNFTStorageHandler(node.ProcessComponentsHolder.ESDTDataStorageHandlerForAPI()),
+		nodePack.WithNativeESDT(configs.GeneralConfig.SovereignConfig.GenesisConfig.NativeESDT),
```
mariusmihaic, line 105: This is not ok.
This option is only needed for sovereign node, not for normal node, which doesn't need to know about nativeESDT.
As I've suggested before, this should be an extra option for sovereign node **only**. This should be passed from main file of sovereign chain as extra option. In case of chain simulator, it also needs to be passed as extra option from `NewSovereignChainSimulator` constructor.

node/nodeHelper.go
```diff
@@ -107,6 +108,7 @@ func CreateNode(
 		WithNodeStopChannel(coreComponents.ChanStopNodeProcess()),
 		WithImportMode(isInImportMode),
 		WithESDTNFTStorageHandler(processComponents.ESDTDataStorageHandlerForAPI()),
+		WithNativeESDT(config.SovereignConfig.GenesisConfig.NativeESDT),
```
mariusmihaic, line 111: This should be deleted, this option should come from `extraOptions`

node/options.go
```diff
@@ -352,3 +353,11 @@ func WithESDTNFTStorageHandler(storageHandler vmcommon.ESDTNFTStorageHandler) Op
 		return nil
 	}
 }
+
+// WithNativeESDT sets the native esdt
+func WithNativeESDT(nativeESDT string) Option {
```
mariusmihaic, line 358: This should set data only for sovereign node

mariusmihaic (changes requested):

node/chainSimulator/components/nodeFacade.go
axenteoctavian: Added `NativeESDT` in sovereign node factory constructor, this way is only in sovereign components

node/sovereignNode_test.go
```diff
@@ -4,11 +4,14 @@ import (
...
 
 	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/stretchr/testify/require"
+
 	"github.com/multiversx/mx-chain-go/errors"
 	"github.com/multiversx/mx-chain-go/node"
-	"github.com/stretchr/testify/require"
 )
 
+var nativeEsdt = "WEGLD-bd4d79"
```
mariusmihaic, line 13:
```suggestion
var nativeESDT = "WEGLD-bd4d79"
```

mariusmihaic (approved):

raduchis (approved):

PR #6311, opened 2024-07-11
Update peer state and sc to protocol

mariusmihaic:
- Update peers state and sc to protocol in sovereign chain block processing
  
## Proposed changes
- Implemented `updateShardDataPeerState` to work **only for current header.** Original Implementation (for metachain) had metaBlock notarized shard headers from cache to be used when updating validators data (current block -1).
- `validatorStatisticsProcessor.UpdatePeerState` in sov chain processing at `processEpochStartMetaBlock` level
- `scToProtocol.UpdateProtocol` called for `postProcessMBs` after processing all sovereign txs

## Testing procedure
- Integration tests with staking new nodes which also go through auction

## Pre-requisites


sasurobert (approved):

cmd/sovereignnode/chainSimulator/tests/epochChange/epochChange_test.go
```diff
@@ -55,13 +64,35 @@ func TestSovereignChainSimulator_EpochChange(t *testing.T) {
 
 	defer cs.Close()
 
-	trie := cs.GetNodeHandler(0).GetStateComponents().TriesContainer().Get([]byte(dataRetriever.PeerAccountsUnit.String()))
+	nodeHandler := cs.GetNodeHandler(0)
```
axenteoctavian, line 67: Would be better to use `core.SovereignChainShardId` instead of `0`

factory/processing/processComponents_test.go
```diff
@@ -1012,6 +1023,8 @@ func getRunTypeComponents(rt runType.RunTypeComponentsHolder) *mainFactoryMocks.
 		GenesisMetaBlockChecker:             rt.GenesisMetaBlockCheckerCreator(),
 		NodesSetupCheckerFactoryField:       rt.NodesSetupCheckerFactory(),
 		EpochStartTriggerFactoryField:       rt.EpochStartTriggerFactory(),
+		LatestDataProviderFactoryField:      rt.LatestDataProviderFactory(),
```
axenteoctavian, line 1026: :+1: 
Can we also have a test for nil `LatestDataProviderFactoryField`? We missed this in the previous PR

cmd/sovereignnode/chainSimulator/tests/epochChange/epochChange_test.go
mariusmihaic: Right, did as suggested

factory/processing/processComponents_test.go
mariusmihaic: Test was added in previous PR

axenteoctavian (approved):

sasurobert (approved):

PR #6315, opened 2024-07-15
[sovereign] Save validators info shard data epoch change

mariusmihaic:
- Save validator info data for peer miniblocks in sovereign chain block processing part
  
## Proposed changes
- No need to broadcast data for epoch start peer mbs and intercept it to be added in pool
- When creating validator infos, we will save them directly in pool inside `sovereignValidatorInfoCreator`, which will be called at processing epoch start block level in `sovereignChainBlock`

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/epochStart/metachain/sovereignValidators_test.go
+++ b/epochStart/metachain/sovereignValidators_test.go
@@ -1,0 +1,81 @@
+package metachain
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/data/block"
+	"github.com/multiversx/mx-chain-go/common"
+	"github.com/multiversx/mx-chain-go/dataRetriever"
+	"github.com/multiversx/mx-chain-go/state"
+	"github.com/multiversx/mx-chain-go/testscommon"
+	dataRetrieverMock "github.com/multiversx/mx-chain-go/testscommon/dataRetriever"
+	vics "github.com/multiversx/mx-chain-go/testscommon/validatorInfoCacher"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSovereignValidatorInfoCreator_CreateMarshalledData(t *testing.T) {
+	t.Parallel()
+
+	arguments := createMockEpochValidatorInfoCreatorsArguments()
+
+	svi1 := &state.ShardValidatorInfo{PublicKey: []byte("x")}
+	marshalledSVI1, _ := arguments.Marshalizer.Marshal(svi1)
+
+	svi2 := &state.ShardValidatorInfo{PublicKey: []byte("y")}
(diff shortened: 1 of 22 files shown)


epochStart/metachain/sovereignValidators_test.go
```diff
@@ -0,0 +1,191 @@
...
+	t.Run("CreateMarshalledData should return empty slice when tx hash does not exist in validator info cacher", func(t *testing.T) {
+		t.Parallel()
+
+		arguments := createMockEpochValidatorInfoCreatorsArguments()
+		localErr := errors.New("local error")
+		arguments.DataPool = &dataRetrieverMock.PoolsHolderStub{
+			CurrEpochValidatorInfoCalled: func() dataRetriever.ValidatorInfoCacher {
+				return &vics.ValidatorInfoCacherStub{
+					GetValidatorInfoCalled: func(validatorInfoHash []byte) (*state.ShardValidatorInfo, error) {
+						return nil, localErr
```
axenteoctavian, line 130: Maybe add another `wasCalled = true` here and `require.True` at line 150

epochStart/metachain/sovereignValidators_test.go
```diff
@@ -0,0 +1,191 @@
...
+						return &state.ShardValidatorInfo{}, nil
+					},
+				}
+			},
+		}
+		localErr := errors.New("local error")
+
+		arguments.Marshalizer = &testscommon.MarshallerStub{
+			MarshalCalled: func(obj interface{}) ([]byte, error) {
+
```
axenteoctavian, line 169: Maybe add another `wasMarshallCalled = true` here and `require.True` at line 188.
Or delete this empty line.

epochStart/metachain/sovereignValidators_test.go
mariusmihaic: No need, since I already check the return to be empty, so that wasn't called.

epochStart/metachain/sovereignValidators_test.go
mariusmihaic: No need, I already check `shardDataCacher.Keys` to be empty

epochStart/metachain/sovereignValidators_test.go
axenteoctavian: This was called every time, that's why marshal was never called and marshalledData is empty. 
I would say makes sense to check this was called at least once, if not for every txHash. But it's not really mandatory.

epochStart/metachain/sovereignValidators_test.go
axenteoctavian: `shardDataCacher.Keys` is always empty and `ValidatorsInfoCalled` is never called in this test.

sasurobert (approved):

axenteoctavian (approved):

PR #6322, opened 2024-07-17
Optimize "DisplayProcessTxDetails": Early exit if log level is not TRACE

andreibancioiu:
- The function `DisplayProcessTxDetails` is called in a few places during transactions processing. Its main concern is to display some data by means of Log.Trace(). In this regard, it performs some computation, which takes some non-negligible time (analyzed with _pprof_).

## Proposed changes
- Skip the computations performed by  `DisplayProcessTxDetails` if it should display nothing, given the current log level.

## Testing procedure
- Standard testing.

## Pre-requisites

### DIFF
--- a/process/common.go
+++ b/process/common.go
@@ -680,6 +680,10 @@ func DisplayProcessTxDetails(
 	txHash []byte,
 	addressPubkeyConverter core.PubkeyConverter,
 ) {
+	if log.GetLevel() > logger.LogTrace {
+		return
+	}
+
 	if !check.IfNil(accountHandler) {
 		account, ok := accountHandler.(state.UserAccountHandler)
 		if ok {


mariusmihaic (approved):

miiu96 (approved):

danidrasovean (approved):
Normal allin test: v1.7.13-dev-config-aebb55e8e0 -> optimize-display-process-t-3b72d1056a

--- Specific errors --- 

block hash does not match 1922
wrong nonce in block 860
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  1
Nr. of all WARNS:  890
Nr. of new ERRORS:  1
Nr. of new WARNS:  10
Nr. of PANICS:  0

/------/

PR #6338, opened 2024-07-29
Optimization: encode hash to hex only on log TRACE level set

cristure:

Reasonable optimization on trie updates. Improved update times by roughly 30%.

Before fix:

```
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkPatriciaMerkleTrie_Update-8   	      50	2007485284 ns/op
ok  	github.com/multiversx/mx-chain-go/trie	131.749s

BenchmarkPatriciaMerkleTrie_Update-8   	      50	1991358120 ns/op
ok  	github.com/multiversx/mx-chain-go/trie	132.290s
```

After fix:

```
BenchmarkPatriciaMerkleTrie_Update-8   	      50	1342203572 ns/op
ok  	github.com/multiversx/mx-chain-go/trie	95.049s

BenchmarkPatriciaMerkleTrie_Update-8   	      50	1381106022 ns/op
ok  	github.com/multiversx/mx-chain-go/trie	97.326s
```

## Proposed changes
- Added if statement before logging.

## Testing procedure
- `go test -run=^$ -bench ^BenchmarkPatriciaMerkleTrie_Update$ -benchtime=10x`
- Observe times with from the target branch and compare with the times from this branch.
- Standard testing

## Pre-requisites


andreibancioiu (dismissed):

trie/patriciaMerkleTrie_test.go
```diff
@@ -1708,3 +1709,32 @@ func BenchmarkPatriciaMerkleTrie_RootHashAfterChanging30000NodesInBatchesOf200(b
 		}
 	}
 }
+
+func TestTrieUpdateTimer(t *testing.T) {
```
andreibancioiu, line 1713: This unit test:
 - is skipped
 - has no assertions
 - does a print line

Instead, let's get the output of this test, or of the existing benchmarks - before & after the fix, and put the output (textual) in the PR description. Then, we can drop this test.

AdoAdoAdo: yes, we can remove the test after we have the results.

andreibancioiu: Fixed (transformed into a benchmark test).

AdoAdoAdo (approved):

miiu96 (approved):

andreibancioiu (approved):

PR #6377, opened 2024-08-06
Integrate sys tests ci2

btc-fan:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
+++ b/.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
@@ -21,7 +21,8 @@ jobs:
   build-and-test:
     if: |
       github.event_name == 'pull_request' ||
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, 'Run Tests:'))
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, 'Run Tests:')) ||
+      github.event_name == 'workflow_dispatch'
 
     strategy:
       matrix:
@@ -35,17 +36,6 @@ jobs:
       MX_CHAIN_TESTING_SUITE_TARGET_BRANCH: ""
 
     steps:
-      - name: Debug Comment
-        if: github.event_name == 'issue_comment'
-        run: |
-          echo "Comment body: ${{ github.event.comment.body }}"
-          echo "Is PR: ${{ github.event.issue.pull_request }}
-
-      - name: Log Trigger Information
-        run: |
-          echo "Event: ${{ github.event_name }}"
-          echo "Action: ${{ github.event.action }}"
-
       - name: Parse Comment for Branches
         if: github.event_name == 'issue_comment'
         run: |
@@ -62,6 +52,10 @@ jobs:
               TESTING_SUITE_BRANCH=$(echo "$COMMENT" | grep "mx-chain-testing-suite:" | awk -F': ' '{print $2}')
               echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${TESTING_SUITE_BRANCH}" >> $GITHUB_ENV
             fi
+          else
+            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+          fi
 
       - name: Set up Go 1.20.7
         uses: actions/setup-go@v3
@@ -91,20 +85,33 @@ jobs:
         id: target_branch
         run: |
           echo "CURRENT_BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF_NAME}}" >> $GITHUB_ENV
-          if [[ "${{ github.event.pull_request.base.ref }}" == "main" ]]; then
-            echo "TARGET_BRANCH=main" >> $GITHUB_ENV
-            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=main" >> $GITHUB_ENV
-            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
-          elif [[ "${{ github.event.pull_request.base.ref }}" == "master" ]]; then
-            echo "TARGET_BRANCH=master" >> $GITHUB_ENV
-            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=main" >> $GITHUB_ENV
-            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
-          elif [[ "${{ github.event.pull_request.base.ref }}" == rc/* || "${{ github.event.pull_request.base.ref }}" == integrate-sys-tests-ci* ]]; then
-            echo "TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
-            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
-            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+
+          # Default to comment branches if they are set, otherwise use existing logic
+          if [ -n "${{ env.MX_CHAIN_SIMULATOR_TARGET_BRANCH }}" ]; then
+            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=${{ env.MX_CHAIN_SIMULATOR_TARGET_BRANCH }}" >> $GITHUB_ENV
+          else
+            if [[ "${{ github.event.pull_request.base.ref }}" == "main" ]]; then
+              echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == "master" ]]; then
+              echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == rc/* || "${{ github.event.pull_request.base.ref }}" == integrate-sys-tests-ci* ]]; then
+              echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+            fi
+          fi
+
+          if [ -n "${{ env.MX_CHAIN_TESTING_SUITE_TARGET_BRANCH }}" ]; then
+            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${{ env.MX_CHAIN_TESTING_SUITE_TARGET_BRANCH }}" >> $GITHUB_ENV
+          else
+            if [[ "${{ github.event.pull_request.base.ref }}" == "main" ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == "master" ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == rc/* || "${{ github.event.pull_request.base.ref }}" == integrate-sys-tests-ci* ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+            fi
           fi
           echo "MX_CHAIN_GO_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+      
 
       - name: Print Target Branches
         run: |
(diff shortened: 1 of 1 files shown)


btc-fan:
Run Tests:
mx-chain-simulator-go: rc/v1.7.next1
mx-chain-testing-suite: rc/v1.7.next1

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,341 @@
...
+          # Extract branch names from the comment
+          if echo "$COMMENT" | grep -q "mx-chain-simulator-go:"; then
+            SIMULATOR_BRANCH=$(echo "$COMMENT" | grep "mx-chain-simulator-go:" | awk -F': ' '{print $2}')
+            echo "MX_CHAIN_SIMULATOR_TARGET_BRANCH=${SIMULATOR_BRANCH}" >> $GITHUB_ENV
+          fi
+
+          if echo "$COMMENT" | grep -q "mx-chain-testing-suite:"; then
+            TESTING_SUITE_BRANCH=$(echo "$COMMENT" | grep "mx-chain-testing-suite:" | awk -F': ' '{print $2}')
+            echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${TESTING_SUITE_BRANCH}" >> $GITHUB_ENV
+          fi
```
cristure, line 79: This looks great, but wouldn't it be more simple to use an input that defaults to master for the branch name? Also why is the SIMULATOR_BRANCH mentioned here? Shouldn't it be taken from the `github.ref`? Or in the `workflow_dispatch` case use an input instead. I think it is a more clearer approach than adding logic for parsing comments on PRs.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,341 @@
...
+            elif [[ "${{ github.event.pull_request.base.ref }}" == "master" ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == rc/* || "${{ github.event.pull_request.base.ref }}" == integrate-sys-tests-ci* ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+            fi
+          fi
+
+          # Always set MX_CHAIN_GO_TARGET_BRANCH based on the PR base branch
+          echo "MX_CHAIN_GO_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+
```
cristure, line 139: I think it'd simpler and cleaner to just retrieve the target branch from an input.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,341 @@
...
+          pip install -r scripts/update_go_mod/requirements.txt
+          python scripts/update_go_mod/update_go_mod.py $LATEST_COMMIT_HASH
+
+
+      - name: Run go mod tidy and go build
+        run: |
+          cd mx-chain-simulator-go/cmd/chainsimulator
+          go mod tidy
+          go build
+          echo "CHAIN_SIMULATOR_BUILD_PATH=$(pwd)" >> $GITHUB_ENV
```
cristure, line 172: is the binary actually executed? or is it started by the mx-chain-testing-suite?

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,341 @@
...
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+
+      - name: Install Python Dependencies and Update go.mod
+        run: |
+          cd mx-chain-simulator-go
+          pip install -r scripts/update_go_mod/requirements.txt
+          python scripts/update_go_mod/update_go_mod.py $LATEST_COMMIT_HASH
```
cristure, line 164: Have you considered using containers instead?

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,341 @@
...
+
+      - name: Fail job if tests failed
+        if: always()
+        run: |
+          if [ "${{ env.PYTEST_EXIT_CODE }}" != "0" ]; then
+            echo "Tests failed with exit code ${{ env.PYTEST_EXIT_CODE }}"
+            exit 1
+          else
+            echo "Tests passed successfully."
+          fi
```
cristure, line 341: Just my 2 cents. I am leaving this here for posterity.

Don't get me wrong, I appreciate your effor into building this and believe it can offer great value to our organizaction, but the overall design is too complex and could be spread out into more parts. CI pipelines shouldn't be too complex in nature and should rely more on the binaries it invokes. Such as enviornment provisioning for example, one could more easily start a container than checking out a repository, building a binary with the correct go version and exposing it as an environment variable. This could be handled in the `mx-chain-testing-suite` repository for instance.

For instance `mx-chain-testing-suite` already has a job that triggers the tests. We can trigger that job from this respective workflow through a webhook and wait for its response. In the webhook's body you can put the dependencies required to build the `chain-simulator` binary needed for the tests. It would be easier with containers, but that requires a bit extra overhead, such as a private container registry, but from my knowledge we do have something that could be used @liviuf-ancas @steliandoruzaharia.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: Both methods are supported, but the target branch is prioritized by default.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: IT is started by the mx-chain-testing-suite

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: No :)

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: It's good to have a containerized testing framework, and we can consider that for future improvements. For now, let's focus on the current implementation and at least start with what we have.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
+name: Build and Smoke Test
```
miiu96, line 1: Maybe you can find a better name.

miiu96 (approved):

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+
+            if (latestComment) {
+              core.setOutput('latest_comment', latestComment.body);
+            } else {
+              core.setOutput('latest_comment', '');
+            }
+        env:
+          LATEST_COMMENT: ${{ steps.fetch_comment.outputs.latest_comment }}
+
+      - name: Parse Comment for Branches
```
andreibancioiu, line 62: Oh, now I see, some parsing is done, directly on the comment.

How about workflow dispatch with input fields (a bit less code)?

https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
andreibancioiu: About workflow inputs - GH also provides a nice user interface (a simple form) to provide them. Structured input vs. PR comments.

> specify branches directly in comments

Branches can also be specified in the simple GH form.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
andreibancioiu: Oh, now I see the discussion below (on GH inputs).

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+        uses: actions/checkout@v4
+        with:
+          repository: 'multiversx/mx-chain-go'
+          ref: ${{ env.MX_CHAIN_GO_TARGET_BRANCH || github.head_ref || github.ref }}
+          fetch-depth: 0
+          path: 'mx-chain-go'
+
+      - name: Get Latest Commit Hash
+        run: |
+          cd mx-chain-go
```
andreibancioiu, line 97: Above, you've used github-script (very nice!). Not applicable here?

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+          else
+            if [[ "${{ github.event.pull_request.base.ref }}" == "main" ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
+            elif [[ "${{ github.event.pull_request.base.ref }}" == "master" ]]; then
+              echo "MX_CHAIN_TESTING_SUITE_TARGET_BRANCH=main" >> $GITHUB_ENV
+            fi
+          fi
+
+          # Always set MX_CHAIN_GO_TARGET_BRANCH based on the PR base branch
+          echo "MX_CHAIN_GO_TARGET_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
```
andreibancioiu, line 134: GH inputs with default values seem nicer, IMHO.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+          repository: 'multiversx/mx-chain-testing-suite'
+          path: 'mx-chain-testing-suite'
+          fetch-depth: 0
+          ref: ${{ env.MX_CHAIN_TESTING_SUITE_TARGET_BRANCH || github.event.pull_request.base.ref }}
+          token: ${{ secrets.MVX_TESTER_GH_TOKEN }}
+
+      - name: Install Dependencies
+        run: |
+          pip install -r mx-chain-testing-suite/requirements.txt
+          echo "PYTHONPATH=mx-chain-testing-suite" >> $GITHUB_ENV
```
andreibancioiu, line 181: Maybe set PYTHONPATH where needed (below), before invoking the scenarios?

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+          fi
+
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: pytest-report-${{ github.run_id }}
+          path: reports/report.html
+
+      - name: Deploy Report to GitHub Pages
```
andreibancioiu, line 207: GH pages are public, and cannot be made private. Could this be a security concern?

andreibancioiu (approved):

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+permissions:
+  issues: write
+  pull-requests: write
+  contents: read
+
+jobs:
+  build-and-test:
+    if: |
+      github.event_name == 'pull_request' ||
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, 'Run Tests:')) ||
```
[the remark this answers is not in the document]
btc-fan: You’re correct that workflow_dispatch can trigger the workflow manually. However, including issue_comment allows for more flexibility, like running specific tests based on comments.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
```diff
@@ -0,0 +1,336 @@
...
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+      - 'master'
+      - 'rc/*'
+  workflow_dispatch:
+  issue_comment:
+    types: [created]
```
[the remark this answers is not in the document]
btc-fan: Yes, this line allows triggering based on specific comments

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: Given that this is the only repo using Python, and the need to append the current path to PYTHONPATH, it might be cleaner to set PYTHONPATH directly where it's needed, right before invoking the scenarios.

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: The reports are public, which is strange as mx-chain-testing-repo is private. I just found out that they are accessible even if you are outside the organization. To make them private we need to host them somewhere else or pay for GitHub Enterprise (which is not a suitable option I believe).
For now, we can leave them as it is, let's test the flow, if everything works fine then we can think about where to store the reports privately.

image

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: It seems easier to get the hash by running some bash commands :)

.github/workflows/build_and_run_chain_simulator_and_execute_system_test.yml
btc-fan: done.

sstanculeanu (approved):

PR #6436, opened 2024-09-04
Mx 15065 fixed order consensus

AdoAdoAdo:
- moving to 400 sized consensus group, it is no longer required to shuffle the order of signing, as every eligible node is required to sign.
- this simplifies the equivalent proof mechanics verification as it can be almost self contained during an epoch.
- this also makes it simpler to create consensus zk proofs
  
## Proposed changes
- make the order of signing fixed during the epoch, and use the ordering of nodes in the eligible list which is defined at the start of the epoch
- explicitly return the consensus leader/proposer
- add activation flag for the transition from dynamic ordering to fixed ordering

## Testing procedure
- regular system tests transitioning from dynamic ordering to fixed ordering, consensus & operation transitions successfully, blocks are added to the shard chains and notarization continues.
- rewards checks - rewards converge to the same value for each node over a longer period of time
- check number of selections as consensus leader for each node over a longer period of time, converges to the same value.

## Pre-requisites


common/constants.go
```diff
@@ -1223,5 +1223,6 @@ const (
 	EGLDInESDTMultiTransferFlag                        core.EnableEpochFlag = "EGLDInESDTMultiTransferFlag"
 	CryptoOpcodesV2Flag                                core.EnableEpochFlag = "CryptoOpcodesV2Flag"
 	EquivalentMessagesFlag                             core.EnableEpochFlag = "EquivalentMessagesFlag"
+	FixedOrderInConsensusEnableEpochFlag               core.EnableEpochFlag = "FixedOrderInConsensusEnableEpoch"
```
sstanculeanu, line 1226:
```suggestion
	FixedOrderInConsensusFlag               core.EnableEpochFlag = "FixedOrderInConsensusFlag"
```

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/beevik/ntp v1.3.0
+	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
```
sstanculeanu, line 7: looks like it is not used anywhere.. `go mod tidy`

integrationTests/multiShard/block/interceptedHeadersSigVerification/interceptedHeadersSigVerification_test.go
```diff
@@ -160,6 +162,7 @@ func TestInterceptedShardBlockHeaderWithLeaderSignatureAndRandSeedChecks(t *test
 		t.Skip("this is not a short test")
 	}
 
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu, line 165: is this needed?

AdoAdoAdo: forgot to remove

sstanculeanu (approved):

miiu96 (approved):

consensus/spos/consensusState.go
```diff
@@ -151,7 +152,7 @@ func (cns *ConsensusState) GetLeader() (string, error) {
 		return "", ErrEmptyConsensusGroup
 	}
 
-	return cns.consensusGroup[IndexOfLeaderInConsensusGroup], nil
```
ssd04, line 154: remove `IndexOfLeaderInConsensusGroup`? as it should not be used anymore, it can be done on next PR

AdoAdoAdo: done in #6448

ssd04 (approved):

PR #6457, opened 2024-09-12
equivalentProofsTopic + interceptor

sstanculeanu:
- equivalent proofs will be sent on a new topic
  
## Proposed changes
- added new topic
- added interceptor

## Testing procedure

## Pre-requisites

### DIFF
--- a/cmd/node/factory/interface.go
+++ b/cmd/node/factory/interface.go
@@ -16,6 +16,7 @@ type HeaderSigVerifierHandler interface {
 	VerifySignature(header data.HeaderHandler) error
 	VerifySignatureForHash(header data.HeaderHandler, hash []byte, pubkeysBitmap []byte, signature []byte) error
 	VerifyPreviousBlockProof(header data.HeaderHandler) error
+	VerifyHeaderProof(headerProof data.HeaderProofHandler) error
 	IsInterfaceNil() bool
 }
 
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -80,7 +80,14 @@ func createEquivalentProof(marshaller marshal.Marshalizer, buff []byte) (*block.
 		return nil, err
 	}
 
-	log.Trace("interceptedEquivalentProof successfully created")
+	log.Trace("interceptedEquivalentProof successfully created",
+		"header hash", logger.DisplayByteSlice(headerProof.HeaderHash),
+		"header shard", headerProof.HeaderShardId,
+		"header epoch", headerProof.HeaderEpoch,
+		"header nonce", headerProof.HeaderNonce,
+		"bitmap", logger.DisplayByteSlice(headerProof.PubKeysBitmap),
+		"signature", logger.DisplayByteSlice(headerProof.AggregatedSignature),
+	)
 
 	return headerProof, nil
 }
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -9,7 +9,6 @@ import (
 	"github.com/multiversx/mx-chain-core-go/data/block"
 	"github.com/multiversx/mx-chain-core-go/marshal"
 	"github.com/multiversx/mx-chain-go/consensus"
-	"github.com/multiversx/mx-chain-go/dataRetriever"
 	"github.com/multiversx/mx-chain-go/process"
 	"github.com/multiversx/mx-chain-go/sharding"
 	logger "github.com/multiversx/mx-chain-logger-go"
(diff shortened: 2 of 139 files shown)


process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func createEquivalentProof(marshaller marshal.Marshalizer, buff []byte) (*block.HeaderProof, error) {
+	headerProof := &block.HeaderProof{}
+	err := marshaller.Unmarshal(headerProof, buff)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Trace("interceptedEquivalentProof successfully created")
```
ssd04, line 83: add more info here? like headerHash

process/factory/interceptorscontainer/metaInterceptorsContainerFactory.go
```diff
@@ -329,6 +336,39 @@ func (micf *metaInterceptorsContainerFactory) generateRewardTxInterceptors() err
 	return micf.addInterceptorsToContainers(keys, interceptorSlice)
 }
 
+func (micf *metaInterceptorsContainerFactory) generateEquivalentProofsInterceptors() error {
+	shardC := micf.shardCoordinator
+	noOfShards := shardC.NumberOfShards()
+
+	keys := make([]string, noOfShards+1)
```
ssd04, line 343: why `+1` here? since for meta it is added next

process/block/interceptedBlocks/interceptedEquivalentProof.go
sstanculeanu: improved this log

process/interceptors/factory/interceptedEquivalentProofsFactory.go
```diff
@@ -0,0 +1,44 @@
...
+	headers           dataRetriever.HeadersPool
+}
+
+// NewInterceptedEquivalentProofsFactory creates a new instance of interceptedEquivalentProofsFactory
+func NewInterceptedEquivalentProofsFactory(args ArgInterceptedDataFactory) *interceptedEquivalentProofsFactory {
+	return &interceptedEquivalentProofsFactory{
+		marshaller:        args.CoreComponents.InternalMarshalizer(),
+		shardCoordinator:  args.ShardCoordinator,
+		headerSigVerifier: args.HeaderSigVerifier,
+		headers:           args.HeadersPool,
```
AdoAdoAdo, line 25: why do we need the headers pool?
With the latest version of the proof we do not need to check any header, just the proof itself.

process/factory/interceptorscontainer/metaInterceptorsContainerFactory.go
ssd04: aa, right

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return err
+	}
+
+	hdr, err := iep.headers.GetHeaderByHash(iep.proof.HeaderHash)
+	if err != nil {
+		return err
+	}
+
+	return iep.headerSigVerifier.VerifySignatureForHash(hdr, iep.proof.HeaderHash, iep.proof.PubKeysBitmap, iep.proof.AggregatedSignature)
```
AdoAdoAdo, line 116: I don't think we need the header to verify the equivalent proof (at least not in its final form)
Is the header only used for the randomness? We no longer need it as we don't care if the leader has signed or not, and we have the bitmap which refers to the fixed order.

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -0,0 +1,170 @@
...
+
+const interceptedEquivalentProofType = "intercepted equivalent proof"
+
+// ArgInterceptedEquivalentProof is the argument used in the intercepted equivalent proof constructor
+type ArgInterceptedEquivalentProof struct {
+	DataBuff          []byte
+	Marshaller        marshal.Marshalizer
+	ShardCoordinator  sharding.Coordinator
+	HeaderSigVerifier consensus.HeaderSigVerifier
+	Headers           dataRetriever.HeadersPool
```
AdoAdoAdo, line 26: I don't think we need the headers pool with the latest changes in the proof structure

process/factory/interceptorscontainer/metaInterceptorsContainerFactory.go
```diff
@@ -99,6 +100,7 @@ func NewMetaInterceptorsContainerFactory(
 		SignaturesHandler:            args.SignaturesHandler,
 		HeartbeatExpiryTimespanInSec: args.HeartbeatExpiryTimespanInSec,
 		PeerID:                       args.MainMessenger.ID(),
+		HeadersPool:                  args.DataPool.Headers(),
```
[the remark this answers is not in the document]
sstanculeanu: removed and updated the signature check as discussed

cmd/node/factory/interface.go
```diff
@@ -16,6 +16,7 @@ type HeaderSigVerifierHandler interface {
 	VerifySignature(header data.HeaderHandler) error
 	VerifySignatureForHash(header data.HeaderHandler, hash []byte, pubkeysBitmap []byte, signature []byte) error
 	VerifyPreviousBlockProof(header data.HeaderHandler) error
```
AdoAdoAdo, line 18: do we still need the VerifyPreviousBlockProof? it seems a bit redundant as it would simply extract the proof from the header and call VerifyHeaderProof on it right? Does it also verify the hash is previousBlockHash?

AdoAdoAdo (approved):

ssd04 (approved):

PR #6461, opened 2024-09-13
cross header notarization fix

axenteoctavian:
- The same first dummy cross header is added in the block tracker until a valid cross chain header was received
- 
- 
  
## Proposed changes
- Only if there's a new cross chain header processed then call the save function
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/sovereignChainBlock.go
+++ b/process/block/sovereignChainBlock.go
@@ -1460,9 +1460,11 @@ func (scbp *sovereignChainBlockProcessor) CommitBlock(headerHandler data.HeaderH
 		return err
 	}
 
-	err = scbp.saveLastNotarizedHeader(core.MainChainShardId, processedExtendedShardHdrs)
-	if err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


process/block/sovereignChainBlock.go
```diff
@@ -1460,9 +1460,11 @@ func (scbp *sovereignChainBlockProcessor) CommitBlock(headerHandler data.HeaderH
 		return err
 	}
 
-	err = scbp.saveLastNotarizedHeader(core.MainChainShardId, processedExtendedShardHdrs)
-	if err != nil {
-		return err
+	if len(processedExtendedShardHdrs) > 0 {
+		err = scbp.saveLastNotarizedHeader(core.MainChainShardId, processedExtendedShardHdrs)
```
sasurobert, line 1464: why don't we have the len check inside the saveLastNotarizedHeader function ?

mariusmihaic (approved):

sasurobert (approved):

PR #6495, opened 2024-09-24
updated core enums in state changes.

cristure:
- Update enums in mx-chain-core-go
  
## Proposed changes
- Fix references to updated enums.

## Testing procedure
- Usual PR checks should suffice.

## Pre-requisites


cristure:
This PR should be taken into consideration too: https://github.com/multiversx/mx-chain-core-go/pull/324

factory/status/export_test.go
```diff
@@ -30,7 +31,7 @@ func ComputeConnectedPeers(
 	computeConnectedPeers(appStatusHandler, netMessenger, suffix)
 }
 
-// MakeHostDriversArgs -
+// MakerHostDriversArgs -
```
ssd04, line 34:
```suggestion
// MakeHostDriversArgs -
```
revert here?

ssd04 (approved):

errors/errors.go
```diff
@@ -601,3 +601,6 @@ var ErrNilEpochSystemSCProcessor = errors.New("nil epoch system SC processor")
 
 // ErrNilRelayedTxV3Processor signals that a nil relayed tx v3 processor has been provided
 var ErrNilRelayedTxV3Processor = errors.New("nil relayed tx v3 processor")
+
+// ErrOnlyReadStateChangesToCollect signals that only read state changes to collect have been configured to be collected.
+var ErrOnlyReadStateChangesToCollect = errors.New("cannot collect only read state changes")
```
BeniaminDrasovean, line 606: This is not used

BeniaminDrasovean (approved):

errors/errors.go
cristure: right. will remove it

BeniaminDrasovean (approved):

config/config.go
```diff
@@ -308,14 +308,14 @@ type FacadeConfig struct {
...
-	PeerStatePruningEnabled            bool
-	CollectStateChangesEnabled         bool
-	CollectStateChangesWithReadEnabled bool
-	MaxStateTrieLevelInMemory          uint
-	MaxPeerTrieLevelInMemory           uint
-	StateStatisticsEnabled             bool
+	SnapshotsEnabled            bool
+	AccountsStatePruningEnabled bool
+	PeerStatePruningEnabled     bool
+	StateChangesDataAnalysis    bool
```
ssd04, line 314: this change needs to be reflected also in config.toml

factory/state/stateParser_test.go
```diff
@@ -0,0 +1,46 @@
+package state
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStateComponents_ParseStateChangesTypesToCollect(t *testing.T) {
```
ssd04, line 9: add tests for error cases?

state/stateChanges/dataAnalysisCollector.go
```diff
@@ -163,22 +165,49 @@ func (scc *dataAnalysisCollector) Reset() {
 	scc.cachedTxs = make(map[string]*transaction.Transaction)
 }
 
-// Publish will export state changes
-func (scc *dataAnalysisCollector) Publish() error {
+// Publish will retrieve the state changes linked with the tx hash.
+func (scc *dataAnalysisCollector) Publish() (map[string]*data.StateChanges, error) {
```
ssd04, line 169: is this correct? this is the same as for normal collector.
my suggestion was to store into levelDB on this Publish, and avoid having `Store` function, since this component is used only for data analysis, not for the production flow, at least currently

ssd04: this will also save normal state changes to outport driver when running in data analysis mode, which i don't think we want

ssd04 (changes requested):

state/stateChanges/dataAnalysisCollector.go
cristure: yes, but in the current implementation if you don't have the outport driver configured you will never `publish`, hence the `store` function. If you have data analysis enabled and the host driver off, you will still have them persisted in leveldb.

config/config.go
cristure: done

factory/state/stateComponents.go
```diff
@@ -122,34 +125,48 @@ func (scf *stateComponentsFactory) Create() (*stateComponents, error) {
 }
 
 func (scf *stateComponentsFactory) createStateChangesCollector() (state.StateChangesCollector, error) {
```
ssd04, line 127: state changes collector is created 2 times in this file, it should be creating only once for main accounts db and then passed to peer accounts

ssd04: not related to this PR, but please fix this here

ssd04 (changes requested):

factory/state/stateComponents.go
BeniaminDrasovean: I think we should have a different collector for PeerAccounts that can be disabled separately from config.

factory/state/stateComponents.go
```diff
@@ -125,6 +125,24 @@ func (scf *stateComponentsFactory) Create() (*stateComponents, error) {
...
+		return nil, fmt.Errorf("failed to parse state changes types to collect: %w", err)
+	}
+
+	var opts []stateChanges.CollectorOption
+
+	if collectRead {
+		opts = append(opts, stateChanges.WithCollectRead())
+	}
+	if collectWrite {
+		opts = append(opts, stateChanges.WithCollectRead())
```
BeniaminDrasovean, line 143: .WithCollectWrite()

cristure: :+1:

BeniaminDrasovean (approved):

ssd04 (approved):

PR #6521, opened 2024-10-04
Add "outcomeEmitted" event

andreibancioiu:


sasurobert (changes requested):
no need for extra log. the solution has to be something else.

PR #6523, opened 2024-10-07
[sovereign] FIX: Save metrics panic

mariusmihaic:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/sovereignChainBlock.go
+++ b/process/block/sovereignChainBlock.go
@@ -727,14 +727,10 @@ func (scbp *sovereignChainBlockProcessor) ProcessBlock(headerHandler data.Header
 		return nil, nil, process.ErrWrongTypeAssertion
 	}
 
-	go getMetricsFromBlockBody(body, scbp.marshalizer, scbp.appStatusHandler)
-
 	txCounts, rewardCounts, unsignedCounts := scbp.txCounter.getPoolCounts(scbp.dataPool)
 	log.Debug("total txs in pool", "counts", txCounts.String())
 	log.Debug("total txs in rewards pool", "counts", rewardCounts.String())
 	log.Debug("total txs in unsigned pool", "counts", unsignedCounts.String())
-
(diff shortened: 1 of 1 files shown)


axenteoctavian (dismissed):

process/block/sovereignChainBlock.go
```diff
@@ -792,6 +788,20 @@ func (scbp *sovereignChainBlockProcessor) ProcessBlock(headerHandler data.Header
...
 	}
 
+	newBody := body
+	defer func() {
+		if err != nil {
+			return
+		}
+
+		if !check.IfNil(shardHeader) {
+			go getMetricsFromHeader(headerHandler.ShallowClone(), uint64(txCounts.GetTotal()), scbp.marshalizer, scbp.appStatusHandler)
```
sasurobert, line 798: shardHeader in the go func ?

how can you have nil shardheader at this point ?

Even the body should not be nil.

mariusmihaic: Right, we can get rid of those.
Fixed as suggested

axenteoctavian (approved):

process/block/sovereignChainBlock.go
```diff
@@ -792,6 +788,16 @@ func (scbp *sovereignChainBlockProcessor) ProcessBlock(headerHandler data.Header
 		return nil, nil, err
 	}
 
+	newBody := body
+	defer func() {
+		if err != nil {
```
sasurobert, line 793: this if does not have a reason to be here.

sasurobert (changes requested):

process/block/sovereignChainBlock.go
mariusmihaic: You are absolutely right here, deleted it.

axenteoctavian (approved):

sasurobert (approved):

PR #6535, opened 2024-10-15
Refactor v2 subround block

AdoAdoAdo:
- Simplification of the v2 consensus (built on equivalent proofs) after adding a proxy for switching between versions. This will remove many edge cases as the logic between v1 and v2 will be separated. 
  
## Proposed changes
- Remove the specific consensus v1 handling from the equivalent proofs consensus (v2) block subround.
- Add Leader signature directly on the proposed header to avoid the tampering of the proposed block, since it now gets propagated on the regular topic, not the consensus one, and there is no longer the originator check which was previously available on the consensus message.
- Fix checks for starting to process a block
- Verify the leader signature over the proposed block header in the interceptor

## Testing procedure
- regular system tests - consensus 
- adversarial system testing - malicious proposer, malicious validators, block tampering
- unit & integration tests

## Pre-requisites


sstanculeanu (approved):

consensus/spos/bls/v2/subroundBlock.go
```diff
@@ -455,22 +375,23 @@ func (sr *subroundBlock) addProofOnHeader(header data.HeaderHandler) bool {
 	// in this case, we set the previous proof as signature and bitmap from the previous header
 	// 2. current node is leader in the first block after sync
 	// in this case, we won't set the proof, return false and wait for the next round to receive a proof
-	currentHeader := sr.Blockchain().GetCurrentBlockHeader()
-	if check.IfNil(currentHeader) {
+	prevBlockHeader := sr.Blockchain().GetCurrentBlockHeader()
+	if check.IfNil(prevBlockHeader) {
 		log.Debug("addProofOnHeader.GetCurrentBlockHeader, nil current header")
```
ssd04, line 380: update maybe also log messages here with prev block header?

AdoAdoAdo: I will take this on the next PR

PR #6544, opened 2024-10-17
full gas on scr of inner of relayed v3

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/transaction/shardProcess.go
+++ b/process/transaction/shardProcess.go
@@ -1078,7 +1078,7 @@ func (txProc *txProcessor) processUserTx(
 			innerTxIdx)
 	}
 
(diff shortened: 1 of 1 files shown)


process/transaction/shardProcess.go
```diff
@@ -1202,6 +1203,17 @@ func (txProc *txProcessor) makeSCRFromUserTx(
 		CallType:       vm.DirectCall,
 	}
 
+	isUserTxOfRelayedV3 := len(originalTx.InnerTransactions) > 0 &&
```
raduchis, line 1206: does this happen also for relayed V1 and V2? could this fix also those? 
if so, I would remove the check isUserTxOfRelayedV3.

sstanculeanu: pushed your suggestion, as it was only applying for v3 (I misunderstood)

PR #6547, opened 2024-10-21
Feat/accounts storage iterator

BeniaminDrasovean:
- We need a way to retrieve tokens from a data trie in a sequential order, but with the possibility of continuing from a checkpoint if the data trie has lots of leaves, thus iterating all of the leaves in multiple accesses.
  
## Proposed changes
- Add new API endpoint `/address/iterate-keys`. This will be a POST request. The request has `numKeys`, `address`, and `iteratorState`. If the iterator state is empty, the iteration will begin from the rootHash. If numKeys is 0, the iteration will stop either when all the leaves have been retrieved or when the request times out.

## Testing procedure
- Benchmark the iteration of the shard 1 system account data trie
- Normal testing procedure
- Testing scenarios

## Pre-requisites


danidrasovean (approved):
Normal allin test: master-37ab7adf91 -> feat/accounts-storage-iterator

--- Specific errors --- 

block hash does not match 19
wrong nonce in block 25
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  10
Nr. of all WARNS:  265
Nr. of new ERRORS:  10
Nr. of new WARNS:  8

/------/ 


--- System test go --- 

/------/

ssd04 (approved):

andreibancioiu (approved):

PR #6562, opened 2024-10-25
[sovereign] Cross chain header storage bootstrap

mariusmihaic:
- Extended header cleanup and apply notarized data from storage on bootstrap
  
## Proposed changes
- Overwrite `applyCrossNotarizedHeaders`, `cleanupNotarizedStorage`, `cleanupNotarizedStorageForHigherNoncesIfExist` for `sovereignChainShardStorageBootstrapper` 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/sync/storageBootstrap/sovereignChainShardStorageBootstrapper.go
axenteoctavian, line 1: In each `log.Debug` you print hash as `[]byte`, isn't it better to use `hex.EncodeToString`?

mariusmihaic: No need, logger takes care of that

process/sync/storageBootstrap/sovereignChainShardStorageBootstrapper.go
```diff
@@ -19,5 +25,118 @@ func NewSovereignChainShardStorageBootstrapper(shardStorageBootstrapper *shardSt
...
+	if !castOk {
+		log.Warn("sovereignChainShardStorageBootstrapper.cleanupNotarizedStorage",
+			"error", process.ErrWrongTypeAssertion,
+			"expected shard header of type", "SovereignChainHeaderHandler",
+		)
+		return
+	}
+
+	for _, extendedHeaderHash := range sovereignHeader.GetExtendedShardHeaderHashes() {
+		var extendedHeader data.HeaderHandler
```
sasurobert, line 76: this line can be moved outside of the for.

process/sync/storageBootstrap/sovereignChainShardStorageBootstrapper.go
```diff
@@ -19,5 +25,118 @@ func NewSovereignChainShardStorageBootstrapper(shardStorageBootstrapper *shardSt
...
+		nonce++
+
+		extendedBlock, extendedBlockHash, err := process.GetExtendedHeaderFromStorageWithNonce(
+			nonce,
+			ssb.store,
+			ssb.uint64Converter,
+			ssb.marshalizer,
+		)
+		if err != nil {
+			log.Debug("extended block is not found in ExtendedShardHeadersNonceHashDataUnit storage",
```
sasurobert, line 118: can you rewrite this ? like only in case of error do we have an exit from the for. This does not look good. Can you rephrase ?

sasurobert (changes requested):

process/sync/storageBootstrap/sovereignChainShardStorageBootstrapper.go
mariusmihaic: Right, moved it outside

process/sync/storageBootstrap/sovereignChainShardStorageBootstrapper.go
mariusmihaic: Yes, it is the same way as in `shardStorageBootstrapper` -> basically you cleanUp anything until there is nothing more to cleanUp.

I will rephrase the log, though

sasurobert (approved):

axenteoctavian (approved):

PR #6568, opened 2024-10-28
[sovereign] Block finality to zero for extended headers

mariusmihaic:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/track/sovereignChainBlockProcessor.go
+++ b/process/track/sovereignChainBlockProcessor.go
@@ -34,6 +34,7 @@ func NewSovereignChainBlockProcessor(blockProcessor *blockProcessor) (*sovereign
 	scbp.doJobOnReceivedCrossNotarizedHeaderFunc = scbp.doJobOnReceivedCrossNotarizedHeader
 	scbp.requestHeaderWithShardAndNonceFunc = scbp.requestHeaderWithShardAndNonce
 	scbp.requestHeadersIfNothingNewIsReceivedFunc = scbp.requestHeadersIfNothingNewIsReceived
+	scbp.getBlockFinalityFunc = scbp.getBlockFinality
 
 	extendedShardHeaderRequester, ok := scbp.requestHandler.(extendedShardHeaderRequestHandler)
 	if !ok {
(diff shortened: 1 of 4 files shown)


axenteoctavian (approved):

process/track/sovereignChainBlockProcessor.go
```diff
@@ -117,3 +118,11 @@ func (scbp *sovereignChainBlockProcessor) requestHeadersIfNothingNewIsReceived(
 
 	scbp.baseRequestHeadersIfNothingNewIsReceived(lastNotarizedHeaderNonce, latestValidHeader, highestRoundInReceivedHeaders, shardID)
 }
+
+func (scbp *sovereignChainBlockProcessor) getBlockFinality(shardID uint32) uint64 {
+	if shardID == core.MainChainShardId {
+		return 0
+	}
+
+	return scbp.blockFinality
```
sasurobert, line 127: we should have instant finality here as well.

process/track/sovereignChainBlockProcessor.go
```diff
@@ -117,3 +118,11 @@ func (scbp *sovereignChainBlockProcessor) requestHeadersIfNothingNewIsReceived(
 
 	scbp.baseRequestHeadersIfNothingNewIsReceived(lastNotarizedHeaderNonce, latestValidHeader, highestRoundInReceivedHeaders, shardID)
 }
+
+func (scbp *sovereignChainBlockProcessor) getBlockFinality(shardID uint32) uint64 {
+	if shardID == core.MainChainShardId {
+		return 0
```
sasurobert, line 124: block finality isn't 1 for mainchain now ?

mariusmihaic: Yes, it is, but notifier only sends finalized headers, w.r.t to this comment:

> Whenever receving a new incoming header, the resulted header will be instantly included in the sovereign block. Reason being that notifier only sends finalized mainchain headers, so we do not need block finality = 1 when notarizing incoming headers

process/track/sovereignChainBlockProcessor.go
mariusmihaic: Right, will simply change the variable to 0 then

axenteoctavian (approved):

sasurobert (approved):

PR #6570, opened 2024-10-29
New relayed v3

sstanculeanu:
- new implementation for relayed transactions v3 was needed, in order to simplify the initial implementation
  
## Proposed changes
- new approach, with relayer signature on the relayed transaction

## Testing procedure
- scenarios + system test

## Pre-requisites


cristure (dismissed):

process/transaction/shardProcess.go
```diff
@@ -613,27 +621,35 @@ func (txProc *txProcessor) finishExecutionOfRelayedTx(
...
+	err = txProc.addFeeAndValueToDest(acntDst, relayedValue, computedFees.remainingFee)
 	if err != nil {
 		return 0, err
 	}
 
-	return txProc.processUserTx(tx, userTx, tx.Value, tx.Nonce, txHash)
+	relayedNonce := tx.Nonce
+	relayerAddr := tx.SndAddr
+	if isUserTxOfRelayedV3 && !check.IfNil(relayerAcnt) {
+		relayedNonce = relayerAcnt.GetNonce() - 1 // nonce already increased inside processTxAtRelayer
```
sasurobert, line 632: yeah. but it is increased for the relayer. which is not really good.

In case of relayedV3 increase nonce should happen only at the user - and it has to be exactly as its transaction nonce has.

process/transaction/shardProcess.go
```diff
@@ -669,6 +685,37 @@ func (txProc *txProcessor) addFeeAndValueToDest(acntDst state.UserAccountHandler
...
+		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAccount, process.ErrShardIdMissmatch)
+	}
+
+	if !check.IfNil(relayerAccount) && relayerAccount.IsGuarded() {
+		return vmcommon.UserError, txProc.executingFailedTransaction(tx, relayerAccount, process.ErrGuardedRelayerNotAllowed)
+	}
+
+	userTx := *tx
+	// remove relayer addr and signature for tx type handler
+	userTx.RelayerAddr = nil
```
sasurobert, line 708: it could work without these 2 lines.

sasurobert (changes requested):

process/economics/economicsData.go
```diff
@@ -488,15 +488,24 @@ func (ed *economicsData) ComputeGasLimit(tx data.TransactionWithFeeHandler) uint
...
+	if ok {
+		if ed.txVersionHandler.IsGuardedTransaction(txInstance) {
+			gasLimit += ed.getExtraGasLimitGuardedTx(epoch)
+		}
+
+		hasValidRelayer := len(txInstance.GetRelayerAddr()) == len(txInstance.GetSndAddr()) && len(txInstance.GetRelayerAddr()) > 0
+		hasValidRelayerSignature := len(txInstance.GetRelayerSignature()) == len(txInstance.GetSignature()) && len(txInstance.GetRelayerSignature()) > 0
+		isRelayedV3 := hasValidRelayer && hasValidRelayerSignature
+		if isRelayedV3 {
+			gasLimit += ed.MinGasLimitInEpoch(epoch)
```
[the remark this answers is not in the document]
sstanculeanu: extracted in a new method

outport/process/transactionsfee/transactionsFeeProcessor.go
```diff
@@ -271,7 +271,15 @@ func (tep *transactionsFeeProcessor) setGasUsedAndFeeBasedOnRefundValue(
 	refund *big.Int,
 	epoch uint32,
 ) {
-	if !check.IfNil(userTx) && tep.enableEpochsHandler.IsFlagEnabledInEpoch(common.FixRelayedBaseCostFlag, epoch) {
+	tx := txWithResults.GetTxHandler()
+	relayedTx, isRelayedV3 := tx.(data.RelayedTransactionHandler)
+	if isRelayedV3 {
```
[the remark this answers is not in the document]
sstanculeanu: applied the suggestion

process/transaction/interceptedTransaction.go
```diff
@@ -223,6 +232,36 @@ func isRelayedTx(funcName string) bool {
 		core.RelayedTransactionV2 == funcName
 }
 
+func isRelayedTxV3(tx *transaction.Transaction) bool {
```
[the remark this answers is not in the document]
sstanculeanu: removed the method and reused the common one

the reason why I've added the extended check was to avoid treating txs that only have relayer address or relayer signature as relayed v3.. of course, these txs would have failed later

process/transaction/shardProcess.go
sstanculeanu: with the current implementation of [ProcessTransactionType](https://github.com/multiversx/mx-chain-go/blob/relayedv3/process/coordinator/transactionType.go#L80) from txTypeHandler, if these fields won't be removed, the user tx will still be considered of type relayed v3

at [this point](https://github.com/multiversx/mx-chain-go/blob/relayedv3/process/transaction/shardProcess.go#L913), we want to get the proper type of the user tx for its processing

process/transaction/baseProcess.go
```diff
@@ -152,10 +152,20 @@ func (txProc *baseTxProcessor) checkTxValues(
...
 	}
 
-	if acntSnd.GetBalance().Cmp(txFee) < 0 {
+	feePayer, isRelayedV3, err := txProc.getFeePayer(tx, acntSnd)
+	if err != nil {
+		return err
+	}
+
+	// early exit for relayed v3, the cost will be compared with the sender balance
+	if isRelayedV3 {
```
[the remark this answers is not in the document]
sstanculeanu: it should be few lines later, indeed

the reason why it is needed is because this method is called from multiple places:
1. for full relayed tx, before moving the funds from relayer to the sender. In this case, we need the early exit as the feePayer will be the relayer, but the txValue will be taken from the sender. So we don't want to check if the relayer also has the tx.Value.
2. only for user tx, when the tx fee was already moved into the relayer account. At this point, the sender of the user tx will be verified if it has both the fee(moved from relayer) and the tx.Value (which is here defined as `cost`)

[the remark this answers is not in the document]
sasurobert: it is too hard to understand the code and a lot of mistakes can happen because of it.

Can you have separate functions called as it should be ?

process/transaction/shardProcess.go
sasurobert: yeah. but this is not ok at all. somebody make a hash for the USER TX - and it gets completely different, links will be lost.

You may create a new function down the line - which works a little bit differently. 

I would also separate into new function because we want to get rid of relayedV1 and relayedV2.

sstanculeanu: pushed the suggestions + improved the comment of mentioned area

sstanculeanu:
Run tests:
mx-chain-simulator-go Target Branch: main
mx-chain-testing-suite Target Branch: main

process/transaction/shardProcess.go
```diff
@@ -598,13 +604,49 @@ func makeUserTxFromRelayedTxV2Args(args [][]byte) *transaction.Transaction {
 	return userTx
 }
 
+func (txProc *txProcessor) finishExecutionOfRelayedTxV3(
+	relayerAcnt, acntDst state.UserAccountHandler,
+	tx *transaction.Transaction,
+	userTx *transaction.Transaction,
+) (vmcommon.ReturnCode, error) {
+	computedFees := txProc.computeRelayedTxV3Fees(tx, userTx)
+	txHash, err := txProc.processTxV3AtRelayer(
```
sasurobert, line 613: this is not necessary processing, but taking out fees from relayer. maybe rename the function.

process/transaction/shardProcess.go
```diff
@@ -655,6 +698,40 @@ func (txProc *txProcessor) processTxAtRelayer(
...
+		if err != nil {
+			return nil, err
+		}
+
+		err = txProc.accounts.SaveAccount(relayerAcnt)
+		if err != nil {
+			return nil, err
+		}
+
+		txProc.txFeeHandler.ProcessTransactionFee(relayerFee, big.NewInt(0), txHash)
```
sasurobert, line 729: isn't the fee the Total Fee in this case ?

sstanculeanu: I think it should stay relayerFee, same as relayed v1 and v2, as the cost of execution is added to txFeeHandler as part of sc processor

process/transaction/shardProcess.go
sstanculeanu: renamed to consumeFeeFromRelayer

process/transaction/shardProcess.go
```diff
@@ -306,7 +308,7 @@ func (txProc *txProcessor) executingFailedTransaction(
 		return err
 	}
 
-	acntSnd.IncreaseNonce(1)
+	relayerAccount.IncreaseNonce(1)
```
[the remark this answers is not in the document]
sstanculeanu: indeed this was wrong.. reverted the changes done for this method and created a new one for relayed v3, thus further separating the code

sstanculeanu:
Run Tests:
mx-chain-simulator-go: main
mx-chain-testing-suite: main

sasurobert (approved):

PR #6585, opened 2024-11-05
[sovereign] Do not request incoming headers upon receiving

mariusmihaic:
- This PR addresses an "edge case" issue that arises when a sovereign chain’s round duration is greater than the main chain’s and incoming headers are quickly added to the pool. Due to go routines notifying the pool and process handlers asynchronously, it’s possible for an extended header with nonce `X+2` to be added before one with nonce `X+1`.
- When the function `doJobOnReceivedCrossNotarizedHeader` receives a new main chain header, it notifies all relevant handlers via `crossBlockNotifier`, which then calls `ComputeLongestChain`. If a header with nonce X+2 is processed first, unnecessary network requests are made to receive headers with nonce X+1.
  
## Proposed changes
- In normal processing chain code, if we have no registered handler inside `crossBlockNotifier`,we  return early
- We also have no registered handler, so returned earlier .

## Testing procedure
- 
- 
- 

## Pre-requisites


sasurobert (dismissed):

cmd/sovereignnode/chainSimulator/common/common.go
```diff
@@ -2,10 +2,38 @@ package common
...
+	sovRunType "github.com/multiversx/mx-chain-go/sovereignnode/runType"
 )
 
 // GetCurrentSovereignHeader returns current sovereign chain block handler from blockchain hook
 func GetCurrentSovereignHeader(nodeHandler process.NodeHandler) data.SovereignChainHeaderHandler {
 	return nodeHandler.GetChainHandler().GetCurrentBlockHeader().(data.SovereignChainHeaderHandler)
 }
+
+// CreateSovereignRunTypeComponents will create sovereign run type components
+func CreateSovereignRunTypeComponents(args runType.ArgsRunTypeComponents, sovereignExtraConfig config.SovereignConfig) (factory.RunTypeComponentsHolder, error) {
```
axenteoctavian, line 18: The cleanest solution would be to move this function somewhere in `/sovereignnode/runtype` and use it in both chainSimualtor and sovereignNodeRunner

axenteoctavian (dismissed):

cmd/sovereignnode/chainSimulator/common/common.go
mariusmihaic: Not quite same usage, just verified.
Will leave it as it is

axenteoctavian (approved):

sasurobert (approved):

PR #6596, opened 2024-11-12
Refactor register functions to be usable in sovereign chain simulator

axenteoctavian:
- 
- 
- 
  
## Proposed changes
- Little refactor for some functions to work in sovereign chain simulator
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/integrationTests/chainSimulator/bridge/esdtSafe_test.go
+++ b/integrationTests/chainSimulator/bridge/esdtSafe_test.go
@@ -522,7 +522,7 @@ func generateAccountsAndTokens(
 	wallets[account] = token
 
 	account, accountAddrBytes = createNewAccount(cs, &accountShardId)
-	collectionId := chainSim.RegisterAndSetAllRoles(t, cs, accountAddrBytes, &account.nonce, issueCost, "NFTV2", "NFTV2", core.NonFungibleESDTv2, 0)
+	collectionId := chainSim.RegisterAndSetAllRoles(t, cs, cs.GetNodeHandler(core.MetachainShardId), accountAddrBytes, &account.nonce, issueCost, "NFTV2", "NFTV2", core.NonFungibleESDTv2, 0)
 	supply = big.NewInt(1)
 	createArgs := createNftArgs(collectionId, supply, "NFTV2 #1")
 	chainSim.SendTransactionWithSuccess(t, cs, accountAddrBytes, &account.nonce, accountAddrBytes, chainSim.ZeroValue, createArgs, uint64(60000000))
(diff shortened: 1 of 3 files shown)


integrationTests/chainSimulator/bridge/esdtSafe_test.go
```diff
@@ -522,7 +522,7 @@ func generateAccountsAndTokens(
 	wallets[account] = token
 
 	account, accountAddrBytes = createNewAccount(cs, &accountShardId)
-	collectionId := chainSim.RegisterAndSetAllRoles(t, cs, accountAddrBytes, &account.nonce, issueCost, "NFTV2", "NFTV2", core.NonFungibleESDTv2, 0)
+	collectionId := chainSim.RegisterAndSetAllRoles(t, cs, cs.GetNodeHandler(core.MetachainShardId), accountAddrBytes, &account.nonce, issueCost, "NFTV2", "NFTV2", core.NonFungibleESDTv2, 0)
```
mariusmihaic, line 525: Let' make a variable and use it instead of calling this everytime, please

integrationTests/chainSimulator/common.go
```diff
@@ -366,6 +366,7 @@ func SetEsdtInWallet(
 func RegisterAndSetAllRoles(
 	t *testing.T,
 	cs ChainSimulator,
+	nodeHandler process.NodeHandler,
```
mariusmihaic, line 369: I think we could call this "metaNodeHandler"

integrationTests/chainSimulator/common.go
```diff
@@ -382,13 +383,14 @@ func RegisterAndSetAllRoles(
...
 
-	return GetIssuedEsdtIdentifier(t, cs.GetNodeHandler(core.MetachainShardId), esdtTicker, tokenType)
+	return GetIssuedEsdtIdentifier(t, nodeHandler, esdtTicker, tokenType)
 }
 
 // RegisterAndSetAllRolesDynamic will issue a dynamic esdt collection with all roles enabled
 func RegisterAndSetAllRolesDynamic(
 	t *testing.T,
 	cs ChainSimulator,
+	nodeHandler process.NodeHandler,
```
mariusmihaic, line 393: Same remark

integrationTests/chainSimulator/common.go
mariusmihaic: Actually, I think a "smarter" way to reuse this would be to use `cs.GetShardID(vm.ESDTSCAddress)`.
What do you think about this approach?

mariusmihaic: This way you also get rid of passing `nodeHandler`

axenteoctavian: Implemented the node handler from esdt sc shard id

mariusmihaic (approved):

sasurobert (approved):

PR #6624, opened 2024-11-21
Mx 15407 sovereign proxy scripts

axenteoctavian:
- 
- 
- 
  
## Proposed changes
- sovereign local testnet: proxy deploy script update
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


scripts/testnet/include/tools.sh
```diff
@@ -5,7 +5,10 @@ startProxy() {
   setTerminalLayout "even-vertical"
 
   setWorkdirForNextCommands "$TESTNETDIR/proxy"
-  runCommandInTerminal "./proxy" $1
+  if [ "$SOVEREIGN_DEPLOY" -eq 1 ]; then
+      ARGS="--sovereign-config"
+  fi
+  runCommandInTerminal "./proxy $ARGS" $1
```
mariusmihaic, line 11: `ARGS` is undefined unless `SOVEREIGN_DEPLOY` is true.

mariusmihaic (changes requested):

scripts/testnet/include/tools.sh
sasurobert: +1

sasurobert (commented):
Only what Marius commented.

scripts/testnet/include/tools.sh
```diff
@@ -6,7 +6,9 @@ startProxy() {
 
   setWorkdirForNextCommands "$TESTNETDIR/proxy"
   if [ "$SOVEREIGN_DEPLOY" -eq 1 ]; then
-      ARGS="--sovereign-config"
+      ARGS="--sovereign"
+  else
```
mariusmihaic, line 10: I think we should define it **above the first if** `ARGS=""`
Then only change it `if [ "$SOVEREIGN_DEPLOY" -eq 1 ]`(and remove else)

mariusmihaic (approved):

PR #6639, opened 2024-11-26
RelayedV3 fixes

sstanculeanu:
- the tx validity checks are always returning insufficient funds for sender, even when there are no funds at relayer 
- relayed v3 calls to meta sc are also processed as a transaction on metachain, resulting in duplicated processing on meta: 1 as a regular transaction to meta, 2 from the scr generated from the original relayed tx
- a check for guardian != relayer should be added in order to avoid same signature
  
## Proposed changes
- fix tx validity wrapped error
- avoid processing of relayed tx v3 on meta as regular tx, but only as scr
- added check on for guardian != relayer

## Testing procedure
- will be tested with feat branch and scenarios

## Pre-requisites

### DIFF
--- a/process/smartContract/processorV2/processV2.go
+++ b/process/smartContract/processorV2/processV2.go
@@ -603,6 +603,9 @@ func (sc *scProcessor) updateDeveloperRewards(
 	}
 
 	moveBalanceGasLimit := sc.economicsFee.ComputeGasLimit(tx)
+	if common.IsValidRelayedTxV3(tx) {
+		moveBalanceGasLimit -= sc.economicsFee.MinGasLimit() // this was already consumed from the relayer
+	}
 	if !isSmartContractResult(tx) {
 		usedGasByMainSC, err = core.SafeSubUint64(usedGasByMainSC, moveBalanceGasLimit)
 		if err != nil {
@@ -744,6 +747,9 @@ func (sc *scProcessor) computeTotalConsumedFeeAndDevRwd(
 	}
 
 	moveBalanceGasLimit := sc.economicsFee.ComputeGasLimit(tx)
+	if common.IsValidRelayedTxV3(tx) {
+		moveBalanceGasLimit -= sc.economicsFee.MinGasLimit() // this was already consumed from the relayer
+	}
 	if !isSmartContractResult(tx) {
 		displayConsumedGas := consumedGas
 		consumedGas, err = core.SafeSubUint64(consumedGas, moveBalanceGasLimit)
@@ -1408,7 +1414,7 @@ func (sc *scProcessor) getOriginalTxHashIfIntraShardRelayedSCR(
 	tx data.TransactionHandler,
 	txHash []byte,
 ) []byte {
-	relayedSCR, isRelayed := isRelayedTx(tx)
+	relayedSCR, isRelayed := isRelayedSCR(tx)
 	if !isRelayed {
 		return txHash
 	}
@@ -1578,7 +1584,7 @@ func (sc *scProcessor) processForRelayerWhenError(
 	txHash []byte,
 	returnMessage []byte,
 ) (*vmcommon.LogEntry, error) {
-	relayedSCR, isRelayed := isRelayedTx(originalTx)
+	relayedSCR, isRelayed := isRelayedSCR(originalTx)
 	if !isRelayed {
 		return nil, nil
 	}
@@ -1664,7 +1670,7 @@ func createNewLogFromSCRIfError(txHandler data.TransactionHandler) *vmcommon.Log
 }
 
 // transaction must be of type SCR and relayed address to be set with relayed value higher than 0
-func isRelayedTx(tx data.TransactionHandler) (*smartContractResult.SmartContractResult, bool) {
+func isRelayedSCR(tx data.TransactionHandler) (*smartContractResult.SmartContractResult, bool) {
 	relayedSCR, ok := tx.(*smartContractResult.SmartContractResult)
 	if !ok {
 		return nil, false
@@ -1686,7 +1692,7 @@ func (sc *scProcessor) addBackTxValues(
 ) error {
 	valueForSnd := big.NewInt(0).Set(scrIfError.Value)
 
-	relayedSCR, isRelayed := isRelayedTx(originalTx)
+	relayedSCR, isRelayed := isRelayedSCR(originalTx)
 	if isRelayed {
 		valueForSnd.Sub(valueForSnd, relayedSCR.RelayedValue)
 		if valueForSnd.Cmp(zero) < 0 {
@@ -1923,6 +1929,11 @@ func (sc *scProcessor) processSCPayment(tx data.TransactionHandler, acntSnd stat
 
 	cost := sc.economicsFee.ComputeTxFee(tx)
 	cost = cost.Add(cost, tx.GetValue())
+
+	if common.IsValidRelayedTxV3(tx) {
+		// for relayed v3, fee was consumed from relayer
+		cost = tx.GetValue()
+	}
 
 	if cost.Cmp(big.NewInt(0)) == 0 {
 		return nil
@@ -2257,10 +2268,16 @@ func createBaseSCR(
 	result.CallType = vmData.DirectCall
 	setOriginalTxHash(result, txHash, tx)
 
-	relayedTx, isRelayed := isRelayedTx(tx)
+	relayedTx, isRelayed := isRelayedSCR(tx)
 	if isRelayed {
 		result.RelayedValue = big.NewInt(0)
 		result.RelayerAddr = relayedTx.RelayerAddr
+	}
+
+	if common.IsValidRelayedTxV3(tx) {
+		relayedTx := tx.(data.RelayedTransactionHandler)
+		result.RelayedValue = big.NewInt(0)
+		result.RelayerAddr = relayedTx.GetRelayerAddr()
 	}
 
 	return result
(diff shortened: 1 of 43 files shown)


cristure (approved):

process/transaction/metaProcess.go
```diff
@@ -145,6 +145,8 @@ func (txProc *metaTxProcessor) ProcessTransaction(tx *transaction.Transaction) (
 		if txProc.enableEpochsHandler.IsFlagEnabled(common.ESDTFlag) {
 			return txProc.processSCInvoking(tx, tx.SndAddr, tx.RcvAddr)
 		}
+	case process.RelayedTxV3:
```
sasurobert, line 148: this is not the best solution.

SCR does not need to be created. You have all the data on the transaction.

sasurobert (changes requested):

process/transaction/metaProcess.go
AdoAdoAdo, line 148: can you confirm that the txType detected here is one of the above : SCDeployment, SCInvoking, BuiltInFunctionCall.

I don't think we need a process.RelayedTxV3 here.

sstanculeanu: updated the execution to follow the normal tx flow

process/smartContract/processorV2/processV2.go
```diff
@@ -2257,12 +2268,18 @@ func createBaseSCR(
 	result.CallType = vmData.DirectCall
 	setOriginalTxHash(result, txHash, tx)
 
-	relayedTx, isRelayed := isRelayedTx(tx)
+	relayedTx, isRelayed := isRelayedSCR(tx)
```
sasurobert, line 2271: duplicated code here and on createAsnyCallBackSCRFromVMOutput. 

Make a new function and use it in both places.

duplicated from 2271 - 2282

process/transaction/metaProcess.go
```diff
@@ -135,7 +135,13 @@ func (txProc *metaTxProcessor) ProcessTransaction(tx *transaction.Transaction) (
 		return 0, err
 	}
 
+	txCopy := *tx
 	txType, _ := txProc.txTypeHandler.ComputeTransactionType(tx)
+	if txType == process.RelayedTxV3 {
+		// extract the inner transaction in order to get the proper user tx type
+		txCopy.RelayerSignature = nil
```
sasurobert, line 142: this is an ugly hack. As we have inShard txType, dest txType already returned by compute Transaction type - you could have a 3rd return, or a txType structure return.

process/transaction/shardProcess.go
```diff
@@ -482,22 +500,77 @@ func (txProc *txProcessor) processTxFee(
...
+		if err != nil {
+			return nil, nil, err
+		}
 	} else {
 		err := acntSnd.SubFromBalance(moveBalanceFee)
 		if err != nil {
 			return nil, nil, err
 		}
+
+		err = txProc.accounts.SaveAccount(acntSnd)
```
sasurobert, line 514: you can move the saveAccount out from the if and else structure. Actually why do you need the saveAccount here in the processTxFee function ? You could do it outside of this, if processTxFee does not return error.

process/transaction/shardProcess.go
```diff
@@ -482,22 +500,77 @@ func (txProc *txProcessor) processTxFee(
...
+		}
+
+		txProc.txFeeHandler.ProcessTransactionFee(moveBalanceFee, big.NewInt(0), txHash)
+
+		err = txProc.accounts.SaveAccount(relayerAccount)
+		if err != nil {
+			return err
+		}
+	}
+
```
sasurobert, line 567: saveAccount can be moved outside the if/else structure.

process/smartContract/processorV2/processV2.go
```diff
@@ -2592,6 +2615,31 @@ func (sc *scProcessor) createSCRForSenderAndRelayer(
 		gasRemaining = 0
 	}
 
+	isRelayedV3 := common.IsValidRelayedTxV3(tx)
+	shouldRefundGasToRelayerSCR = isRelayedV3 && callType != vmData.AsynchronousCall && gasRefund.Cmp(zero) > 0
+	if shouldRefundGasToRelayerSCR {
+		senderForRelayerRefund := tx.GetRcvAddr()
+		if !sc.isSelfShard(tx.GetRcvAddr()) {
```
sasurobert, line 2622: this seems like a big chunk of duplicated code with the If above.

Can you move the creation of refundGasToRelayer into a separate function and make the code smaller?

sasurobert (changes requested):

process/transaction/baseProcess.go
```diff
@@ -119,6 +119,15 @@ func (txProc *baseTxProcessor) checkTxValues(
 	acntSnd, acntDst state.UserAccountHandler,
 	isUserTxOfRelayed bool,
 ) error {
+	if common.IsValidRelayedTxV3(tx) {
+		relayerAccount, _, err := txProc.getAccounts(tx.RelayerAddr, tx.RelayerAddr)
```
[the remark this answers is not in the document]
sstanculeanu: switched to `getAccountFromAddress`

process/smartContract/processorV2/processV2.go
sstanculeanu: refactored

process/smartContract/processorV2/processV2.go
```diff
@@ -603,6 +603,9 @@ func (sc *scProcessor) updateDeveloperRewards(
 	}
 
 	moveBalanceGasLimit := sc.economicsFee.ComputeGasLimit(tx)
+	if common.IsValidRelayedTxV3(tx) {
```
AdoAdoAdo, line 606: I don't think we need this, as the moveBalanceGasLimit completely needs to not be considered for dev fees, not only part of it.

From what I understand, here you are removing the extra part for relayer signature validation from the base cost, and will be using that instead for dev fees (by not removing it from the usedGasByMainSC).

You will also need to consider this for the preparation for SC execution, there the gas limit you provide in vmInput needs to not contain this part.

AdoAdoAdo (changes requested):

process/smartContract/processorV2/vmInputV2.go
```diff
@@ -68,6 +73,10 @@ func (sc *scProcessor) prepareGasProvided(tx data.TransactionHandler) (uint64, e
 	}
 
 	gasForTxData := sc.economicsFee.ComputeGasLimit(tx)
```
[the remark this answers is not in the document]
sstanculeanu: removed completely, not needed

process/smartContract/processorV2/processV2.go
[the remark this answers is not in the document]
sstanculeanu: removed it

sstanculeanu: removed as discussed

process/transaction/metaProcess.go
sstanculeanu: refactored with extra return, simplifies a lot

process/transaction/shardProcess.go
```diff
@@ -664,34 +807,85 @@ func makeUserTxFromRelayedTxV2Args(args [][]byte) *transaction.Transaction {
 }
 
 func (txProc *txProcessor) finishExecutionOfRelayedTxV3(
-	relayerAcnt, acntDst state.UserAccountHandler,
```
AdoAdoAdo, line 667: I don't think we need any "finishExecutionOfRelayedTxV3" this was mainly used for the different execution flow of relayed transactions that were composed by 2 transactions relay tx and user tx.

I would suggest to just go on the flow of regular transactions and change the fee consumption in case the transaction is relayed v3 (create a function for that)

process/transaction/shardProcess.go
```diff
@@ -1380,6 +1429,43 @@ func (txProc *txProcessor) executeFailedRelayedUserTx(
 	return nil
 }
 
+func (txProc *txProcessor) executeFailedRelayedV3UserTx(
```
AdoAdoAdo, line 1432: I don't think we need to execute any failed relayv3 usertx, as there is only one tx, where the fee is paid by a relayer.

process/transaction/shardProcess.go
sstanculeanu: updated

process/transaction/shardProcess.go
sstanculeanu: updated

sasurobert (dismissed):

process/transaction/interceptedTransaction.go
```diff
@@ -238,9 +244,21 @@ func (inTx *InterceptedTransaction) verifyIfRelayedTxV3(tx *transaction.Transact
 		return nil
 	}
 
+	if !inTx.enableEpochsHandler.IsFlagEnabled(common.RelayedTransactionsV3Flag) {
```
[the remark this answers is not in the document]
sstanculeanu: updated everywhere

process/transaction/baseProcess.go
```diff
@@ -189,23 +185,25 @@ func (txProc *baseTxProcessor) checkTxValues(
 
 func (txProc *baseTxProcessor) checkUserTxOfRelayedV3Values(
 	tx *transaction.Transaction,
-	acntSnd, acntDst state.UserAccountHandler,
+	senderAccount state.UserAccountHandler,
+	destinationAccount state.UserAccountHandler,
+	relayerAccount state.UserAccountHandler,
 ) error {
-	err := txProc.verifyGuardian(tx, acntSnd)
+	err := txProc.verifyGuardian(tx, senderAccount)
```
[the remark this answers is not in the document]
sstanculeanu: updated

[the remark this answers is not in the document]
sstanculeanu: reverted, in order to keep the same validation and returned errors order

AdoAdoAdo (approved):

cristure (approved):

PR #6643, opened 2024-11-29
Implement an ephemeral accounts cache within a selection session

andreibancioiu:
- `accountsAdapter.GetExistingAccount()` has a non-trivial implementation, with a non-negligible performance impact. 
- We have to optimize the transactions selection flow as much as possible.
  
## Proposed changes
- Since both `SelectionSession.GetAccountState()` and `SelectionSession.IsIncorrectlyGuarded()` fetch an account object from the blockchain state - as it was when the selection session started, we can use an (ephemeral) cache for these accounts.
- Thus, if before we had `2 * S` calls to `accountsAdapter.GetExistingAccount()`, where `S` is _the number of unique senders of selected transactions within a session_, we now have `S` calls.

## Testing procedure
- Standard testing.

## Pre-requisites

### DIFF
--- a/process/block/preprocess/selectionSession.go
+++ b/process/block/preprocess/selectionSession.go
@@ -9,11 +9,16 @@ import (
 	"github.com/multiversx/mx-chain-go/process"
 	"github.com/multiversx/mx-chain-go/state"
 	"github.com/multiversx/mx-chain-go/storage/txcache"
+	vmcommon "github.com/multiversx/mx-chain-vm-common-go"
 )
 
 type selectionSession struct {
 	accountsAdapter       state.AccountsAdapter
 	transactionsProcessor process.TransactionProcessor
+
+	// Cache of accounts, held in the scope of a single selection session.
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/multiversx/mx-chain-es-indexer-go v1.7.10
 	github.com/multiversx/mx-chain-logger-go v1.0.15
 	github.com/multiversx/mx-chain-scenario-go v1.4.4
-	github.com/multiversx/mx-chain-storage-go v1.0.18-0.20241128201607-ea3ee2dfd9eb
+	github.com/multiversx/mx-chain-storage-go v1.0.18-0.20241129141109-31826611a4b5
```
andreibancioiu, line 23: https://github.com/multiversx/mx-chain-storage-go/pull/62/commits

AdoAdoAdo (approved):

AdoAdoAdo (approved):

sstanculeanu (approved):

PR #6665, opened 2024-12-10
Fixes header sig verification

AdoAdoAdo:
- The proof verification (aggregated signature) with equivalent proofs is for the unaltered header, so it also needs to be verified on the unaltered header
- Add more safety and integrity checks
  
## Proposed changes
- Leader signature is not removed from the header before verifying the proof aggregated signature
- Make sure new header fields (proof related) are validated for both before activation - throw an error if the new fields are present, and after activation - throw an error if the fields are not present

## Testing procedure
- Unit tests
- Regular system test with transitioning from old to new consensus will be done on the feat branch
   
## Pre-requisites

### DIFF
--- a/consensus/spos/bls/v1/subroundBlock.go
+++ b/consensus/spos/bls/v1/subroundBlock.go
@@ -415,9 +415,14 @@ func (sr *subroundBlock) receivedBlockBodyAndHeader(ctx context.Context, cnsDta 
 		return false
 	}
 
+	header := sr.BlockProcessor().DecodeBlockHeader(cnsDta.Header)
+	if headerHasProof(header) {
+		return false
+	}
+
 	sr.SetData(cnsDta.BlockHeaderHash)
 	sr.SetBody(sr.BlockProcessor().DecodeBlockBody(cnsDta.Body))
-	sr.SetHeader(sr.BlockProcessor().DecodeBlockHeader(cnsDta.Header))
+	sr.SetHeader(header)
 
 	isInvalidData := check.IfNil(sr.GetBody()) || sr.isInvalidHeaderOrData()
 	if isInvalidData {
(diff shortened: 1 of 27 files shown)


integrationTests/consensus/consensusSigning_test.go
```diff
@@ -69,6 +71,8 @@ func TestConsensusWithInvalidSigners(t *testing.T) {
 		t.Skip("this is not a short test")
 	}
 
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu, line 74: is this still needed?

AdoAdoAdo: no, forgot this

consensus/spos/bls/v1/subroundBlock.go
```diff
@@ -535,6 +545,13 @@ func (sr *subroundBlock) receivedBlockHeader(ctx context.Context, cnsDta *consen
 	return blockProcessedWithSuccess
 }
 
+func headerHasProof(headerHandler data.HeaderHandler) bool {
+	if check.IfNil(headerHandler) {
+		return true
```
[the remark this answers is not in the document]
AdoAdoAdo: meant to return false

consensus/spos/bls/v1/subroundBlock.go
```diff
@@ -535,6 +545,13 @@ func (sr *subroundBlock) receivedBlockHeader(ctx context.Context, cnsDta *consen
...
+	}
+	return headerHandler.GetPreviousProof() != nil
```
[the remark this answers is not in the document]
AdoAdoAdo: right, forgot the Get was returning an interface not a pointer to the proof

integrationTests/consensus/consensusSigning_test.go
```diff
@@ -103,7 +107,7 @@ func TestConsensusWithInvalidSigners(t *testing.T) {
 		go checkBlockProposedEveryRound(numCommBlock, nonceForRoundMap, mutex, chDone, t)
 
 		extraTime := uint64(2)
-		endTime := time.Duration(roundTime)*time.Duration(numCommBlock+extraTime)*time.Millisecond + time.Minute
+		endTime := time.Duration(roundTime)*time.Duration(numCommBlock+extraTime)*time.Millisecond + 10*time.Minute
```
ssd04, line 110: do we want to extend end time so much?

sstanculeanu (approved):

PR #6667, opened 2024-12-11
Test pruning storer

BeniaminDrasovean:
- There is an edgecase in which if a node starts with `dbLookupExtension` set to `true`, it will not remove any old dbs. In this case the node will create a `FullHistoryPruningStorer` that can create a new db for a new epoch if some methods are called. When the `changeEpoch` trigger is activated, the new db already exists, thus not executing the code that should close older dbs.
  
## Proposed changes
- On `changeEpoch`, even if a db was already created, also check if older dbs need to be removed

## Testing procedure
- Normal testing procedure
- Test with `full-archive`, `db-lookup-extension`, `historical-balances`, `snapshotless-observer` node operation mode and check that the dbs will be kept or removed as they should be.

## Pre-requisites


sasurobert (dismissed):

storage/pruning/pruningStorer.go
```diff
@@ -802,6 +802,10 @@ func (ps *PruningStorer) changeEpoch(header data.HeaderHandler) error {
 	ps.activePersisters = append(singleItemPersisters, ps.activePersisters...)
 	ps.persistersMapByEpoch[epoch] = newPersister
 
+	return ps.removeOldPersistersIfNeeded(header, epoch)
+}
+
+func (ps *PruningStorer) removeOldPersistersIfNeeded(header data.HeaderHandler, epoch uint32) error {
```
AdoAdoAdo, line 808: can we pass only one argument, either the epoch or the header?

BeniaminDrasovean: Yes

danidrasovean (approved):
Normal allin test: v1.8.4-dev-config-171a31edab -> test-pruning-storer-fee796fb44

--- Specific errors --- 

block hash does not match 1907
wrong nonce in block 829
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  0
Nr. of all WARNS:  416
Nr. of new ERRORS:  0
Nr. of new WARNS:  15
Nr. of PANICS:  0
Nodes with panic:  []

/------/ 


--- System test go --- 

/------/ 



--- ERRORS ---  

/------/ 

--- WARNINGS ---  

/------/

sasurobert (approved):

PR #6675, opened 2024-12-13
sovereign chain block unit tests

axenteoctavian:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/sovereignChainBlock_test.go
+++ b/process/block/sovereignChainBlock_test.go
@@ -1,8 +1,17 @@
 package block_test
 
 import (
+	"math/big"
 	"testing"
 	"time"
+
+	"github.com/multiversx/mx-chain-core-go/core"
+	atomicCore "github.com/multiversx/mx-chain-core-go/core/atomic"
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data"
+	"github.com/multiversx/mx-chain-core-go/data/block"
+	sovereignCore "github.com/multiversx/mx-chain-core-go/data/sovereign"
+	"github.com/stretchr/testify/require"
 
 	"github.com/multiversx/mx-chain-go/dataRetriever/requestHandlers"
 	"github.com/multiversx/mx-chain-go/errors"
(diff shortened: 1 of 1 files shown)


process/block/sovereignChainBlock_test.go
```diff
@@ -343,4 +400,283 @@ func TestSovereignChainBlockProcessor_createAndSetOutGoingMiniBlock(t *testing.T
...
+				Epoch: nextEpoch,
+			},
+			IsStartOfEpoch: true,
+		}
+		doesHaveTime := func() bool {
+			return true
+		}
+
+		hdr, bodyHandler, err := scbp.CreateBlock(sovHeader, doesHaveTime)
+		require.False(t, check.IfNil(bodyHandler))
```
mariusmihaic, line 628: Check directly with miniblocks and skip using these False.Nil.
Applies also for expected header. If you require it to be equal to your expected object, there is no need to also check it is not nil

process/block/sovereignChainBlock_test.go
```diff
@@ -343,4 +400,283 @@ func TestSovereignChainBlockProcessor_createAndSetOutGoingMiniBlock(t *testing.T
...
+			},
+		}
+
+		hdr, bodyHandler, err := scbp.CreateBlock(expectedSovHeader, doesHaveTime)
+		require.True(t, check.IfNil(bodyHandler))
+		require.True(t, check.IfNil(hdr))
+		require.Equal(t, expectedErr, err)
+		require.Equal(t, expectedBusyIdleSequencePerCall, busyIdleCalled)
+	})
+	t.Run("should work with sovereign header and epoch start rewriting the epoch value", func(t *testing.T) {
```
mariusmihaic, line 579: Please rename this func name, we do not rewrite, I think you mean we change epoch

process/block/sovereignChainBlock_test.go
```diff
@@ -343,4 +400,283 @@ func TestSovereignChainBlockProcessor_createAndSetOutGoingMiniBlock(t *testing.T
...
+				Nonce: 37,
+				Round: 38,
+				Epoch: currentEpoch,
+			},
+		}
+		doesHaveTime := func() bool {
+			return true
+		}
+
+		hdr, bodyHandler, err := scbp.CreateBlock(expectedSovHeader, doesHaveTime)
```
mariusmihaic, line 672: Same remarks as above

process/block/sovereignChainBlock_test.go
axenteoctavian: I can't get the miniblocks from `bodyHandler` that's why is being checked for nil and cast to `Body` type

process/block/sovereignChainBlock_test.go
```diff
@@ -432,55 +420,47 @@ func TestSovereignShardProcessor_CreateNewHeaderValsOK(t *testing.T) {
...
+		Header: &block.Header{
+			Nonce:           5,
+			Round:           7,
+			AccumulatedFees: big.NewInt(0),
+			DeveloperFees:   big.NewInt(0),
+		},
+		AccumulatedFeesInEpoch: big.NewInt(0),
+	}, h)
 	require.Nil(t, h.GetValidatorStatsRootHash())
 	require.Nil(t, h.GetRootHash())
```
mariusmihaic, line 439: I think these could be removed

PR #6677, opened 2024-12-13
Accept relayed tx v3 with sender account non-existent

sstanculeanu:
- relayed txs v3 with sender account not-existent are rejected by interceptors, despite the zero value of transaction
  
## Proposed changes
- allow relayed txs v3 with new senders, in case of zero value transactions

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/process/dataValidators/txValidator.go
+++ b/process/dataValidators/txValidator.go
@@ -2,7 +2,6 @@ package dataValidators
 
 import (
 	"fmt"
-	"math/big"
 
 	"github.com/multiversx/mx-chain-core-go/core"
 	"github.com/multiversx/mx-chain-core-go/core/check"
@@ -95,7 +94,7 @@ func hasTxValue(interceptedTx process.InterceptedTransactionHandler) bool {
 		return false
 	}
 
-	return big.NewInt(0).Cmp(txValue) < 0
+	return txValue.Sign() > 0
 }
 
 func (txv *txValidator) checkAccount(
--- a/process/dataValidators/txValidator.go
+++ b/process/dataValidators/txValidator.go
@@ -2,6 +2,7 @@ package dataValidators
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/multiversx/mx-chain-core-go/core"
 	"github.com/multiversx/mx-chain-core-go/core/check"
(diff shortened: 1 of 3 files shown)


process/dataValidators/txValidator.go
```diff
@@ -76,14 +77,27 @@ func (txv *txValidator) CheckTxValidity(interceptedTx process.InterceptedTransac
 		return nil
 	}
 
+	// for relayed v3, we allow sender accounts that do not exist
+	isRelayedV3 := common.IsRelayedTxV3(interceptedTx.Transaction())
+	hasValue := hasTxValue(interceptedTx)
+	shouldAllowMissingSenderAccount := isRelayedV3 && !hasValue
 	accountHandler, err := txv.getSenderAccount(interceptedTx)
-	if err != nil {
+	if err != nil && !shouldAllowMissingSenderAccount {
```
andreibancioiu, line 85: Should we check for `ErrAccountNotFound` (explicitly)? Just a question.

process/dataValidators/txValidator.go
```diff
@@ -76,14 +77,27 @@ func (txv *txValidator) CheckTxValidity(interceptedTx process.InterceptedTransac
...
 		return err
 	}
 
 	return txv.checkAccount(interceptedTx, accountHandler)
 }
 
+func hasTxValue(interceptedTx process.InterceptedTransactionHandler) bool {
```
andreibancioiu, line 92: Indeed, we don't handle `MultiESDTTransfer` with EGLD in interceptor (we never did, but that is OK).

process/dataValidators/txValidator.go
```diff
@@ -76,14 +77,27 @@ func (txv *txValidator) CheckTxValidity(interceptedTx process.InterceptedTransac
...
+	txValue := interceptedTx.Transaction().GetValue()
+	if check.IfNilReflect(txValue) {
+		return false
+	}
+
+	return big.NewInt(0).Cmp(txValue) < 0
```
andreibancioiu, line 98: Maybe directly use `Sign()` method of `txValue`, without instantiating a new big int (zero)?

andreibancioiu (approved):

process/dataValidators/txValidator.go
sstanculeanu: I don't think it is necessary, as `ErrAccountNotFound` is the only one returned from `getSenderAccount` method

process/dataValidators/txValidator.go
sstanculeanu: pushed

AdoAdoAdo (approved):

andreibancioiu (approved):

PR #6723, opened 2025-01-21
refactor trie constructor

BeniaminDrasovean:
- The trie constructor needs a lot of parameter
  
## Proposed changes
- Instead of passing each parameter for the trie constructor, pass an `args` struct
- Pass a `goRoutinesThrottler` to the trie constructor. On trie recreate, use the same throttler. This will prevent us from opening an unlimited number of goroutines for trie operations
- Add a new config val to the `[StateTriesConfig]` - `MaxNumGoroutines`

## Testing procedure
- This will be tested when testing the feature branch it is targeted to

## Pre-requisites

### DIFF
--- a/trie/disabledTrieGoRoutinesThrottler.go
+++ b/trie/disabledTrieGoRoutinesThrottler.go
@@ -1,0 +1,27 @@
+package trie
+
+type disabledTrieGoRoutinesThrottler struct {
+}
+
+// NewDisabledTrieGoRoutinesThrottler returns a new instance of a disabledTrieGoRoutinesThrottler
+func NewDisabledTrieGoRoutinesThrottler() *disabledTrieGoRoutinesThrottler {
+	return &disabledTrieGoRoutinesThrottler{}
+}
+
+// CanProcess will always return false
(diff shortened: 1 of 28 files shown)


sasurobert (approved):

trie/disabledTrieGoRoutinesThrottler.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// NewDisabledTrieGoRoutinesThrottler returns a new instance of a disabledTrieGoRoutinesThrottler
+func NewDisabledTrieGoRoutinesThrottler() *disabledTrieGoRoutinesThrottler {
+	return &disabledTrieGoRoutinesThrottler{}
+}
+
+// CanProcess will always return false
+func (d *disabledTrieGoRoutinesThrottler) CanProcess() bool {
+	return false
```
ssd04, line 13: return true here?

BeniaminDrasovean: If I return true here the goroutines manager will open an unlimited ammount of goroutines. This component should be used if you want to process sequentially. `CanProcess` =  false means that you can not process on a goroutine

ssd04 (approved):

PR #6724, opened 2025-01-22
Fix transition block

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.2.5
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/multiversx/mx-chain-communication-go v1.0.15-0.20240508074652-e128a1c05c8e
-	github.com/multiversx/mx-chain-core-go v1.2.21-0.20250121162639-a5fdb010d9ac
+	github.com/multiversx/mx-chain-core-go v1.2.21-0.20250122100317-798de8d35458
 	github.com/multiversx/mx-chain-crypto-go v1.2.12-0.20240508074452-cc21c1b505df
 	github.com/multiversx/mx-chain-es-indexer-go v1.7.2-0.20240619122842-05143459c554
 	github.com/multiversx/mx-chain-logger-go v1.0.15-0.20240508072523-3f00a726af57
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.2.5
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/multiversx/mx-chain-communication-go v1.0.15-0.20240508074652-e128a1c05c8e
-	github.com/multiversx/mx-chain-core-go v1.2.21-0.20250120115427-6f987e761979
+	github.com/multiversx/mx-chain-core-go v1.2.21-0.20250121162639-a5fdb010d9ac
```
AdoAdoAdo, line 18: don't forget to update to the latest commit (also copy IsStartOfEpoch flag)

process/track/blockProcessor.go
```diff
@@ -336,6 +336,16 @@ func (bp *blockProcessor) checkHeaderFinality(
 			continue
 		}
 
+		// if the currentHeader(the one that should confirm the finality of the prev)
```
AdoAdoAdo, line 339: probably not needed if we handle inside the block tracker the check that there is a proof.
Will take the change of block tracker in a separate PR.

Can you add a check somewhere on the processing (for both leader and other nodes - in create block for leader & process block for others) that every used block has proof

sstanculeanu: this should already be covered:
- create block
   - shard: https://github.com/multiversx/mx-chain-go/blob/fix_transition_block/process/block/shardblock.go#L1975
   - meta: https://github.com/multiversx/mx-chain-go/blob/fix_transition_block/process/block/metablock.go#L1162
- process block:
   - shard: https://github.com/multiversx/mx-chain-go/blob/fix_transition_block/process/block/shardblock.go#L581
   - meta: https://github.com/multiversx/mx-chain-go/blob/fix_transition_block/process/block/metablock.go#L2026

AdoAdoAdo: when verifying the proof for the header not the prev proof, we should *not* use `common.IsFlagEnabledAfterEpochsStartBlock(header, sp.enableEpochsHandler, common.EquivalentMessagesFlag)`

so I think `func (sp *shardProcessor) checkMetaHdrFinality(header data.HeaderHandler) error` has the wrong check

AdoAdoAdo (approved):

ssd04 (approved):

PR #6757, opened 2025-01-30
fix previous block proof saving

AdoAdoAdo:
- fixes for internal testnets 

  
## Proposed changes
- on bootstrap from storage properly set the final block in case of headers with proofs  
- fix previous proof check for the transition block
- fix consensus size check fixes for the transition shard block

## Testing procedure
- 
- 
- 

## Pre-requisites


AdoAdoAdo:
closing this, will be taken in a different branch, together with another fix

PR #6760, opened 2025-01-30
Fix consensus check

AdoAdoAdo:
- previous proof verification fixes
  
## Proposed changes
- for the transition block prev proof should not be checked
- the transition block for shard has a different consensus size (it follows the previous epoch rules), so the check of keys in consensus needs to be adapted.

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/headerCheck/headerSignatureVerify.go
+++ b/process/headerCheck/headerSignatureVerify.go
@@ -35,6 +35,7 @@ type ArgsHeaderSigVerifier struct {
 	FallbackHeaderValidator process.FallbackHeaderValidator
 	EnableEpochsHandler     common.EnableEpochsHandler
 	HeadersPool             dataRetriever.HeadersPool
+	StorageService          dataRetriever.StorageService
 }
 
 // HeaderSigVerifier is component used to check if a header is valid
@@ -48,6 +49,7 @@ type HeaderSigVerifier struct {
 	fallbackHeaderValidator process.FallbackHeaderValidator
 	enableEpochsHandler     common.EnableEpochsHandler
 	headersPool             dataRetriever.HeadersPool
+	storageService          dataRetriever.StorageService
 }
 
 // NewHeaderSigVerifier will create a new instance of HeaderSigVerifier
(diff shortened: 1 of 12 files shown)


process/headerCheck/headerSignatureVerify.go
```diff
@@ -290,25 +294,43 @@ func (hsv *HeaderSigVerifier) VerifySignatureForHash(header data.HeaderHandler,
 	return multiSigVerifier.VerifyAggregatedSig(pubKeysSigners, hash, signature)
 }
 
-// VerifyHeaderWithProof checks if the proof on the header is correct
-func (hsv *HeaderSigVerifier) VerifyHeaderWithProof(header data.HeaderHandler) error {
+// VerifyHeaderWithPrevProof checks if the proof on the header is correct
```
sstanculeanu, line 297: update comment as well.. or maybe consider unexporting this method? it is only used internally now

process/headerCheck/headerSignatureVerify.go
```diff
@@ -290,25 +294,43 @@ func (hsv *HeaderSigVerifier) VerifySignatureForHash(header data.HeaderHandler,
...
+func (hsv *HeaderSigVerifier) VerifyHeaderWithPrevProof(header data.HeaderHandler) error {
 	// first block for transition to equivalent proofs consensus does not have a previous proof
```
sstanculeanu, line 299: this comment can be removed

process/headerCheck/headerSignatureVerify.go
```diff
@@ -290,25 +294,43 @@ func (hsv *HeaderSigVerifier) VerifySignatureForHash(header data.HeaderHandler,
...
-	}
-
 	err := verifyPrevProofForHeaderIntegrity(header)
 	if err != nil {
 		return err
 	}
 
 	prevProof := header.GetPreviousProof()
+	if common.IsEpochChangeBlockForFlagActivation(header, hsv.enableEpochsHandler, common.EquivalentMessagesFlag) {
+		return hsv.verifyHeaderProofAtTransition(header, prevProof)
```
sstanculeanu, line 307: I don't think this would ever be called, as this method, `VerifyHeaderWithPrevProof`, is called only if `ShouldBlockHavePrevProof`, but if block should have proof, it means it is not the epoch change block for flag activation, thus `VerifySignature` would go through L238-L250 for the epoch start block

sstanculeanu (approved):

ssd04 (approved):

PR #6771, opened 2025-02-03
fix wrong size bitmap error on transition block

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/headerCheck/headerSignatureVerify.go
+++ b/process/headerCheck/headerSignatureVerify.go
@@ -310,7 +310,7 @@ func (hsv *HeaderSigVerifier) VerifyHeaderWithProof(header data.HeaderHandler) e
 	}
 
 	prevProof := header.GetPreviousProof()
-	if prevProof.GetIsStartOfEpoch() {
+	if common.IsEpochStartProofForFlagActivation(prevProof, hsv.enableEpochsHandler) {
 		return hsv.verifyHeaderProofAtTransition(prevProof)
 	}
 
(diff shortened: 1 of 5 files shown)


process/headerCheck/headerSignatureVerify.go
```diff
@@ -370,6 +365,10 @@ func (hsv *HeaderSigVerifier) VerifyHeaderProof(proofHandler data.HeaderProofHan
 		return fmt.Errorf("%w for flag %s", process.ErrFlagNotActive, common.EquivalentMessagesFlag)
 	}
 
+	if proofHandler.GetIsStartOfEpoch() {
+		return hsv.verifyHeaderProofAtTransition(proofHandler)
```
ssd04, line 369: do we need to trigger this flow if previous epoch is with equivalent proofs, and the transition is also to an epoch with equivalent proofs?

ssd04: maybe IsStartOfEpoch and previous epoch not with proofs?

ssd04: we can also have an unit test case for this scenario

sstanculeanu: very good catch, updated in 2 places

ssd04 (approved):

AdoAdoAdo (approved):

ssd04 (approved):

PR #6774, opened 2025-02-04
vm integration for get all transfers value hook, get code hash hook and internal errors masking

laurci:
## Proposed changes
- Integrate new version of VM
- Added new VM hook (managedGetAllTransfersCallValue)
- Added new VM hook (managedGetCodeHash)
- Masked errors from libraries (that should remain internal to the VM) with activation epoch
  
## Related PRs:
- https://github.com/multiversx/mx-chain-vm-go/pull/895
- https://github.com/multiversx/mx-chain-vm-go/pull/898
- https://github.com/multiversx/mx-chain-vm-go/pull/897

### DIFF
--- a/cmd/node/config/enableEpochs.toml
+++ b/cmd/node/config/enableEpochs.toml
@@ -337,7 +337,7 @@
     RelayedTransactionsV3EnableEpoch = 1
 
     # MaskVMInternalDependenciesErrorsEnableEpoch represents the epoch when the additional internal erorr masking in vm is enabled
-    MaskVMInternalDependenciesErrorsEnableEpoch = 10
+    MaskVMInternalDependenciesErrorsEnableEpoch = 1
 
     # BLSMultiSignerEnableEpoch represents the activation epoch for different types of BLS multi-signers
     BLSMultiSignerEnableEpoch = [
--- a/cmd/node/config/enableEpochs.toml
+++ b/cmd/node/config/enableEpochs.toml
@@ -337,7 +337,7 @@
     RelayedTransactionsV3EnableEpoch = 1
 
     # MaskVMInternalDependenciesErrorsEnableEpoch represents the epoch when the additional internal erorr masking in vm is enabled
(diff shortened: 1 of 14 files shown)


sasurobert (dismissed):

sasurobert (dismissed):

sasurobert (dismissed):

danidrasovean (approved):
Normal allin test: v1.8.9-dev-config-373fab6fb9 -> integration-vm-all-call-va-a5e427ab8f

--- Specific errors --- 

block hash does not match 441
wrong nonce in block 226
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  8
Nr. of all WARNS:  4082
Nr. of new ERRORS:  8
Nr. of new WARNS:  3847

/------/ 


--- System test go --- 

/------/

cmd/node/config/enableEpochs.toml
```diff
@@ -336,6 +336,9 @@
     # RelayedTransactionsV3EnableEpoch represents the epoch when the relayed transactions v3 will be enabled
     RelayedTransactionsV3EnableEpoch = 1
 
+    # MaskVMInternalDependenciesErrorsEnableEpoch represents the epoch when the additional internal erorr masking in vm is enabled
+    MaskVMInternalDependenciesErrorsEnableEpoch = 10
```
AdoAdoAdo, line 340: can we have this activated sooner ? 
we have the maximum epoch for activation of the other flags as 4
if there is no other dependency I would set this to 4 as well or even earlier

laurci: this was set for the system test (we had to do some manual testing before and after the activation epoch). before we ship barnard, this should be set to an epoch in the future.

sasurobert (approved):

AdoAdoAdo (approved):

PR #6788, opened 2025-02-10
trigger activation fixes

AdoAdoAdo:
Several fixes to Andromeda identified while testing
- there were some missed blocks on shuffling into a new shard or bootstrapping from storage, due to probable highest block nonce not being set.
- checking finality for meta blocks at transition when there is a mix of old consensus blocks and new consensus blocks being referenced, fails in case of blocks that did not have proofs (were eventually reverted) so they should not have been considered
- not receiving in due time the epoch change data when transitioning from one consensus version to the next would cause the shard nodes to reach consensus on a block with old epoch outside of the grace period for transition 
  
## Proposed changes
- set probable highest nonce on bootstrap
- filter out the attesting blocks that should have proofs but don't
- use the delayed broadcaster component to propagate the block data, so we don't end up with each node requesting it and wasting more time.

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -8,7 +8,6 @@ import (
 	"github.com/multiversx/mx-chain-core-go/data"
 
 	"github.com/multiversx/mx-chain-go/consensus"
-	"github.com/multiversx/mx-chain-go/sharding"
 )
 
 // IsValidRelayedTxV3 returns true if the provided transaction is a valid transaction of type relayed v3
@@ -88,9 +87,9 @@ func VerifyProofAgainstHeader(proof data.HeaderProofHandler, header data.HeaderH
 }
 
 // GetShardIDs returns a map of shard IDs based on the provided shard coordinator
-func GetShardIDs(shardCoordinator sharding.Coordinator) map[uint32]struct{} {
+func GetShardIDs(numShards uint32) map[uint32]struct{} {
 	shardIdentifiers := make(map[uint32]struct{})
-	for i := uint32(0); i < shardCoordinator.NumberOfShards(); i++ {
+	for i := uint32(0); i < numShards; i++ {
 		shardIdentifiers[i] = struct{}{}
 	}
 	shardIdentifiers[core.MetachainShardId] = struct{}{}
--- a/common/common.go
+++ b/common/common.go
@@ -86,7 +86,7 @@ func VerifyProofAgainstHeader(proof data.HeaderProofHandler, header data.HeaderH
 	return nil
 }
 
-// GetShardIDs returns a map of shard IDs based on the provided shard coordinator
+// GetShardIDs returns a map of shard IDs based on the number of shards
 func GetShardIDs(numShards uint32) map[uint32]struct{} {
 	shardIdentifiers := make(map[uint32]struct{})
 	for i := uint32(0); i < numShards; i++ {
--- a/common/common.go
+++ b/common/common.go
@@ -6,7 +6,9 @@ import (
 	"github.com/multiversx/mx-chain-core-go/core"
 	"github.com/multiversx/mx-chain-core-go/core/check"
 	"github.com/multiversx/mx-chain-core-go/data"
+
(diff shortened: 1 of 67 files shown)


consensus/broadcast/shardChainMessenger.go
```diff
@@ -205,6 +227,29 @@ func (scm *shardChainMessenger) PrepareBroadcastHeaderValidator(
 	}
 }
 
+func (scm *shardChainMessenger) PrepareBroadcastBlockDataWithEquivalentProofs(
```
sstanculeanu, line 230: missing comment

consensus/broadcast/shardChainMessenger.go
```diff
@@ -142,34 +142,56 @@ func (scm *shardChainMessenger) BroadcastBlockDataLeader(
...
+	if err != nil {
+		return err
+	}
+	err = scm.delayedBlockBroadcaster.SetLeaderData(dtb.delayedBroadcastData)
 	if err != nil {
 		return err
 	}
 
-	metaMiniBlocks, metaTransactions := scm.extractMetaMiniBlocksAndTransactions(miniBlocks, transactions)
+	go scm.BroadcastBlockData(dtb.metaMiniBlocks, dtb.metaTransactions, pkBytes, common.ExtraDelayForBroadcastBlockInfo)
```
sstanculeanu, line 157: maybe add a TODO to analyze how we can improve this, similar to equivalent proofs broadcast

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -248,11 +248,28 @@ func (sr *subroundEndRound) doEndRoundJobByNode() bool {
...
 }
 
+func (sr *subroundEndRound) prepareBroadcastBlockData() error {
+	miniBlocks, transactions, err := sr.BlockProcessor().MarshalizedDataToBroadcast(sr.GetHeader(), sr.GetBody())
+	if err != nil {
+		return err
+	}
+	if err != nil {
+		return err
+	}
```
sstanculeanu, line 265: duplicate error check

consensus/broadcast/shardChainMessenger.go
AdoAdoAdo: right, added in our checklist, will add the todo here as well.

common/common.go
```diff
@@ -84,3 +85,14 @@ func VerifyProofAgainstHeader(proof data.HeaderProofHandler, header data.HeaderH
 
 	return nil
 }
+
+// GetShardIDs returns a map of shard IDs based on the provided shard coordinator
```
ssd04, line 89:
```suggestion
// GetShardIDs returns a map of shard IDs based on the number of shards
```
?

consensus/broadcast/shardChainMessenger.go
```diff
@@ -205,6 +227,29 @@ func (scm *shardChainMessenger) PrepareBroadcastHeaderValidator(
...
+	if err != nil {
+		return
+	}
+	// everyone broadcasts as if they were the leader
+	err = scm.delayedBlockBroadcaster.SetLeaderData(dtb.delayedBroadcastData)
+	if err != nil {
+		return
+	}
+
+	// todo: consider moving this to the initial block broadcast - optimization
```
ssd04, line 251:
```suggestion
	// TODO: consider moving this to the initial block broadcast - optimization
```
use upper case to match common grep pattern?

consensus/spos/bls/v2/subroundEndRound.go
AdoAdoAdo: done

consensus/broadcast/shardChainMessenger.go
AdoAdoAdo: done

sstanculeanu (approved):

sstanculeanu (approved):

ssd04 (approved):

PR #6793, opened 2025-02-11
Automatic activation of nodes disabled

sasurobert:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


andreibancioiu (approved):

vm/systemSmartContracts/delegation.go
```diff
@@ -810,6 +814,10 @@ func (d *delegation) setAutomaticActivation(args *vmcommon.ContractCallInput) vm
 
 	switch string(args.Arguments[0]) {
 	case "true":
+		if d.enableEpochsHandler.IsFlagEnabled(common.AutomaticActivationOfNodesDisableFlag) {
```
AdoAdoAdo, line 817: so we return user error if the transaction is with automatic activation but the automatic activation is no longer supported by the protocol, enforcing the user to explicitly set it.

I think it would be better to simply ignore the argument value without returning an error in this case.

AdoAdoAdo (approved):

statusHandler/statusMetricsProvider_test.go
```diff
@@ -537,8 +543,12 @@ func TestStatusMetrics_EnableEpochMetrics(t *testing.T) {
 		common.MetricMultiESDTNFTTransferAndExecuteByUserEnableEpoch:          uint64(4),
 		common.MetricFixRelayedMoveBalanceToNonPayableSCEnableEpoch:           uint64(4),
 		common.MetricRelayedTransactionsV3EnableEpoch:                         uint64(4),
+		common.MetricRelayedTransactionsV3FixESDTTransferEnableEpoch:          uint64(4),
```
andreibancioiu, line 546: :+1:

andreibancioiu: We should bring these to Andromeda maybe, as well.

andreibancioiu (approved):

danidrasovean (approved):
Normal allin test: v1.8.12-dev-config-5558df2bd0 -> automatic-activation-of-no-379b186f35

--- Specific errors --- 

block hash does not match 10
wrong nonce in block 27
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  12
Nr. of all WARNS:  208
Nr. of new ERRORS:  12
Nr. of new WARNS:  0
Nr. of PANICS:  0
Nodes with panic:  []

/------/ 


--- System test go --- 

/------/

AdoAdoAdo (approved):

PR #6873, opened 2025-03-12
Invalid signers cache

sstanculeanu:
- as now all participants can aggregate signatures and send final info, the network might get flooded with the invalid signers messages
  
## Proposed changes
- added a mechanism similar to equivalent proofs to avoid further broadcast in case of known invalid signers message

## Testing procedure
- with rc branch

## Pre-requisites

### DIFF
--- a/consensus/spos/bls/v2/subroundEndRound.go
+++ b/consensus/spos/bls/v2/subroundEndRound.go
@@ -150,13 +150,12 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 		return false
 	}
 
-	invalidSignersHash := sr.Hasher().Compute(string(cnsDta.InvalidSigners))
 	invalidSignersCache := sr.InvalidSignersCache()
-	if invalidSignersCache.HasInvalidSigners(string(invalidSignersHash)) {
+	if invalidSignersCache.HasInvalidSigners(cnsDta.BlockHeaderHash, cnsDta.InvalidSigners) {
 		return false
 	}
 
-	err := sr.verifyInvalidSigners(cnsDta.InvalidSigners)
+	invalidSignersPubKeys, err := sr.verifyInvalidSigners(cnsDta.InvalidSigners)
 	if err != nil {
 		log.Trace("receivedInvalidSignersInfo.verifyInvalidSigners", "error", err.Error())
 		return false
@@ -164,7 +163,7 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 
 	log.Debug("step 3: invalid signers info has been evaluated")
 
-	invalidSignersCache.AddInvalidSigners(string(invalidSignersHash))
+	invalidSignersCache.AddInvalidSigners(cnsDta.BlockHeaderHash, cnsDta.InvalidSigners, invalidSignersPubKeys)
 
 	sr.PeerHonestyHandler().ChangeScore(
 		messageSender,
@@ -175,32 +174,37 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 	return true
 }
 
-func (sr *subroundEndRound) verifyInvalidSigners(invalidSigners []byte) error {
+func (sr *subroundEndRound) verifyInvalidSigners(invalidSigners []byte) ([]string, error) {
 	messages, err := sr.MessageSigningHandler().Deserialize(invalidSigners)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
+	pubKeys := make([]string, 0, len(messages))
 	for _, msg := range messages {
-		err = sr.verifyInvalidSigner(msg)
-		if err != nil {
-			return err
+		pubKey, errVerify := sr.verifyInvalidSigner(msg)
+		if errVerify != nil {
+			return nil, errVerify
+		}
+
+		if len(pubKey) > 0 {
+			pubKeys = append(pubKeys, pubKey)
 		}
 	}
 
-	return nil
+	return pubKeys, nil
 }
 
-func (sr *subroundEndRound) verifyInvalidSigner(msg p2p.MessageP2P) error {
+func (sr *subroundEndRound) verifyInvalidSigner(msg p2p.MessageP2P) (string, error) {
 	err := sr.MessageSigningHandler().Verify(msg)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	cnsMsg := &consensus.Message{}
 	err = sr.Marshalizer().Unmarshal(cnsMsg, msg.Data())
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = sr.SigningHandler().VerifySingleSignature(cnsMsg.PubKey, cnsMsg.BlockHeaderHash, cnsMsg.SignatureShare)
@@ -211,9 +215,11 @@ func (sr *subroundEndRound) verifyInvalidSigner(msg p2p.MessageP2P) error {
 			"error", err.Error(),
 		)
 		sr.applyBlacklistOnNode(msg.Peer())
+
+		return string(cnsMsg.PubKey), nil
 	}
 
-	return nil
+	return "", nil
 }
 
 func (sr *subroundEndRound) applyBlacklistOnNode(peer core.PeerID) {
@@ -520,7 +526,7 @@ func (sr *subroundEndRound) handleInvalidSignersOnAggSigFail() ([]byte, []byte, 
 	}
 
 	if len(invalidSigners) > 0 {
-		sr.createAndBroadcastInvalidSigners(invalidSigners)
+		sr.createAndBroadcastInvalidSigners(invalidSigners, invalidPubKeys)
 	}
 
 	bitmap, sig, err := sr.computeAggSigOnValidNodes()
@@ -625,7 +631,7 @@ func (sr *subroundEndRound) getRandomManagedKeyProofSender() string {
 	return randManagedKey
 }
 
-func (sr *subroundEndRound) createAndBroadcastInvalidSigners(invalidSigners []byte) {
+func (sr *subroundEndRound) createAndBroadcastInvalidSigners(invalidSigners []byte, invalidSignersPubKeys []string) {
 	if !sr.ShouldConsiderSelfKeyInConsensus() {
 		return
 	}
@@ -653,8 +659,7 @@ func (sr *subroundEndRound) createAndBroadcastInvalidSigners(invalidSigners []by
 		invalidSigners,
 	)
 
-	invalidSignersHash := sr.Hasher().Compute(string(invalidSigners))
-	sr.InvalidSignersCache().AddInvalidSigners(string(invalidSignersHash))
+	sr.InvalidSignersCache().AddInvalidSigners(sr.GetData(), invalidSigners, invalidSignersPubKeys)
 
 	err = sr.BroadcastMessenger().BroadcastConsensusMessage(cnsMsg)
 	if err != nil {
--- a/consensus/spos/bls/v2/subroundEndRound.go
+++ b/consensus/spos/bls/v2/subroundEndRound.go
@@ -150,6 +150,12 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 		return false
 	}
 
+	invalidSignersHash := sr.Hasher().Compute(string(cnsDta.InvalidSigners))
+	invalidSignersCache := sr.InvalidSignersCache()
+	if invalidSignersCache.HasInvalidSigners(string(invalidSignersHash)) {
+		return false
+	}
+
 	err := sr.verifyInvalidSigners(cnsDta.InvalidSigners)
 	if err != nil {
 		log.Trace("receivedInvalidSignersInfo.verifyInvalidSigners", "error", err.Error())
@@ -157,6 +163,8 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 	}
 
 	log.Debug("step 3: invalid signers info has been evaluated")
+
+	invalidSignersCache.AddInvalidSigners(string(invalidSignersHash))
 
 	sr.PeerHonestyHandler().ChangeScore(
 		messageSender,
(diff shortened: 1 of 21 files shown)


consensus/spos/errors.go
```diff
@@ -276,3 +276,9 @@ var ErrNilEpochNotifier = errors.New("nil epoch notifier")
 
 // ErrNilEpochStartNotifier signals that nil epoch start notifier has been provided
 var ErrNilEpochStartNotifier = errors.New("nil epoch start notifier")
+
+// ErrInvalidSignersAlreadyReceived signals that an invalid signers message has been already received
+var ErrInvalidSignersAlreadyReceived = errors.New("invalid signers already received")
+
+// ErrNilInvalidSignersCache signals that nil invalid signers has been provided
```
andreibancioiu, line 283: Typo `nil invalid signers` :upside_down_face:

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -501,24 +509,23 @@ func (sr *subroundEndRound) handleInvalidSignersOnAggSigFail() ([]byte, []byte,
...
+		log.Debug("handleInvalidSignersOnAggSigFail.getFullMessagesForInvalidSigners", "error", err.Error())
 		return nil, nil, err
 	}
 
-	// TODO: handle invalid signers broadcast without flooding the network
-	// if len(invalidSigners) > 0 {
-	// 	sr.createAndBroadcastInvalidSigners(invalidSigners)
-	// }
+	if len(invalidSigners) > 0 {
+		sr.createAndBroadcastInvalidSigners(invalidSigners)
```
andreibancioiu, line 523: If it was cached previously, we had the early return, thus no broadcasting. All good?

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -501,24 +509,23 @@ func (sr *subroundEndRound) handleInvalidSignersOnAggSigFail() ([]byte, []byte,
 	invalidPubKeys, err := sr.verifyNodesOnAggSigFail(ctx)
 	cancel()
```
raduchis, line 510: not related to the changes from this PR, but this cancel() doesn't seem right to call it after the "intensive". Maybe move it up 1 line and call it on a defer?

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -150,6 +150,12 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 		return false
 	}
 
+	invalidSignersHash := sr.Hasher().Compute(string(cnsDta.InvalidSigners))
+	invalidSignersCache := sr.InvalidSignersCache()
+	if invalidSignersCache.HasInvalidSigners(string(invalidSignersHash)) {
```
raduchis, line 155: if there are different number of invalidSigners: first time 2: A and B, and in on a second message it gets only A, it seems that the message will be propagated 2 times?

consensus/spos/worker.go
```diff
@@ -463,6 +470,13 @@ func (wrk *Worker) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedP
 		wrk.doJobOnMessageWithSignature(cnsMsg, message)
 	}
 
+	if isMessageWithInvalidSigners {
```
raduchis, line 473: no need for a activation epoch here? If a message with invalid signers comes before the activation, the old handling of "ProcessReceivedMessage" will be different.

consensus/spos/worker.go
```diff
@@ -485,7 +499,8 @@ func (wrk *Worker) shouldBlacklistPeer(err error) bool {
 		errors.Is(err, errorsErd.ErrSignatureMismatch) ||
 		errors.Is(err, nodesCoordinator.ErrEpochNodesConfigDoesNotExist) ||
 		errors.Is(err, ErrMessageTypeLimitReached) ||
-		errors.Is(err, ErrEquivalentMessageAlreadyReceived) {
+		errors.Is(err, ErrEquivalentMessageAlreadyReceived) ||
```
raduchis, line 502: also here, maybe epoch activation check. If some malicious node sends EquivalentMessages or InvalidSigners before the activation, it might blacklist the entire network.

raduchis (changes requested):

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,49 @@
+package spos
+
+import "sync"
+
+type invalidSignersCache struct {
+	sync.RWMutex
+	invalidSignersMap map[string]struct{}
```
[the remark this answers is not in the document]
sstanculeanu: applied

consensus/spos/worker.go
sstanculeanu: added early exit, indeed it may happen

consensus/spos/worker.go
sstanculeanu: added

consensus/spos/invalidSignersCache.go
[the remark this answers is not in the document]
sstanculeanu: back to map + mut due to reset method

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+type ArgInvalidSignersCache struct {
+	Hasher         hashing.Hasher
+	SigningHandler p2p.P2PSigningHandler
+	Marshaller     marshal.Marshalizer
+}
+
+type invalidSignersCache struct {
+	sync.RWMutex
+	invalidSignersHashesMap    map[string]struct{}
+	invalidSignersForHeaderMap map[string]map[string]struct{}
```
andreibancioiu, line 23: So, for each header, we retain the public keys of the "bad" signers (as a map, for easy lookup), correct?

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+
+// HasInvalidSigners check whether the provided hash exists in the internal map or not
+func (cache *invalidSignersCache) HasInvalidSigners(headerHash []byte, invalidSigners []byte) bool {
+	cache.RLock()
+	defer cache.RUnlock()
+
+	invalidSignersHash := cache.hasher.Compute(string(invalidSigners))
+	_, hasSameInvalidSigners := cache.invalidSignersHashesMap[string(invalidSignersHash)]
+	if hasSameInvalidSigners {
+		return true
```
andreibancioiu, line 89: Early answer if we've previously received the exact same message, correct? I mean, this is just an optimization - is it a crucial one?

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return false
+	}
+
+	knownInvalidSigners := 0
+	for _, msg := range messages {
+		cnsMsg := &consensus.Message{}
+		err = cache.marshaller.Unmarshal(cnsMsg, msg.Data())
+		if err != nil {
+			return false
```
andreibancioiu, line 107: Unfortunate / rare circumstance? If so, maybe we should log it (as warning perhaps)?

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+	if hasSameInvalidSigners {
+		return true
+	}
+
+	_, isHeaderKnown := cache.invalidSignersForHeaderMap[string(headerHash)]
+	if !isHeaderKnown {
+		return false
+	}
+
+	messages, err := cache.signingHandler.Deserialize(invalidSigners)
```
andreibancioiu, line 97: Maybe can be renamed from `messages` to `signers`. Or even `invalidSigners`, while the existing variable can be renamed to `serializedInvalidSigners`.

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+			return false
+		}
+
+		_, isKnownInvalidSigner := cache.invalidSignersForHeaderMap[string(headerHash)][string(cnsMsg.PubKey)]
+		if isKnownInvalidSigner {
+			knownInvalidSigners++
+		}
+	}
+
+	return knownInvalidSigners == len(messages)
```
andreibancioiu, line 116: Opinion (sorry if mistaken, I might have missed something): we can also return early if we find an unknown one, and skip the counting logic.

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Reset clears the internal map
```
andreibancioiu, line 119: `maps`.

consensus/spos/invalidSignersCache.go
andreibancioiu: I am thinking out loud - if this node already knows about this exact set of invalid signers, for the given block, it will not forward the message / it will stop the gossip. Also, if the node does not know about this exact set, but knows about a superset of invalid signers (which include this set), it will also stop the gossip.

Correct?

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -167,32 +174,37 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
...
-		err = sr.verifyInvalidSigner(msg)
-		if err != nil {
-			return err
+		pubKey, errVerify := sr.verifyInvalidSigner(msg)
+		if errVerify != nil {
+			return nil, errVerify
+		}
+
+		if len(pubKey) > 0 {
+			pubKeys = append(pubKeys, pubKey)
```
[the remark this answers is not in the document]
sstanculeanu: not really, this method checks whether the invalid signers provided an invalid signature.. so this appends the  invalid signers that we confirmed they had an invalid signature

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -167,32 +174,37 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
...
 	}
 
+	pubKeys := make([]string, 0, len(messages))
 	for _, msg := range messages {
-		err = sr.verifyInvalidSigner(msg)
-		if err != nil {
-			return err
+		pubKey, errVerify := sr.verifyInvalidSigner(msg)
+		if errVerify != nil {
+			return nil, errVerify
```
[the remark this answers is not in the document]
sstanculeanu: yes, it means we were not able to verify if the provided invalid signers indeed had invalid signatures

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -167,32 +174,37 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
...
-	return nil
+	return pubKeys, nil
 }
 
-func (sr *subroundEndRound) verifyInvalidSigner(msg p2p.MessageP2P) error {
+func (sr *subroundEndRound) verifyInvalidSigner(msg p2p.MessageP2P) (string, error) {
 	err := sr.MessageSigningHandler().Verify(msg)
 	if err != nil {
-		return err
+		return "", err
```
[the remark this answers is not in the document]
sstanculeanu: correct

consensus/spos/invalidSignersCache.go
sstanculeanu: for the first question, it indeed is just an optimization to avoid the Deserialization

for the second one, correct

consensus/spos/invalidSignersCache.go
sstanculeanu: hm.. not sure it is needed. worst case, the node will process twice the same message

consensus/spos/invalidSignersCache.go
sstanculeanu: applied suggestion, renamed to `invalidSignersP2PMessages`

andreibancioiu (dismissed):

consensus/spos/worker.go
raduchis: is it added somewhere else?

consensus/spos/worker.go
raduchis: is it added somewhere else?

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -501,24 +515,23 @@ func (sr *subroundEndRound) handleInvalidSignersOnAggSigFail() ([]byte, []byte,
...
 
-	// TODO: handle invalid signers broadcast without flooding the network
-	// if len(invalidSigners) > 0 {
-	// 	sr.createAndBroadcastInvalidSigners(invalidSigners)
-	// }
+	if len(invalidSigners) > 0 {
+		sr.createAndBroadcastInvalidSigners(invalidSigners, invalidPubKeys)
+	}
 
 	bitmap, sig, err := sr.computeAggSigOnValidNodes()
```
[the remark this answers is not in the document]
sstanculeanu: they are not.. opened thread, perhaps we tackle this in a different PR

consensus/spos/invalidSignersCache.go
```diff
@@ -0,0 +1,130 @@
...
+		cache.invalidSignersForHeaderMap[string(headerHash)] = make(map[string]struct{})
+	}
+
+	for _, pk := range invalidPublicKeys {
+		cache.invalidSignersForHeaderMap[string(headerHash)][pk] = struct{}{}
+	}
+}
+
+// HasInvalidSigners check whether the provided hash exists in the internal map or not
+func (cache *invalidSignersCache) HasInvalidSigners(headerHash []byte, serializedInvalidSigners []byte) bool {
```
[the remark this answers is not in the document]
sstanculeanu: updated the comment and renamed to CheckKnownInvalidSigners

consensus/spos/worker.go
sstanculeanu: removed it, forgot to update the comment

as discussed, even though a message would be received before activation, the cache should not have any data, thus this verification should always return nil

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -150,14 +150,21 @@ func (sr *subroundEndRound) receivedInvalidSignersInfo(_ context.Context, cnsDta
 		return false
 	}
 
-	err := sr.verifyInvalidSigners(cnsDta.InvalidSigners)
+	invalidSignersCache := sr.InvalidSignersCache()
+	if invalidSignersCache.HasInvalidSigners(cnsDta.BlockHeaderHash, cnsDta.InvalidSigners) {
+		return false
+	}
+
+	invalidSignersPubKeys, err := sr.verifyInvalidSigners(cnsDta.InvalidSigners)
```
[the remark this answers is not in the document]
sstanculeanu: correct, maybe we can add an early exit here.. can be done in a different PR

andreibancioiu (approved):

consensus/spos/bls/v2/subroundEndRound.go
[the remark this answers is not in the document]
raduchis: ok

consensus/spos/invalidSignersCache.go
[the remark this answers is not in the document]
raduchis: ok

consensus/spos/worker.go
raduchis: ok

consensus/spos/worker.go
raduchis: ok

raduchis (approved):

PR #6928, opened 2025-04-01
use KeyRWMutexHandler for CheckValidity of equivalent proofs

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/multiversx/mx-chain-core-go/core"
 	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/core/sync"
 	"github.com/multiversx/mx-chain-core-go/data"
 	"github.com/multiversx/mx-chain-core-go/data/block"
 	"github.com/multiversx/mx-chain-core-go/hashing"
@@ -41,6 +42,7 @@ type interceptedEquivalentProof struct {
 	marshaller        marshaling.Marshalizer
 	hasher            hashing.Hasher
 	hash              []byte
+	km                sync.KeyRWMutexHandler
 }
 
 // NewInterceptedEquivalentProof returns a new instance of interceptedEquivalentProof
@@ -66,6 +68,7 @@ func NewInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) (*interce
 		marshaller:        args.Marshaller,
 		hasher:            args.Hasher,
 		hash:              hash,
+		km:                sync.NewKeyRWMutex(),
 	}, nil
 }
 
@@ -136,6 +139,10 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 	if err != nil {
 		return err
 	}
+
+	headerHash := string(iep.proof.GetHeaderHash())
+	iep.km.Lock(headerHash)
+	defer iep.km.Unlock(headerHash)
 
 	ok := iep.proofsPool.HasProof(iep.proof.GetHeaderShardId(), iep.proof.GetHeaderHash())
 	if ok {
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -36,6 +36,7 @@ type ArgInterceptedEquivalentProof struct {
 	KeyRWMutexHandler  sync.KeyRWMutexHandler
 	EligibleNodesCache process.EligibleNodesCache
 	MessageOriginator  core.PeerID
+	WhiteListHandler   process.WhiteListHandler
 }
 
 type interceptedEquivalentProof struct {
@@ -51,6 +52,7 @@ type interceptedEquivalentProof struct {
 	km                 sync.KeyRWMutexHandler
 	eligibleNodesCache process.EligibleNodesCache
 	messageOriginator  core.PeerID
+	whiteListHandler   process.WhiteListHandler
 }
 
 // NewInterceptedEquivalentProof returns a new instance of interceptedEquivalentProof
@@ -80,6 +82,7 @@ func NewInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) (*interce
 		km:                 args.KeyRWMutexHandler,
 		eligibleNodesCache: args.EligibleNodesCache,
 		messageOriginator:  args.MessageOriginator,
+		whiteListHandler:   args.WhiteListHandler,
 	}, nil
 }
 
@@ -113,6 +116,9 @@ func checkArgInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) erro
 	}
 	if check.IfNil(args.EligibleNodesCache) {
 		return process.ErrNilEligibleNodesCache
+	}
+	if check.IfNil(args.WhiteListHandler) {
+		return process.ErrNilWhiteListHandler
 	}
 
 	return nil
@@ -160,10 +166,11 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 		return err
 	}
 
-	// TODO: avoid this check if the intercepted proof was requested by the current node,
-	// in order to avoid dropping valid messages from not-eligible nodes
-	if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), common.GetEpochForConsensus(iep.proof)) {
-		return fmt.Errorf("%w, proof originator must be an eligible node", process.ErrInvalidHeaderProof)
+	shouldSkipEligibleVerification := iep.whiteListHandler.IsWhiteListed(iep)
+	if !shouldSkipEligibleVerification {
+		if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), common.GetEpochForConsensus(iep.proof)) {
+			return fmt.Errorf("%w, proof originator must be an eligible node", process.ErrInvalidHeaderProof)
+		}
 	}
 
 	headerHash := string(iep.proof.GetHeaderHash())
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -31,6 +31,7 @@ type ArgInterceptedEquivalentProof struct {
 	HeaderSigVerifier consensus.HeaderSigVerifier
 	Proofs            dataRetriever.ProofsPool
 	Headers           dataRetriever.HeadersPool
+	KeyRWMutexHandler sync.KeyRWMutexHandler
 }
 
 type interceptedEquivalentProof struct {
@@ -68,7 +69,7 @@ func NewInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) (*interce
 		marshaller:        args.Marshaller,
 		hasher:            args.Hasher,
 		hash:              hash,
-		km:                sync.NewKeyRWMutex(),
+		km:                args.KeyRWMutexHandler,
 	}, nil
 }
 
@@ -93,6 +94,9 @@ func checkArgInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) erro
 	}
 	if check.IfNil(args.Hasher) {
 		return process.ErrNilHasher
+	}
+	if check.IfNil(args.KeyRWMutexHandler) {
+		return process.ErrNilKeyRWMutexHandler
 	}
 
 	return nil
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -160,7 +160,7 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 		return err
 	}
 
-	if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), iep.getEpochForEligibleList()) {
+	if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), common.GetEpochForConsensus(iep.proof)) {
 		return fmt.Errorf("%w, proof originator must be an eligible node", process.ErrInvalidHeaderProof)
 	}
 
@@ -186,14 +186,6 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 	}
 
 	return nil
-}
-func (iep *interceptedEquivalentProof) getEpochForEligibleList() uint32 {
-	epochForEligibleList := iep.proof.GetHeaderEpoch()
-	if iep.proof.GetIsStartOfEpoch() && epochForEligibleList > 0 {
-		return epochForEligibleList - 1
-	}
-
-	return epochForEligibleList
 }
 
 func (iep *interceptedEquivalentProof) integrity() error {
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -160,6 +160,8 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 		return err
 	}
 
+	// TODO: avoid this check if the intercepted proof was requested by the current node,
+	// in order to avoid dropping valid messages from not-eligible nodes
 	if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), common.GetEpochForConsensus(iep.proof)) {
 		return fmt.Errorf("%w, proof originator must be an eligible node", process.ErrInvalidHeaderProof)
 	}
--- a/process/block/interceptedBlocks/interceptedEquivalentProof.go
+++ b/process/block/interceptedBlocks/interceptedEquivalentProof.go
@@ -25,28 +25,32 @@ const interceptedEquivalentProofType = "intercepted equivalent proof"
 
 // ArgInterceptedEquivalentProof is the argument used in the intercepted equivalent proof constructor
 type ArgInterceptedEquivalentProof struct {
-	DataBuff          []byte
-	Marshaller        marshal.Marshalizer
-	Hasher            hashing.Hasher
-	ShardCoordinator  sharding.Coordinator
-	HeaderSigVerifier consensus.HeaderSigVerifier
-	Proofs            dataRetriever.ProofsPool
-	Headers           dataRetriever.HeadersPool
-	ProofSizeChecker  common.FieldsSizeChecker
-	KeyRWMutexHandler sync.KeyRWMutexHandler
+	DataBuff           []byte
+	Marshaller         marshal.Marshalizer
+	Hasher             hashing.Hasher
+	ShardCoordinator   sharding.Coordinator
+	HeaderSigVerifier  consensus.HeaderSigVerifier
+	Proofs             dataRetriever.ProofsPool
+	Headers            dataRetriever.HeadersPool
+	ProofSizeChecker   common.FieldsSizeChecker
+	KeyRWMutexHandler  sync.KeyRWMutexHandler
+	EligibleNodesCache process.EligibleNodesCache
+	MessageOriginator  core.PeerID
 }
 
 type interceptedEquivalentProof struct {
-	proof             *block.HeaderProof
-	isForCurrentShard bool
-	headerSigVerifier consensus.HeaderSigVerifier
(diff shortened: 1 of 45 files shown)


process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -66,6 +68,7 @@ func NewInterceptedEquivalentProof(args ArgInterceptedEquivalentProof) (*interce
 		marshaller:        args.Marshaller,
 		hasher:            args.Hasher,
 		hash:              hash,
+		km:                sync.NewKeyRWMutex(),
```
andreibancioiu, line 71: I didn't know we had such a component :+1:

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -137,6 +140,10 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 		return err
 	}
 
+	headerHash := string(iep.proof.GetHeaderHash())
+	iep.km.Lock(headerHash)
```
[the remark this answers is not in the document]
andreibancioiu: If everyone (every node behaves), this should be fine. Otherwise (malevolent nodes spamming), can it be problematic.

andreibancioiu (dismissed):

ssd04 (dismissed):

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -31,6 +31,7 @@ type ArgInterceptedEquivalentProof struct {
 	HeaderSigVerifier consensus.HeaderSigVerifier
 	Proofs            dataRetriever.ProofsPool
 	Headers           dataRetriever.HeadersPool
+	KeyRWMutexHandler sync.KeyRWMutexHandler
```
andreibancioiu, line 34: Do we have to receive it as a parameter, is it a shared resource? If it's only an internal component that helps us serialize some proof checks, we can simply instantiate it here, in the constructor.

andreibancioiu (approved):

process/interceptors/processor/equivalentProofsInterceptorProcessor.go
```diff
@@ -40,14 +48,38 @@ func checkArgsEquivalentProofs(args ArgEquivalentProofsInterceptorProcessor) err
...
+// Validate checks if the intercepted data was sent by an eligible node
+func (epip *equivalentProofsInterceptorProcessor) Validate(data process.InterceptedData, pid core.PeerID) error {
+	interceptedProof, ok := data.(interceptedEquivalentProof)
+	if !ok {
+		return process.ErrWrongTypeAssertion
+	}
+
+	proof := interceptedProof.GetProof()
+
+	eligibleList, err := epip.nodesCoordinator.GetAllEligibleValidatorsPublicKeysForShard(proof.GetHeaderEpoch(), proof.GetHeaderShardId())
```
andreibancioiu, line 70: Is `GetAllEligibleValidatorsPublicKeysForShard` a "fast-enough" operation?

process/interceptors/processor/equivalentProofsInterceptorProcessor.go
```diff
@@ -40,14 +48,38 @@ func checkArgsEquivalentProofs(args ArgEquivalentProofsInterceptorProcessor) err
...
+	if err != nil {
+		return err
+	}
+
+	peerInfo := epip.peerShardMapper.GetPeerInfo(pid)
+	for _, eligibleValidator := range eligibleList {
+		if string(peerInfo.PkBytes) == eligibleValidator {
+			return nil
```
andreibancioiu, line 78: If the originator is a validator, we say "all good" (in interceptors) :+1:

andreibancioiu (dismissed):

process/interceptors/processor/equivalentProofsInterceptorProcessor.go
ssd04: it is taken from nodes config, which is in memory, in a map in nodesCoordinator, so it should be fast-enough

ssd04 (approved):

process/interceptors/processor/equivalentProofsInterceptorProcessor.go
AdoAdoAdo, line 70: the epoch for the start of epoch block is not OK, should be epoch-1.

AdoAdoAdo: I think we can keep the check for eligible nodes, but cache it in a map inside the interceptor, so for the entire epoch we only do the computation once, and then just access the map.

sstanculeanu: pushed the suggestion

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -144,12 +160,40 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
...
+
+	return nil
+}
+func (iep *interceptedEquivalentProof) getEpochForEligibleList() uint32 {
+	epochForEligibleList := iep.proof.GetHeaderEpoch()
+	if iep.proof.GetIsStartOfEpoch() && epochForEligibleList > 0 {
+		return epochForEligibleList - 1
+	}
+
+	return epochForEligibleList
```
ssd04, line 196: with the latest merged PR with fields checker we have in common `GetEpochForConsensus` which can be used for this

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+type eligibleNodesCache struct {
+	// eligibleNodesMap holds a map with key shard id and value
+	//		a map with key epoch and value
+	//		a map with key eligible node pk and value struct{}
+	// this is needed because:
+	// 	- a shard node will intercept proof for self shard and meta
+	// 	- a meta node will intercept proofs from all shards
+	eligibleNodesMap    map[uint32]map[uint32]map[string]struct{}
```
ssd04, line 18: can we use a combined shard+epoch key in the map here?

sstanculeanu: hm.. not sure if this would work in terms of cleanup. when a new epoch is received, with the current implementation it is easy to clean the entries for that shard and other epochs.. using a combined key, that would not be possible

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,102 @@
...
+	}, nil
+}
+
+// IsPeerEligible returns true if the provided peer is eligible
+func (cache *eligibleNodesCache) IsPeerEligible(pid core.PeerID, shard uint32, epoch uint32) bool {
+	cache.mutEligibleNodesMap.Lock()
+	defer cache.mutEligibleNodesMap.Unlock()
+
+	var err error
+	eligibleNodesForShard, hasShardCached := cache.eligibleNodesMap[shard]
```
AdoAdoAdo, line 49: can you use instead read lock and only use Lock when you need to update the map?
then most of the accesses could be done without waiting.

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,102 @@
...
+	if !hasEpochCachedForShard {
+		// get the eligible list for the new epoch from nodesCoordinator
+		eligibleNodesForShardInEpoch, err = cache.getEligibleNodesForShardInEpoch(epoch, shard)
+		if err != nil {
+			return false
+		}
+
+		// if no eligible list was cached for this epoch, it means that it is a new one
+		// it is safe to clean the cached eligible list for other epochs for this shard
+		cache.eligibleNodesMap[shard] = map[uint32]map[string]struct{}{
```
AdoAdoAdo, line 73: why not cache.eligibleNodesMap[shard][epoch] = eligibleNodesForShardInEpoch?

AdoAdoAdo: is this in order to cleanup also the other epochs data? 
I think in case of syncing node in previous epoch-1, you will do cleanups of epoch data back and every round, replacing epoch with epoch -1 data back and forth.

I think we should keep at least a few epochs config (like we do in nodes coordinator)

process/interceptors/factory/eligibleNodesCache_test.go
```diff
@@ -0,0 +1,278 @@
...
+		// should not return from cache, new shard
+		require.True(t, cache.IsPeerEligible(providedPid, 1, 0))
+
+		// should return from cache, same epoch and shard
+		require.True(t, cache.IsPeerEligible(providedPid, 1, 0))
+
+		// should not return from cache, new epoch
+		require.True(t, cache.IsPeerEligible(providedPid, 1, 1))
+
+		// expecting to calls to nodesCoordinator:
```
AdoAdoAdo, line 164: expecting to calls -> expecting calls?

AdoAdoAdo (changes requested):

process/interceptors/factory/eligibleNodesCache.go
sstanculeanu: updated as suggested

process/interceptors/factory/eligibleNodesCache.go
sstanculeanu: indeed that was the reason, to delete old epochs.. enhanced the implementation with a epochsDelta of 3

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -144,12 +160,32 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
 		return err
 	}
 
+	if !iep.eligibleNodesCache.IsPeerEligible(iep.messageOriginator, iep.proof.GetHeaderShardId(), common.GetEpochForConsensus(iep.proof)) {
```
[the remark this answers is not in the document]
sstanculeanu: added TODO for another PR as it may require bigger changes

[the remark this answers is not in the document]
sstanculeanu: pushed

andreibancioiu (approved):

process/block/interceptedBlocks/interceptedEquivalentProof.go
```diff
@@ -144,12 +160,32 @@ func (iep *interceptedEquivalentProof) CheckValidity() error {
...
+	err = iep.headerSigVerifier.VerifyHeaderProof(iep.proof)
+	if err != nil {
+		return err
+	}
+
+	// also save the proof here in order to complete the flow under mutex lock
+	wasAdded := iep.proofsPool.AddProof(iep.proof)
+	if !wasAdded {
+		// with the current implementation, this should never happen
+		return common.ErrAlreadyExistingEquivalentProof
```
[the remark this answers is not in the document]
sstanculeanu: with the current implementation, with the critical section, if the proof already exist we will still return this error but from L182.. I don't think there is a use case where we return on L185, but just for safety, added return  err

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+		return isPeerEligible
+	}
+
+	cache.mutEligibleNodesMap.Lock()
+	defer cache.mutEligibleNodesMap.Unlock()
+
+	// get the eligible list for the new epoch from nodesCoordinator
+	eligibleNodesForShardInEpoch, err := cache.getEligibleNodesForShardInEpoch(epoch, shard)
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
sstanculeanu: added log trace

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+}
+
+func (cache *eligibleNodesCache) isPeerEligible(pid core.PeerID, shard uint32, epoch uint32) (bool, bool) {
+	cache.mutEligibleNodesMap.RLock()
+	defer cache.mutEligibleNodesMap.RUnlock()
+
+	eligibleNodesForShard, hasShardCached := cache.eligibleNodesMap[shard]
+	if !hasShardCached {
+		return false, true
```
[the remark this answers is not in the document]
sstanculeanu: added comments

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+	// there are different situations for the new epoch:
+	//	1. the epoch is higher than the min cached epoch + epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
+	//  2. the epoch is lower than the max cached epoch - epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
+	// 	3. the new epoch is higher than the min cached but still in the delta limits ->
+	//		cache it and update the max cached if needed
+	//	4. the new epoch is lower than the max cached but still in the delta limits ->
+	//		cache it and update the min cached if needed
+	isHigherInRange := newEpoch > limitEpochsForShard.min && newEpoch < limitEpochsForShard.min+epochsDelta
```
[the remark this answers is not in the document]
sstanculeanu: good point.. still, it has 100% coverage

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+	isHigherOutOfRange := newEpoch >= limitEpochsForShard.min+epochsDelta
+	isLowerInRange := newEpoch < limitEpochsForShard.max && newEpoch > limitEpochsForShard.max-epochsDelta && limitEpochsForShard.max >= epochsDelta
+	isLowerOutOfRange := newEpoch <= limitEpochsForShard.max-epochsDelta && limitEpochsForShard.max >= epochsDelta
+
+	// case 1.
+	if isHigherOutOfRange {
+		tmpMin := newEpoch
```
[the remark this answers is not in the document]
sstanculeanu: exactly, the reason why I initialized it with newEpoch was the case where we iterate over nothing

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+		for cachedEpoch := range cache.eligibleNodesMap[shard] {
+			shouldDeleteEpoch := cachedEpoch <= newEpoch-epochsDelta
+			if shouldDeleteEpoch {
```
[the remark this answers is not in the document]
sstanculeanu: tried to keep it as tight as possible

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,225 @@
...
+	//		store the new + all cached ones that ate still in delta limit
+	//  2. the epoch is lower than the max cached epoch - epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
+	// 	3. the new epoch is higher than the min cached but still in the delta limits ->
+	//		cache it and update the max cached if needed
+	//	4. the new epoch is lower than the max cached but still in the delta limits ->
+	//		cache it and update the min cached if needed
+	isHigherInRange := newEpoch > limitEpochsForShard.min && newEpoch < limitEpochsForShard.min+epochsDelta
+	isHigherOutOfRange := newEpoch >= limitEpochsForShard.min+epochsDelta
+	isLowerInRange := newEpoch < limitEpochsForShard.max && newEpoch > limitEpochsForShard.max-epochsDelta && limitEpochsForShard.max >= epochsDelta
```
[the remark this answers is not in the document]
sstanculeanu: extracted into functions as suggested

andreibancioiu (approved):

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,243 @@
...
+	//		store the new + all cached ones that ate still in delta limit
+	//  2. the epoch is lower than the max cached epoch - epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
+	// 	3. the new epoch is higher than the min cached but still in the delta limits ->
+	//		cache it and update the max cached if needed
+	//	4. the new epoch is lower than the max cached but still in the delta limits ->
+	//		cache it and update the min cached if needed
+
+	// case 1.
+	if limitEpochsForShard.isHigherOutOfRange(newEpoch) {
```
ssd04, line 130: separate functions for these cases? at least for the first 2

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,243 @@
...
+				continue
+			}
+
+			if cachedEpoch < tmpMin {
+				tmpMin = cachedEpoch
+			}
+		}
+
+		// store the limits
+		// if nothing was saved from previous cached ones, they would be the new one
```
ssd04, line 145:
```suggestion
		// if nothing was saved from previous cached ones, they would be the new ones
```
?

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,243 @@
...
+
+		return
+	}
+
+	// reaching this point means that the shard is cached, but we have a new epoch
+	// there are different situations for the new epoch:
+	//	1. the epoch is higher than the min cached epoch + epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
+	//  2. the epoch is lower than the max cached epoch - epochs delta ->
+	//		store the new + all cached ones that ate still in delta limit
```
ssd04, line 123:
```suggestion
	//	1. the epoch is higher than the min cached epoch + epochs delta ->
	//		store the new + all cached ones that are still in delta limit
	//  2. the epoch is lower than the max cached epoch - epochs delta ->
	//		store the new + all cached ones that are still in delta limit
```
?

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,243 @@
...
+type eligibleNodesCache struct {
+	// eligibleNodesMap holds a map with key shard id and value
+	//		a map with key epoch and value
+	//		a map with key eligible node pk and value struct{}
+	// this is needed because:
+	// 	- a shard node will intercept proof for self shard and meta
+	// 	- a meta node will intercept proofs from all shards
+	eligibleNodesMap map[uint32]map[uint32]map[string]struct{}
+	// epochsLimitsForShardMap holds map with key shard id and value the limits for that shard
+	epochsLimitsForShardMap map[uint32]*epochLimits
```
ssd04, line 47: we can have a separate component for `eligibleNodesMap` per shard, coupled with `epochsLimitsForShardMap`, because we already do most of the operations per shard

ssd04: but it can be also considered later on with a TODO

sstanculeanu: added todo

process/interceptors/factory/eligibleNodesCache.go
sstanculeanu: extracted all cases into separate methods

andreibancioiu (commented):
Only Adi's comment (maybe separate PR?).

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -0,0 +1,275 @@
...
+func (limits *epochLimits) isHigherOutOfRange(newEpoch uint32) bool {
+	return newEpoch >= limits.min+epochsDelta
+}
+
+func (limits *epochLimits) isLowerInRange(newEpoch uint32) bool {
+	return newEpoch < limits.max && newEpoch > limits.max-epochsDelta && limits.max >= epochsDelta
+}
+
+func (limits *epochLimits) isLowerOutOfRange(newEpoch uint32) bool {
+	return newEpoch <= limits.max-epochsDelta && limits.max >= epochsDelta
```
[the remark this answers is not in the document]
sstanculeanu: pushed

process/interceptors/factory/eligibleNodesCache.go
```diff
@@ -255,17 +91,45 @@ func (cache *eligibleNodesCache) isPeerEligible(pid core.PeerID, shard uint32, e
...
+func (cache *eligibleNodesCache) refreshCache(epoch uint32, shardID uint32) (map[string]struct{}, error) {
 	eligibleList, err := cache.nodesCoordinator.GetAllEligibleValidatorsPublicKeysForShard(epoch, shardID)
 	if err != nil {
 		return nil, err
 	}
 
+	cachedEpochs := cache.nodesCoordinator.GetCachedEpochs()
+	epochsToBeDeleted := make([]uint32, 0)
+	for cachedEpoch := range cache.eligibleNodesMap {
+		_, isEpochCachedByNodesCoordinator := cachedEpochs[cachedEpoch]
```
AdoAdoAdo, line 103: I think it would be safe to directly delete while iterating

ssd04 (approved):

andreibancioiu (approved):

PR #6935, opened 2025-04-03
Equivalent proofs resolver and requester

sstanculeanu:
- equivalent proofs resolver and resolver
  
## Proposed changes
- equivalent proofs resolver and resolver

## Testing procedure
- with rc

## Pre-requisites

### DIFF
--- a/dataRetriever/requestHandlers/requesters/equivalentProofsRequester.go
+++ b/dataRetriever/requestHandlers/requesters/equivalentProofsRequester.go
@@ -1,0 +1,33 @@
+package requesters
+
+// ArgEquivalentProofsRequester is the argument structure used to create a new equivalent proofs requester instance
+type ArgEquivalentProofsRequester struct {
+	ArgBaseRequester
+}
+
+type equivalentProofsRequester struct {
+	*baseRequester
+}
+
+// NewEquivalentProofsRequester returns a new instance of equivalent proofs requester
+func NewEquivalentProofsRequester(args ArgEquivalentProofsRequester) (*equivalentProofsRequester, error) {
+	err := checkArgBase(args.ArgBaseRequester)
+	if err != nil {
+		return nil, err
+	}
+
+	return &equivalentProofsRequester{
+		baseRequester: createBaseRequester(args.ArgBaseRequester),
+	}, nil
+}
+
+// RequestDataFromHashArray requests equivalent proofs data from other peers by having multiple header hashes and the epoch as input
+// all headers must be from the same epoch
+func (requester *equivalentProofsRequester) RequestDataFromHashArray(hashes [][]byte, epoch uint32) error {
+	return requester.requestDataFromHashArray(hashes, epoch)
+}
+
+// IsInterfaceNil returns true if there is no value under the interface
+func (requester *equivalentProofsRequester) IsInterfaceNil() bool {
+	return requester == nil
+}
--- a/dataRetriever/resolvers/equivalentProofsResolver.go
+++ b/dataRetriever/resolvers/equivalentProofsResolver.go
@@ -1,0 +1,196 @@
+package resolvers
+
+import (
+	"fmt"
+
+	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data/batch"
+	"github.com/multiversx/mx-chain-go/dataRetriever"
+	"github.com/multiversx/mx-chain-go/p2p"
+	"github.com/multiversx/mx-chain-go/process/interceptors/processor"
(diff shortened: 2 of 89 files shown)


miiu96 (approved):

miiu96 (dismissed):

dataRetriever/requestHandlers/requesters/equivalentProofsRequester.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &equivalentProofsRequester{
+		baseRequester: createBaseRequester(args.ArgBaseRequester),
+	}, nil
+}
+
+// RequestDataFromHashArray requests equivalent proofs data from other peers by having multiple header hashes and the epoch as input
```
AdoAdoAdo, line 24: do we not have also requests for nonce?
I think when syncing we request headers in advance by nonce rather than by hash
So the same should be done for proofs

dataRetriever/resolvers/equivalentProofsResolver.go
```diff
@@ -0,0 +1,196 @@
...
+
+	rd, err := res.parseReceivedMessage(message, fromConnectedPeer)
+	if err != nil {
+		return nil, err
+	}
+
+	switch rd.Type {
+	case dataRetriever.HashType:
+		return nil, res.resolveHashRequest(rd.Value, rd.Epoch, message.Peer(), source)
+	case dataRetriever.HashArrayType:
```
AdoAdoAdo, line 98: we would require also by nonce type and nonce array

AdoAdoAdo (changes requested):

dataRetriever/requestHandlers/requesters/equivalentProofsRequester.go
sstanculeanu: implemented by nonce, will be integrated in the next PR

dataRetriever/resolvers/equivalentProofsResolver.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return fmt.Errorf("resolveMultipleHashesRequest.fetchEquivalentProofsSlicesForHeaders error %w", err)
+	}
+
+	return res.sendEquivalentProofsForHashes(equivalentProofsForHashes, pid, source)
+}
+
+// resolveNonceRequest sends the response for a nonce request
+func (res *equivalentProofsResolver) resolveNonceRequest(nonceShardKey []byte, epoch uint32, pid core.PeerID, source p2p.MessageHandler) error {
+	data, err := res.fetchEquivalentProofFromNonceAsByteSlice(nonceShardKey, epoch)
```
ssd04, line 160: i think we can keep all proofs into one map for all shards, so that we don't have to consider shardId here, it's not really needed (by shard) in proofs pool (we should have some benchmarks for this)

miiu96 (approved):

dataRetriever/unitType.go
```diff
@@ -51,6 +51,8 @@ const (
 	ScheduledSCRsUnit UnitType = 22
 	// ProofsUnit is the header proofs unit identifier
 	ProofsUnit UnitType = 23
+	// ProofsNonceHashDataUnit is the proofs nonce-hash pair data unit identifier
+	ProofsNonceHashDataUnit UnitType = 24
```
AdoAdoAdo, line 55: can you check if we can use the same header nonce to hash storer instead of a new one?

dataRetriever/resolvers/equivalentProofsResolver.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch rd.Type {
+	case dataRetriever.HashType:
+		return nil, res.resolveHashRequest(rd.Value, rd.Epoch, message.Peer(), source)
+	case dataRetriever.HashArrayType:
+		return nil, res.resolveMultipleHashesRequest(rd.Value, rd.Epoch, message.Peer(), source)
+	case dataRetriever.NonceType:
```
AdoAdoAdo, line 114: do we also have a nonce array type?

dataRetriever/unitType.go
sstanculeanu: updated to reuse them

dataRetriever/resolvers/equivalentProofsResolver.go
sstanculeanu: no, might not be needed for proofs either

miiu96 (approved):

AdoAdoAdo (approved):

ssd04 (approved):

miiu96 (approved):

PR #6985, opened 2025-05-12
remove hash from trie nodes and use hash from parent nodes

BeniaminDrasovean:
- Each trie node saves it's own hash. But intermediary nodes also save the hash for each child. This means that some hashes are duplicated in memory. 
  
## Proposed changes
- Remove the `hash` field from each trie node. When that hash is needed, it can be accessed from the parent node. 
- The trie insert/delete now happens in batches. Because of this we can compute the hash for each trie node when a batch is commited to the trie, thus removing the need to traverse the trie and compute the rootHash when it is needed.

## Testing procedure
- Normal testing procedure

## Pre-requisites

### DIFF
--- a/trie/branchNode_test.go
+++ b/trie/branchNode_test.go
@@ -63,6 +63,7 @@ func getBnAndCollapsedBn(marshalizer marshal.Marshalizer, hasher hashing.Hasher)
 	ChildrenHashes[13], _ = encodeNodeAndGetHash(children[13])
 	collapsedBn, _ := newBranchNode(marshalizer, hasher)
 	collapsedBn.ChildrenHashes = ChildrenHashes
+	bn.ChildrenHashes = ChildrenHashes
 
 	return bn, collapsedBn
 }
@@ -139,13 +140,6 @@ func getEncodedTrieNodesAndHashes(tr common.Trie) ([][]byte, [][]byte) {
 	return nodes, hashes
 }
 
-func TestBranchNode_getHash(t *testing.T) {
-	t.Parallel()
-
-	bn := &branchNode{baseNode: &baseNode{hash: []byte("test hash")}}
-	assert.Equal(t, bn.hash, bn.getHash())
-}
-
 func TestBranchNode_isDirty(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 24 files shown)


trie/interface.go
```diff
@@ -18,9 +18,12 @@ type nodeData struct {
 	hexKey      []byte
 }
 
+type nodeWithHash struct {
```
ssd04, line 21: maybe add a comment describing the usage of this struct

ssd04: same for `nodeData`

trie/errors.go
```diff
@@ -144,3 +144,6 @@ var ErrInvalidTypeConversion = errors.New("invalid type conversion")
 
 // ErrNodeHashIsNotSet signals that the node hash is not set
 var ErrNodeHashIsNotSet = errors.New("node hash is not set")
+
+// ErrInvalidNodeState indicates an erroneous condition where a node's child pointer is set but the child hash is empty.
+var ErrInvalidNodeState = errors.New("invalid node state, child pointer is set but the child hash is empty ")
```
ssd04, line 149:
```suggestion
var ErrInvalidNodeState = errors.New("invalid node state, child pointer is set but the child hash is empty")
```

trie/branchNode_test.go
```diff
@@ -247,30 +206,56 @@ func TestBranchNode_getEncodedNodeNil(t *testing.T) {
...
+		bn, _ := getBnAndCollapsedBn(getTestMarshalizerAndHasher())
+		bn.ChildrenHashes = make([][]byte, nrOfChildren)
+
+		childNode, childHash, err := bn.resolveIfCollapsed(2, nil)
+		assert.Equal(t, ErrInvalidNodeState, err)
+		assert.Nil(t, childNode)
+		assert.Nil(t, childHash)
+	})
+	t.Run("node is not collapsed", func(t *testing.T) {
+		bn, _ := getBnAndCollapsedBn(getTestMarshalizerAndHasher())
```
ssd04, line 249: add `t.Parallel()`?

sasurobert (approved):

trie/interface.go
BeniaminDrasovean: Added.

trie/errors.go
BeniaminDrasovean: Done.

trie/branchNode_test.go
BeniaminDrasovean: Added.

ssd04 (approved):

sasurobert (approved):

PR #6999, opened 2025-05-21
Supply corrrection db lookup extension

miiu96:
- Added a mechanism in dblookup extension module to alter the information related to the supply of an ESDT token
  
## Proposed changes
- Extended the dblookup extension config with an array of supply corrections
- Implemented a mechanism to update the information about the supply of an ESDT token based on the provided config.
## Testing procedure
- System test and check on mainnet on tokens with negative supply

## Pre-requisites


dblookupext/esdtSupply/corrections.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		correctionValue, ok := big.NewInt(0).SetString(supplyCorrection.Value, 10)
+		if !ok {
+			return errors.New("failed to parse supplyCorrection value")
+		}
+
+		switch supplyCorrection.Operation {
+		case addOperation:
+			tokenSupply.Supply.Add(tokenSupply.Supply, correctionValue)
```
AdoAdoAdo, line 59: seems like no need for the type of operation add, or subtract, if we allow both negative values and positive values.
then depending on the sign of the value we need to apply either mint or burn

Also we could probably use a check not to be left with negative supply (on burn) and log a warning if so.

raduchis: yes, cleaner without an operation... we cand have only add or subtract, right? that means positive or negative value in the correction

dblookupext/esdtSupply/esdtSuppliesProcessor.go
```diff
@@ -39,11 +42,19 @@ func NewSuppliesProcessor(
...
+	supplyCorrectionProc := newSupplyCorrectionProcessor(shardID, logsProc)
 
-	return &suppliesProcessor{
+	sp := &suppliesProcessor{
 		logsProc: logsProc,
 		logsGet:  logsGet,
-	}, nil
+	}
+
+	err := supplyCorrectionProc.applySupplyCorrections(supplyCorrections)
```
raduchis, line 52: This works only if a restart happens after the processing of th BlockNonce set in the  SupplyCorrections. 
If this new version is started before the BlockNonce, another restart is required, right?

dblookupext/esdtSupply/corrections.go
miiu96: removed the `operation` field. 
and also these settings from `config.toml`

dblookupext/esdtSupply/corrections.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (scp *supplyCorrectionProcessor) processSupplyCorrection(supplyCorrection config.SupplyCorrection, supplies map[string]*SupplyESDT) error {
+	tokenSupply, err := scp.logsProc.getESDTSupply([]byte(supplyCorrection.Token))
+	if err != nil {
+		return err
+	}
+
+	// if token supply was recomputed should not apply supply correction
+	if tokenSupply.RecomputedSupply {
+		return nil
```
raduchis, line 92: could you start before the supplyCorrectionBlock with the RecomputedSupply = true, the "bug" manifests after and when it is time to do the supplyCorrection, you think it was recompted?

raduchis (approved):

AdoAdoAdo (approved):

PR #6993, opened 2025-05-15
move constants to config file

danielradu10:
## Reasoning behind the pull request
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

Based on the [Contributing Guidelines](https://github.com/multiversx/mx-chain-go/blob/master/.github/CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:
- was the PR targeted to the correct branch?
- if this is a larger feature that probably needs more than one PR, is there a `feat` branch created?
- if this is a `feat` branch merging, do all satellite projects have a proper tag inside `go.mod`?

PR #7004, opened 2025-05-22
moved upper bounds to config

danielradu10:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -423,6 +423,7 @@
     SelectionGasBandwidthIncreasePercent = 400
     SelectionGasBandwidthIncreaseScheduledPercent = 260
     TxCacheSelectionMaxNumTxs = 30000
+    MaxNumBytesPerSenderUpperBound = 33_554_432
 
 [TrieNodesChunksDataPool]
     Name = "TrieNodesDataPool"
--- a/dataRetriever/constants.go
+++ b/dataRetriever/constants.go
@@ -8,3 +8,5 @@ const RewardTxPoolName = "rewardTxPool"
 
 // ValidatorsInfoPoolName defines the name of the validators info pool
 const ValidatorsInfoPoolName = "validatorsInfoPool"
+
+const MaxNumBytesPerSenderUpperBoundTest = 33_554_432 // 32 MB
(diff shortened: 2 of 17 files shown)


cmd/node/config/config.toml
```diff
@@ -423,6 +423,7 @@
     SelectionGasBandwidthIncreasePercent = 400
     SelectionGasBandwidthIncreaseScheduledPercent = 260
     TxCacheSelectionMaxNumTxs = 30000
+    MaxNumBytesPerSenderUpperBound = 33_554_432
```
AdoAdoAdo, line 426: do you have a test for loading this?
I am not sure the result of loading this is as expected.

dataRetriever/constants.go
```diff
@@ -8,3 +8,5 @@ const RewardTxPoolName = "rewardTxPool"
 
 // ValidatorsInfoPoolName defines the name of the validators info pool
 const ValidatorsInfoPoolName = "validatorsInfoPool"
+
+const MaxNumBytesPerSenderUpperBoundTest = 33_554_432 // 32 MB
```
AdoAdoAdo, line 12: We do not need a constant if we have already a config value for it.

If this is used only in tests please move it to the test file

txcache/constants.go
```diff
@@ -3,3 +3,6 @@ package txcache
 const diagnosisMaxTransactionsToDisplay = 10000
 const initialCapacityOfSelectionSlice = 30000
 const selectionLoopDurationCheckInterval = 10
+
+// maxNumBytesPerSenderUpperBoundTest is used for setting the MaxNumBytesPerSenderUpperBoundTest from ConfigSourceMe in tests
+const maxNumBytesPerSenderUpperBoundTest = 33_554_432 // 32 MB
```
AdoAdoAdo, line 8: we do not need a constant here if we already have a config value for it.

cmd/node/config/config.toml
danielradu10: fixed

dataRetriever/constants.go
danielradu10: moved

andreibancioiu (approved):

AdoAdoAdo (commented):
no new findings but needs to be updated after https://github.com/multiversx/mx-chain-go/pull/6993 is merged.

PR #7007, opened 2025-05-26
Fix backup number of inactive rounds computation for multikey nodes

sstanculeanu:
- With the full consensus, there might be an edge case where backups will step in without being the need when managing multiple bls keys
  
## Proposed changes
- updated both config and increment of number of rounds without received message

## Testing procedure
- standard system test + backups + chaos setup

## Pre-requisites

### DIFF
--- a/consensus/spos/bls/v2/subroundBlock.go
+++ b/consensus/spos/bls/v2/subroundBlock.go
@@ -519,12 +519,23 @@ func (sr *subroundBlock) receivedBlockHeader(headerHandler data.HeaderHandler) {
 	ctx, cancel := context.WithTimeout(context.Background(), sr.RoundHandler().TimeDuration())
 	defer cancel()
 
-	_ = sr.processReceivedBlock(ctx, int64(headerHandler.GetRound()), []byte(sr.Leader()))
+	blockProcessedWithSuccess := sr.processReceivedBlock(ctx, int64(headerHandler.GetRound()), []byte(sr.Leader()))
 	sr.PeerHonestyHandler().ChangeScore(
 		sr.Leader(),
 		spos.GetConsensusTopicID(sr.ShardCoordinator()),
 		spos.LeaderPeerHonestyIncreaseFactor,
 	)
+
+	pubKeys := sr.ConsensusGroup()
+	if blockProcessedWithSuccess {
+		// increment the number of rounds without received message
+		// at this point, it is certain that the leader is not managed by self and
+		// the block was processed, so signatures are expected from the current managed nodes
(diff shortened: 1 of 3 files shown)


AdoAdoAdo (approved):

ssd04 (approved):

danidrasovean (approved):
Normal allin test: fix-backup-edge-case-ff41222f60 -> fix-backup-edge-case-ff41222f60

--- Specific errors --- 

block hash does not match 0
wrong nonce in block 0
miniblocks does not match 0
block bodies does not match 0
receipts hash missmatch 0

/------/ 


--- Statistics --- 

Nr. of all ERRORS:  0
Nr. of all WARNS:  100
Nr. of new ERRORS:  0
Nr. of new WARNS:  25
Nr. of PANICS:  0
Nodes with panic:  []

/------/ 


--- System test go --- 

/------/

consensus/spos/bls/v2/subroundStartRound.go
```diff
@@ -224,18 +224,22 @@ func (sr *subroundStartRound) computeNumManagedKeysInConsensusGroup(pubKeys []st
 	numMultiKeysInConsensusGroup := 0
 	for _, pk := range pubKeys {
 		pkBytes := []byte(pk)
+
+		if sr.IsMultiKeyLeaderInCurrentRound() {
```
AdoAdoAdo, line 228: this check here is introducing a sort of circular dependency
it cannot increment if it is not already incremented, due to the inner check of "backup machine needs to start acting as validator"

consensus/spos/bls/v2/subroundBlock.go
```diff
@@ -519,12 +519,23 @@ func (sr *subroundBlock) receivedBlockHeader(headerHandler data.HeaderHandler) {
...
 	)
+
+	pubKeys := sr.ConsensusGroup()
+	if blockProcessedWithSuccess {
+		// increment the number of rounds without received message
+		// at this point, it is certain that the leader is not managed by self and
+		// the block was processed, so signatures are expected from the current managed nodes
+		for _, pk := range pubKeys {
+			pkBytes := []byte(pk)
+			sr.IncrementRoundsWithoutReceivedMessages(pkBytes)
```
AdoAdoAdo, line 536: I think this is not working as expected because from the perspective of a backup machine, it does not exit at line 469, and then it could increment the rounds even if it has already increased on subround start round, so ending up with a double increase.

AdoAdoAdo (changes requested):

sstanculeanu:
Closing this one, proper fix should be done on #7010

PR #7023, opened 2025-06-02
use keyMutex on trie nodes

BeniaminDrasovean:
- This PR reduces the amount of memory needed to keep a trie in RAM.  
  
## Proposed changes
- Each trie node had one or more mutexes. In order to reduce the size in memory of a trie, remove the mutex pointers from trie nodes and use the `keyMutex` component. Thus mutexes will be allocated only when necessary and only for the trie nodes that are directly affected by a certain operation.

## Testing procedure
- Normal testing & import-db

## Pre-requisites


trie/branchNode.go
```diff
@@ -185,9 +189,6 @@ func (bn *branchNode) resolveIfCollapsed(pos byte, trieCtx common.TrieContext) (
 		return nil, nil, ErrChildPosOutOfRange
 	}
 
-	bn.childrenMutexes[pos].Lock()
-	defer bn.childrenMutexes[pos].Unlock()
-
```
ssd04, line 190: this was removed from here since it is under mutex from the caller, right?
- it seems that in `print` method it is not called under mutex
- maybe add a comment on this function to indicate that it has to be called under mutex

trie/branchNode.go
```diff
@@ -605,24 +622,26 @@ func (bn *branchNode) revertChildrenVersionSliceIfNeeded() {
...
 }
 
 func getChildPosition(n *branchNode) (numChildren int, childPos int) {
 	for i := 0; i < nrOfChildren; i++ {
-		n.childrenMutexes[i].RLock()
 		if n.children[i] != nil || len(n.ChildrenHashes[i]) != 0 {
 			numChildren++
 			childPos = i
 		}
-		n.childrenMutexes[i].RUnlock()
```
ssd04, line 625: mutex was removed here, is this function always called under mutex?

trie/extensionNode.go
```diff
@@ -542,8 +564,6 @@ func (en *extensionNode) isEmptyOrNil() error {
 		return ErrNilExtensionNode
 	}
 
-	en.childMutex.RLock()
-	defer en.childMutex.RUnlock()
 	if en.child == nil && len(en.ChildHash) == 0 {
```
ssd04, line 562: why removed the mutex here? what fields have to be covered by mutex in `extensionNode`?

trie/leafNode.go
```diff
@@ -213,20 +225,22 @@ func (ln *leafNode) insertInNewBn(
...
 	_ common.TrieGoroutinesManager,
 	_ common.AtomicBytesSlice,
-	_ common.TrieContext,
+	trieCtx common.TrieContext,
 ) (bool, node) {
-	ln.mutex.RLock()
-	defer ln.mutex.RUnlock()
+	mutexKey := getMutexKeyFromPath(pathKey)
+	trieCtx.Lock(mutexKey)
+	defer trieCtx.Unlock(mutexKey)
```
ssd04: change to `RLock`?

trie/interface.go
```diff
@@ -24,20 +24,21 @@ type nodeWithHash struct {
 	hash []byte
 }
 
-type baseTrieNode interface {
-	setDirty(bool)
-	isDirty() bool
+type keyData struct {
+	keyRemainder []byte
+	pathKey      []byte
```
ssd04: add comments here on these fields? what `keyRemainder` means and how it will be used?

trie/branchNode_test.go
```diff
@@ -1770,11 +1768,238 @@ func TestBranchNode_deleteBatch(t *testing.T) {
...
+		dirty, newNode := bn.delete(keyBuilder.NewKeyBuilder(), data, goRoutinesManager, modifiedHashes, trieCtx)
 		assert.Nil(t, goRoutinesManager.GetError())
 		assert.True(t, dirty)
 		assert.Nil(t, newNode)
 		expectedNumTrieNodesChanged := 5
 		assert.Equal(t, expectedNumTrieNodesChanged, len(modifiedHashes.Get()))
 	})
 }
+
+func TestBranchNode_concurrency(t *testing.T) {
```
ssd04, line 1780: can we have tests with multiple concurrent calls?

ssd04 (commented):
can we have some benchmarks/tests before and after these changes to see the difference in size

trie/branchNode.go
BeniaminDrasovean: Yes, the caller should handle the mutex protection. 
The `print` method is used only for testing purposes, and that is why I did not use mutex protection. To solve the concurrency problem, I added a new trie pointer in the `ToString` trie method. This is the only place where `print` is called. Also refactored the `print` to always get trie nodes from db, because this is almost always the use case.
Added a comment to explain the exception to using mutex protection for this func call.

trie/branchNode.go
BeniaminDrasovean: Yes, the only place where this is called from is `reduceNodeIfNeeded`, and that method has mutex protection.

trie/extensionNode.go
BeniaminDrasovean: `en.child` & `en.ChildHash` should be protected by mutex. However, from all places where this is called, a new pointer for `en` is created, so no concurrent calls will take place. Removed `isEmptyOrNil` from trie node interface to avoid misuse.

trie/leafNode.go
BeniaminDrasovean: Done

trie/interface.go
BeniaminDrasovean: Added

trie/branchNode_test.go
BeniaminDrasovean: There are tests with multiple concurrent calls in `patriciaMerkleTrie_test.go`: 

- TestPatriciaMerkleTrie_UpdateAndGetConcurrently
- TestPatriciaMerkleTrie_ConcurrentOperations
- TestPatriciaMerkleTrie_AddBatchedDataToTrie

trie/sync_test.go
```diff
@@ -3,6 +3,7 @@ package trie
 import (
 	"context"
 	"errors"
+	"github.com/multiversx/mx-chain-go/trie/keyBuilder"
```
sasurobert: goimports

trie/patriciaMerkleTrie_test.go
```diff
@@ -997,84 +997,141 @@ func TestPatriciaMerkleTree_GetValueReturnsTrieDepth(t *testing.T) {
...
-					initialRootHash,
-					keyBuilder.NewKeyBuilder(),
-					parsers.NewMainTrieLeafParser(),
-				)
-				assert.Nil(t, err)
-			case 9:
-				_, _, _ = tr.GetProof(initialRootHash, initialRootHash) // this might error due to concurrent operations that change the roothash
-			case 10:
-				// extremely hard to compute an existing hash due to concurrent changes.
-				_, _ = tr.VerifyProof([]byte("dog"), []byte("puppy"), [][]byte{[]byte("proof1")}) // this might error due to concurrent operations that change the roothash
```
ssd04, line 1061: are these comments still valid now with the latest changes?

BeniaminDrasovean: No, removed the comments and added some asserts

ssd04 (approved):

sasurobert (approved):

PR #7035, opened 2025-06-10
skip meta data entries when iterating esdts on API calls

BeniaminDrasovean:
- When a creator updates the metaData of an Dynamic NFT which is not in its balance, that entry will be saved in that account's data trie. When an API call iterates over the esdts from that account, it will also try to print that metaData.
  
## Proposed changes
- Skip data trie entries for esdts if the esdt is not in the balance of the account

## Testing procedure
- Normal testing procedure
- Create an dynamic NFT, transfer it to another account, call `ESDTSetNewURIs` from the creator account, and then call `/address/:address:/esdt` for the creator address
- Call `/:address/nft/:tokenIdentifier/nonce/:nonce` with a nonce that has > 20 characters, should return error

## Pre-requisites

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,1 @@
+sonar.exclusions=**/*_test.go, integrationTests/**, testcommon/**, **/mock/**, **/disabled/**, **/disabled*.go,  examples/**, scripts/**, **/*.pb.go


node/node.go
```diff
@@ -679,6 +679,10 @@ func (n *Node) GetAllESDTTokens(address string, options api.AccountQueryOptions,
 			continue
 		}
 
+		if esdtToken.Value.Cmp(big.NewInt(0)) <= 0 {
```
andreibancioiu: :+1: 

However, maybe have a single (reusable) instance for the big int zero (just an idea, maybe it does not worth it)?

andreibancioiu: How about using `Sign()`?

https://pkg.go.dev/math/big#Int.Sign

BeniaminDrasovean: Good idea, changed.

api/groups/addressGroup.go
```diff
@@ -634,6 +637,9 @@ func extractGetESDTNFTDataParams(c *gin.Context) (string, string, *big.Int, api.
 	if nonceAsStr == "" {
 		return "", "", nil, api.AccountQueryOptions{}, errors.ErrNonceInvalid
 	}
+	if len(nonceAsStr) > maxNumCharsForNonceAsString {
+		return "", "", nil, api.AccountQueryOptions{}, fmt.Errorf("%w: nonce too long, num chars %v, max num chars %v", errors.ErrNonceInvalid, len(nonceAsStr), maxNumCharsForNonceAsString)
```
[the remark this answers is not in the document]
BeniaminDrasovean: Not true

PR #7097, opened 2025-07-08
Added round duration for supernova + refactor nft syncer args

ssd04:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/factory/core/coreComponents.go
+++ b/factory/core/coreComponents.go
@@ -229,7 +229,19 @@ func (ccf *coreComponentsFactory) Create() (*coreComponents, error) {
 		return nil, err
 	}
 
-	syncer := ntp.NewSyncTime(ccf.config.NTPConfig, nil, genesisRoundDuration)
+	syncTimeArgs := ntp.SyncTimeArgs{
+		NtpConfig:              ccf.config.NTPConfig,
+		CustomQueryFunc:        nil,
+		RoundDuration:          genesisRoundDuration,
+		SupernovaRoundDuration: supernovaRoundDuration,
+		EnableEpochsHandler:    enableEpochsHandler,
+		EnableRoundsHandler:    enableRoundsHandler,
+	}
+	syncer, err := ntp.NewSyncTime(syncTimeArgs)
+	if err != nil {
+		return nil, err
+	}
+
 	syncer.StartSyncingTime()
 	log.Debug("NTP average clock offset", "value", syncer.ClockOffset())
 
--- a/ntp/syncTime.go
+++ b/ntp/syncTime.go
@@ -12,10 +12,13 @@ import (
 
 	"github.com/beevik/ntp"
 	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/core/check"
 	"github.com/multiversx/mx-chain-core-go/core/closing"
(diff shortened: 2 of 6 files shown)


factory/core/coreComponents.go
```diff
@@ -229,7 +229,19 @@ func (ccf *coreComponentsFactory) Create() (*coreComponents, error) {
 		return nil, err
 	}
 
-	syncer := ntp.NewSyncTime(ccf.config.NTPConfig, nil, genesisRoundDuration)
+	syncTimeArgs := ntp.SyncTimeArgs{
```
AdoAdoAdo: I think we don't need to over engineer this component.
I would simply give it the supernova round duration, and no need for transitions.
The supernova configuration is more restrictive, so it would apply also for previous period.

Also no need to give it the enable epoch or enable round for the same reasons

ssd04: the only change is to get the new round duration, the other code changes are due to adding SyncTime args struct

ssd04: i would keep the transition to have the flow dynamic, in care, for example, we change to a higher round duration instead of a lower one

ssd04: changed to get out of bounds threshold from config

ntp/syncTime.go
```diff
@@ -61,11 +57,12 @@ type NTPOptions struct {
...
-		Port:                123,
-		Version:             0,
-		TimeoutMilliseconds: 100,
-		SyncPeriodSeconds:   3600,
+		Hosts:                []string{"time.google.com", "time.cloudflare.com", "time.apple.com", "time.windows.com"},
+		Port:                 123,
+		Version:              0,
+		TimeoutMilliseconds:  100,
+		SyncPeriodSeconds:    3600,
+		OutOfBoundsThreshold: 120,
```
AdoAdoAdo: can you move this instantiation into the test file using it if it is not used by production code.

ntp/syncTime.go
```diff
@@ -219,14 +219,11 @@ func (s *syncTime) sync() {
...
-	thresholdOutOfBonds := s.roundDuration * time.Duration(outOfBoundsRoundDurationPercentage) / time.Duration(100)
-	isOutOfBounds := core.AbsDuration(clockOffsetHarmonicMean) > thresholdOutOfBonds
+	isOutOfBounds := core.AbsDuration(clockOffsetHarmonicMean) > time.Duration(s.outOfBoundsThreshold)
 	if isOutOfBounds {
 		log.Error("syncTime.sync: clock offset is out of expected bounds",
 			"clock offset harmonic mean", clockOffsetHarmonicMean,
-			"thresholdOutOfBonds", thresholdOutOfBonds,
-			"outOfBoundsRoundDurationPercentage", outOfBoundsRoundDurationPercentage,
-			"roundDuration", s.roundDuration,
+			"thresholdOutOfBonds", s.outOfBoundsThreshold,
```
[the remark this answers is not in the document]
ssd04: changed

ntp/syncTime.go
ssd04: moved to testscommon

sstanculeanu (approved):

PR #7104, opened 2025-07-10
Fix round handler transition

ssd04:
- handle supernova activation in round handler based on timestamp if node not fully initiated
- check supernova activation by epoch in fork detector
- fetch round duration from chain parameters instead of round handler

## Testing procedure
- TBD

## Pre-requisites

### DIFF
--- a/consensus/round/round_test.go
+++ b/consensus/round/round_test.go
@@ -102,7 +102,7 @@ func TestRound_UpdateRoundShouldAdvanceOneRound(t *testing.T) {
 
 		args := createDefaultRoundArgs()
 		args.EnableRoundsHandler = &testscommon.EnableRoundsHandlerStub{
-			IsFlagEnabledCalled: func(flag common.EnableRoundFlag) bool {
+			IsFlagEnabledInRoundCalled: func(flag common.EnableRoundFlag, round uint64) bool {
 				return flag != common.SupernovaRoundFlag
 			},
 		}
@@ -123,28 +123,6 @@ func TestRound_UpdateRoundShouldAdvanceOneRound(t *testing.T) {
 
 		genesisTime := time.Now()
 
-		args := createDefaultRoundArgs()
-		args.EnableRoundsHandler = &testscommon.EnableRoundsHandlerStub{
-			IsFlagEnabledCalled: func(flag common.EnableRoundFlag) bool {
-				return flag == common.SupernovaRoundFlag
-			},
-		}
-
-		args.SupernovaGenesisTimeStamp = genesisTime
-
-		rnd, _ := round.NewRound(args)
-		oldIndex := rnd.Index()
-		rnd.UpdateRound(genesisTime, genesisTime.Add(roundTimeDuration))
-		newIndex := rnd.Index()
-
-		assert.Equal(t, oldIndex, newIndex-1)
-	})
-
-	t.Run("after supernova, without flag activated, but after supernova genesis time", func(t *testing.T) {
-		t.Parallel()
-
-		genesisTime := time.Now()
-
 		roundDuration := 10 * time.Millisecond
 
 		supernovaRoundDuration := 5 * time.Millisecond
@@ -154,7 +132,11 @@ func TestRound_UpdateRoundShouldAdvanceOneRound(t *testing.T) {
 		args := createDefaultRoundArgs()
 		args.RoundTimeDuration = roundDuration
 		args.SupernovaTimeDuration = supernovaRoundDuration
-		args.EnableRoundsHandler = &testscommon.EnableRoundsHandlerStub{}
+		args.EnableRoundsHandler = &testscommon.EnableRoundsHandlerStub{
+			IsFlagEnabledInRoundCalled: func(flag common.EnableRoundFlag, round uint64) bool {
+				return flag == common.SupernovaRoundFlag && round >= uint64(supernovaStartRond)
+			},
+		}
 
 		args.SupernovaStartRound = supernovaStartRond
 		args.GenesisTimeStamp = genesisTime
@@ -170,6 +152,7 @@ func TestRound_UpdateRoundShouldAdvanceOneRound(t *testing.T) {
 
 		index0 := rnd.Index()
 		timestamp0 := rnd.TimeStamp()
+		assert.Equal(t, roundDuration, rnd.TimeDuration())
 
 		rnd.UpdateRound(genesisTime, supernovaGenesisTime)
 
@@ -178,6 +161,7 @@ func TestRound_UpdateRoundShouldAdvanceOneRound(t *testing.T) {
 
 		diffTime := timestamp1.Sub(timestamp0)
 
+		assert.Equal(t, supernovaRoundDuration, rnd.TimeDuration())
 		assert.Equal(t, index0, index1-1)
 		assert.Equal(t, int64(5), index1)
 		assert.Equal(t, roundDuration, diffTime)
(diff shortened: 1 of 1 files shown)


process/sync/baseForkDetector.go
```diff
@@ -716,6 +711,12 @@ func (bfd *baseForkDetector) checkGenesisTimeForHeaderBeforeSupernova(
 	genesisTime := int64(headerHandler.GetTimeStamp()) - roundDifference*roundDuration
 
 	if genesisTime != bfd.genesisTime {
+		log.Error("checkGenesisTimeForHeaderBeforeSupernova: genesis time missmatch",
```
sstanculeanu:
```suggestion
		log.Error("checkGenesisTimeForHeaderBeforeSupernova: genesis time mismatch",
```

process/sync/baseForkDetector.go
```diff
@@ -746,6 +747,11 @@ func (bfd *baseForkDetector) checkGenesisTimeForHeaderAfterSupernovaWithoutRound
 	}
 
 	if genesisTime != bfd.genesisTime {
+		log.Error("checkGenesisTimeForHeaderAfterSupernovaWithoutRoundActivation: genesis time missmatch",
```
sstanculeanu:
```suggestion
		log.Error("checkGenesisTimeForHeaderAfterSupernovaWithoutRoundActivation: genesis time mismatch",
```

process/sync/baseForkDetector.go
```diff
@@ -781,6 +790,11 @@ func (bfd *baseForkDetector) checkGenesisTimeForHeaderAfterSupernovaWithRoundAct
 	)
 
 	if genesisTime != bfd.supernovaGenesisTime {
+		log.Error("checkGenesisTimeForHeaderAfterSupernovaWithRoundActivation: genesis time missmatch",
```
sstanculeanu:
```suggestion
		log.Error("checkGenesisTimeForHeaderAfterSupernovaWithRoundActivation: genesis time mismatch",
```

consensus/round/round.go
```diff
@@ -84,27 +84,54 @@ func NewRound(args ArgsRound) (*round, error) {
...
+		baseTimeStamp = genesisTimeStamp
+		roundDuration = rnd.timeDuration
+		startRound = rnd.startRound
+	}
+
+	rnd.updateRound(baseTimeStamp, currentTimeStamp, startRound, roundDuration)
+}
+
+func (rnd *round) isSupernovaActivated(currentTimeStamp time.Time) bool {
+	supernovaActivated := common.IsSupernovaRoundActivated(rnd.enableEpochsHandler, rnd.enableRoundsHandler)
```
sstanculeanu: perhaps add an initial early exit after this line if !supernovaActivated

ssd04: added an early check if supernovaActivated; if not activated we have to check based on current time if node just started and current epoch and round is not set

sstanculeanu (dismissed):

consensus/round/round.go
```diff
@@ -84,27 +84,57 @@ func NewRound(args ArgsRound) (*round, error) {
...
 // UpdateRound updates the index and the time stamp of the round depending on the genesis time and the current time given
 func (rnd *round) UpdateRound(genesisTimeStamp time.Time, currentTimeStamp time.Time) {
-	if !common.IsSupernovaRoundActivated(rnd.enableEpochsHandler, rnd.enableRoundsHandler) {
-		rnd.updateRound(genesisTimeStamp, currentTimeStamp, rnd.startRound)
-		return
+	baseTimeStamp := rnd.supernovaGenesisTimeStamp
+	roundDuration := rnd.supernovaTimeDuration
+	startRound := rnd.supernovaStartRound
+
+	supernovaActivated := rnd.isSupernovaActivated(currentTimeStamp)
```
AdoAdoAdo, line 85: possibly not ok if node is in sync with header that has a round < supernovaRoundActivation, but the actual (chronology) round is after the supernova round.

common/interface.go
```diff
@@ -313,6 +313,7 @@ type EnableEpochsHandler interface {
 // EnableRoundsHandler defines the operations of an entity that manages round activation flags
 type EnableRoundsHandler interface {
 	RoundConfirmed(round uint64, timestamp uint64)
+	GetCurrentRound() uint64
```
AdoAdoAdo, line 316: why do we need to get a current round from the EnableRoundsHandler?

consensus/round/round.go
```diff
@@ -84,27 +84,57 @@ func NewRound(args ArgsRound) (*round, error) {
...
+		log.Debug("isSupernovaActivated: force set supernovaActivated",
+			"currentTimeAfterSupernova", currentTimeAfterSupernova,
+			"enableEpochsHandler current epoch", rnd.enableEpochsHandler.GetCurrentEpoch(),
+			"enableRoundsHandler current round", rnd.enableRoundsHandler.GetCurrentRound(),
+		)
+		supernovaActivated = true
 	}
 
-	rnd.updateRound(rnd.supernovaGenesisTimeStamp, currentTimeStamp, rnd.supernovaStartRound)
+	return supernovaActivated
```
AdoAdoAdo, line 115: in the case of node syncing headers from before supernova round activation, it will return false, but this will then be used to initialize the new rounds (after supernova round activation) with apparently previous (wrong) round config/duration.

consensus/round/round.go
```diff
@@ -143,6 +173,7 @@ func (rnd *round) TimeDuration() time.Duration {
 
 // this should be called under mutex protection
 func (rnd *round) getTimeDuration() time.Duration {
+	// TODO: analysize adding here also forced activation based on current timestamp
 	if common.IsSupernovaRoundActivated(rnd.enableEpochsHandler, rnd.enableRoundsHandler) {
```
AdoAdoAdo: not sure, but probably not ok due to the same reason: the round is assumed from the enableRoundsHandler internal round, but we don't know the call intention for which round it was meant, if it is for the current round duration or for the round duration of an older block.

integrationTests/testFullNode.go
```diff
@@ -632,7 +634,9 @@ func (tfn *TestFullNode) createForkDetector(
 			tfn.GenesisTimeField.UnixMilli(),
 			tfn.EnableEpochsHandler,
 			tfn.EnableRoundsHandler,
-			tfn.DataPool.Proofs())
+			tfn.DataPool.Proofs(),
+			&chainParameters.ChainParametersHandlerStub{},
```
AdoAdoAdo: don't we have a tfn.ChainsParametersHandler() ? we should add one, otherwise we will end up with multiple versions in different subcomponents.

common/interface.go
ssd04: it's needed when we check current round for enable rounds handler at start in consensus round handler

integrationTests/testFullNode.go
ssd04: changed to use a common component in tests

consensus/round/round.go
ssd04: changed to check activation per round

consensus/round/round.go
ssd04: changed to use activation per current consensus round

consensus/round/round.go
ssd04: changed to use activation per current consensus round, in round handler we return round duration from chronology perspective

consensus/round/round.go
```diff
@@ -84,27 +78,57 @@ func NewRound(args ArgsRound) (*round, error) {
...
+	return rnd.enableRoundsHandler.IsFlagEnabledInRound(common.SupernovaRoundFlag, uint64(rnd.index))
+}
+
+func (rnd *round) isSupernovaActivated(currentTimeStamp time.Time) bool {
+	supernovaActivated := rnd.isSupernovaRoundActivated()
+	if supernovaActivated {
+		return supernovaActivated
+	}
+
+	currentTimeAfterSupernova := currentTimeStamp.UnixMilli() > rnd.supernovaGenesisTimeStamp.UnixMilli()
```
AdoAdoAdo: can you check if this should be instead `>=` ?
maybe create a unit test for that

sstanculeanu (approved):

consensus/round/round.go
ssd04: for `currentTimeStamp` as supernova genesis timestamp, it will get to the same results on both flows:
- if `>` it will calculate from genesis time to supernova activation timestamp, which should get to supernova activation round index
- if `>=` it will get supernova timestamp and round as required

kept it with `>=`

sstanculeanu (approved):

PR #7080, opened 2025-06-25
added compiling method

danielradu10:
## Reasoning behind the pull request
 modified selection tracker for adding supernova support in tx pool
- use block.Body for OnProposedBlock to get the tx hashes
- added method for collecting the txHashes from miniblocks
- added method for compiling the breadcrumbs
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

Based on the [Contributing Guidelines](https://github.com/multiversx/mx-chain-go/blob/master/.github/CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:
- was the PR targeted to the correct branch?
- if this is a larger feature that probably needs more than one PR, is there a `feat` branch created?
- if this is a `feat` branch merging, do all satellite projects have a proper tag inside `go.mod`?

PR #7133, opened 2025-07-21
optimize compile breadcrumb flow

danielradu10:
- solved a **TODO** related to [this comment](https://github.com/multiversx/mx-chain-go/pull/7080#discussion_r2174380195)
  
## Proposed changes
- avoid double search in map by **checking if the sender is also the fee payer** and accumulate in its breadcrumb
- added a unit test for this case

## Pre-requisites

### DIFF
--- a/txcache/trackedBlock.go
+++ b/txcache/trackedBlock.go
@@ -65,17 +65,19 @@ func (tb *trackedBlock) compileBreadcrumb(tx *WrappedTransaction) {
 	})
 
 	// compile for fee payer
-	if feePayer != nil && bytes.Equal(sender, feePayer) {
+	if feePayer == nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


sstanculeanu (approved):

txcache/trackedBlock.go
```diff
@@ -65,6 +65,12 @@ func (tb *trackedBlock) compileBreadcrumb(tx *WrappedTransaction) {
 	})
 
 	// compile for fee payer
+	if feePayer != nil && bytes.Equal(sender, feePayer) {
```
AdoAdoAdo: you can extract the conditional feePayer == nil and return on that

this will simplify the conditions below

danielradu10: fixed

PR #7137, opened 2025-07-22
Gas consumption

sstanculeanu:
- a new component is needed to handle the max gas limits, as this will be completely used at block creation time, with no further estimation
  
## Proposed changes
- implement a new component that keeps track of the gas consumed throughout tx selection process

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -294,7 +294,11 @@ func (gc *gasConsumption) checkOutgoingTransaction(
 	}
 	// TODO: double check these limits
 	txsLimitReachedForSelfShard := gc.totalGasConsumed[outgoingSelfKey]+gasConsumedInSenderShard > bandwidthForOutgoingTransactions
-	txsLimitReachedForDestShard := gc.totalGasConsumed[outgoingDestKey]+gasConsumedInReceiverShard > bandwidthForOutgoingTransactions
+	txsLimitReachedForDestShard := false
+	if isCrossShard {
+		bandwidthForOutgoingCrossTransactions := gc.getGasLimitForOneDirection(receiverShard)
+		txsLimitReachedForDestShard = gc.totalGasConsumed[outgoingDestKey]+gasConsumedInReceiverShard > bandwidthForOutgoingCrossTransactions
+	}
 	gasConsumedByOutgoingTransactions := gc.getTotalOutgoingGas()
 	txsLimitReachedForTotalOutgoing := gasConsumedByOutgoingTransactions+gasConsumedInSenderShard > bandwidthForOutgoingTransactions
 	txsLimitReachedInAnyShard := txsLimitReachedForSelfShard || txsLimitReachedForDestShard || txsLimitReachedForTotalOutgoing
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -27,7 +27,6 @@ const (
 
 type txCheckResult struct {
 	shouldSkipSender bool
-	shouldStop       bool
 	err              error
 }
 
@@ -270,9 +269,9 @@ func (gc *gasConsumption) CheckOutgoingTransactions(transactions []data.Transact
 			skippedSenders[string(transactions[i].GetSndAddr())] = struct{}{}
 			continue
 		}
-		if result.err != nil || result.shouldStop {
-			gc.isTransactionSelectionDone = true
-			return addedTransactions, result.err
+		if result.err != nil {
+			log.Warn("checkOutgoingTransaction failed", "error", result.err.Error())
+			continue
 		}
 
 		addedTransactions = append(addedTransactions, transactions[i])
@@ -292,7 +291,6 @@ func (gc *gasConsumption) checkOutgoingTransaction(
 	if check.IfNil(tx) {
 		return txCheckResult{
 			shouldSkipSender: false,
-			shouldStop:       false,
 			err:              nil,
 		}
 	}
@@ -302,8 +300,7 @@ func (gc *gasConsumption) checkOutgoingTransaction(
 	gasConsumedInSenderShard, gasConsumedInReceiverShard, err := gc.checkGasConsumedByTx(senderShard, receiverShard, tx)
 	if err != nil {
 		return txCheckResult{
-			shouldSkipSender: false,
-			shouldStop:       true,
+			shouldSkipSender: true,
 			err:              err,
 		}
 	}
@@ -312,14 +309,12 @@ func (gc *gasConsumption) checkOutgoingTransaction(
 	if shouldSkipSender {
 		return txCheckResult{
 			shouldSkipSender: true,
-			shouldStop:       false,
 			err:              nil,
 		}
 	}
 
 	return txCheckResult{
 		shouldSkipSender: false,
-		shouldStop:       false,
 		err:              nil,
 	}
 }
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -180,7 +180,7 @@ func (gc *gasConsumption) checkIncomingMiniBlock(
 		gasConsumedByOutgoingTransactions := gc.getTotalOutgoingGas()
 		bandwidthForIncomingMiniBlocks += bandwidthForOutgoingTransactions - gasConsumedByOutgoingTransactions
 	}
-	mbsLimitReached := gc.totalGasConsumed[string(incoming)]+gasConsumedByMB > blockGasLimitForOneDirection
+	mbsLimitReached := gc.totalGasConsumed[string(incoming)]+gasConsumedByMB > bandwidthForIncomingMiniBlocks
 	if !mbsLimitReached {
 		// limit not reached, continue
 		// simply increment the lastMiniBlockIndex as this method might be called either from
@@ -292,6 +292,7 @@ func (gc *gasConsumption) checkOutgoingTransaction(
 		gasConsumedByIncomingMiniBlocks := gc.totalGasConsumed[string(incoming)]
 		bandwidthForOutgoingTransactions += bandwidthForIncomingMiniBlocks - gasConsumedByIncomingMiniBlocks
 	}
+	// TODO: double check these limits
 	txsLimitReachedForSelfShard := gc.totalGasConsumed[outgoingSelfKey]+gasConsumedInSenderShard > bandwidthForOutgoingTransactions
 	txsLimitReachedForDestShard := gc.totalGasConsumed[outgoingDestKey]+gasConsumedInReceiverShard > bandwidthForOutgoingTransactions
 	gasConsumedByOutgoingTransactions := gc.getTotalOutgoingGas()
@@ -365,8 +366,8 @@ func (gc *gasConsumption) GetLastMiniBlockIndexIncluded() int {
 	return gc.lastMiniBlockIndex
 }
 
-// GetLasTransactionIndexIncluded returns the last transactions index added
-func (gc *gasConsumption) GetLasTransactionIndexIncluded() int {
+// GetLastTransactionIndexIncluded returns the last transactions index added
+func (gc *gasConsumption) GetLastTransactionIndexIncluded() int {
 	gc.mut.RLock()
 	defer gc.mut.RUnlock()
 
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -25,11 +25,6 @@ const (
 	initialLastIndex       = -1
 )
 
-type txCheckResult struct {
-	shouldSkipSender bool
-	err              error
-}
-
 // ArgsGasConsumption holds the arguments needed to create a gasConsumption instance
 type ArgsGasConsumption struct {
 	EconomicsFee                      process.FeeHandler
@@ -264,13 +259,9 @@ func (gc *gasConsumption) CheckOutgoingTransactions(transactions []data.Transact
 			continue
 		}
 
-		result := gc.checkOutgoingTransaction(transactions[i])
-		if result.shouldSkipSender {
+		shouldSkipSender = gc.checkOutgoingTransaction(transactions[i])
+		if shouldSkipSender {
 			skippedSenders[string(transactions[i].GetSndAddr())] = struct{}{}
-			continue
-		}
-		if result.err != nil {
-			log.Warn("checkOutgoingTransaction failed", "error", result.err.Error())
 			continue
 		}
 
@@ -287,36 +278,25 @@ func (gc *gasConsumption) CheckOutgoingTransactions(transactions []data.Transact
 // must be called under mutex protection
 func (gc *gasConsumption) checkOutgoingTransaction(
 	tx data.TransactionHandler,
-) txCheckResult {
+) bool {
 	if check.IfNil(tx) {
-		return txCheckResult{
-			shouldSkipSender: false,
-			err:              nil,
-		}
+		return false
 	}
 
 	senderShard := gc.shardCoordinator.SelfId()
 	receiverShard := gc.shardCoordinator.ComputeId(tx.GetRcvAddr())
 	gasConsumedInSenderShard, gasConsumedInReceiverShard, err := gc.checkGasConsumedByTx(senderShard, receiverShard, tx)
 	if err != nil {
-		return txCheckResult{
-			shouldSkipSender: true,
-			err:              err,
-		}
+		log.Warn("checkOutgoingTransaction.checkGasConsumedByTx failed", "error", err)
+		return true
 	}
 
 	shouldSkipSender := gc.checkShardsLimits(senderShard, receiverShard, gasConsumedInSenderShard, gasConsumedInReceiverShard)
 	if shouldSkipSender {
-		return txCheckResult{
-			shouldSkipSender: true,
-			err:              nil,
-		}
+		return true
 	}
 
-	return txCheckResult{
-		shouldSkipSender: false,
-		err:              nil,
-	}
+	return false
 }
 
 func (gc *gasConsumption) checkGasConsumedByTx(
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -48,7 +48,9 @@ type gasConsumption struct {
 	isMiniBlockSelectionDone         bool
 	isTransactionSelectionDone       bool
 	miniBlockLimitFactor             uint64
-	blockLimitFactor                 uint64
+	incomingLimitFactor              uint64
+	outgoingLimitFactor              uint64
+	decreaseStep                     uint64
 }
 
 // NewGasConsumption creates a new instance of gasConsumption
@@ -71,9 +73,11 @@ func NewGasConsumption(args ArgsGasConsumption) (*gasConsumption, error) {
 		gasConsumedByMiniBlock:           make(map[string]uint64),
 		transactionsForPendingMiniBlocks: make(map[string][]data.TransactionHandler),
 		miniBlockLimitFactor:             initialLimitsFactor,
-		blockLimitFactor:                 initialLimitsFactor,
+		incomingLimitFactor:              initialLimitsFactor,
+		outgoingLimitFactor:              initialLimitsFactor,
 		lastMiniBlockIndex:               initialLastIndex,
 		lastTransactionIndex:             initialLastIndex,
+		decreaseStep:                     initialLimitsFactor * percentDecreaseLimitsStep / 100,
 	}, nil
 }
 
@@ -94,7 +98,7 @@ func (gc *gasConsumption) CheckIncomingMiniBlocks(
 		return initialLastIndex, 0, process.ErrMiniBlocksAlreadyProcessed
 	}
 
-	blockGasLimitForOneDirection := gc.getGasLimitForOneDirection(gc.shardCoordinator.SelfId())
+	blockGasLimitForOneDirection := gc.getGasLimitForOneDirection(incoming, gc.shardCoordinator.SelfId())
 
 	for i := 0; i < len(miniBlocks); i++ {
 		shouldSavePending, shouldStop, err := gc.checkIncomingMiniBlock(miniBlocks[i], transactions, blockGasLimitForOneDirection)
@@ -176,7 +180,7 @@ func (gc *gasConsumption) checkIncomingMiniBlock(
 	bandwidthForIncomingMiniBlocks := blockGasLimitForOneDirection
 	if gc.isTransactionSelectionDone {
 		// if transactions are already handled, use the space left
-		bandwidthForOutgoingTransactions := gc.getGasLimitForOneDirection(gc.shardCoordinator.SelfId())
+		bandwidthForOutgoingTransactions := gc.getGasLimitForOneDirection(outgoing, gc.shardCoordinator.SelfId())
 		gasConsumedByOutgoingTransactions := gc.getTotalOutgoingGas()
 		bandwidthForIncomingMiniBlocks += bandwidthForOutgoingTransactions - gasConsumedByOutgoingTransactions
 	}
@@ -214,7 +218,7 @@ func (gc *gasConsumption) checkPendingIncomingMiniBlocks() error {
 		return nil
 	}
 
-	blockGasLimitForOneDirection := gc.getGasLimitForOneDirection(gc.shardCoordinator.SelfId())
+	blockGasLimitForOneDirection := gc.getGasLimitForOneDirection(incoming, gc.shardCoordinator.SelfId())
 	for i := 0; i < len(gc.pendingMiniBlocks); i++ {
 		mb := gc.pendingMiniBlocks[i]
 		_, shouldStop, err := gc.checkIncomingMiniBlock(mb, gc.transactionsForPendingMiniBlocks, blockGasLimitForOneDirection)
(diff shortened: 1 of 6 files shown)


process/block/gasConsumption.go
```diff
@@ -0,0 +1,158 @@
...
+func (gc *gasConsumption) CheckTransactionsForMiniBlock(
+	miniBlockHash []byte,
+	txs []data.TransactionHandler,
+) (uint32, error) {
+	gc.Lock()
+	defer gc.Unlock()
+
+	lastIndexAdded := uint32(0)
+
+	senderShard := gc.shardCoordinator.ComputeId(txs[0].GetSndAddr())
```
ssd04: check len(txs) > 0

process/block/gasConsumption.go
```diff
@@ -0,0 +1,158 @@
...
+	percentSplitBlock         = uint64(50) // 50%
+	percentDecreaseLimitsStep = uint64(10) // 10%
+)
+
+type gasConsumption struct {
+	sync.RWMutex
+	economicsFee            process.FeeHandler
+	shardCoordinator        sharding.Coordinator
+	totalGasConsumedInShard map[uint32]uint64
+	gasConsumedByMiniBlock  map[string]uint64
```
ssd04: could we end up having a huge map here? or some mechanism to cleanup old miniblocks, instead of just `Reset` all?

sstanculeanu: not needed, this component should be reset for each block

process/block/gasConsumption.go
sstanculeanu: pushed

process/block/gasConsumption_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+	// decrease limits 5 times, so they will be half
+	gc.DecreaseLimits()
+	gc.DecreaseLimits()
+	gc.DecreaseLimits()
+	gc.DecreaseLimits()
+	gc.DecreaseLimits()
+
+	// gas exceeded per mb
+	txsToExceedMb := getTxs(maxGasLimitPerTx, 6) // 6 should be enough as the limit is nou 50 / mb
```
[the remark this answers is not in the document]
sstanculeanu: fixed

process/block/gasConsumption.go
```diff
@@ -0,0 +1,162 @@
...
+		if tx.GetGasLimit() > maxGasLimitPerTx {
+			return lastIndexAdded, process.ErrInvalidMaxGasLimitPerTx
+		}
+
+		gasConsumedByMB += tx.GetGasLimit()
+		if gasConsumedByMB > maxGasLimitPerMB {
+			return lastIndexAdded, process.ErrMaxGasLimitPerMiniBlockIsReached
+		}
+
+		gasLeftForShard := gc.getMaxGasLeftForShard(senderShard)
```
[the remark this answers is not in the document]
sstanculeanu: fixed

process/block/gasConsumption.go
```diff
@@ -0,0 +1,162 @@
...
+	return gc.gasConsumedByMiniBlock[string(hash)]
+}
+
+// DecreaseLimits lowers the block limits
+func (gc *gasConsumption) DecreaseLimits() {
+	gc.Lock()
+	defer gc.Unlock()
+
+	decreaseStep := initialLimitsFactor * percentDecreaseLimitsStep / 100
+	gc.limitsFactor = gc.limitsFactor - decreaseStep
```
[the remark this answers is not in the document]
sstanculeanu: fixed

process/block/gasConsumption.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/multiversx/mx-chain-go/sharding"
+)
+
+const (
+	initialLimitsFactor = uint64(100)
+	// TODO: move these to config
+	percentSplitBlock         = uint64(50) // 50%
+	percentDecreaseLimitsStep = uint64(10) // 10%
+)
+
```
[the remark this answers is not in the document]
sstanculeanu: will stay as a TODO for now

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+	return gc.gasConsumedByMiniBlock[string(hash)]
+}
+
+// DecreaseLimits lowers the block limits
+func (gc *gasConsumption) DecreaseLimits() {
+	gc.Lock()
+	defer gc.Unlock()
+
+	decreaseStep := initialLimitsFactor * percentDecreaseLimitsStep / 100
+	gc.limitsFactor = gc.limitsFactor - decreaseStep
```
AdoAdoAdo: a check needs to be done before the decrease, that the decreaseStep<= gc.limitsFctor, as this is an uint64, and it can underflow (it will never be negative).

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if gc.limitsFactor <= minPercentLimitsFactor {
+		gc.limitsFactor = minPercentLimitsFactor
+	}
+}
+
+// Reset resets the accumulated values to the initial state
+func (gc *gasConsumption) Reset() {
```
AdoAdoAdo: The reset would be called on every block proposal, but it will probably not need to reset the limitsFactor as well.
e.g the execution still needs to catch up with the proposals.

So maybe we need a separate method to reset the gas limit factor to the default.

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return int64(limit) - consumedInOtherShards
+}
+
+func (gc *gasConsumption) maxGasLimitPerBlock(shardID uint32) uint64 {
+	return gc.economicsFee.MaxGasLimitPerBlock(shardID) * gc.limitsFactor / 100
+}
+
+func (gc *gasConsumption) maxGasLimitPerMiniBlock(shardID uint32) uint64 {
+	return gc.economicsFee.MaxGasLimitPerMiniBlock(shardID) * gc.limitsFactor / 100
```
AdoAdoAdo: I think the factor should apply only on block gas limit.

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return gc.gasConsumedByMiniBlock[string(hash)]
+}
+
+// DecreaseLimits lowers the block limits
+func (gc *gasConsumption) DecreaseLimits() {
+	gc.Lock()
+	defer gc.Unlock()
+
+	decreaseStep := initialLimitsFactor * percentDecreaseLimitsStep / 100
```
AdoAdoAdo: can we have the step computed in constructor?

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+	"sync"
+
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data"
+	"github.com/multiversx/mx-chain-go/process"
+	"github.com/multiversx/mx-chain-go/sharding"
+)
+
+const (
+	initialLimitsFactor = uint64(100)
```
AdoAdoAdo: maybe this one we will set to 200% for block gas overestimation (to compensate tx gas limit overestimations)

process/block/gasConsumption.go
```diff
@@ -0,0 +1,173 @@
...
+		totalGasConsumedInShard: make(map[uint32]uint64),
+		gasConsumedByMiniBlock:  make(map[string]uint64),
+		limitsFactor:            initialLimitsFactor,
+	}, nil
+}
+
+// CheckTransactionsForMiniBlock checks if the provided transactions can be added in a mini block
+// for the current selection session
+func (gc *gasConsumption) CheckTransactionsForMiniBlock(
+	miniBlockHash []byte,
```
AdoAdoAdo: for the proposed intrashard miniblock, we don't have yet a mbHash, we need to first see which transactions can fit into the current limits.

process/block/gasConsumption.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	return gc.economicsFee.MaxGasLimitPerBlock(shardID) * gc.blockLimitFactor / 100
+}
+
+// must be called under mutex protection as it access miniBlockLimitFactor
+func (gc *gasConsumption) maxGasLimitPerMiniBlock(shardID uint32) uint64 {
+	isCrossShard := shardID != gc.shardCoordinator.SelfId()
+	if isCrossShard {
+		return gc.economicsFee.MaxGasLimitPerMiniBlockForSafeCrossShard() * gc.miniBlockLimitFactor / 100
```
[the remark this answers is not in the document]
sstanculeanu: not really

process/block/gasConsumption.go
```diff
@@ -0,0 +1,493 @@
...
+)
+
+// gasType defines the type of gas consumption
+type gasType string
+
+const (
+	incoming gasType = "incoming"
+	outgoing gasType = "outgoing"
+
+	// TODO: move these to config
```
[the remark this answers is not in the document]
sstanculeanu: we'll be handled later

process/block/gasConsumption.go
```diff
@@ -0,0 +1,493 @@
...
+	outgoingSelfKey := getOutgoingKey(senderShard)
+	outgoingDestKey := getOutgoingKey(receiverShard)
+	bandwidthForOutgoingTransactions := gc.getGasLimitForOneDirection(senderShard)
+	if gc.isMiniBlockSelectionDone {
+		// if mini blocks are already handled, use the space left
+		bandwidthForIncomingMiniBlocks := gc.getGasLimitForOneDirection(senderShard)
+		gasConsumedByIncomingMiniBlocks := gc.totalGasConsumed[string(incoming)]
+		bandwidthForOutgoingTransactions += bandwidthForIncomingMiniBlocks - gasConsumedByIncomingMiniBlocks
+	}
+	// TODO: double check these limits
```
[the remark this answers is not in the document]
sstanculeanu: exactly, will be discussed

process/block/gasConsumption.go
```diff
@@ -0,0 +1,493 @@
...
+}
+
+// CheckIncomingMiniBlocks verifies if an incoming mini block and its transactions can be included within gas limits
+// returns the last mini block index included, the number of pending mini blocks left and error if needed
+func (gc *gasConsumption) CheckIncomingMiniBlocks(
+	miniBlocks []data.MiniBlockHeaderHandler,
+	transactions map[string][]data.TransactionHandler,
+) (int, int, error) {
+	if len(miniBlocks) == 0 || len(transactions) == 0 {
+		return gc.lastMiniBlockIndex, 0, nil
```
miiu96, line 103: Maybe this section should be protected by mutex ?

sstanculeanu: right, moved the lock/unlock before this

process/block/gasConsumption_test.go
```diff
@@ -0,0 +1,672 @@
...
+					gc.ResetOutgoingLimit()
+				case 6:
+					gc.ResetIncomingLimit()
+				case 7:
+					gc.GetLastMiniBlockIndexIncluded()
+				case 8:
+					gc.GetLastTransactionIndexIncluded()
+				case 9:
+					gc.TotalGasConsumed()
+				default:
```
ssd04, line 604: you can also add `ResetMiniBlockLimit` and `DecreaseMiniBlockLimit`

process/block/gasConsumption.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	if gc.isMiniBlockSelectionDone {
+		return initialLastIndex, 0, process.ErrMiniBlocksAlreadyProcessed
+	}
+
+	blockGasLimitForOneDirection := gc.getGasLimitForOneDirection(incoming, gc.shardCoordinator.SelfId())
+
```
ssd04: maybe defer with `gc.isMiniBlockSelectionDone = true`?, not a strong opinion though

process/block/gasConsumption.go
```diff
@@ -0,0 +1,524 @@
...
+	if gc.isTransactionSelectionDone {
+		return initialLastIndex, process.ErrTransactionsAlreadyProcessed
+	}
+
+	for i := 0; i < len(transactions); i++ {
+		shouldSavePending, shouldStop, err := gc.checkOutgoingTransaction(transactions[i])
+		if shouldSavePending {
+			gc.pendingTransactions = append(gc.pendingTransactions, transactions[i:]...)
+			gc.isTransactionSelectionDone = true
```
ssd04: same here with defer

process/block/gasConsumption.go
```diff
@@ -0,0 +1,524 @@
...
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data"
+	"github.com/multiversx/mx-chain-go/process"
+)
+
+// gasType defines the type of gas consumption
+type gasType string
+
+const (
+	incoming gasType = "incoming"
```
AdoAdoAdo: can you switch from string type to integer?
then comparisons will be simpler

process/block/gasConsumption.go
sstanculeanu: it won't work, as the call checkPendingOutgoingTransactions from L121 checks isMiniBlockSelectionDone

process/block/gasConsumption.go
sstanculeanu: same here, checkPendingIncomingMiniBlocks from L64 checks if isTransactionSelectionDone

process/block/gasConsumption.go
```diff
@@ -0,0 +1,529 @@
...
+	err := gc.checkPendingOutgoingTransactions()
+	return gc.lastMiniBlockIndex, 0, err
+}
+
+func (gc *gasConsumption) checkIncomingMiniBlock(
+	mb data.MiniBlockHeaderHandler,
+	transactions map[string][]data.TransactionHandler,
+	blockGasLimitForOneDirection uint64,
+) (bool, bool, error) {
+	if mb == nil {
```
AdoAdoAdo, line 141: can you extract some smaller methods/functions?

process/block/gasConsumption.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+// must be called under mutex protection as it access miniBlockLimitFactor
+func (gc *gasConsumption) maxGasLimitPerMiniBlock(shardID uint32) uint64 {
+	isCrossShard := shardID != gc.shardCoordinator.SelfId()
+	if isCrossShard {
+		return gc.economicsFee.MaxGasLimitPerMiniBlockForSafeCrossShard() * gc.miniBlockLimitFactor / 100
+	}
+
+	return gc.economicsFee.MaxGasLimitPerMiniBlock(shardID) * gc.miniBlockLimitFactor / 100
```
AdoAdoAdo: I don't think we need to adjust the limit for miniBlock dynamically, just the limit for block, and maybe gas limit per destination (cross shard incoming, intra shard, outgoing to ShardX)

Did you have a specific scenario in mind?

process/block/gasConsumption.go
```diff
@@ -0,0 +1,529 @@
...
+
+// must be called under mutex protection
+func (gc *gasConsumption) checkOutgoingTransaction(
+	tx data.TransactionHandler,
+) (bool, bool, error) {
+	if check.IfNil(tx) {
+		return false, true, nil
+	}
+
+	senderShard := gc.shardCoordinator.SelfId()
```
AdoAdoAdo, line 286: can you extract some smaller methods/functions?

process/block/gasConsumption.go
```diff
@@ -0,0 +1,529 @@
...
+// GetLastTransactionIndexIncluded returns the last transactions index added
+func (gc *gasConsumption) GetLastTransactionIndexIncluded() int {
+	gc.mut.RLock()
+	defer gc.mut.RUnlock()
+
+	return gc.lastTransactionIndex
+}
+
+// DecreaseMiniBlockLimit reduces the mini block gas limit by a configured percentage
+func (gc *gasConsumption) DecreaseMiniBlockLimit() {
```
AdoAdoAdo: I am not sure if we need to decrease the miniBlock limit.
Maybe we can discuss this.

process/block/gasConsumption.go
sstanculeanu: updated the method to return without limit factor

process/block/gasConsumption.go
sstanculeanu: removed the methods

process/block/gasConsumption.go
sstanculeanu: this would be later on used as a base for outgoing_dest inside method getOutgoingKey, thus the choice of strings

process/block/gasConsumption.go
sstanculeanu: done

process/block/gasConsumption.go
sstanculeanu: done

ssd04 (approved):

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+		totalGasConsumed:                 make(map[string]uint64),
+		gasConsumedByMiniBlock:           make(map[string]uint64),
+		transactionsForPendingMiniBlocks: make(map[string][]data.TransactionHandler),
+		incomingLimitFactor:              args.InitialLimitsFactor,
+		outgoingLimitFactor:              args.InitialLimitsFactor,
+		lastMiniBlockIndex:               initialLastIndex,
+		lastTransactionIndex:             initialLastIndex,
+		decreaseStep:                     args.InitialLimitsFactor * args.PercentDecreaseLimitsStep / 100,
+		percentDecreaseLimitsStep:        args.PercentDecreaseLimitsStep,
+		initialLimitsFactor:              args.InitialLimitsFactor,
```
AdoAdoAdo: The args.InitialLimitsFactor is a bit confusing maybe something related to block capacity overestimation factor

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+	if err != nil {
+		return false, true, err
+	}
+
+	bandwidthForIncomingMiniBlocks := blockGasLimitForOneDirection
+	if gc.isTransactionSelectionDone {
+		// if transactions are already handled, use the space left
+		bandwidthForOutgoingTransactions := gc.getGasLimitForOneDirection(outgoing, gc.shardCoordinator.SelfId())
+		gasConsumedByOutgoingTransactions := gc.getTotalOutgoingGas()
+		bandwidthForIncomingMiniBlocks += bandwidthForOutgoingTransactions - gasConsumedByOutgoingTransactions
```
AdoAdoAdo: here it is possible to have an overflow (bandwidthForOutgoingTransactions < gasConsumedByOutgoingTransactions) as the operation is on uints.

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+	gasConsumedInReceiverShard uint64,
+) bool {
+	isCrossShard := senderShard != receiverShard
+
+	bandwidthForOutgoingTransactions := gc.getGasLimitForOneDirection(outgoing, senderShard)
+	if gc.isMiniBlockSelectionDone {
+		// if mini blocks are already handled, use the space left
+		bandwidthForIncomingMiniBlocks := gc.getGasLimitForOneDirection(incoming, senderShard)
+		gasConsumedByIncomingMiniBlocks := gc.totalGasConsumed[string(incoming)]
+		bandwidthForOutgoingTransactions += bandwidthForIncomingMiniBlocks - gasConsumedByIncomingMiniBlocks
```
AdoAdoAdo: I think there is a risk of uint overflow here as well if gasConsumedByIncomingMiniBlocks is > bandwidthForIncomingMiniBlocks

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+	if check.IfNil(args.EconomicsFee) {
+		return nil, process.ErrNilEconomicsFeeHandler
+	}
+	if check.IfNil(args.ShardCoordinator) {
+		return nil, process.ErrNilShardCoordinator
+	}
+	if check.IfNil(args.GasHandler) {
+		return nil, process.ErrNilGasHandler
+	}
+	if args.InitialLimitsFactor <= minPercentLimitsFactor {
```
AdoAdoAdo: maybe add an initial limits factor upper bound as well?

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+	for i := 0; i < len(miniBlocks); i++ {
+		shouldSavePending, shouldStop, err := gc.checkIncomingMiniBlock(miniBlocks[i], transactions, blockGasLimitForOneDirection)
+		if shouldSavePending {
+			gc.pendingMiniBlocks = append(gc.pendingMiniBlocks, miniBlocks[i:]...)
+			gc.transactionsForPendingMiniBlocks = transactions
+			gc.isMiniBlockSelectionDone = true
+
+			return gc.lastMiniBlockIndex, len(gc.pendingMiniBlocks), err
+		}
+		if err != nil || shouldStop {
```
AdoAdoAdo: maybe move the error check before shouldSavePending check?
I don't think we save any pending in case of error.

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+
+	senderShard := gc.shardCoordinator.SelfId()
+	receiverShard := gc.shardCoordinator.ComputeId(tx.GetRcvAddr())
+	gasConsumedInSenderShard, gasConsumedInReceiverShard, err := gc.checkGasConsumedByTx(senderShard, receiverShard, tx)
+	if err != nil {
+		return false, true, err
+	}
+
+	shouldContinue := gc.checkShardsLimits(senderShard, receiverShard, gasConsumedInSenderShard, gasConsumedInReceiverShard)
```
AdoAdoAdo: I think the condition here is a bit too strict, as if we reach any limit we no longer check to fit any other transactions.
e.g if we reach the limit for one destination shard but not the one in our shard and the remaining other shards, we no longer add transactions.

I would suggest to iterate through all transactions, and skip the ones that do not fit into the remaining gas bandwidth (make sure to skip the next transactions from the same senders as well to avoid nonce gaps).

AdoAdoAdo: can we split this outgoing gas limit into 2 buckets?
1. what we estimate to consume intrashard
2. what we estimate to consume in all destination shards (other than own shard) together.

then have for each of these buckets the same limit.
e.g if we start with a 2*k "inflated" block gas capacity to account for transactions gas overestimation, then this will be split in the following buckets, where k - the real gas used capacity - e.g. 600 mil gas, then:
- cross shard incoming transactions gas bucket (estimated gas) - 1*k
- in shard gas usage from intra-shard and cross shard outgoing transactions  - 1*k (this one and the one above will use in own shard the full block slot for execution if transactions are ~2x overestimated)
- destination other shard gas usage from outgoing transactions - 1*k (this one does not take time from own shard to execute, but will take from the destinations)

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+// DecreaseIncomingLimit reduces the gas limit for incoming mini blocks by a configured percentage
+func (gc *gasConsumption) DecreaseIncomingLimit() {
+	gc.mut.Lock()
+	defer gc.mut.Unlock()
+
+	if gc.incomingLimitFactor == minPercentLimitsFactor {
+		return
+	}
+
+	gc.incomingLimitFactor = gc.incomingLimitFactor - gc.decreaseStep
```
AdoAdoAdo: possible uint overflow if decreaseStep > incomingLimitFactor

process/block/gasConsumption.go
```diff
@@ -0,0 +1,540 @@
...
+// DecreaseOutgoingLimit reduces the gas limit for outgoing transactions by a configured percentage
+func (gc *gasConsumption) DecreaseOutgoingLimit() {
+	gc.mut.Lock()
+	defer gc.mut.Unlock()
+
+	if gc.outgoingLimitFactor == minPercentLimitsFactor {
+		return
+	}
+
+	gc.outgoingLimitFactor = gc.outgoingLimitFactor - gc.decreaseStep
```
AdoAdoAdo: possible uint overflow if decreaseStep > outgoingLimitFactor

process/block/gasConsumption.go
sstanculeanu: renamed to BlockCapacityOverestimationFactor

process/block/gasConsumption.go
sstanculeanu: added

process/block/gasConsumption.go
sstanculeanu: removed this section, as discussed, won't be the case.. first we handle the incoming miniblocks and after that the outgoing txs

process/block/gasConsumption.go
sstanculeanu: left it after shouldSavePending check in order to stop for both conditions: either error or shouldStop

process/block/gasConsumption.go
sstanculeanu: extracted into a new method and tackled this case

miiu96 (dismissed):

miiu96 (dismissed):

AdoAdoAdo (approved):

process/block/gasConsumption.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	mbsLimitReached := gc.totalGasConsumed[incoming]+gasConsumedByMB > bandwidthForIncomingMiniBlocks
+	if !mbsLimitReached {
+		// limit not reached, continue
+		// simply increment the lastMiniBlockIndex as this method might be called either from
+		// handling all mini blocks, either from handling pending, where the pending ones
+		// should have continuos indexes after the ones already included
```
ssd04, line 169:
```suggestion
		// should have continuous indexes after the ones already included
```

PR #7145, opened 2025-07-24
Handling for chronology watchdog alarm for supernova

ssd04:
- Handler transition for chronology number of rounds to wait before an alarms

## Testing procedure
- TBD

## Pre-requisites

### DIFF
--- a/consensus/chronology/chronology.go
+++ b/consensus/chronology/chronology.go
@@ -26,7 +26,9 @@ var log = logger.GetOrCreate("consensus/chronology")
 // srBeforeStartRound defines the state which exist before the start of the round
 const srBeforeStartRound = -1
 
+// TODO: add variables to config
 const numRoundsToWaitBeforeSignalingChronologyStuck = 10
+const numRoundsToWaitBeforeSignalingChronologyStuckSupernova = 100
 const chronologyAlarmID = "chronology"
 
 // ArgChronology holds all dependencies required by the chronology component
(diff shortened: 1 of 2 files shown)


consensus/chronology/chronology.go
```diff
@@ -26,7 +26,9 @@ var log = logger.GetOrCreate("consensus/chronology")
 // srBeforeStartRound defines the state which exist before the start of the round
 const srBeforeStartRound = -1
 
+// TODO: add variables to config
 const numRoundsToWaitBeforeSignalingChronologyStuck = 10
+const numRoundsToWaitBeforeSignalingChronologyStuckSupernova = 100
 const chronologyAlarmID = "chronology"
```
ssd04, line 32: maybe add a component similar to `EpochChangeGracePeriod` (or maybe rename and expand it to handle more params for other components as well) with parameters per epoch

ssd04: we also have other constants that might need this treatment

AdoAdoAdo (approved):

consensus/chronology/chronology_test.go
```diff
@@ -400,22 +402,53 @@ func TestChronology_Close(t *testing.T) {
...
+		}
+
+		chr, err := chronology.NewChronology(arg)
+		require.Nil(t, err)
+
+		chr.StartRounds()
+
+		time.Sleep(5 * time.Millisecond)
+
+		require.True(t, updateRoundCalled)
```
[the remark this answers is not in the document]
ssd04: not really necessary for this small test case

PR #7147, opened 2025-07-25
Txpool profile

danielradu10:
**- use pprof for txpool benchmarks**
  
## Proposed changes
- added an env variable for the option of generating the pprof files or not
- added the cpu profile for each test

## Pre-requisites


integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -353,8 +358,25 @@ func testOnProposed(t *testing.T, sw *core.StopWatch, numTxs int, numAddresses i
...
+		err = pprof.StartCPUProfile(f)
+		require.Nil(t, err)
+
+		defer func() {
+			// stop profiling
+			pprof.StopCPUProfile()
+			err = f.Close()
+			require.NoError(t, err)
+		}()
+	}
```
sstanculeanu: extract this in a new method and reuse it

perhaps return the file and keep only the defer part

sstanculeanu: also inlcude initEnvVariables() in the same method

maybe also rename initEnvVariables() to checkEnvVariables

danielradu10: fixed

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -353,8 +356,18 @@ func testOnProposed(t *testing.T, sw *core.StopWatch, numTxs int, numAddresses i
 	require.Equal(t, numTxs, len(selectedTransactions))
 
 	proposedBlock1 := createProposedBlock(selectedTransactions)
-	sw.Start(t.Name())
 
+	f := createPprofFiles(t)
```
andreibancioiu: Now I see that this function does not belong to the common test utilities. I think it belongs in the file that implements the benchmarks.

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -353,8 +356,18 @@ func testOnProposed(t *testing.T, sw *core.StopWatch, numTxs int, numAddresses i
...
+	if f != nil {
+		defer func() {
+			// stop profiling
```
andreibancioiu: Redundant comment :see_no_evil:

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -523,10 +553,45 @@ func testSecondSelectionWithManyTxsInPool(t *testing.T, sw *core.StopWatch, numT
...
+	f, err := os.Create(fileName)
+	require.Nil(t, err)
+
+	err = pprof.StartCPUProfile(f)
+	require.Nil(t, err)
+
+	return f
+}
+
+func checkEnvVariables() bool {
```
andreibancioiu: Function has a generic name, does something specific. Maybe rename it to `shouldCreatePprofFiles()`? `shouldDoCPUProfiling`?

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -402,10 +413,19 @@ func testFirstSelection(t *testing.T, sw *core.StopWatch, numTxs int, numTxsToBe
 
 	blockchainInfo := holders.NewBlockchainInfo([]byte("blockHash0"), 1)
 
+	f := createPprofFiles(t)
```
andreibancioiu: I think we should explicitly call the start profiling & stop profiling commands not here (in the test utilities), but in the benchmark tests themselves. So that we can individually (for each benchmark) measure what we really want to measure.

andreibancioiu (changes requested):

integrationTests/chainSimulator/mempool/testutils_test.go
danielradu10: moved the profiling call inside the benchmarks

integrationTests/chainSimulator/mempool/testutils_test.go
danielradu10: fixed

integrationTests/chainSimulator/mempool/testutils_test.go
danielradu10: moved it inside benchmarks file

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -523,10 +553,45 @@ func testSecondSelectionWithManyTxsInPool(t *testing.T, sw *core.StopWatch, numT
...
+
+func createPprofFiles(t *testing.T) *os.File {
+	shouldCreatePprofFiles := checkEnvVariables()
+	if !shouldCreatePprofFiles {
+		return nil
+	}
+
+	// start profiling
+	testName := strings.ReplaceAll(t.Name(), "/", "_")
+	fileName := fmt.Sprintf("%s.pprof", testName)
```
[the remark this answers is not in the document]
danielradu10: fixed

integrationTests/chainSimulator/mempool/testutils_test.go
```diff
@@ -523,10 +553,45 @@ func testSecondSelectionWithManyTxsInPool(t *testing.T, sw *core.StopWatch, numT
...
+	f, err := os.Create(fileName)
+	require.Nil(t, err)
+
+	err = pprof.StartCPUProfile(f)
+	require.Nil(t, err)
```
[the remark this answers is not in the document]
danielradu10: fixed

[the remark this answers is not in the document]
danielradu10: fixed

integrationTests/chainSimulator/mempool/testutils_test.go
danielradu10: fixed

andreibancioiu (commented):
Idea for refactoring:

```
func setupProfilingIfNecessary() - returns a tuple of two functions, the one that starts profiling, the one that stops profiling
```

In benchmarks:

```
startProfiling, stopProfiling, shouldDoProfiling := setupProfilingIfNecessary()

if shouldDoProfiling then startProfiling() and defer stopProfiling
```

Advantages - less (duplicated) code, a bit more expressive benchmarks.

integrationTests/chainSimulator/mempool/benchmarks_test.go
```diff
@@ -27,59 +31,156 @@ var (
 	}
 )
 
+func setUpPprofFilesIfNecessary(t *testing.T) (bool, func(t *testing.T) *os.File, func(t *testing.T, f *os.File)) {
+	return shouldCreatePprofFiles(), createPprofFiles, shouldStopProfiling
```
sstanculeanu: I would handle here the logic and only return 2 methods

suggested changes:
- rename `shouldStopProfiling` to `stopProfiling`
- `setUpPprofFilesIfNecessary` only returns 2 functions, `func(t *testing.T) io.Closer`, `func(t *testing.T, closable io.Closer))`
- if `shouldCreatePprofFiles` returns false, then simply return empty methods that do nothing/call mocked/disabled components
- `stopProfiling` receives a Closer interface parameter, instead of pointer to file
- `createPprofFiles` returns a Closer interface, instead of file pointer. If env var is not found, then return a CloserMock/CloserStub that do nothing

This way we reduce the check for `shouldDoProfiling`

sstanculeanu (approved):

andreibancioiu (approved):

PR #7194, opened 2025-08-18
MX-16926 Execution result inclusion estimator

mradian1:
- First version of Supernova will need to estimate when to notarize execution results, so that the block with the notarization passes consensus.
  
## Proposed changes
- Add a deterministic component **Execution‑Result Inclusion Estimator (EIE)** for *Supernova* to determine, at proposal‑time and at validation‑time, whether one or more pending execution results can be safely embedded in the block that is being produced / verified. Safety means that a validator running on **minimum‑spec hardware** will have completed execution of every referenced block *before* it must validate the new block.

## Testing procedure
- Unit tests: prefix acceptance / rejection, boundary at exactly `t_done == t_now`, max‑results cap, overflow protection                                                                                
- Property‑based: Randomised gas vectors & timestamps maintain invariants (no unsafe inclusion)
- Performance: Ensure ≤100µs at 12 results, linear scaling.                                                         |
- Chaos / Fault‑injection: Clock‑skew headers, malformed gas values, out‑of‑order nonces; Full proposer / validator run; inject artificial slow executor and verify blocks with/without results.

## Pre-requisites


process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+
+var log = logger.GetOrCreate("process/executionResultInclusionEstimator")
+
+// ExecutionResultMeta is a lightweight summary EIE requires.
+type ExecutionResultMeta struct {
+	HeaderHash   [32]byte // Link to full header in DB / cache
+	HeaderNonce  uint64   // Monotonic within shard
+	HeaderTimeMs uint64   // Milliseconds since Unix epoch, from header
+	GasUsed      uint64   // Units actually consumed (post‑execution)
+}
```
sstanculeanu: this should not be needed here, but instead reference the core-go implementation from [here](https://github.com/multiversx/mx-chain-core-go/pull/378/files#diff-566ebf8d66d38733f14c59795044870aa8f49e5d604931c5fcebdac628b9d169R55)

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Config supplied at construction, read‑only thereafter.
+// TODO add also max estimated block gas capacity  // used gas must be lower than this
+type Config struct {
```
sstanculeanu: we usually define configs inside `config` package. Also, it needs an entry inside `config.toml`. Also, not necessarily read-only thereafter, perhaps comment should be updated.

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	SafetyMargin       uint64 // default 110
+	MaxResultsPerBlock uint64 // 0 = unlimited
+	GenesisTimeMs      uint64 // required if lastNotarised == nil
+}
+
+type ExecutionResultInclusionEstimator struct {
```
sstanculeanu, line 26: missing comment on exported

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	cfg  Config // immutable after construction
+	tGas uint64 // time per gas unit on **minimum‑spec** hardware - 1 ns per gas unit
+	// TODO add also max estimated block gas capacity  // used gas must be lower than this
+}
+
+func NewExecutionResultInclusionEstimator(cfg Config) *ExecutionResultInclusionEstimator {
+	return &ExecutionResultInclusionEstimator{
+		cfg:  cfg,
+		tGas: uint64(1),
```
sstanculeanu: tGas should also come from config or a defined constant(considering that Setting it later is also allowed)

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+}
+
+func (erie *ExecutionResultInclusionEstimator) SetTimePerGasUnit(tGas uint64) {
```
sstanculeanu: missing comment on exported

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	if tGas == 0 {
+		log.Warn("ExecutionResultInclusionEstimator: SetTimePerGasUnit called with zero value, using default 1 ns per gas unit")
+		tGas = 1
+	}
```
sstanculeanu: perhaps check if tGas == 0 and simply return, without setting to default but keep the previous value with a log.Debug

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+
+func NewExecutionResultInclusionEstimator(cfg Config) *ExecutionResultInclusionEstimator {
+	return &ExecutionResultInclusionEstimator{
+		cfg:  cfg,
+		tGas: uint64(1),
+	}
+}
+
+func (erie *ExecutionResultInclusionEstimator) SetTimePerGasUnit(tGas uint64) {
+	if tGas == 0 {
```
sstanculeanu: we need a mutex on this component in order to protect tGas, as this method might be called concurrently

suggestion: add a concurrency unittest

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		log.Debug("ExecutionResultInclusionEstimator: SetTimePerGasUnit called",
+			"oldValue", erie.tGas,
+			"newValue", tGas)
+		erie.tGas = tGas
+	}
+}
+
+// Decide returns the prefix of `pending` that may be inserted into the block currently being built / verified.
+// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]`and embeds them.
+func (erie *ExecutionResultInclusionEstimator) Decide(cfg Config,
```
sstanculeanu: cfg param is not needed/used, we already have it as member of the structure

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	lastNotarised *ExecutionResultMeta,
+	pending []ExecutionResultMeta,
+	currentHdrTsNs uint64) (allowed int) {
```
sstanculeanu:
```suggestion
	currentHdrTsNs uint64,
) (allowed int) {
```

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+	allowed = 0
+
+	// time per gas unit on **minimum‑spec** hardware - 1 ns per gas unit
```
sstanculeanu: is this comment needed?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+			"oldValue", erie.tGas,
+			"newValue", tGas)
+		erie.tGas = tGas
+	}
+}
+
+// Decide returns the prefix of `pending` that may be inserted into the block currently being built / verified.
+// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]`and embeds them.
+func (erie *ExecutionResultInclusionEstimator) Decide(cfg Config,
+	lastNotarised *ExecutionResultMeta,
```
sstanculeanu: we usually use notari**z**ed across the project

sstanculeanu: also, referencing the core-go PR would allow using interfaces here as parameters, instead of pointers. Each new structure has an interface defined in core-go

as this component should work for shard as well, probably `BaseExecutionResultHandler` is the most appropriate

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		tBase = convertMsToNs(erie.cfg.GenesisTimeMs)
+	} else {
+		tBase = convertMsToNs(lastNotarised.HeaderTimeMs)
+	}
+
+	// accumulated execution time in ns (1 gas = 1ns)
+	estimatedTime := uint64(0)
+	for i, executionResultMeta := range pending {
+		// Check for nonce monotonicity
+		if i > 0 && executionResultMeta.HeaderNonce <= pending[i-1].HeaderNonce {
```
sstanculeanu: should this check be updated to verify exactly `executionResultMeta.HeaderNonce != pending[i-1].HeaderNonce + 1`?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+			log.Debug("ExecutionResultInclusionEstimator: non-monotonic HeaderNonce detected",
+				"currentHeaderNonce", executionResultMeta.HeaderNonce,
+				"previousHeaderNonce", pending[i-1].HeaderNonce,
+				"currentHeaderTimeMs", executionResultMeta.HeaderTimeMs,
+				"previousHeaderTimeMs", pending[i-1].HeaderTimeMs,
+			)
+			return i
```
sstanculeanu: i think we should return i-1 here, as the current i won't be included

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		// Check for monotonicity in time
+		if i > 0 && executionResultMeta.HeaderTimeMs < pending[i-1].HeaderTimeMs {
+			log.Debug("ExecutionResultInclusionEstimator: non-monotonic HeaderTimeMs detected",
+				"currentHeaderTimeMs", executionResultMeta.HeaderTimeMs,
+				"previousHeaderTimeMs", pending[i-1].HeaderTimeMs,
+			)
+			return i
```
sstanculeanu: same note here, although I'm not sure this check is needed

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		// Check for time before genesis time
+		if executionResultMeta.HeaderTimeMs < erie.cfg.GenesisTimeMs {
+			log.Debug("ExecutionResultInclusionEstimator: HeaderTimeMs before genesis detected",
+				"headerNonce", executionResultMeta.HeaderNonce,
+				"headerTimeMs", executionResultMeta.HeaderTimeMs,
+				"genesisTimeMs", erie.cfg.GenesisTimeMs,
+			)
+			return i
```
sstanculeanu: same note here

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		// Check for time before last notarised
+		if lastNotarised != nil && executionResultMeta.HeaderTimeMs < lastNotarised.HeaderTimeMs {
+			log.Debug("ExecutionResultInclusionEstimator: HeaderTimeMs before last notarised detected",
+				"headerNonce", executionResultMeta.HeaderNonce,
+				"headerTimeMs", executionResultMeta.HeaderTimeMs,
+				"lastNotarisedTimeMs", lastNotarised.HeaderTimeMs,
+			)
+			return i
```
sstanculeanu: same note here

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		// Check for results in the future
+		if convertMsToNs(executionResultMeta.HeaderTimeMs) > currentHdrTsNs {
+			log.Debug("ExecutionResultInclusionEstimator: HeaderTimeMs in the future detected",
+				"headerNonce", executionResultMeta.HeaderNonce,
+				"headerTimeMs", executionResultMeta.HeaderTimeMs,
+				"currentHdrTsNs", currentHdrTsNs,
+			)
+			return i
```
sstanculeanu, line 77: same and valid on all further error cases

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in currentEstimatedTime",
+				"currentEstimatedTime", currentEstimatedTime,
+				"gasUsed", executionResultMeta.GasUsed,
+				"tGas", erie.tGas,
+			)
+			return i
+		}
+
+		estimatedTime, overflow := bits.Add64(estimatedTime, currentEstimatedTime, 0)
```
sstanculeanu:
```suggestion
		estimatedTime, overflow = bits.Add64(estimatedTime, currentEstimatedTime, 0)
```

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in block tranzactions time estimation",
```
sstanculeanu:
```suggestion
			log.Debug("ExecutionResultInclusionEstimator: overflow detected in block transactions time estimation",
```

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		tDone, overflow := bits.Add64(tBase, estimatedTimeWithMargin, 0)
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in total estimated time",
+				"tBase", tBase,
+				"estimatedTimeWithMargin", estimatedTimeWithMargin)
+			return i
+		}
+
+		// check for time cap reached, cannot include current pending item or anything after
+		if tDone >= currentHdrTsNs {
```
sstanculeanu:
```suggestion
		if tDone > currentHdrTsNs {
```
?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+			log.Debug("ExecutionResultInclusionEstimator: estimated time exceeds current header timestamp",
+				"tDone", tDone,
+				"currentHdrTsNs", currentHdrTsNs)
+			return i
+		}
+
+		// check for number of results cap reached, including current pending item
+		if erie.cfg.MaxResultsPerBlock != 0 && uint64(i+1) >= erie.cfg.MaxResultsPerBlock {
```
sstanculeanu, line 125: if erie.cfg.MaxResultsPerBlock != 0 should be a check made on constructor and return error

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+			log.Debug("ExecutionResultInclusionEstimator: reached MaxResultsPerBlock cap",
+				"maxResultsPerBlock", erie.cfg.MaxResultsPerBlock,
+				"currentIndex", i+1)
+			return i + 1
```
sstanculeanu, line 129: shouldn't this be still i-1?

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -0,0 +1,297 @@
...
+	"fmt"
+	"math"
+	"testing"
+
+	logger "github.com/multiversx/mx-chain-logger-go"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEstimatorCreation(t *testing.T) {
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu: should not be needed.. also t.Paralel can be added on each individual t.Run

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -0,0 +1,297 @@
...
+		cfg := Config{SafetyMargin: 110}
+		erie := NewExecutionResultInclusionEstimator(cfg)
+		erie.SetTimePerGasUnit(0)
+		require.NotNil(t, erie, "NewExecutionResultInclusionEstimator should not return nil")
+		require.Equal(t, uint64(1), erie.tGas, "tGas should default to 1 ns per gas unit when set to zero")
+	})
+}
+
+func TestDecide(t *testing.T) {
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu: not needed

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+		got := erie.Decide(cfg, lastNotarised, pending, currentHdrTsNs)
+		if got != wantAllowed {
+			t.Errorf("Decide() = %d, want %d", got, wantAllowed)
+		}
+	})
+}
+
+func TestOverflowProtection(t *testing.T) {
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu: not needed

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+		num_accepted := erie.Decide(cfg, nil, pending, currentTime)
+		t.Log("num_accepted:", num_accepted)
+		require.Equal(t, 0, num_accepted, "should not accept any results due to overflow")
+	})
+
+}
+
+func TestDecide_EdgeCases(t *testing.T) {
+	_ = logger.SetLogLevel("*:DEBUG")
```
sstanculeanu: not needed

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: since we are returning the number of acceptable results, and i starts from 0, the current result is not included if we return i, in order for it to be included we would return i+1

process/estimator/executionResultInclusionEstimator.go
mradian1: as before, returning i doesn't include current record. will be removed if considered unnecessary

process/estimator/executionResultInclusionEstimator.go
mradian1: removed

process/estimator/executionResultInclusionEstimator.go
mradian1: in this case we are keeping the current result, but not taking any further ones, therefore i+1, since i starts from 0, so if we keep the element at index i=0, we return 1 etc.

process/estimator/executionResultInclusionEstimator.go
mradian1: returning i doesn't include current execution result: for example if i=0 and we want to reject it, returning i means 0 allowed execution results etc

process/estimator/executionResultInclusionEstimator.go
mradian1: explained earlier

process/estimator/executionResultInclusionEstimator.go
mradian1: explained earlier

process/estimator/executionResultInclusionEstimator.go
mradian1: it depends on whether or not we include results for blocks without transactions. It should be discussed further, probably

process/estimator/executionResultInclusionEstimator.go
mradian1: changed

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,184 @@
...
+		estimatedTime, overflow := bits.Add64(estimatedTime, currentEstimatedTime, 0)
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in block tranzactions time estimation",
+				"estimatedTime", estimatedTime,
+				"currentEstimatedTime", currentEstimatedTime)
+			return i
+		}
+
+		// Apply safety margin
+		overflow, estimatedTimeWithMargin := bits.Mul64(estimatedTime, erie.cfg.SafetyMargin/100)
```
AdoAdoAdo: safety margin is not correctly applied, as there is an integer division (e.g. 110/100 = 1)

process/estimator/executionResultInclusionEstimator.go
mradian1: this is temporary. it will be correctly implemented once the algorithm is approved upon

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: removed

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: A value of 0 for MaxResultsPerBlock means unlimited, so this check shouldn't be made. Changed the comment to reflect this

process/estimator/executionResultInclusionEstimator.go
AdoAdoAdo: here meta means metadata, not necessarily the metablock execution result.
I think we can rename to ExecutionResultMetaData to make this less confusing.

I would suggest to have this for now as a self-contained component (so we can keep it as it is for now) and adapt if needed upon integration (on the execution results we are missing the original block timestamp, that one needs to be taken from the block so we can use this struct as an intermediate DTO)

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+// NewExecutionResultInclusionEstimator returns a new instance of EIE
+func NewExecutionResultInclusionEstimator(cfg Config) *ExecutionResultInclusionEstimator {
+	return &ExecutionResultInclusionEstimator{
+		cfg:  cfg,
+		tGas: uint64(1),
+	}
+}
+
+// SetTimePerGasUnit sets the time per gas unit on minimum‑spec hardware.
+func (erie *ExecutionResultInclusionEstimator) SetTimePerGasUnit(tGas uint64) {
```
AdoAdoAdo: I think we can remove SetTimePerGasUnit, and have tGas in the config on init, or as a const (I don't think we will change it any time soon), then there will also be no need to protect access to it with a mutex as it will not change.

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+		tBase = convertMsToNs(erie.cfg.GenesisTimeMs)
+	} else {
+		tBase = convertMsToNs(lastNotarised.HeaderTimeMs)
+	}
+
+	// accumulated execution time in ns (1 gas = 1ns)
+	estimatedTime := uint64(0)
+	for i, executionResultMeta := range pending {
+		// Check for nonce monotonicity
+		// TODO confirm if we are including execution results for empty blocks, in which case we should check strict continuity
```
AdoAdoAdo: yes, there will be also execution results for empty blocks, so we can check for strict continuity

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+				"currentHeaderNonce", executionResultMeta.HeaderNonce,
+				"previousHeaderNonce", pending[i-1].HeaderNonce,
+				"currentHeaderTimeMs", executionResultMeta.HeaderTimeMs,
+				"previousHeaderTimeMs", pending[i-1].HeaderTimeMs,
+			)
+			return i
+		}
+		// Check for monotonicity in time
+		// TODO confirm if we are keeping this check
+		if i > 0 && executionResultMeta.HeaderTimeMs < pending[i-1].HeaderTimeMs {
```
AdoAdoAdo: yes consecutive header need to heave increasing timestamps.

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+			"newValue", tGas)
+		erie.tGas = tGas
+	}
+}
+
+// Decide returns the prefix of `pending` that may be inserted into the block currently being built / verified.
+// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]`and embeds them.
+func (erie *ExecutionResultInclusionEstimator) Decide(lastNotarised *ExecutionResultMeta,
+	pending []ExecutionResultMeta,
+	currentHdrTsNs uint64,
```
AdoAdoAdo: I think it would be better to have this as ms. The header timestamp will be in milli seconds, and I think it would be better that it is simply passed and all further conversions needed are done inside the estimation component.
(we need to update also the specs)

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+	// lastNotarised is nil if genesis.
+	if lastNotarised == nil {
+		tBase = convertMsToNs(erie.cfg.GenesisTimeMs)
+	} else {
+		tBase = convertMsToNs(lastNotarised.HeaderTimeMs)
+	}
+
+	// accumulated execution time in ns (1 gas = 1ns)
+	estimatedTime := uint64(0)
+	for i, executionResultMeta := range pending {
```
AdoAdoAdo, line 71: because of all the checks, this method became too large.
I would suggest to extract all "sanity checks" before doing any computation of time using the gas into a separate method.

Maybe everything from L81 to L128 could be extracted into a sanity check method, then have another method that does the time computation based on gas and compares it against current header timestamp to see if it is allowed?

process/estimator/executionResultInclusionEstimator.go
mradian1: defined as constant

process/estimator/executionResultInclusionEstimator.go
mradian1: removed

process/estimator/executionResultInclusionEstimator.go
mradian1: removed method

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+	if erie.tGas != tGas {
+		log.Debug("ExecutionResultInclusionEstimator: SetTimePerGasUnit called",
+			"oldValue", erie.tGas,
+			"newValue", tGas)
+		erie.tGas = tGas
+	}
+}
+
+// Decide returns the prefix of `pending` that may be inserted into the block currently being built / verified.
+// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]`and embeds them.
```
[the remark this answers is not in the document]
mradian1: changed

process/estimator/executionResultInclusionEstimator.go
mradian1: changed

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,189 @@
...
+		estimatedTime, overflow = bits.Add64(estimatedTime, currentEstimatedTime, 0)
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in block transactions time estimation",
+				"estimatedTime", estimatedTime,
+				"currentEstimatedTime", currentEstimatedTime)
+			return i
+		}
+
+		// Apply safety margin
+		overflow, estimatedTimeWithMargin := bits.Mul64(estimatedTime, erie.cfg.SafetyMargin/100)
```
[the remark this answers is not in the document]
mradian1: changed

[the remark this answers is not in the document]
mradian1: marginFactor would be 1

process/estimator/executionResultInclusionEstimator.go
mradian1: done, added checkSanity

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: changed

process/estimator/executionResultInclusionEstimator.go
mradian1: removed method

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: kept the check

process/estimator/executionResultInclusionEstimator.go
mradian1: moved to config package

config/config.go
```diff
@@ -25,6 +25,13 @@ type ProofsPoolConfig struct {
 	BucketSize        int
 }
 
+// ExecutionResultInclusionEstimatorConfig will map the EIE configuration - supplied at construction, read‑only thereafter.
+// TODO add also max estimated block gas capacity
```
miiu96, line 44: fix todo ?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+var log = logger.GetOrCreate("process/executionResultInclusionEstimator")
+
+// Time per gas unit on minimum‑spec hardware (1 gas = 1ns)
+var tGas = uint64(1)
+
+// ExecutionResultMetaData is a lightweight summary EIE requires.
+type ExecutionResultMetaData struct {
+	HeaderHash   [32]byte // Link to full header in DB / cache
```
miiu96: maybe use []byte instead of [32], since we generally don’t define fixed byte array sizes in this project.

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+// safely embedded in the block that is being produced / verified.
+type ExecutionResultInclusionEstimator struct {
+	cfg           config.ExecutionResultInclusionEstimatorConfig // immutable after construction
+	tGas          uint64                                         // time per gas unit on minimum‑spec hardware - 1 ns per gas unit
+	GenesisTimeMs uint64                                         // required if lastNotarised == nil
+	// TODO add also max estimated block gas capacity - used gas must be lower than this
+}
+
+// NewExecutionResultInclusionEstimator returns a new instance of EIE
+func NewExecutionResultInclusionEstimator(cfg config.ExecutionResultInclusionEstimatorConfig, GenesisTimeMs uint64) *ExecutionResultInclusionEstimator {
```
miiu96: GenesisTimeMs --> genesisTimeMs or

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+	HeaderHash   [32]byte // Link to full header in DB / cache
+	HeaderNonce  uint64   // Monotonic within shard
+	HeaderTimeMs uint64   // Milliseconds since Unix epoch, from header
+	GasUsed      uint64   // Units actually consumed (post‑execution)
+}
+
+// ExecutionResultInclusionEstimator (EIE) is a deterministic component shipped with the MultiversX *Supernova*
+// node. It determines, at proposal‑time and at validation‑time, whether one or more pending execution results can be
+// safely embedded in the block that is being produced / verified.
+type ExecutionResultInclusionEstimator struct {
```
miiu96, line 26: do we need to export this structure, or can it remain unexported?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+	cfg           config.ExecutionResultInclusionEstimatorConfig // immutable after construction
+	tGas          uint64                                         // time per gas unit on minimum‑spec hardware - 1 ns per gas unit
+	GenesisTimeMs uint64                                         // required if lastNotarised == nil
```
miiu96: why this field is exported ?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+				"currentIndex", i+1)
+			return i + 1
+		}
+	}
+
+	// If we reach here, all pending items are safe to include
+	return len(pending)
+}
+
+func (erie ExecutionResultInclusionEstimator) checkSanity(currentExecutionResultMeta ExecutionResultMetaData,
```
miiu96: use pointer receiver in all the places 
`(erie *ExecutionResultInclusionEstimator)`

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,191 @@
...
+// ExecutionResultInclusionEstimator (EIE) is a deterministic component shipped with the MultiversX *Supernova*
+// node. It determines, at proposal‑time and at validation‑time, whether one or more pending execution results can be
+// safely embedded in the block that is being produced / verified.
+type ExecutionResultInclusionEstimator struct {
+	cfg           config.ExecutionResultInclusionEstimatorConfig // immutable after construction
+	tGas          uint64                                         // time per gas unit on minimum‑spec hardware - 1 ns per gas unit
+	GenesisTimeMs uint64                                         // required if lastNotarised == nil
+	// TODO add also max estimated block gas capacity - used gas must be lower than this
+}
+
```
miiu96, line 32: define method 
func (erie *ExecutionResultInclusionEstimator) IsInterfaceNil() bool {
	return erie == nil
}

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: changed

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: done

config/config.go
mradian1: it will be solved in a future pr

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -0,0 +1,196 @@
...
+	} else {
+		tBase = convertMsToNs(lastNotarised.HeaderTimeMs)
+	}
+
+	currentHdrTsNs := convertMsToNs(currentHdrTsMs)
+
+	// accumulated execution time in ns (1 gas = 1ns)
+	estimatedTime := uint64(0)
+	for i, executionResultMeta := range pending {
+		var previousExecutionResultMeta *ExecutionResultMetaData
```
AdoAdoAdo: can you move this declaration outside of the loop, to avoid multiple allocations?

mradian1: moved declaration

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -0,0 +1,313 @@
...
+		t.Parallel()
+		cfg := config.ExecutionResultInclusionEstimatorConfig{SafetyMargin: 110}
+		erie := NewExecutionResultInclusionEstimator(cfg, 500)
+		lastNotarised := &ExecutionResultMetaData{HeaderTimeMs: 1000}
+		pending := []ExecutionResultMetaData{
+			{HeaderNonce: 1, HeaderTimeMs: 1010, GasUsed: 100_000_000},
+			{HeaderNonce: 2, HeaderTimeMs: 1011, GasUsed: 100_000_000},
+			{HeaderNonce: 3, HeaderTimeMs: 1012, GasUsed: 100_000_000},
+		}
+		currentHdrTsMs := uint64(1010 * 110 / 100)
```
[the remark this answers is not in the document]
mradian1: changed

process/estimator/executionResultInclusionEstimator_test.go
mradian1: done

process/estimator/executionResultInclusionEstimator_test.go
mradian1: removed

process/estimator/executionResultInclusionEstimator_test.go
mradian1: removed

process/estimator/executionResultInclusionEstimator_test.go
mradian1: removed

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -3,122 +3,77 @@ package estimator
...
 }
 
 // Decide returns the prefix of `pending` that may be inserted into the block currently being built / verified.
-// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]`and embeds them.
-func (erie *ExecutionResultInclusionEstimator) Decide(cfg Config,
-	lastNotarised *ExecutionResultMeta,
-	pending []ExecutionResultMeta,
-	currentHdrTsNs uint64) (allowed int) {
+// Return value: `allowed` is the count of leading entries in `pending` deemed safe. The caller slices `pending[:allowed]` and embeds them.
+func (erie *ExecutionResultInclusionEstimator) Decide(lastNotarised *ExecutionResultMetaData,
```
sstanculeanu, line 49:
```suggestion
func (erie *ExecutionResultInclusionEstimator) Decide(
	lastNotarised *ExecutionResultMetaData,
```

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -178,6 +134,61 @@ func (erie *ExecutionResultInclusionEstimator) Decide(cfg Config,
 	return len(pending)
 }
 
+func (erie *ExecutionResultInclusionEstimator) checkSanity(currentExecutionResultMeta ExecutionResultMetaData,
```
sstanculeanu, line 137:
```suggestion
func (erie *ExecutionResultInclusionEstimator) checkSanity(
	currentExecutionResultMeta ExecutionResultMetaData,
```

PR #7189, opened 2025-08-12
Refactor handling of "account not found". Simplify interface of "selection session"

andreibancioiu:
## Reasoning behind the pull request
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

Based on the [Contributing Guidelines](https://github.com/multiversx/mx-chain-go/blob/master/.github/CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:
- was the PR targeted to the correct branch?
- if this is a larger feature that probably needs more than one PR, is there a `feat` branch created?
- if this is a `feat` branch merging, do all satellite projects have a proper tag inside `go.mod`?

PR #7206, opened 2025-08-25
Supernova TxPool: On "OnProposedBlock", pass an accounts provider (not the selection session)

andreibancioiu:
This is the continuation of https://github.com/multiversx/mx-chain-go/pull/7189.

- The caller of `TxCache.OnProposedBlock()` does not manage transaction selection, or a selection session for that matter. Thus, we have to pass something else to the `OnProposedBlock()` function, something able to provide the downstream pieces with nonces and balances of accounts.
- A `selectionSession` requires a `transactionProcessor` when constructed - required in the implementation of `selectionSession.IsIncorrectlyGuarded()`.
  
## Proposed changes
- Extract most of the logic from `process/block/preprocess/selectionSession.go` to `state/accountsEphemeralProvider.go`.
- Refactor `selectionSession` to rely on the new `accountsEphemeralProvider` (no change in logic).
- Adjust `TxCache.OnProposedBlock()` to receive a `nonce and balance provider`, instead of a `selection session`. This provider will actually be an `accountsEphemeralProvider` at runtime.
- Moved tests, added some more.
- Bits of adjacent refactoring.

## Testing procedure
- Standard testing.

## Pre-requisites

### DIFF
--- a/dataRetriever/txpool/shardedTxPool_test.go
+++ b/dataRetriever/txpool/shardedTxPool_test.go
@@ -275,6 +275,7 @@ func Test_RemoveSetOfDataFromPool(t *testing.T) {
 }
 
 func TestCleanupSelfShardTxCache_NilMempool(t *testing.T) {
+	t.Parallel()
 	t.Run("with nil self shard pool", func(t *testing.T) {
 		poolAsInterface, _ := newTxPoolToTest()
 		txPool := poolAsInterface.(*shardedTxPool)
@@ -289,6 +290,39 @@ func TestCleanupSelfShardTxCache_NilMempool(t *testing.T) {
 
 		ok := txPool.CleanupSelfShardTxCache(session, 7, math.MaxInt, cleanupLoopMaximumDuration)
 		t.Logf("shardedTxPool.CleanupSelfShardTxCache() starting cleanup %d", txPool.selfShardID)
+		require.True(t, ok)
+	})
+}
+
+func Test_Parallel_CleanupSelfShardTxCache(t *testing.T) {
+	t.Parallel()
+	t.Run("with lower nonces", func(t *testing.T) {
+		t.Parallel()
+		poolAsInterface, _ := newTxPoolToTest()
+		pool := poolAsInterface.(*shardedTxPool)
+		session := txcachemocks.NewSelectionSessionMock()
+		session.SetNonce([]byte("alice"), 2)
+		session.SetNonce([]byte("bob"), 42)
+		session.SetNonce([]byte("carol"), 7)
+
+		// One lower nonce
+		pool.AddData([]byte("hash-alice-1"), createTx("alice", 1), 0, "0")
+		pool.AddData([]byte("hash-alice-2"), createTx("alice", 2), 0, "0")
+		pool.AddData([]byte("hash-alice-3"), createTx("alice", 3), 0, "0")
+
+		// A few with lower nonce
+		pool.AddData([]byte("hash-bob-40"), createTx("bob", 40), 0, "0")
+		pool.AddData([]byte("hash-bob-41"), createTx("bob", 41), 0, "0")
+		pool.AddData([]byte("hash-bob-42"), createTx("bob", 42), 0, "0")
+
+		// Good
+		pool.AddData([]byte("hash-carol-7"), createTx("carol", 7), 0, "0")
+		pool.AddData([]byte("hash-carol-8"), createTx("carol", 8), 0, "0")
+
+		cleanupLoopMaximumDuration := time.Millisecond * 100
+
+		ok := pool.CleanupSelfShardTxCache(session, 7, math.MaxInt, cleanupLoopMaximumDuration)
+		t.Logf("shardedTxPool.CleanupSelfShardTxCache() starting cleanup %d", pool.selfShardID)
 		require.True(t, ok)
 	})
 }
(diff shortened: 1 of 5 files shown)


integrationTests/chainSimulator/mempool/mempool_test.go
```diff
@@ -507,16 +508,19 @@ func Test_Selection_ShouldNotSelectSameTransactionsWithSameSender(t *testing.T)
...
 		"receiver": {
-			balance: big.NewInt(0),
-			nonce:   0,
+			Balance: big.NewInt(0),
+			Nonce:   0,
 		},
-	})
+	}
+
+	selectionSession := txcachemocks.NewSelectionSessionMockWithAccounts(accounts)
```
andreibancioiu, line 522: For tests, we ground the _selection session_ and the _nonce and balance provider_ on the same accounts adapter.

integrationTests/chainSimulator/mempool/mempool_test.go
```diff
@@ -729,7 +736,7 @@ func Test_Selection_ShouldNotSelectSameTransactionsWithDifferentSenders(t *testi
 			PrevHash: []byte("blockHash0"),
 			RootHash: []byte(fmt.Sprintf("rootHash%d", 0)),
 		},
-		selectionSession,
+		accountsProvider,
```
andreibancioiu, line 739: `OnProposedBlock()` now receives an accounts provider, instead of a selection session.

testscommon/txcachemocks/accountNonceAndBalanceProviderMock.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"math/big"
+	"sync"
+
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	stateMock "github.com/multiversx/mx-chain-go/testscommon/state"
+)
+
+// AccountNonceAndBalanceProviderMock -
+type AccountNonceAndBalanceProviderMock struct {
```
andreibancioiu, line 12: This mock is extremely similar to `SelectionSessionMock`.

process/block/preprocess/selectionSession_test.go
```diff
@@ -92,54 +92,6 @@ func TestSelectionSession_GetAccountNonceAndBalance(t *testing.T) {
 	require.False(t, existing)
 }
 
-func TestSelectionSession_getCachedUserAccount(t *testing.T) {
```
andreibancioiu, line 95: Tests were moved to `state/accountsEphemeralProvider_test.go`.

process/block/preprocess/selectionSession.go
```diff
@@ -12,13 +12,8 @@ import (
...
 
 type selectionSession struct {
-	accountsAdapter       state.AccountsAdapter
 	transactionsProcessor process.TransactionProcessor
-
-	// Cache of accounts, held in the scope of a single selection session.
-	// Not concurrency-safe, but never accessed concurrently.
-	// Entries for new (unknown) accounts are "nil" values.
-	ephemeralAccountsCache map[string]state.UserAccountHandler
+	accountsProvider      *state.AccountsEphemeralProvider
```
andreibancioiu, line 16: `selectionSession` was heavily refactored to rely on the new `AccountsEphemeralProvider`. No change in logic / functionality :crossed_fingers:

state/accountsEphemeralProvider.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/multiversx/mx-chain-core-go/core/check"
+)
+
+// AccountsEphemeralProvider acts as an "ephemeral" provider for accounts.
+// Make sure such a provider isn't reused among multiple transactions selections or multiple processing (of blocks) phases,
+// since it contains a **never-invalidating cache** (deliberate, by design).
+// Create it privately (don't receive it in constructors), use it privately, make sure it's forgotten afterwards. Don't keep lasting references to it.
+// Note: this structure is exported on purpose (less ceremonious code where it's being used, no extra interfaces needed).
+type AccountsEphemeralProvider struct {
```
andreibancioiu, line 14: Most of the logic extracted from `process/block/preprocess/selectionSession.go`.

process/block/preprocess/selectionSession.go
```diff
@@ -29,81 +24,36 @@ type ArgsSelectionSession struct {
...
 func NewSelectionSession(args ArgsSelectionSession) (*selectionSession, error) {
-	if check.IfNil(args.AccountsAdapter) {
-		return nil, process.ErrNilAccountsAdapter
-	}
 	if check.IfNil(args.TransactionsProcessor) {
 		return nil, process.ErrNilTxProcessor
 	}
 
+	// Provider is not concurrency-safe, but it's never accessed concurrently.
+	accountsProvider, err := state.NewAccountsEphemeralProvider(args.AccountsAdapter)
```
ssd04, line 32: how do we instantiate `NewSelectionSession`? if possible i would pass `accountsProvider` as argument, it's easier to test and better to have all initialization in factories

process/block/preprocess/selectionSession.go
```diff
@@ -29,81 +24,36 @@ type ArgsSelectionSession struct {
...
-		return nil, process.ErrWrongTypeAssertion
-	}
-
-	return userAccount, nil
+	return session.accountsProvider.GetAccountNonceAndBalance(address)
 }
 
 // IsIncorrectlyGuarded checks if a transaction is incorrectly guarded (not executable).
 // Will be called by mempool during transaction selection.
+// See: MX-16157, MX-16772.
```
ssd04: suggestion to not have private IDs here, they are not useful for others

state/accountsEphemeralProvider.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if check.IfNil(account) {
+		// New (unknown) account.
+		return 0, big.NewInt(0), false, nil
+	}
+
+	return account.GetNonce(), account.GetBalance(), true, nil
+}
+
+// GetUserAccount returns the user account, as found on blockchain. If missing (account not found), nil is returned (with no error).
```
ssd04, line 52: why not returning error on missing account?

process/block/preprocess/selectionSession.go
ssd04: ok, i see now that it was intended like this

andreibancioiu: Indeed, the accounts provider it's more like a "ephemeral accounts holder" - instantiated on the spot, where needed. Furthermore, the "selection session" is also ephemeral:
 - https://github.com/multiversx/mx-chain-go/blob/v1.10.6/process/block/preprocess/transactions.go#L1414

That is, the selection session object has a short lifetime, as well, which doesn't cross the "round boundary".

This accounts provider is (and will be) used when:
 - creating the selection session, within its constructor
 - calling the "OnProposedBlock" callback of the TxCache (who needs to inspect the _current_ nonce and balance of accounts).

process/block/preprocess/selectionSession.go
andreibancioiu: Removed.

state/accountsEphemeralProvider.go
andreibancioiu: Indeed, `ErrAccNotFound` is encapsulated, and treated as a non-error. Other errors are propagated.

Primary reason:
 - all components that are clients of the `AccountsEphemeralProvider` do not want (need) to inspect and compare returned errors. Generally speaking, they'd want to avoid handling such detailed information.
 - if we were to propagate `ErrAccNotFound`, as well, then all client code would have to do comparison against `ErrAccNotFound`.
 - in the `AccountsEphemeralProvider`, we want to cache the information "this account is missing now". Propagating the error would imply a more complex caching mechanism, or no caching at all for this case. Though, in practice, caching on this case should be useful (think of a new user sending 100 transactions through a relayer).

On the other hand, yes, the only client code of `AccountsEphemeralProvider` which needs to call `GetUserAccount` is `SelectionSession`. So, the first two points from above are not very strong.

All good to have it this way?

ssd04: yes, all good if it's used only in selection session

PR #7222, opened 2025-09-01
Tx coordinator requests

AdoAdoAdo:
- This PR refactors the transaction coordinator by extracting block data requesting logic into a separate component called BlockDataRequester. 
- The main purpose is to isolate data requesting functionality so it can be reused in different contexts, particularly for proposal processing and validation.
  
## Proposed changes
- Create a new BlockDataRequester component that handles all data requesting logic
- Integrate the new component into the existing TransactionCoordinator to maintain backward compatibility
- Move utility functions to a common location for shared use

## Testing procedure
- check that sync and consensus processes work as before
- 

## Pre-requisites

### DIFF
--- a/process/common_test.go
+++ b/process/common_test.go
@@ -2579,3 +2579,21 @@ func Test_SetBaseExecutionResult(t *testing.T) {
 		require.True(t, called)
 	})
 }
+
+func TestTransactionCoordinator_SeparateBody(t *testing.T) {
+	t.Parallel()
+
+	body := &block.Body{}
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+
+	separated := process.SeparateBodyByType(body)
+	assert.Equal(t, 2, len(separated))
+	assert.Equal(t, 3, len(separated[block.TxBlock].MiniBlocks))
+	assert.Equal(t, 4, len(separated[block.SmartContractResultBlock].MiniBlocks))
+}
--- a/process/common_test.go
+++ b/process/common_test.go
@@ -2591,9 +2591,17 @@ func TestTransactionCoordinator_SeparateBody(t *testing.T) {
 	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
 	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
 	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.InvalidBlock}) // invalid blocks will go into the TxBlock bucket
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.PeerBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.PeerBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.RewardsBlock})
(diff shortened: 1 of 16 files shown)


process/common_test.go
```diff
@@ -2579,3 +2579,21 @@ func Test_SetBaseExecutionResult(t *testing.T) {
...
+	t.Parallel()
+
+	body := &block.Body{}
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.TxBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
+	body.MiniBlocks = append(body.MiniBlocks, &block.MiniBlock{Type: block.SmartContractResultBlock})
```
ssd04, line 2593: add test case also with `InvalidBlock` type?

process/coordinator/blockDataRequest.go
```diff
@@ -0,0 +1,282 @@
...
+type BlockDataRequestArgs struct {
+	RequestHandler      process.RequestHandler
+	MiniBlockPool       storage.Cacher
+	PreProcessors       process.PreProcessorsContainer
+	ShardCoordinator    process.ShardCoordinator
+	EnableEpochsHandler common.EnableEpochsHandler
+}
+
+// BlockDataRequest implements the BlockDataRequester interface
+type BlockDataRequest struct {
```
ssd04, line 29: does this needs to be exported?

process/common_test.go
ssd04: i see now that it's old code, it can be updated later on in another PR

process/coordinator/blockDataRequest.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/multiversx/mx-chain-core-go/data"
+	"github.com/multiversx/mx-chain-core-go/data/block"
+
+	"github.com/multiversx/mx-chain-go/common"
+	"github.com/multiversx/mx-chain-go/process"
+	"github.com/multiversx/mx-chain-go/storage"
+	"github.com/multiversx/mx-chain-go/storage/cache"
+)
+
+// BlockDataRequestArgs holds the arguments needed to create a CoordinatorRequest
```
BeniaminDrasovean: ...to create a `BlockDataRequest`

process/coordinator/process.go
```diff
@@ -981,7 +823,7 @@ func (tc *transactionCoordinator) CreateBlockStarted() {
 	tc.transactionsLogProcessor.Clean()
 }
 
-func (tc *transactionCoordinator) getPreProcessor(blockType block.Type) process.PreProcessor {
+func (tc *transactionCoordinator) GetPreProcessor(blockType block.Type) process.PreProcessor {
```
BeniaminDrasovean: Why did you export this? As far as I can tell, it is used only in this package. Is this for a future case?

process/coordinator/process_test.go
```diff
@@ -319,17 +352,6 @@ func TestNewTransactionCoordinator_NilDataPool(t *testing.T) {
 	assert.Equal(t, process.ErrNilMiniBlockPool, err)
 }
 
-func TestNewTransactionCoordinator_NilRequestHandler(t *testing.T) {
```
BeniaminDrasovean, line 322: Replace this test with a nil check for `blockDataRequester`

process/coordinator/process_test.go
```diff
@@ -4539,82 +3956,30 @@ func TestTransactionCoordinator_getIndexesOfLastTxProcessed(t *testing.T) {
...
-
-	wg.Wait()
-
-	mutMap.RLock()
-	assert.Equal(t, 3, mapRequestedMiniBlocksPerShard[0])
-	assert.Equal(t, 2, mapRequestedMiniBlocksPerShard[1])
-	assert.Equal(t, 1, mapRequestedMiniBlocksPerShard[2])
-	assert.Equal(t, 2, numTxsRequested)
-	mutMap.RUnlock()
+func createDefaultTxCoordinatorArgs() ArgTransactionCoordinator {
```
BeniaminDrasovean, line 3970: :+1:

process/common_test.go
AdoAdoAdo: added the missing types

process/coordinator/blockDataRequest.go
AdoAdoAdo: I think instead of unexported struct type, it would be more useful to use exported types but with unexported fields. This will avoid the warnings of unexported return types for exported methods and still keep the internals unexported (and it's easier to construct interfaces or change them without breaking the point of use)

process/coordinator/blockDataRequest.go
AdoAdoAdo: done

process/coordinator/process.go
AdoAdoAdo: left over after a refactor, reverted

process/coordinator/process_test.go
AdoAdoAdo: done

PR #7237, opened 2025-09-09
Supernova TxPool: Propagate error from SelectTransactions

danielradu10:
- Currently, an error that might appear in the SelectTransactions method i**s only logged** and it is not propagated upwards, which means that the caller component is not aware of a possible error. This issue is fixed in this PR.
  
## Proposed changes
- propagate the error upwards from **SelectTransactions**.
- check for an error in the **computeSortedTxs** method and propagate it.
- bug fix in an integration test.
- added unit tests.

## Pre-requisites

### DIFF
--- a/process/block/preprocess/transactions.go
+++ b/process/block/preprocess/transactions.go
@@ -1468,6 +1468,7 @@ func (txs *transactions) computeSortedTxs(
 	blockchainInfo := holders.NewBlockchainInfo(nil, nil, 0)
 	sortedTxs, _, err := txCache.SelectTransactions(session, selectionOptions, blockchainInfo)
 	if err != nil {
+		// TODO re-brainstorm if this error should be propagated or just logged
 		return nil, nil, err
 	}
 


integrationTests/chainSimulator/mempool/mempool_test.go
```diff
@@ -1250,9 +1269,10 @@ func Test_Selection_ProposeEmptyBlocksAndExecutedBlockNotification(t *testing.T)
 	})
 	require.Nil(t, err)
 
-	blockchainInfo = holders.NewBlockchainInfo([]byte("blockHash1"), []byte("blockHash4"), 5)
+	blockchainInfo = holders.NewBlockchainInfo([]byte("blockHash3"), []byte("blockHash4"), 5)
```
danielradu10, line 1272: Checking the returned error from **SelectTransactions** helped in discovering a bug in this test:
 the **latestExecutedBlockHash** parameter of the **blockchainInfo** was not updated after the **OnExecuted** notification. 

Even though the returned number of selected transactions was as expected (zero), **it was because of an error.**
 However, we want here to be sure the transactions **are not re-selected**, not that we don't select the transactions because of an error.

txcache/txCache.go
```diff
@@ -112,10 +112,10 @@ func (cache *TxCache) SelectTransactions(
 	session SelectionSession,
 	options common.TxSelectionOptions,
 	blockchainInfo common.BlockchainInfo,
-) ([]*WrappedTransaction, uint64) {
+) ([]*WrappedTransaction, uint64, error) {
```
danielradu10, line 115: this is the important change of the PR.

process/block/preprocess/transactions.go
```diff
@@ -1466,7 +1466,11 @@ func (txs *transactions) computeSortedTxs(
 	)
 
 	blockchainInfo := holders.NewBlockchainInfo(nil, nil, 0)
-	sortedTxs, _ := txCache.SelectTransactions(session, selectionOptions, blockchainInfo)
+	sortedTxs, _, err := txCache.SelectTransactions(session, selectionOptions, blockchainInfo)
+	if err != nil {
+		return nil, nil, err
```
danielradu10, line 1472: should we return it here or should we only log it?

andreibancioiu: I think in Supernova, this should be an error, indeed. Though, please add a TODO to re-brainstorm this, once processing components are ready etc.

integrationTests/chainSimulator/mempool/mempool_test.go
andreibancioiu: Very good!

andreibancioiu (approved):

process/block/preprocess/transactions.go
danielradu10: added

andreibancioiu (approved):

sstanculeanu (approved):

PR #7261, opened 2025-09-16
Supernova TxPool: Improvements-16-09

danielradu10:
- Another round of improvements for TxPool.
  
## Proposed changes

- renamings, reordering
- removed some duplicated code
- fixes, logs, comments

## Pre-requisites


txcache/virtualAccountRecord.go
```diff
@@ -28,43 +28,54 @@ func (virtualRecord *virtualAccountRecord) updateVirtualRecord(breadcrumb *accou
...
 
-func (virtualRecord *virtualAccountRecord) updateInitialNonce(breadcrumb *accountBreadcrumb) {
-	// this solves the next scenario:
-	// two different breadcrumbs for the same sender
-	// in the first tracked block, the last tx of that sender has nonce X -> virtual record, for the moment, has nonce X
-	// in the second tracked block, the last tx of that sender has nonce X + N
-	// the virtual record has to be updated with the maximum between current value (X) and (X+N) value
-	// increasing by 1 helps knowing that we can select txs starting from this nonce of the virtual record
+func (virtualRecord *virtualAccountRecord) setInitialNonceOnSubsequentUpdate(breadcrumb *accountBreadcrumb) {
+	// This solves the common scenario when we have at least two different breadcrumbs for the same sender.
```
andreibancioiu, line 56: :saluting_face:

txcache/virtualSelectionSession.go
```diff
@@ -53,37 +53,13 @@ func (virtualSession *virtualSelectionSession) createAccountRecord(address []byt
...
-	}
-
-	return account.initialNonce.Value, nil
+func (virtualSession *virtualSelectionSession) getNonceForAccountRecord(accountRecord *virtualAccountRecord) (uint64, error) {
+	return accountRecord.getInitialNonce()
 }
 
-func (virtualSession *virtualSelectionSession) accumulateConsumedBalance(tx *WrappedTransaction) error {
-	sender := tx.Tx.GetSndAddr()
+func (virtualSession *virtualSelectionSession) accumulateConsumedBalance(tx *WrappedTransaction, senderRecord *virtualAccountRecord) error {
```
andreibancioiu, line 60: Usually (most frequently), the fee payer is the same as the sender (regular transactions, not relayed). In this case, `virtualSession.getRecord` is called twice for the same account.

However, maybe this is perfectly fine. Maybe we can even add a new function on `WrappedTransaction`, such as `isFeePayerSameAsSender()`, and do a quick check below, around `getRecord`.

Just an opinion, can stay as it is (for now or for the future).

txcache/selection.go
```diff
@@ -115,13 +122,14 @@ func selectTransactionsFromBunches(
...
 // Note (future micro-optimization): we can merge "detectSkippableSender()" and "detectSkippableTransaction()" into a single function,
-// any share the result of "sessionWrapper.getNonce()".
-func detectSkippableSender(virtualSession *virtualSelectionSession, item *transactionsHeapItem) bool {
-	nonce, err := virtualSession.getNonce(item.sender)
+// any share the result of "sessionWrapper.getNonceForAccountRecord()".
+func detectSkippableSender(virtualSession *virtualSelectionSession, item *transactionsHeapItem, virtualRecord *virtualAccountRecord) bool {
+	nonce, err := virtualRecord.getInitialNonce()
 	if err != nil {
-		log.Debug("detectSkippableSender", "err", err)
+		log.Debug("detectSkippableTransaction", "err", err)
```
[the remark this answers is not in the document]
danielradu10: fixed

txcache/virtualSelectionSession.go
danielradu10: i did a little fix here, hope is fine

txcache/wrappedTransaction.go
```diff
@@ -69,3 +72,7 @@ func (wrappedTx *WrappedTransaction) isTransactionMoreValuableForNetwork(otherTr
 	// In the end, compare by transaction hash
 	return bytes.Compare(wrappedTx.TxHash, otherTransaction.TxHash) < 0
 }
+
+func (wrappedTx *WrappedTransaction) isFeePayerSameAsSender() bool {
+	return bytes.Equal(wrappedTx.FeePayer, wrappedTx.Tx.GetSndAddr())
```
andreibancioiu, line 77: Maybe `len(asRelayed.GetRelayerAddr())` (see above) can be checked to skip the comparison on the most common flow. But then again, perhaps this is a micro-optimization, sorry. Can also be kept in mind for a future PR.

txcache/virtualSelectionSession.go
```diff
@@ -60,23 +60,32 @@ func (virtualSession *virtualSelectionSession) getNonceForAccountRecord(accountR
...
+		}
+
+		// should affect the record of fee payer
+		affectedRecord = feePayerRecord
+	}
+
 	fee := tx.Fee
 	if fee != nil {
-		consumedBalance := feePayerRecord.getConsumedBalance()
+		consumedBalance := affectedRecord.getConsumedBalance()
```
andreibancioiu: I remember we had a comment about `accumulateConsumedBalance`, but maybe that's on a different set of TODOs, correct? Here, it's very tricky that we mutate (with impact on the future) the result of `getConsumedBalance`.

txcache/virtualSelectionSession.go
```diff
@@ -60,23 +60,32 @@ func (virtualSession *virtualSelectionSession) getNonceForAccountRecord(accountR
...
 
 	transferredValue := tx.TransferredValue
 	if transferredValue != nil {
-		consumedBalance := senderRecord.getConsumedBalance()
+		consumedBalance := affectedRecord.getConsumedBalance()
 		consumedBalance.Add(consumedBalance, transferredValue)
 	}
 
+	// check if there's a need to search for another record
+	if !tx.isFeePayerSameAsSender() {
```
andreibancioiu: Maybe we can have a slightly more readable flow (opinion):
 - don't use a new variable called `affectedRecord`.
 - here, do as follows: initialize feePayerRecord := none.
 - then, if fee payer is same as sender (positive check), feePayerRecord = senderRecord
 - else, feePayerRecord = ...

danielradu10: fixed

txcache/virtualSelectionSession.go
danielradu10: fixed

txcache/accountBreadcrumbsValidator.go
```diff
@@ -124,6 +124,6 @@ func (validator *breadcrumbsValidator) validateBalance(
 		validator.virtualBalancesByAddress[address] = balance
 	}
 
-	virtualBalance.accumulateConsumedBalance(breadcrumb)
+	virtualBalance.accumulateConsumedBalance(breadcrumb.consumedBalance)
```
andreibancioiu, line 127: Missing discarded return value?

txcache/virtualSelectionSession.go
```diff
@@ -80,13 +77,14 @@ func (virtualSession *virtualSelectionSession) accumulateConsumedBalance(tx *Wra
 		}
 
 		// should affect the record of fee payer
-		affectedRecord = feePayerRecord
+		feePayerRecord = record
+	} else {
+		feePayerRecord = senderRecord
```
andreibancioiu: If we have to write both `if` and `else`, generally speaking, using the positive condition is preferred. Thus, if `tx.isFeePayerSameAsSender` etc. Opinion.

txcache/accountBreadcrumbsValidator.go
danielradu10: no, accumulateConsumedBalance from virtualAccountBalance does not return any value :thinking:

txcache/virtualSelectionSession.go
danielradu10: fixed

txcache/virtualSelectionSession.go
```diff
@@ -67,7 +67,9 @@ func (virtualSession *virtualSelectionSession) accumulateConsumedBalance(tx *Wra
 	}
 
 	// check if there's a need to search for another record
-	if !tx.isFeePayerSameAsSender() {
+	if tx.isFeePayerSameAsSender() {
+		feePayerRecord = senderRecord
+	} else {
```
andreibancioiu, line 71: Line 62 - can it be removed / used where needed (in the else _branch_)?

Maybe move `var feePayerRecord *virtualAccountRecord` right before the if-else (our current convention).

danielradu10: true, solved

andreibancioiu (approved):

sstanculeanu (approved):

PR #7270, opened 2025-09-18
export original key and val in dataTrieChanges

BeniaminDrasovean:
- In case of delete from a data trie, the exported `dataTrieChange` contained the trie key (which in some cases is a hash of the original key) and an empty val (to signal the deletion). In this case, there is no way to recompute the original key.
  
## Proposed changes
- Export the original key and the value without the added metadata on `dataTrieChange`
- Add another field in `dataTrieChange` to signal when a certain operation is a `delete`

## Testing procedure
- Test with the created tools

## Pre-requisites

### DIFF
--- a/state/trackableDataTrie/trackableDataTrie.go
+++ b/state/trackableDataTrie/trackableDataTrie.go
@@ -471,12 +471,7 @@ func (tdt *trackableDataTrie) modifyTrie(key []byte, dataEntry dirtyData, oldVal
 	newKey := tdt.getKeyForVersion(key, version)
 
 	if len(dataEntry.value) == 0 {
-		deletedKey, err := tdt.deleteFromTrie(oldVal, key, dtr)
-		if err != nil {
-			return nil, err
-		}
-
-		return deletedKey, nil
+		return tdt.deleteFromTrie(oldVal, key, dtr)
 	}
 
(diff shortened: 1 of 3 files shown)


ssd04 (dismissed):

state/trackableDataTrie/trackableDataTrie.go
```diff
@@ -449,30 +466,30 @@ func (tdt *trackableDataTrie) deleteOldEntryIfMigrated(key []byte, newData dirty
...
 
 	if len(dataEntry.value) == 0 {
 		deletedKey, err := tdt.deleteFromTrie(oldVal, key, dtr)
 		if err != nil {
-			return nil, nil, err
+			return nil, err
 		}
 
-		return deletedKey, nil, nil
+		return deletedKey, nil
```
sstanculeanu: return tdt.deleteFromTrie(oldVal, key, dtr)

sstanculeanu (approved):

state/trackableDataTrie/trackableDataTrie.go
BeniaminDrasovean: Done

ssd04 (approved):

sstanculeanu (approved):

PR #7272, opened 2025-09-19
Supernova TxPool: Fixed failing integration tests 09-19

danielradu10:
- Fixed some failing integration tests.
  
## Proposed changes
- used the right AccountsAdapter from nodeFacade, deleted the ones that are not used.
## Pre-requisites

### DIFF
--- a/facade/nodeFacade.go
+++ b/facade/nodeFacade.go
@@ -51,7 +51,7 @@ type ArgNodeFacade struct {
 	WsAntifloodConfig      config.WebServerAntifloodConfig
 	FacadeConfig           config.FacadeConfig
 	ApiRoutesConfig        config.ApiRoutesConfig
-	AccountsState          state.AccountsAdapter
+	AccountsStateAPI       state.AccountsAdapter
 	Blockchain             chainData.ChainHandler
 }
 
@@ -65,7 +65,7 @@ type nodeFacade struct {
 	endpointsThrottlers    map[string]core.Throttler
 	wsAntifloodConfig      config.WebServerAntifloodConfig
 	restAPIServerDebugMode bool
-	accountsState          state.AccountsAdapter
+	accountsStateAPI       state.AccountsAdapter
 	blockchain             chainData.ChainHandler
 }
 
@@ -84,7 +84,7 @@ func NewNodeFacade(arg ArgNodeFacade) (*nodeFacade, error) {
 	if err != nil {
 		return nil, err
 	}
-	if check.IfNil(arg.AccountsState) {
+	if check.IfNil(arg.AccountsStateAPI) {
 		return nil, ErrNilAccountState
 	}
 	if check.IfNil(arg.Blockchain) {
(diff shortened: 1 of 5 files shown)


facade/nodeFacade.go
```diff
@@ -369,7 +357,8 @@ func (nf *nodeFacade) GetSelectedTransactions(fields string) (*common.Transactio
 		fields,
 	)
 
-	return nf.apiResolver.GetSelectedTransactions(selectionOptionsAPI, nf.blockchain, nf.accountStateAPI)
+	// TODO brainstorm if this could be handled by the node
```
miiu96, line 360: maybe you should create a JIRA task for this TODO ?

process/block/preprocess/selectionSession.go
```diff
@@ -23,6 +23,7 @@ type ArgsSelectionSession struct {
 }
 
 // NewSelectionSession creates a new selection session.
+// TODO minimalize the interface used for AccountsAdapter
```
miiu96, line 26: also here

danielradu10: created

facade/nodeFacade.go
danielradu10: created

facade/nodeFacade.go
```diff
@@ -369,7 +357,8 @@ func (nf *nodeFacade) GetSelectedTransactions(fields string) (*common.Transactio
...
+	return nf.apiResolver.GetSelectedTransactions(selectionOptionsAPI, nf.blockchain, nf.accountsState)
```
andreibancioiu: Can we rename the field `nf.accountsState` to be obvious that it's `accountStateAPI`?

facade/nodeFacade.go
```diff
@@ -111,9 +101,7 @@ func NewNodeFacade(arg ArgNodeFacade) (*nodeFacade, error) {
 		config:                 arg.FacadeConfig,
 		apiRoutesConfig:        arg.ApiRoutesConfig,
 		endpointsThrottlers:    throttlersMap,
-		accountStateAPI:        arg.AccountsStateAPI,
```
andreibancioiu, line 114: I think you should have removed the other one, actually, as discussed. Now, the change looks dangerous to me.

facade/nodeFacade.go
andreibancioiu: The change looks dangerous to me. Shouldn't we have used the other one (as discussed)?

andreibancioiu (changes requested):
I think the change is not correct, can you please double check?

facade/nodeFacade.go
danielradu10: I think it's not a difference in which one should be deleted, both are of type AccountsAdapter. What is important, though, is to assure we initialize correctly with the AccountsAdapterAPI from the state components of the node, as we do in the file nodeRunner.go at line 763.

I will rename, to indicate that is the AccountsAdapterAPI and avoid misunderstanding.

facade/nodeFacade.go
danielradu10: renamed. the second question is answered above.

PR #7289, opened 2025-09-30
Unbond period for supernova

ssd04:
- Changed system smart contract setup to use unbonding period based on timestamp, instead of round difference

## Testing procedure
- To be tested with the feature branch

## Pre-requisites

### DIFF
--- a/integrationTests/testProcessorNode.go
+++ b/integrationTests/testProcessorNode.go
@@ -1042,7 +1042,7 @@ func (tpn *TestProcessorNode) createFullSCQueryService(gasMap map[string]map[str
 			ChanceComputer:      tpn.NodesCoordinator,
 			ShardCoordinator:    tpn.ShardCoordinator,
 			EnableEpochsHandler: tpn.EnableEpochsHandler,
-			EnableRoundsHandler: tpn.EnableRoundsHandler,
+			RoundHandler:        tpn.RoundHandler,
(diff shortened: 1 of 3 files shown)


node/chainSimulator/components/manualRoundHandler.go
```diff
@@ -65,6 +65,13 @@ func (handler *manualRoundHandler) RemainingTime(_ time.Time, maxTime time.Durat
 	return maxTime
 }
 
+// GetTimeStampForRound -
+func (handler *manualRoundHandler) GetTimeStampForRound(round uint64) uint64 {
+	timeFromGenesis := handler.roundDuration * time.Duration(round)
```
AdoAdoAdo, line 70: This is seems like it is not updated for Supernova

config/systemSmartContractsConfig.go
```diff
@@ -18,6 +18,7 @@ type StakingSystemSCConfig struct {
 	UnJailValue                          string
 	MinStepValue                         string
 	UnBondPeriod                         uint64
+	UnBondPeriodSec                      uint64
```
AdoAdoAdo, line 21: how is this getting updated from the config?
the config name for this field is different in the toml and seems to represent rounds for Supernova, not the time in seconds.

ssd04:
we do not follow this path for now

PR #7288, opened 2025-09-29
Supernova TxPool Part2: Added global account breadcrumb

danielradu10:
## Reasoning behind the pull request
- At the moment, each time we need the virtual records **we recreate the chain of tracked blocks.** However, this can be optimized by using a **global state of each breadcrumb's account,** which we maintain **with each added / removed tracked block**.
 In this first PR of this change, a new component (**kind of similar with accountBreadcrumb**, but different in methods) was created with its methods **to maintain a global breadcrumb.**
 
 The changes in this PR are related to this discussions:
 1. [discussion 1](https://multiversx.slack.com/archives/C07FKN2MC69/p1758524545618719?thread_ts=1758524519.367009&cid=C07FKN2MC69)
 2. [discussion 2](https://multiversx.slack.com/archives/C07FKN2MC69/p1758798023886399?thread_ts=1758786472.340799&cid=C07FKN2MC69)
 
  
## Proposed changes
- new component **globalAccountBreadcrumb**
- created method for each needed operation: update a global breadcrumb **when a tracked block is added**, update a global account breadcrumb when a **tracked block is removed on the OnProposedBlock**, update a global account breadcrumb **when a tracked block is removed on the OnExecutedBlock**.
- _**added TODOs for what should be done after integrating the new component.**_
- added unit tests for the new component

## Pre-requisites

Based on the [Contributing Guidelines](https://github.com/multiversx/mx-chain-go/blob/master/.github/CONTRIBUTING.md#branches-management) the PR author and the reviewers must check the following requirements are met:
- was the PR targeted to the correct branch?
- if this is a larger feature that probably needs more than one PR, is there a `feat` branch created?
- if this is a `feat` branch merging, do all satellite projects have a proper tag inside `go.mod`?

PR #7294, opened 2025-10-01
Supernova TxPool Part2: Integration of Global Account Breadcrumbs

danielradu10:
- Make use of the new component **globalAccountBreadcrumb**. 

related to:
[globalAccountBreadcrumb PR](https://github.com/multiversx/mx-chain-go/pull/7288)
[discussion 1](https://multiversx.slack.com/archives/C07FKN2MC69/p1758524545618719?thread_ts=1758524519.367009&cid=C07FKN2MC69)
[discussion 2](https://multiversx.slack.com/archives/C07FKN2MC69/p1758798023886399?thread_ts=1758786472.340799&cid=C07FKN2MC69)
   
## Proposed changes
- created a **globalAccountBreadcrumbCompiler**, which keeps the state of each address **(a map, where the key is the address and the value is a globalAccountBreadcrumb)**.
- integrated the new compiler component into the **selection tracker.**

- **modified OnProposedBlock flow** from selection tracker:

1. first we check if we have to replace a certain block.
2. when replacing a block on the OnProposed flow we remove all the blocks with **nonce greater or equal** than the received nonce and call **updateGlobalBreadcrumbsOnRemovedBlockOnProposed**.
3. only after, call the **updateGlobalBreadcrumbsOnAddedBlockOnProposed** method.

- when removing a block on the **OnExecutedBlock**, call **updateGlobalBreadcrumbsOnRemovedBlockOnExecuted**.

- **long unit tests to cover different scenarios.**

## Pre-requisites

### DIFF
--- a/txcache/globalAccountBreadcrumbsCompiler.go
+++ b/txcache/globalAccountBreadcrumbsCompiler.go
@@ -15,7 +15,7 @@ func newGlobalAccountBreadcrumbsCompiler() *globalAccountBreadcrumbsCompiler {
 	}
 }
 
-func (gabc *globalAccountBreadcrumbsCompiler) updateGlobalBreadcrumbsOnAddedBlock(tb *trackedBlock) {
+func (gabc *globalAccountBreadcrumbsCompiler) updateOnAddedBlock(tb *trackedBlock) {
 	gabc.mutCompiler.Lock()
 	defer gabc.mutCompiler.Unlock()
 
--- a/txcache/globalAccountBreadcrumbsCompiler_test.go
+++ b/txcache/globalAccountBreadcrumbsCompiler_test.go
@@ -55,7 +55,7 @@ func Test_shouldWorkConcurrently(t *testing.T) {
 		go func(tb *trackedBlock) {
 			defer wg.Done()
 
-			gabc.updateGlobalBreadcrumbsOnAddedBlock(tb)
+			gabc.updateOnAddedBlock(tb)
 		}(tb)
 
 		go func(tb *trackedBlock) {
@@ -91,7 +91,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// update the global state
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb1)
+		gabc.updateOnAddedBlock(tb1)
 
 		expectedGlobalBreadcrumbs := map[string]*globalAccountBreadcrumb{
 			"alice": {
@@ -146,7 +146,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// update the global state
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb2)
+		gabc.updateOnAddedBlock(tb2)
 
 		expectedGlobalBreadcrumbs = map[string]*globalAccountBreadcrumb{
 			"alice": {
@@ -248,7 +248,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// update the global state
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb3)
+		gabc.updateOnAddedBlock(tb3)
 
 		expectedGlobalBreadcrumbs := map[string]*globalAccountBreadcrumb{
 			"eve": {
@@ -304,7 +304,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// update the global state
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb4)
+		gabc.updateOnAddedBlock(tb4)
 
 		expectedGlobalBreadcrumbs = map[string]*globalAccountBreadcrumb{
 			"eve": {
@@ -359,7 +359,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// propose
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb5)
+		gabc.updateOnAddedBlock(tb5)
 
 		expectedGlobalBreadcrumbs = map[string]*globalAccountBreadcrumb{
 			"eve": {
@@ -522,7 +522,7 @@ func Test_shouldWorkOnDifferentScenarios(t *testing.T) {
 		require.NoError(t, err)
 
 		// update the global state
-		gabc.updateGlobalBreadcrumbsOnAddedBlock(tb6)
+		gabc.updateOnAddedBlock(tb6)
 
 		expectedGlobalBreadcrumbs = map[string]*globalAccountBreadcrumb{
 			"eve": {
--- a/txcache/selectionTracker.go
+++ b/txcache/selectionTracker.go
@@ -266,7 +266,7 @@ func (st *selectionTracker) addNewTrackedBlockNoLock(blockToBeAddedHash []byte, 
 
 	// add the new block
 	st.blocks[string(blockToBeAddedHash)] = blockToBeAdded
-	st.globalBreadcrumbsCompiler.updateGlobalBreadcrumbsOnAddedBlock(blockToBeAdded)
+	st.globalBreadcrumbsCompiler.updateOnAddedBlock(blockToBeAdded)
 
 	return nil
 }


txcache/globalAccountBreadcrumb.go
```diff
@@ -132,6 +132,8 @@ func (gab *globalAccountBreadcrumb) reduceConsumedBalance(receivedBreadcrumb *ac
 	_ = gab.consumedBalance.Sub(gab.consumedBalance, receivedBreadcrumb.consumedBalance)
 	if gab.consumedBalance.Sign() == -1 {
 		return errNegativeBalanceForBreadcrumb
+	} else if gab.consumedBalance.Sign() == 0 {
```
danielradu10: I'm not sure if this is a good practice. However, here is why I added it:
when subtracting from consumedBalance exactly the consumedBalance, the result, internally, is not  big.Int(0), but instead len(abs) is 0. I needed big.Int(0) for comparison, that's why I did this.

txcache/globalAccountBreadcrumbsCompiler.go
```diff
@@ -0,0 +1,79 @@
+package txcache
+
+import "sync"
+
+type globalAccountBreadcrumbsCompiler struct {
+	mutCompiler              sync.RWMutex
```
danielradu10, line 7: should we keep this mutex? we already have one in selectionTracker

txcache/globalAccountBreadcrumbsCompiler_test.go
```diff
@@ -0,0 +1,578 @@
...
+			require.NoError(t, err)
+
+			time.Sleep(100 * time.Millisecond)
+		}(tb)
+	}
+
+	wg.Wait()
+}
+
+func Test_shouldWork(t *testing.T) {
```
danielradu10: long tests, I have tried to cover the important scenarios

txcache/testutils_test.go
```diff
@@ -260,7 +260,13 @@ func (wrappedTx *WrappedTransaction) withTransferredValue(value *big.Int) *Wrapp
 	return wrappedTx
 }
```
danielradu10, line 262: modified / added some utils to be easier to create wrapped txs for testing

txcache/trackedBlock.go
```diff
@@ -33,8 +33,8 @@ func (tb *trackedBlock) sameNonceOrBelow(otherBlock *trackedBlock) bool {
 	return tb.nonce <= otherBlock.nonce
 }
 
-func (tb *trackedBlock) sameNonce(otherBlock *trackedBlock) bool {
-	return tb.nonce == otherBlock.nonce
+func (tb *trackedBlock) sameNonceOrHigher(otherBlock *trackedBlock) bool {
+	return tb.nonce >= otherBlock.nonce
```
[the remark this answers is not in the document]
danielradu10: Proper blocks replacement means that we remove each block that has the nonce equal with the new proposed block, or greater than the new proposed block. In the addNewTrackedBlockNoLock we iterate over each tracked block, and for each one we call **sameNonceOrHigher**, which is implemented as it should (verifies that the caller - an already tracked block -  has nonce greater or equal than the receiver - the new proposed block). 

I think the implementation is right, maybe I am missing something. :thinking: 

However, added some more unit tests to validate this.

txcache/selectionTracker.go
```diff
@@ -239,28 +245,30 @@ func (st *selectionTracker) validateBreadcrumbsOfTrackedBlocks(
...
-func (st *selectionTracker) addNewTrackedBlockNoLock(blockToBeAddedHash []byte, blockToBeAdded *trackedBlock) {
-	// search if in the tracked block we already have one with same nonce
+func (st *selectionTracker) addNewTrackedBlockNoLock(blockToBeAddedHash []byte, blockToBeAdded *trackedBlock) error {
+	// search if in the tracked blocks we already have one with same nonce or greater
 	for bHash, b := range st.blocks {
-		// TODO should delete also blocks with nonce greater than the nonce of the blockToBeAdded
-		// TODO should iterate over the breadcrumbs of each tracked block, and call updateOnRemoveAccountBreadcrumbOnProposedBlock.
-		if b.sameNonce(blockToBeAdded) {
-			// delete that block and break because there should be maximum one tracked block with that nonce
+		if b.sameNonceOrHigher(blockToBeAdded) {
```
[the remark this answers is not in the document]
danielradu10: Proper blocks replacement means that we remove each block that has the nonce equal with the new proposed block, or greater than the new proposed block. In the addNewTrackedBlockNoLock we iterate over each tracked block, and for each one we call **sameNonceOrHigher**, which is implemented as it should (verifies that the caller - an already tracked block -  has nonce greater or equal than the receiver - the new proposed block). 

I think the implementation is right, maybe I am missing something. :thinking: 

However, added some more unit tests to validate this.

txcache/globalAccountBreadcrumb.go
andreibancioiu: I think this is acceptable.

> for comparison

Though, for which one?

txcache/trackedBlock.go
```diff
@@ -33,8 +33,8 @@ func (tb *trackedBlock) sameNonceOrBelow(otherBlock *trackedBlock) bool {
 	return tb.nonce <= otherBlock.nonce
 }
 
-func (tb *trackedBlock) sameNonce(otherBlock *trackedBlock) bool {
-	return tb.nonce == otherBlock.nonce
+func (tb *trackedBlock) sameNonceOrHigher(otherBlock *trackedBlock) bool {
```
andreibancioiu: Can use "has" prefix.

txcache/selectionTracker.go
```diff
@@ -17,6 +17,7 @@ type selectionTracker struct {
 	latestNonce      uint64
 	latestRootHash   []byte
 	blocks           map[string]*trackedBlock
+	gabc             *globalAccountBreadcrumbsCompiler
```
andreibancioiu: Maybe "breadcrumbsCompiler"?

txcache/selectionTracker.go
```diff
@@ -307,6 +322,7 @@ func (st *selectionTracker) removeFromTrackedBlocksNoLock(searchedBlock *tracked
 		"searched block prevHash", searchedBlock.prevHash,
 		"removed blocks", removedBlocks,
 	)
+	return nil
```
andreibancioiu, line 326: Nitpicking / formatting - here, and in other place in this PR - can add newline before `return ...`.

txcache/globalAccountBreadcrumbsCompiler.go
andreibancioiu: Is this component's state ever used (read / written) concurrently / from multiple routines, or it's already protected (all access within a critical section) in upper layers?

txcache/globalAccountBreadcrumbsCompiler_test.go
andreibancioiu: I think (not sure) that function name pollutes the test namespace of the package. Maybe find a more specific name?

txcache/globalAccountBreadcrumb.go
danielradu10: unit tests comparisons 😢

txcache/globalAccountBreadcrumb.go
```diff
@@ -132,6 +132,8 @@ func (gab *globalAccountBreadcrumb) reduceConsumedBalance(receivedBreadcrumb *ac
...
+		gab.consumedBalance = big.NewInt(0)
```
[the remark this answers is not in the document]
danielradu10: to be discussed

txcache/globalAccountBreadcrumbsCompiler.go
danielradu10: It can be protected in the upper layer at the moment. I will leave a TODO for this, to be analyzed.

txcache/globalAccountBreadcrumbsCompiler_test.go
danielradu10: fixed

txcache/selectionTracker.go
danielradu10: renamed

txcache/selectionTracker.go
danielradu10: fixed

txcache/trackedBlock.go
danielradu10: fixed

txcache/globalAccountBreadcrumb.go
AdoAdoAdo: This should not be needed, the subtraction should work properly, if it is already zero it doesn't need a new allocation and assignment of a zero value.

txcache/globalAccountBreadcrumbsCompiler.go
AdoAdoAdo: I would keep the mutex, as there are two separate events being triggered and needing the mutex, and the txpool should not know the logic of how those are being called from outside of the txpool.

txcache/globalAccountBreadcrumb.go
danielradu10: fixed

andreibancioiu (dismissed):

txcache/globalAccountBreadcrumbsCompiler_test.go
```diff
@@ -0,0 +1,690 @@
...
+	wg.Add(2 * numOfBlocks)
+
+	for i := 1; i <= numOfBlocks; i++ {
+		tb := newTrackedBlock(uint64(i), []byte(fmt.Sprintf("hash%d", i-1)), []byte("rootHash0"), []byte(fmt.Sprintf("prevHash%d", i-1)))
+		go func(tb *trackedBlock) {
+			defer wg.Done()
+
+			gabc.updateGlobalBreadcrumbsOnAddedBlockOnProposed(tb)
+
+			time.Sleep(100 * time.Millisecond)
```
AdoAdoAdo: why do we need to wait here? it is not doing anything after the wait, just delaying the wg.Done, which is only part of the test code.

txcache/globalAccountBreadcrumbsCompiler_test.go
```diff
@@ -0,0 +1,690 @@
...
+		}(tb)
+
+		go func(tb *trackedBlock) {
+			defer wg.Done()
+
+			err := gabc.updateGlobalBreadcrumbsOnRemovedBlockOnExecuted(tb)
+			require.NoError(t, err)
+
+			time.Sleep(100 * time.Millisecond)
```
AdoAdoAdo: same as the other comment, why do we need to wait here?

txcache/selectionTracker.go
```diff
@@ -284,17 +292,24 @@ func (st *selectionTracker) OnExecutedBlock(blockHeader data.HeaderHandler) erro
...
+	if err != nil {
+		return err
+	}
 
+	st.updateLatestRootHashNoLock(nonce, rootHash)
 	return nil
 }
 
-func (st *selectionTracker) removeFromTrackedBlocksNoLock(searchedBlock *trackedBlock) {
+func (st *selectionTracker) removeFromTrackedBlocksNoLock(searchedBlock *trackedBlock) error {
```
AdoAdoAdo: this method name is misleading:

removeFromTrackedBlocks with the argument searched block seems (from the naming) that it wants to remove this specific block, but it actually removes other blocks with the same nonce or lower than the searched block.

maybe use something like `removeUpToBlock(searchedBlock)`

txcache/selectionTracker.go
```diff
@@ -284,17 +292,24 @@ func (st *selectionTracker) OnExecutedBlock(blockHeader data.HeaderHandler) erro
...
 	removedBlocks := 0
 	for blockHash, b := range st.blocks {
 		if b.sameNonceOrBelow(searchedBlock) {
-			// TODO should call updateOnRemoveAccountBreadcrumbOnExecutedBlock for each breadcrumb of the deleted tracked block
+			err := st.globalBreadcrumbsCompiler.updateGlobalBreadcrumbsOnRemovedBlockOnExecuted(b)
```
AdoAdoAdo: I find the method name a bit unclear.

this is updating the global breadcrumbs after removing tracked blocks from the front of the tracked blocks chain.

maybe use something like `globalBreadcrumbsCompiler.updateAfterRemoveBlockAndBelow(b)`

txcache/selectionTracker.go
```diff
@@ -239,28 +245,30 @@ func (st *selectionTracker) validateBreadcrumbsOfTrackedBlocks(
...
-	// search if in the tracked block we already have one with same nonce
+func (st *selectionTracker) addNewTrackedBlockNoLock(blockToBeAddedHash []byte, blockToBeAdded *trackedBlock) error {
+	// search if in the tracked blocks we already have one with same nonce or greater
 	for bHash, b := range st.blocks {
-		// TODO should delete also blocks with nonce greater than the nonce of the blockToBeAdded
-		// TODO should iterate over the breadcrumbs of each tracked block, and call updateOnRemoveAccountBreadcrumbOnProposedBlock.
-		if b.sameNonce(blockToBeAdded) {
-			// delete that block and break because there should be maximum one tracked block with that nonce
+		if b.hasSameNonceOrHigher(blockToBeAdded) {
+			err := st.globalBreadcrumbsCompiler.updateGlobalBreadcrumbsOnRemovedBlockOnProposed(b)
```
AdoAdoAdo: this method name is unclear how it updates the breadcrumbs.

e.g maybe use something like `globalBreadcrumbsCompiler.updateAfterRemoveBlockAndAbove(b)`

txcache/globalAccountBreadcrumbsCompiler_test.go
danielradu10: fixed.

txcache/globalAccountBreadcrumbsCompiler_test.go
danielradu10: fixed

txcache/selectionTracker.go
danielradu10: renamed

txcache/selectionTracker.go
danielradu10: renamed

txcache/selectionTracker.go
danielradu10: renamed

andreibancioiu (approved):

PR #7304, opened 2025-10-02
added syncBlockV3

sstanculeanu:
- blocks synchronization should be modified for header v3
  
## Proposed changes
- added a new method for v3 that warms up the tx pool for the synced block, then verifies it and appends it into the execution queue

## Testing procedure
- with feat branch

## Pre-requisites


process/block/shardblockProposal.go
```diff
@@ -212,7 +218,27 @@ func (sp *shardProcessor) VerifyBlockProposal(
 	// 	return nil, nil, err
 	// }
 
-	return nil
+	return sp.onProposedBlock(body, header)
+}
+
```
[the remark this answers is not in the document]
sstanculeanu: intended

process/sync/baseSync.go
```diff
@@ -823,6 +846,133 @@ func (boot *baseBootstrap) syncBlock() error {
...
+		return err
+	}
+
+	if !header.IsHeaderV3() {
+		return process.ErrInvalidHeader
+	}
+
+	err = boot.prepareTxPoolToSyncBlock()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
sstanculeanu: should be ok, it only handles one header

process/asyncExecution/queue/blocksQueue.go
```diff
@@ -24,14 +24,14 @@ type blocksQueue struct {
...
-func NewBlocksQueue() (*blocksQueue, error) {
+func NewBlocksQueue() *blocksQueue {
 	mutex := &sync.Mutex{}
 
 	return &blocksQueue{
 		mutex:           mutex,
 		headerBodyPairs: make([]HeaderBodyPair, 0),
 		notifyCh:        make(chan struct{}, 1), // buffered so send won't block if not read yet
-	}, nil
+	}
```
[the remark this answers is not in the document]
sstanculeanu: no error to be returned here

common/common.go
```diff
@@ -322,3 +323,38 @@ func prettifyBigNumbers(val reflect.Value) (string, bool) {
 	}
 	return "", false
 }
+
+// GetLastBaseExecutionResultHandler returns the BaseExecutionResultHandler of the provided header, based on its type
```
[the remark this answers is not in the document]
sstanculeanu: updated a bit

common/common_test.go
```diff
@@ -418,3 +418,86 @@ func TestPrettifyStruct(t *testing.T) {
 
 	})
 }
+
+func TestGetLastBaseExecutionResultHandler(t *testing.T) {
+	t.Parallel()
+
+	t.Run("nil header, should return error", func(t *testing.T) {
```
ssd04, line 489: add parallel run for the inner tests?

factory/processing/processComponents.go
```diff
@@ -2060,6 +2063,9 @@ func (pc *processComponents) Close() error {
 	if !check.IfNil(pc.blockProcessor) {
 		log.LogIfError(pc.blockProcessor.Close())
 	}
+	if !check.IfNil(pc.blocksQueue) {
+		pc.blocksQueue.Close()
```
ssd04, line 2094: use `log.LogIfError` as for the other fields?

process/sync/baseSync.go
```diff
@@ -823,6 +846,133 @@ func (boot *baseBootstrap) syncBlock() error {
...
+			boot.doJobOnSyncBlockFail(body, header, err)
+		}
+	}()
+
+	header, err = boot.getNextHeaderRequestingIfMissing()
+	if err != nil {
+		return err
+	}
+
+	go boot.requestHeadersFromNonceIfMissing(header.GetNonce() + 1)
```
ssd04: should we consider modifying `MaxHeadersToRequestInAdvance` here? we might need to request more headers in advance

process/sync/baseSync.go
```diff
@@ -823,6 +846,133 @@ func (boot *baseBootstrap) syncBlock() error {
...
+
+			boot.doJobOnSyncBlockFail(body, header, err)
+		}
+	}()
+
+	header, err = boot.getNextHeaderRequestingIfMissing()
+	if err != nil {
+		return err
+	}
+
```
ssd04: move `IsHeaderV3` check here?

factory/processing/processComponents.go
sstanculeanu: blocksQueue.Close() does not return error

process/sync/baseSync.go
sstanculeanu: moved

process/sync/baseSync.go
sstanculeanu: tbd
cc @AdoAdoAdo

common/common_test.go
sstanculeanu: added

process/sync/baseSync.go
```diff
@@ -731,13 +745,22 @@ func (boot *baseBootstrap) incrementSyncedWithErrorsForNonce(nonce uint64) uint3
 	return numSyncedWithErrors
 }
 
-// syncBlock method actually does the synchronization. It requests the next block header from the pool
+func (boot *baseBootstrap) syncBlock() error {
+	// TODO[Sorin]: this should check the round, not the epoch, once the PRs are merged
+	if boot.enableEpochsHandler.IsFlagEnabled(common.SupernovaFlag) {
```
ssd04: here we might need to determine activation status based on header so we might need to have the split for header v3 later on in `syncBLock`

process/sync/baseSync.go
```diff
@@ -731,13 +745,22 @@ func (boot *baseBootstrap) incrementSyncedWithErrorsForNonce(nonce uint64) uint3
...
+		return boot.syncBlockV3()
+	}
+
+	return boot.syncBlockLegacy()
```
ssd04: we need to analyse if we should on `OnProposed` on txPool on the legacy flow, maybe a TODO for this

process/sync/baseSync.go
sstanculeanu: applied the suggestion and solved the TODO

process/sync/baseSync.go
sstanculeanu: added todo

process/sync/baseSync.go
```diff
@@ -792,11 +800,28 @@ func (boot *baseBootstrap) syncBlock() error {
 		return err
 	}
 
+	if boot.enableRoundsHandler.IsFlagEnabledInRound(common.SupernovaRoundFlag, header.GetRound()) {
```
ssd04: add also a check for epoch?

ssd04: or leave a TODO and evaluate also in all other places where we check for supernova flag

process/block/baseProcess.go
AdoAdoAdo: remove it from here, since it is now in common?

process/block/export_test.go
```diff
@@ -755,7 +750,7 @@ func (sp *shardProcessor) CheckMetaHeadersValidityAndFinalityProposal(header dat
 
 // VerifyGasLimit -
 func (sp *shardProcessor) VerifyGasLimit(header data.ShardHeaderHandler) error {
-	return sp.verifyGasLimit(header)
+	return sp.verifyGasLimit(&block.Body{}, header)
```
AdoAdoAdo: don't we need also the real body?

process/block/shardblockProposal.go
```diff
@@ -232,7 +238,33 @@ func (sp *shardProcessor) verifyGasLimit(header data.ShardHeaderHandler) error {
 		return fmt.Errorf("%w, incoming mini blocks exceeded the limit", process.ErrInvalidMaxGasLimitPerMiniBlock)
 	}
 
-	return nil
+	// TODO: the verification needs a cutoff handler as well - e.g stop after the block including the execution results for
```
AdoAdoAdo: why is this TODO here in the verification of the gas limit?
this was previously in the VerifyBlockProposal and removed as no longer needed, it is only needed on processing.

also onProposed should not be part of the verification of the gas limit, but should be called separately in the verify block proposal.

(CreateBlockProposal also calls verifyGasLimit, and afterwards calls also onProposedBlock(), so there this would be a double call)

process/block/shardblockProposal.go
```diff
@@ -232,7 +238,33 @@ func (sp *shardProcessor) verifyGasLimit(header data.ShardHeaderHandler) error {
...
+
+func (sp *shardProcessor) onProposedBlock(body *block.Body, header data.HeaderHandler) error {
+	// TODO: proper accounts db should be used and its roothash must be updated first through a new method
+	//  SetRootHashIfNeeded which should recreate the trie if needed
+	accountsProvider, err := state.NewAccountsEphemeralProvider(sp.accountsDB[state.UserAccountsState])
+	if err != nil {
+		return err
+	}
+
+	lastExecResHandler, err := common.GetLastBaseExecutionResultHandler(header)
```
AdoAdoAdo: this should be the last execution result taken from the previous block that has passed consensus, not the last execution result added with this block.

process/sync/baseSync.go
```diff
@@ -792,11 +800,28 @@ func (boot *baseBootstrap) syncBlock() error {
...
+		body, err = boot.syncBlockV3(header)
+		return err
+	}
+
+	// TODO: analyze if OnProposed block should be also called on syncBlockLegacy()
```
AdoAdoAdo: no need to call onProposed, as there should be no tracked blocks on legacy, and onProposed adds tracked blocks.

process/sync/baseSync.go
AdoAdoAdo: yes I would check both together, to avoid round flag being set due to missing config for supernova flag (in tests for example activation round = 0 if not set, which will cause a bad config)

AdoAdoAdo: or we could use the header.IsHeaderV3 instead of the enable epoch + enable round flags check, which is more costly.

process/sync/baseSync.go
```diff
@@ -792,11 +800,28 @@ func (boot *baseBootstrap) syncBlock() error {
 		return err
 	}
 
+	if boot.enableRoundsHandler.IsFlagEnabledInRound(common.SupernovaRoundFlag, header.GetRound()) {
+		body, err = boot.syncBlockV3(header)
```
AdoAdoAdo: why is the body being returned?

process/sync/baseSync.go
```diff
@@ -792,11 +800,28 @@ func (boot *baseBootstrap) syncBlock() error {
...
+	return err
+}
+
+// syncBlockLegacy method actually does the synchronization. It requests the next block header from the pool
+// and if it is not found there it will be requested from the network. After the header is received,
+// it requests the block body in the same way(pool and then, if it is not found in the pool, from network).
+// If either header and body are received the ProcessBlock and CommitBlock method will be called successively.
+// These methods will execute the block and its transactions. Finally, if everything works, the block will be committed
+// in the blockchain, and all this mechanism will be reiterated for the next block.
+func (boot *baseBootstrap) syncBlockLegacy(header data.HeaderHandler) (body data.BodyHandler, err error) {
```
AdoAdoAdo: is the body needed for something on the caller?

process/sync/baseSync.go
```diff
@@ -792,11 +800,28 @@ func (boot *baseBootstrap) syncBlock() error {
...
 	go boot.requestHeadersFromNonceIfMissing(header.GetNonce() + 1)
 
 	body, err = boot.blockBootstrapper.getBlockBodyRequestingIfMissing(header)
 	if err != nil {
-		return err
+		return
```
AdoAdoAdo: can you have the return values explicitly?

process/sync/baseSync.go
```diff
@@ -836,7 +861,79 @@ func (boot *baseBootstrap) syncBlock() error {
...
+	}
+
+	err = boot.prepareTxPoolToSyncBlock(header.GetNonce())
+	if err != nil {
+		return
+	}
+
+	startTime := time.Now()
+	waitTime := boot.processWaitTime
```
AdoAdoAdo: getProcessWaitTime would need to be adjusted as well as it doesn't need to be 10 times shorter on Supernova. (this is also the time for verify block proposal & syncing missing data, so I think we can have a different constant for Supernova)

process/sync/baseSync.go
AdoAdoAdo: I would leave this for now at the same number (I think it requests 10 blocks in advance?)

process/sync/baseSync.go
```diff
@@ -836,7 +861,79 @@ func (boot *baseBootstrap) syncBlock() error {
...
+// And all this mechanism will be reiterated for the next block.
+func (boot *baseBootstrap) syncBlockV3(header data.HeaderHandler) (body data.BodyHandler, err error) {
+	if !header.IsHeaderV3() {
+		err = process.ErrInvalidHeader
+		return
+	}
+
+	go boot.requestHeadersFromNonceIfMissing(header.GetNonce() + 1)
+
+	body, err = boot.blockBootstrapper.getBlockBodyRequestingIfMissing(header)
```
AdoAdoAdo: I think this would no longer be needed, as the VerifyBlockProposal already does the request for any missing data for the block.

process/sync/baseSync.go
```diff
@@ -846,6 +943,68 @@ func (boot *baseBootstrap) syncBlock() error {
...
+		return boot.blocksQueue.AddOrReplace(queue.HeaderBodyPair{
+			Header: currentHeader,
+			Body:   currentBody,
+		})
+	}
+
+	// if there are multiple headers in between the syncing header and the last one executed,
+	// add them into the queue
+	for i := lastExecutedNonce; i < syncingNonce; i++ {
+		hdr, _, errGetHdr := boot.getHeaderFromPoolWithNonce(i)
```
AdoAdoAdo: Is this header in the pool still?
Need to adapt for headerv3 the removal from pool, and only remove from pool if the header results have been committed.

process/sync/baseSync.go
```diff
@@ -846,6 +943,68 @@ func (boot *baseBootstrap) syncBlock() error {
...
+
+	txPool := boot.poolsHolder.Transactions()
+	err = txPool.OnExecutedBlock(lastExecutedHeader)
+	if err != nil {
+		// TODO: reset the txPool context in case of error, once this will be implemented
+		return err
+	}
+
+	lastExecutedNonce := lastExecutedHeader.GetNonce()
+	if syncingNonce == lastExecutedNonce+2 {
```
AdoAdoAdo, line 977: This method is called prepareTxPoolToSyncBlock, but it is also preparing the execution queue. Maybe call it prepareSync.

Also this prepare sync should be called once before starting the block sync loop, not be called for every synced block.

AdoAdoAdo, line 977: can we merge this if with the for loop below?

We need to get the headers between the last executed header nonce notarized on currentBlock (blockchain.currentBlock) and the syncingNonce, and add them to the execution queue (if not already added)

and also for each call the OnProposedBlock()

process/block/baseProcess.go
sstanculeanu: removed

process/block/shardblockProposal.go
sstanculeanu: I think this was a merge issue. removed and updated as suggested.

process/sync/baseSync.go
sstanculeanu: it is returned for the defer func from syncBlock

process/sync/baseSync.go
sstanculeanu: yes, the defer func

process/block/export_test.go
sstanculeanu: removed it completely

process/sync/baseSync.go
sstanculeanu: added new constant

process/sync/baseSync.go
sstanculeanu: still needed, VerifyBlockProposal does the check for nils and the body is passed as parameter

process/sync/baseSync.go
sstanculeanu: added a flag to avoid calling multiple times on the same loop

sstanculeanu: added OnProposedBlock

as for merging, the reason why this separate if branch was added was to use the currentHeader already fetched previously.

process/block/shardblockProposal.go
sstanculeanu: updated

process/block/shardblockProposal_test.go
```diff
@@ -503,7 +503,7 @@ func TestShardProcessor_CreateBlockProposal(t *testing.T) {
 		}
 		dataComponents.BlockChain = &testscommon.ChainHandlerStub{
 			GetCurrentBlockHeaderCalled: func() data.HeaderHandler {
-				return &block.HeaderV2{} // using V2 for simplicity
+				return getSimpleHeaderV3Mock() // using V2 for simplicity
```
ssd04: update comment? or remove it

process/sync/baseSync.go
```diff
@@ -849,6 +889,195 @@ func (boot *baseBootstrap) syncBlock() error {
...
+// And all this mechanism will be reiterated for the next block.
+func (boot *baseBootstrap) syncBlockV3(body data.BodyHandler, header data.HeaderHandler) error {
+
+	err := boot.prepareForSyncIfNeeded(header.GetNonce())
+	if err != nil {
+		return err
+	}
+
+	startTime := time.Now()
+	waitTime := boot.processWaitTime
```
ssd04: use `getProcessWaitTime()` ?

ssd04: also we can remove the TODO for this func as it was handled in this PR

process/sync/baseSync.go
```diff
@@ -849,6 +889,195 @@ func (boot *baseBootstrap) syncBlock() error {
...
 }
 
+// syncBlockV3 method actually does the synchronization. It requests the next block header from the pool
+// and if it is not found there it will be requested from the network. After the header is received,
+// it requests the block body in the same way(pool and then, if it is not found in the pool, from network).
+// Once received, the header is verified through VerifyBlockProposal, but not before warming up the tx pool.
+// Finally, if everything works, the block will be committed and added into the processing queue.
+// And all this mechanism will be reiterated for the next block.
+func (boot *baseBootstrap) syncBlockV3(body data.BodyHandler, header data.HeaderHandler) error {
+
```
ssd04: remove empty line

process/sync/baseSync.go
sstanculeanu: indeed, updated

process/sync/baseSync.go
```diff
@@ -849,6 +888,194 @@ func (boot *baseBootstrap) syncBlock() error {
...
+		// only add the current header into the queue and pool
+		currentBody, errGetBody := boot.blockBootstrapper.getBlockBody(currentHeader)
+		if errGetBody != nil {
+			return errGetBody
+		}
+
+		currentHash := boot.getCurrentBlockHash()
+
+		errOnProposedBlock := boot.onProposedBlock(
+			currentBody,
```
AdoAdoAdo, line 988: the proposed cannot be the same with last committed.

process/sync/baseSync.go
```diff
@@ -849,6 +888,194 @@ func (boot *baseBootstrap) syncBlock() error {
...
+		lastCommittedHeader = hdr
+		lastCommittedHash = hdrHash
+	}
+
+	boot.preparedForSync = true
+
+	return nil
+}
+
+func (boot *baseBootstrap) onProposedBlock(
```
AdoAdoAdo: can we maybe move this to blockProcessor and export it from there (I think we have something similar there) ?
then we can use it here and in the shardBlock as well.

process/sync/baseSync.go
sstanculeanu: deleted, should not be needed for this case

process/sync/baseSync.go
sstanculeanu: moved as suggested

process/block/baseProcess.go
```diff
@@ -2336,12 +2334,17 @@ func (bp *baseProcessor) OnProposedBlock(
...
+		return err
+	}
+
 	lastExecResHandler, err := common.GetLastBaseExecutionResultHandler(lastCommittedHeader)
 	if err != nil {
 		return err
 	}
 
-	blockChainInfo := holders.NewBlockchainInfo(lastExecResHandler.GetHeaderHash(), lastCommittedHash, lastCommittedHeader.GetNonce())
+	blockChainInfo := holders.NewBlockchainInfo(lastExecResHandler.GetHeaderHash(), proposedHeader.GetPrevHash(), lastCommittedHeader.GetNonce()+1)
```
AdoAdoAdo: why not proposedHeader.GetNonce() ?

process/block/shardblockProposal.go
```diff
@@ -222,25 +220,21 @@ func (sp *shardProcessor) VerifyBlockProposal(
...
 // OnProposedBlock calls the OnProposedBlock from transactions pool
 func (sp *shardProcessor) OnProposedBlock(
 	proposedBody data.BodyHandler,
 	proposedHeader data.HeaderHandler,
 	proposedHash []byte,
-	lastCommittedHeader data.HeaderHandler,
-	lastCommittedHash []byte,
 ) error {
-	return sp.baseProcessor.OnProposedBlock(proposedBody, proposedHeader, proposedHash, lastCommittedHeader, lastCommittedHash)
+	return sp.baseProcessor.OnProposedBlock(proposedBody, proposedHeader, proposedHash)
```
AdoAdoAdo: I think you can remove the method from shardProcessor if it does not change the base processor implementation.

process/sync/baseSync.go
```diff
@@ -130,16 +132,15 @@ type baseBootstrap struct {
...
 	processWaitTime              time.Duration
+	processWaitTimeSupernova     time.Duration
+	preparedForSync              bool
 
 	repopulateTokensSupplies bool
 }
 
-// TODO: remove this handling after async exec
 func (boot *baseBootstrap) getProcessWaitTime() time.Duration {
 	if boot.enableRoundsHandler.IsFlagEnabled(common.SupernovaRoundFlag) {
```
ssd04: i think we should check here if enabled by round; where we call this, we have access to header, so we can check by round

sstanculeanu: pushed

process/block/baseProcess.go
sstanculeanu: updated

process/block/shardblockProposal.go
sstanculeanu: removing it leads to shardProcessor not implementing BlockProcessor interface

ssd04 (approved):

AdoAdoAdo (approved):

PR #7312, opened 2025-10-07
Executed miniblocks and transactions cache

sstanculeanu:
- executed mini blocks and the txs resulted after execution must be cached up until commit time, then moved into the storage
  
## Proposed changes
- added new caches and implemented the cache update and cleanup

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -798,7 +798,6 @@ func (bp *baseProcessor) sortHeaderHashesForCurrentBlockByNonce(usedInBlock bool
 func (bp *baseProcessor) createMiniBlockHeaderHandlers(
 	body *block.Body,
 	processedMiniBlocksDestMeInfo map[string]*processedMb.ProcessedMiniBlockInfo,
-	isHeaderV3 bool,
 ) (int, []data.MiniBlockHeaderHandler, error) {
 	if len(body.MiniBlocks) == 0 {
 		return 0, nil, nil
@@ -828,25 +827,6 @@ func (bp *baseProcessor) createMiniBlockHeaderHandlers(
 		if err != nil {
 			return 0, nil, err
 		}
-
-		// the following only applies for header v3
-		if !isHeaderV3 {
-			continue
-		}
-		// do not cache the cross-shard incoming mini blocks
-		selfShardID := bp.shardCoordinator.SelfId()
-		isCrossShardIncoming := miniBlockHeaderHandlers[i].GetReceiverShardID() == selfShardID &&
-			miniBlockHeaderHandlers[i].GetSenderShardID() != selfShardID
-		if isCrossShardIncoming {
-			continue
-		}
-
-		marshalledMiniBlock, err := bp.marshalizer.Marshal(body.MiniBlocks[i])
-		if err != nil {
-			return 0, nil, err
-		}
-
-		bp.dataPool.ExecutedMiniBlocks().Put(miniBlockHash, marshalledMiniBlock, len(marshalledMiniBlock))
 	}
 
 	return totalTxCount, miniBlockHeaderHandlers, nil
@@ -2499,6 +2479,16 @@ func (bp *baseProcessor) putMiniBlocksIntoStorage(miniBlockHeaderHandlers []data
 	executedMiniBlocksCache := bp.dataPool.ExecutedMiniBlocks()
 	for _, miniBlockHeaderHandler := range miniBlockHeaderHandlers {
 		mbHash := miniBlockHeaderHandler.GetHash()
+		// do not cache the cross-shard incoming mini blocks
+		selfShardID := bp.shardCoordinator.SelfId()
+		isCrossShardIncoming := miniBlockHeaderHandler.GetReceiverShardID() == selfShardID &&
+			miniBlockHeaderHandler.GetSenderShardID() != selfShardID
+		if isCrossShardIncoming {
+			// no need to move into storer, should be there already
+			executedMiniBlocksCache.Remove(mbHash)
+			continue
+		}
+
 		cachedMiniBlock, found := executedMiniBlocksCache.Get(mbHash)
 		if !found {
 			log.Warn("mini block from execution result not cached after execution",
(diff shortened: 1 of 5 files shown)


cmd/node/config/config.toml
```diff
@@ -516,6 +516,18 @@
...
+    Name = "ExecutedMiniBlocksCache"
+    Capacity = 1000
+    Type = "SizeLRU"
+    SizeInBytes = 104857600 # 100MB
+
+[PostProcessTransactionsCache]
+    Name = "PostProcessTransactionsCache"
+    Capacity = 500000
+    Type = "SizeLRU"
+    SizeInBytes = 209715200 #200MB
```
ssd04:
```suggestion
    SizeInBytes = 209715200 # 200MB
```

process/block/baseProcess.go
```diff
@@ -2395,3 +2415,185 @@ func (bp *baseProcessor) getMaxRoundsWithoutBlockReceived(round uint64) uint64 {
...
+
+	return nil
+}
+
+func (bp *baseProcessor) saveIntermediateTxs(headerHash []byte) error {
+	postProcessTxsCache := bp.dataPool.PostProcessTransactions()
+	cachedIntermediateTxs, ok := postProcessTxsCache.Get(headerHash)
+	if !ok {
+		log.Warn("saveIntermediateTxs: intermediateTxs not found in dataPool", "hash", headerHash)
+		return process.ErrMissingHeader
```
ssd04: extend this error to include info related to intermediate txs not found?

process/block/baseProcess.go
```diff
@@ -798,6 +798,7 @@ func (bp *baseProcessor) sortHeaderHashesForCurrentBlockByNonce(usedInBlock bool
 func (bp *baseProcessor) createMiniBlockHeaderHandlers(
 	body *block.Body,
 	processedMiniBlocksDestMeInfo map[string]*processedMb.ProcessedMiniBlockInfo,
+	isHeaderV3 bool,
```
ssd04: so this function is called only for processed miniblocks, right?

process/block/baseProcessHeaderV3_test.go
```diff
@@ -0,0 +1,765 @@
...
+			},
+			GetExecutionResultsHandlersCalled: func() []data.BaseExecutionResultHandler {
+				return []data.BaseExecutionResultHandler{}
+			},
+		}
+
+		err := bp.saveExecutedData(header, headerHash)
+		require.NoError(t, err)
+	})
+	t.Run("saveMiniBlocksFromExecutionResults path", func(t *testing.T) {
```
ssd04, line 194: not a strong opinion, but i think it might be better to have a separate main test for these flow, instead of keeping them in one big main test

process/block/shardblock.go
```diff
@@ -972,6 +972,11 @@ func (sp *shardProcessor) CommitBlock(
 
 	sp.saveBody(body, header, headerHash)
 
+	err = sp.saveExecutedData(header, headerHash)
```
ssd04: I'm thinking if we should move this call at the end of the `CommitBlock` function. Especially related to the cleanup for miniblocks and transactions, there might be some places where we might still need this info.
For example on `removeTxsFromPools` where need to remove txs based on miniblocks from execution results. Or we can remove txs from pools here at this step.

process/block/baseProcess.go
sstanculeanu: yes, for old headers it is called from applyBodyToHeader and for v3 it is called while collection execution results

process/block/baseProcessHeaderV3_test.go
sstanculeanu: should be ok

process/block/shardblock.go
sstanculeanu: moved

cmd/node/config/config.toml
```diff
@@ -515,6 +515,18 @@
...
+    Name = "ExecutedMiniBlocksCache"
+    Capacity = 1000
+    Type = "SizeLRU"
+    SizeInBytes = 104857600 # 100MB
+
+[PostProcessTransactionsCache]
+    Name = "PostProcessTransactionsCache"
+    Capacity = 500000
+    Type = "SizeLRU"
+    SizeInBytes = 209715200 # 200MB
```
AdoAdoAdo: I think both of these caches are a bit large

AdoAdoAdo: is there an explicit clean done for the saved items?

process/block/baseProcess.go
```diff
@@ -827,6 +828,25 @@ func (bp *baseProcessor) createMiniBlockHeaderHandlers(
 		if err != nil {
 			return 0, nil, err
 		}
+
+		// the following only applies for header v3
```
AdoAdoAdo: can we create a separate function from this similar to what we have for intermediate results?
We have the creation of the miniblock headers done from shardBlockProposal.go while creating the execution results, and we can call the function from there as well.

process/block/baseProcess.go
```diff
@@ -2395,3 +2415,185 @@ func (bp *baseProcessor) getMaxRoundsWithoutBlockReceived(round uint64) uint64 {
...
+
+	miniBlockStorer, err := bp.store.GetStorer(dataRetriever.MiniBlockUnit)
+	if err != nil {
+		return err
+	}
+
+	executedMiniBlocksCache := bp.dataPool.ExecutedMiniBlocks()
+	for _, miniBlockHeaderHandler := range miniBlockHeaderHandlers {
+		mbHash := miniBlockHeaderHandler.GetHash()
+		cachedMiniBlock, found := executedMiniBlocksCache.Get(mbHash)
```
AdoAdoAdo, line 2492: not all miniBlocks will be found in cache, as there are also the cross shard incoming miniblocks, those need to be ignored here (they were saved on proposal of the block).

process/block/shardblock.go
AdoAdoAdo, line 1116: can you add a TODO here and for meta, that this would probably need to be adjusted

process/block/shardblockProposal.go
```diff
@@ -815,7 +812,7 @@ func (sp *shardProcessor) collectExecutionResults(headerHash []byte, header data
 	// giving an empty processedMiniBlockInfo would cause all miniBlockHeaders to be created as fully processed.
 	processedMiniBlockInfo := make(map[string]*processedMb.ProcessedMiniBlockInfo)
 
-	totalTxCount, miniBlockHeaderHandlers, err := sp.createMiniBlockHeaderHandlers(sanitizedBodyAfterExecution, processedMiniBlockInfo)
+	totalTxCount, miniBlockHeaderHandlers, err := sp.createMiniBlockHeaderHandlers(sanitizedBodyAfterExecution, processedMiniBlockInfo, header.IsHeaderV3())
```
AdoAdoAdo: here we would instead need to call another function to cache the execution results data (mbs and txs)

cmd/node/config/config.toml
sstanculeanu: halved them

process/block/baseProcess.go
sstanculeanu: updated to only remove them and not check the cache, as discussed

process/block/shardblockProposal.go
sstanculeanu: updated

process/block/baseProcess.go
```diff
@@ -2499,6 +2479,16 @@ func (bp *baseProcessor) putMiniBlocksIntoStorage(miniBlockHeaderHandlers []data
 	executedMiniBlocksCache := bp.dataPool.ExecutedMiniBlocks()
 	for _, miniBlockHeaderHandler := range miniBlockHeaderHandlers {
 		mbHash := miniBlockHeaderHandler.GetHash()
+		// do not cache the cross-shard incoming mini blocks
```
AdoAdoAdo, line 2482: do not cache -> do not save ... ?

sstanculeanu: will be fixed in the next pr

PR #7315, opened 2025-10-07
fix for verifyGasLimit: it should not check for pending mini blocks after CheckIncomingMiniBlocks

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -221,7 +221,7 @@ func (gc *gasConsumption) checkPendingIncomingMiniBlocks() ([]data.MiniBlockHead
 		lastIndexAdded = i
 	}
 
-	gc.pendingMiniBlocks = gc.pendingMiniBlocks[lastIndexAdded:]
+	gc.pendingMiniBlocks = gc.pendingMiniBlocks[lastIndexAdded+1:]
 
 	return addedMiniBlocks, nil
 }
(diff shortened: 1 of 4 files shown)


process/block/gasConsumption.go
```diff
@@ -217,8 +218,11 @@ func (gc *gasConsumption) checkPendingIncomingMiniBlocks() ([]data.MiniBlockHead
 		}
 
 		addedMiniBlocks = append(addedMiniBlocks, mb)
+		lastIndexAdded = i
```
andreibancioiu, line 221: Not sure, maybe log the error a few lines above, along with the index?

process/block/shardblockProposal.go
```diff
@@ -215,14 +215,10 @@ func (sp *shardProcessor) verifyGasLimit(header data.ShardHeaderHandler) error {
 	}
 
 	sp.gasComputation.Reset()
-	lastMiniBlockIndex, pendingMiniBlocks, err := sp.gasComputation.CheckIncomingMiniBlocks(incomingMiniBlocks, incomingTransactions)
+	_, pendingMiniBlocks, err := sp.gasComputation.CheckIncomingMiniBlocks(incomingMiniBlocks, incomingTransactions)
```
andreibancioiu: `numPendingMiniblocks`?

process/block/gasConsumption.go
```diff
@@ -217,8 +218,11 @@ func (gc *gasConsumption) checkPendingIncomingMiniBlocks() ([]data.MiniBlockHead
...
 	}
 
+	gc.pendingMiniBlocks = gc.pendingMiniBlocks[lastIndexAdded:]
```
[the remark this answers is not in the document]
andreibancioiu: Could be a good catch?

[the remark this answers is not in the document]
AdoAdoAdo: a test for this scenario would also be good.

process/block/gasConsumption.go
sstanculeanu: should not be needed. any error would be propagated further and will stop the proposal creation process

process/block/shardblockProposal.go
sstanculeanu: applied the suggestion

process/block/gasConsumption.go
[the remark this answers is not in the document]
sstanculeanu: applied the suggestion + added helper get method + added test

andreibancioiu (approved):

PR #7419, opened 2025-11-11
MX-17306 Tooling for incorrect cross shard execution

mradian1:

Using the prettify block logging, improve the logs analysis tool to validate that cross shard miniblocks are executed (and proposed) in strict order, without gaps or duplications.
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


consensus/spos/bls/v1/subroundBlock.go
```diff
@@ -229,6 +229,12 @@ func (sr *subroundBlock) sendHeaderAndBlockBody(
 	sr.SetBody(bodyHandler)
 	sr.SetHeader(headerHandler)
 
+	// log the header output for debugging purposes
+	headerOutput, err := common.PrettifyStruct(headerHandler)
+	if err == nil {
+		log.Debug("Proposed header sent v1", "header", headerOutput)
+	}
```
axenteoctavian: I suggest we should also print something also on error case to know if some header was not printed.
```suggestion
	if err != nil {
		log.Error("Proposed header sent v1", "error", err)
	} else {
		log.Debug("Proposed header sent v1", "header", headerOutput)
	}
```

axenteoctavian: Same comment for all logs

axenteoctavian (commented):
+ Copilot suggestions

consensus/spos/bls/v1/subroundBlock.go
mradian1: done

common/common_test.go
```diff
@@ -665,3 +695,183 @@ func TestPrepareLogEventsKey(t *testing.T) {
...
+				HeaderHash:  []byte("header hash"),
+				HeaderNonce: hdrNonce - 1,
+				HeaderRound: hdrRound - 1,
+				HeaderEpoch: hdrEpoch,
+				RootHash:    []byte("root hash"),
+			},
+			NotarizedInRound: hdrRound - 1,
+		},
+		ExecutionResults: []*block.ExecutionResult{
+			&block.ExecutionResult{
```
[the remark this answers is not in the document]
mradian1: done

common/common_test.go
```diff
@@ -666,6 +694,185 @@ func TestPrepareLogEventsKey(t *testing.T) {
...
+	var hdrNonce = uint64(56)
+	var hdrShardId = uint32(1)
+	var hdrRound = uint64(67)
+	var hdrEpoch = uint32(78)
+	return &block.HeaderV3{
+		Nonce:            hdrNonce,
+		PrevHash:         []byte("prev hash"),
+		PrevRandSeed:     []byte("prev rand seed"),
+		RandSeed:         []byte("rand seed"),
+		ShardID:          hdrShardId,
```
mariusmihaic: You could use "1" here directly, without extra variable declaration

consensus/spos/bls/v1/subroundBlock.go
```diff
@@ -229,6 +229,14 @@ func (sr *subroundBlock) sendHeaderAndBlockBody(
 	sr.SetBody(bodyHandler)
 	sr.SetHeader(headerHandler)
 
+	// log the header output for debugging purposes
+	headerOutput, err := common.PrettifyStruct(headerHandler)
+	if err != nil {
+		log.Error("Proposed header sent v1", "error", err)
+	} else {
+		log.Debug("Proposed header sent v1", "header", headerOutput)
+	}
```
mariusmihaic: 1. Perhaps we should make a func here with input param being the `message string`
2. This should not be logged with log.Error, since it is not critical for debugging. Debug level should work too

mariusmihaic (commented):
Is this extra log intended/expected to be here forever?
All this extra data will increase the log size pretty much

consensus/spos/bls/v1/subroundBlock.go
mradian1: done, added a new method in common to log a header

common/common_test.go
mradian1: done

mariusmihaic (dismissed):

AdoAdoAdo (dismissed):

consensus/spos/bls/v2/subroundEndRound.go
```diff
@@ -362,6 +362,9 @@ func (sr *subroundEndRound) finalizeConfirmedBlock() bool {
 	msg := fmt.Sprintf("Added proposed block with nonce  %d  in blockchain", sr.GetHeader().GetNonce())
 	log.Debug(display.Headline(msg, sr.SyncTimer().FormattedCurrentTime(), "+"))
 
+	// log the header output for debugging purposes
+	common.LogPrettifiedHeader(sr.GetHeader(), "committed", "v2")
```
ssd04, line 366: this will print the same header info as in subroundBlock

ssd04: or this was intended for some extra tool?

AdoAdoAdo: this will be used by our logs checker for alarms trigger for issues on committed blocks during consensus, not on the failed consensuses (can generate false positives otherwise).

AdoAdoAdo (approved):

ssd04 (approved):

PR #7474, opened 2025-11-25
Removed solved TODOs

danielradu10:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


sstanculeanu (approved):

process/block/preprocess/transactions.go
```diff
@@ -1510,10 +1509,8 @@ func (txs *transactions) selectTransactionsFromTxPool(
 		txs.txCacheSelectionConfig.SelectionLoopDurationCheckInterval,
 	)
 
-	// TODO should use the right information for the nonce
```
andreibancioiu, line 1513: Please double check if this TODO should be removed. `SelectTransactions` still receives `currentBlockNonce = 0` below.

txcache/selectionTracker.go
```diff
@@ -28,7 +28,6 @@ func NewSelectionTracker(txCache txCacheForSelectionTracker, maxTrackedBlocks ui
 	if check.IfNil(txCache) {
 		return nil, errNilTxCache
 	}
-	// TODO compare with the maximum allowed offset between proposing a block and actually executing it
```
andreibancioiu, line 31: Not needed anymore / already implemented?

process/block/preprocess/transactions.go
danielradu10: yes, the given nonce is still 0 here, but on this flow we don't have any tracked blocks, because the OnProposedBlock is not called and only the OnExecutedBlock is called. so the given nonce is not important on this flow.
cc @AdoAdoAdo

txcache/selectionTracker.go
danielradu10: this offset is already verified in other place 
cc @AdoAdoAdo

process/block/preprocess/transactions.go
AdoAdoAdo: this is for the legacy, where we have execution and commit in sequence.

mariusmihaic (approved):

andreibancioiu (approved):

PR #7481, opened 2025-11-26
Meta process notarization fixes

ssd04:
- Fix unmarshall for meta headers
- Fix getting/setting shard data proposal
- Add cleanup constant when removing cross notarized headers from pool

## Testing procedure
- Local system testing

## Pre-requisites

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -455,7 +455,7 @@ func CreateLastExecutionResultFromPrevHeader(prevHeader data.HeaderHandler, prev
 		}, nil
 	}
 
-	prevMetaHeader, ok := prevHeader.(*block.MetaBlock)
+	prevMetaHeader, ok := prevHeader.(data.MetaHeaderHandler)
 	if !ok {
 		return nil, ErrWrongTypeAssertion
 	}
--- a/common/common.go
+++ b/common/common.go
@@ -544,7 +544,7 @@ func GetLastExecutionResultNonce(
 }
 
 // GetFirstExecutionResultNonce returns first execution result nonce if header v3 enabled, otherwise it returns provided header nonce.
-// It returns first execution result if there are any, otherwise it returns last execution results on the header
+// For header v3, it returns first execution result if there are any, otherwise it returns last execution results on the header
 func GetFirstExecutionResultNonce(
 	header data.HeaderHandler,
 ) uint64 {
--- a/common/common.go
+++ b/common/common.go
@@ -543,6 +543,23 @@ func GetLastExecutionResultNonce(
 	return lastExecutionResult.GetHeaderNonce()
 }
 
+func GetFirstExecutionResultNonce(
+	header data.HeaderHandler,
+) uint64 {
+	nonce := header.GetNonce()
+
+	if !header.IsHeaderV3() {
+		return nonce
+	}
+
+	if len(header.GetExecutionResultsHandlers()) > 0 {
+		firstExecResult := header.GetExecutionResultsHandlers()[0]
+		return firstExecResult.GetHeaderNonce()
+	}
+
+	return GetLastExecutionResultNonce(header)
+}
+
 // GetMiniBlockHeadersFromExecResult returns mb headers from meta header if v3, otherwise, returns mini block headers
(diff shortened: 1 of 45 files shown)


common/common.go
```diff
@@ -543,6 +543,25 @@ func GetLastExecutionResultNonce(
 	return lastExecutionResult.GetHeaderNonce()
 }
 
+// GetFirstExecutionResultNonce returns first execution result nonce if header v3 enabled, otherwise it returns provided header nonce.
+// It returns first execution result if there are any, otherwise it returns last execution results on the header
```
mariusmihaic:
```suggestion
// For header v3, it returns first execution result if there are any, otherwise it returns last execution results on the header
```

process/block/baseProcess.go
```diff
@@ -47,6 +47,10 @@ import (
 	"github.com/multiversx/mx-chain-go/storage/storageunit"
 )
 
+const (
+	clenaupCrossShardHeadersDelta = 5
```
mariusmihaic, line 51: [question]
Is this const value safe?
I imagine it's here so we don't lose some data while we got execution results accumulating, but can those exceed this value?

process/block/baseProcess.go
```diff
@@ -1205,7 +1209,10 @@ func (bp *baseProcessor) cleanupPoolsForCrossShard(
 		return
 	}
 
-	crossNotarizedHeaderNonce := common.GetLastExecutionResultNonce(crossNotarizedHeader)
+	crossNotarizedHeaderNonce := common.GetFirstExecutionResultNonce(crossNotarizedHeader)
+	if crossNotarizedHeaderNonce > clenaupCrossShardHeadersDelta {
+		crossNotarizedHeaderNonce -= clenaupCrossShardHeadersDelta
+	}
```
mariusmihaic: For nonce 1->5, this will still delete previous nonce instead of keeping a `delta`.
Perhaps we can skip removal if `crossNotarizedHeaderNonce` < `clenaupCrossShardHeadersDelta`?

process/block/metablock.go
```diff
@@ -513,6 +513,13 @@ func (mp *metaProcessor) getAllMiniBlockDstMeFromShards(metaHdr data.MetaHeaderH
 			return nil, err
 		}
 
+		log.Debug("SSS getAllMiniBlockDstMeFromShards",
```
mariusmihaic: Perhaps delete this?

process/peer/process_test.go
```diff
@@ -2951,3 +2952,85 @@ func TestValidatorStatisticsProcessor_getActualList(t *testing.T) {
...
 	assert.Equal(t, jailedList, computedJailedList)
 }
+
+func TestValidatorStatistics_SearchInMap(t *testing.T) {
+	t.Parallel()
+
+	t.Run("should work when found in cache", func(t *testing.T) {
+		t.Parallel()
+
+		cache := createMockCache()
```
mariusmihaic, line 2962: Can we "compress" a bit some of these lines in these tests so we don't have a new line after each code line?

common/common.go
ssd04: done

process/block/baseProcess.go
ssd04: the nonce is considered based on execution results, not based on proposed/committed headers, so this should be fine

process/block/baseProcess.go
ssd04: done

process/peer/process_test.go
```diff
@@ -2951,3 +2952,83 @@ func TestValidatorStatisticsProcessor_getActualList(t *testing.T) {
...
+		require.Nil(t, err)
+		require.Equal(t, header, retHeader)
+	})
+
+	t.Run("should work when found in headers pool", func(t *testing.T) {
+		t.Parallel()
+
+		cache := createMockCache()
+
+		_, header := generateTestShardBlockHeaders(cache)
```
mariusmihaic, line 2982: Can we also "compress" lines in these tests:
- `should fail when not found in headers pool`
- `should work when found in headers pool`

ssd04: i think it's ok for better visibility

AdoAdoAdo (approved):

mariusmihaic (approved):

PR #7492, opened 2025-11-28
Meta registry trigger

mariusmihaic:
- Meta trigger will work with meta registry handler to handle both v1 and v3 meta blocks
  
## Proposed changes
- Epoch change proposed bool flag is also saved

## Testing procedure
- 
- 
- 

## Pre-requisites


epochStart/metachain/triggerRegistry.go
```diff
@@ -27,50 +28,90 @@ func (t *trigger) LoadState(key []byte) error {
...
 	}
+
 	return state, nil
 }
 
 // saveState saves the trigger state. Needs to be called under mutex
 func (t *trigger) saveState(key []byte) error {
 	// TODO: add new structures for the new headers and update this component
-	metaHeader, ok := t.epochStartMeta.(*block.MetaBlock)
+	metaHeader, ok := t.epochStartMeta.(data.MetaHeaderHandler)
```
sstanculeanu, line 100: remove the above TODO

epochStart/metachain/triggerRegistry.go
```diff
@@ -27,50 +28,90 @@ func (t *trigger) LoadState(key []byte) error {
...
+	t.epochChangeProposed = state.GetEpochChangeProposed()
 	t.mutTrigger.Unlock()
 
 	return nil
 }
 
 // UnmarshalTrigger unmarshalls the trigger with json, for backwards compatibility
-func UnmarshalTrigger(marshaller marshal.Marshalizer, data []byte) (*block.MetaTriggerRegistry, error) {
+func UnmarshalTrigger(marshaller marshal.Marshalizer, data []byte) (data.MetaTriggerRegistryHandler, error) {
+	trig, err := unmarshallTriggerV3(marshaller, data)
```
sasurobert, line 45: could you rewrite this ? Like I do not like that in case of error we try the next one. that is not good.

Can you find some other way ? 

Is unmarshal failing only if the structure is wrong ?

epochStart/metachain/triggerRegistry.go
mariusmihaic: I think that is a TODO for something else. Same comment is found on shard trigger, so I woulnd't touch it

epochStart/metachain/triggerRegistry.go
mariusmihaic: Agree, I'm not a big fan of this approach either, but it is the way the whole code has been done since `ShardHeaderV2`.

You can see this is the approach we use for every type of unmarshalling like:

- `UnmarshalTrigger` for shard trigger
- `UnmarshalMetaHeader`
- `UnmarshalShardHeader`
- `UnmarshalExecutionResult`
...

epochStart/metachain/triggerRegistry.go
sstanculeanu: no, I've added the comment for the specific functionality you've added here 😄 

we should add similar structure for shard as well, thus the second TODO

mariusmihaic: My bad, wasn't aware, deleted it

sstanculeanu (approved):

AdoAdoAdo (approved):

sasurobert (approved):

epochStart/metachain/triggerRegistry.go
sasurobert: yeah. saw that. But still :D

PR #7501, opened 2025-12-03
Fix RootHash Mismatch on Leader Bootstrap

danielradu10:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -509,10 +509,26 @@ func (st *storageBootstrapper) setCurrentBlockInfo(
...
+		err = st.blkExecutor.OnExecutedBlock(header, lastExecutionResult.GetRootHash())
+		if err != nil {
+			return err
+		}
+
 		return st.setCurrentBlockInfoV3(header, headerHash)
 	}
 
-	err := st.blkc.SetCurrentBlockHeaderAndRootHash(header, rootHash)
+	err := st.blkExecutor.OnExecutedBlock(header, rootHash)
```
mariusmihaic, line 526: [question]
Is it ok that you call this on non header v3?

danielradu10: yes, OnExecutedBlock has to be called also on non v3  headers, in order to keep the tracker updated with the latest root hash and the latest nonce. 

the OnProposedBlock is the one that has to be called only on v3 headers.

hopefully I am not wrong, cc @AdoAdoAdo

ssd04 (approved):

process/sync/storageBootstrap/baseStorageBootstrapper.go
AdoAdoAdo: The updated txpool will be initialized since the beginning, it should be backwards compatible, but requires for that to have knowledge on the previous block roothash before any selections.

we do not have a transition from the previous txpool to the new txpool, this is why even for old epochs we need this intitialization.

mariusmihaic (approved):

PR #7503, opened 2025-12-03
updated to reset the blocks queue's chan when resuming after a processing pause

sstanculeanu:
- after an execution break, the headersExecutor is able to pop only one header, then the loop is closed due to internal chan
  
## Proposed changes
- reset the buffered channel on resume execution

## Testing procedure
- with feat branch

## Pre-requisites


sasurobert (dismissed):

process/asyncExecution/headersExecutor.go
```diff
@@ -106,12 +106,16 @@ func (he *headersExecutor) start(ctx context.Context) {
...
 			// blocking operation
-			headerBodyPair, ok := he.blocksQueue.Pop()
-			if !ok {
+			headerBodyPair, valueOk, shouldContinue := he.blocksQueue.Pop()
+			if !shouldContinue {
 				// close event
 				return
 			}
 
+			if !valueOk {
```
AdoAdoAdo: can we check here if the header body pair is valid instead of returning another bool?
e.g in case of valueOk false, both the pair.Header and pair.Body are nil

sstanculeanu: pushed

sasurobert (dismissed):

process/block/baseProcess.go
```diff
@@ -3318,13 +3319,15 @@ func (bp *baseProcessor) verifyGasLimit(header data.HeaderHandler, miniBlocks bl
 
 	bp.gasComputation.Reset()
 	_, numPendingMiniBlocks, err := bp.gasComputation.AddIncomingMiniBlocks(splitRes.incomingMiniBlocks, splitRes.incomingTransactions)
-	if err != nil {
+	// if limits are reached, do not propagate the error so consensus can continue but with no blocks
```
AdoAdoAdo: I think on the zero limit condition we should instead not do any selection for incoming and outgoing miniblocks/transactions, then we should not need to ignore this error.

The block production should continue, while there is a zero limit, but with no transactions.

process/asyncExecution/queue/blocksQueue.go
```diff
@@ -164,7 +164,9 @@ func (bq *blocksQueue) Pop() (HeaderBodyPair, bool) {
 		bq.headerBodyPairs = bq.headerBodyPairs[1:]
 		return item, true
 	}
-	return HeaderBodyPair{}, false
+
+	// allow further Pops
+	return HeaderBodyPair{}, true
```
AdoAdoAdo: I think this should still return false.

Maybe there is an issue with the pause resume, it should not cause the notify channel to be written to without having a new block in queue there.

process/block/baseProcess.go
sstanculeanu: fixed

process/asyncExecution/queue/blocksQueue.go
sstanculeanu: fixed

AdoAdoAdo (dismissed):

AdoAdoAdo (dismissed):

AdoAdoAdo (dismissed):

sasurobert (dismissed):

AdoAdoAdo (approved):

sasurobert (approved):

PR #7504, opened 2025-12-03
Execution results handling fixes

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/asyncExecution/executionManager/executionManager.go
+++ b/process/asyncExecution/executionManager/executionManager.go
@@ -90,14 +90,53 @@ func (em *executionManager) AddPairForExecution(pair queue.HeaderBodyPair) error
 	defer em.mut.Unlock()
 
 	lastExecutedBlock := em.blockChain.GetLastExecutedBlockHeader()
+	// TODO: analyze if we just need to re-execute instead of ignoring
 	if areSameHeaders(pair.Header, lastExecutedBlock) {
 		log.Warn("header already executed", "nonce", pair.Header.GetNonce(), "round", pair.Header.GetRound())
 		return nil
 	}
-	// todo: remove pending execution result on same nonce if any
-	// todo: make sure the lastExecutedBlockHeader from blockchain is only set in blockchain if
-	// the block has passed consensus (was committed)
+
+	if lastExecutedBlock.GetNonce() >= pair.Header.GetNonce() {
+		err := em.updateContextForReplacedHeader(pair.Header)
+		if err != nil {
+			return err
+		}
+	}
+
 	return em.blocksQueue.AddOrReplace(pair)
+}
+
+func (em *executionManager) updateContextForReplacedHeader(header data.HeaderHandler) error {
+	prevNonce := header.GetNonce() - 1
+	pendingExecutionResults, err := em.GetPendingExecutionResults()
+	if err != nil {
+		return err
+	}
+
+	lastExecutionResult, err := em.executionResultsTracker.GetLastNotarizedExecutionResult()
+	if err != nil {
+		return err
+	}
+
+	headerHashToSet := lastExecutionResult.GetHeaderHash()
+	executionResultToSet := lastExecutionResult
+	for i := len(pendingExecutionResults) - 1; i >= 0; i-- {
+		if pendingExecutionResults[i].GetHeaderNonce() <= prevNonce {
+			headerHashToSet = pendingExecutionResults[i].GetHeaderHash()
+			executionResultToSet = pendingExecutionResults[i]
+			break
+		}
+	}
+
+	headerToSet, err := em.headers.GetHeaderByHash(headerHashToSet)
+	if err != nil {
+		return err
+	}
+
+	em.blockChain.SetLastExecutedBlockHeaderAndRootHash(headerToSet, headerHashToSet, executionResultToSet.GetRootHash())
+	em.blockChain.SetLastExecutionResult(executionResultToSet)
+
+	return nil
 }
 
 func areSameHeaders(header1, header2 data.HeaderHandler) bool {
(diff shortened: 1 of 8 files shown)


process/asyncExecution/executionManager/executionManager.go
```diff
@@ -90,26 +89,48 @@ func (em *executionManager) AddPairForExecution(pair queue.HeaderBodyPair) error
...
-	sameNonce := header1.GetNonce() == header2.GetNonce()
-	sameRound := header1.GetRound() == header2.GetRound()
-	samePreviousHash := bytes.Equal(header1.GetPrevHash(), header2.GetPrevHash())
+	lastExecutionResult, err := em.executionResultsTracker.GetLastNotarizedExecutionResult()
+	if err != nil {
+		return err
+	}
 
-	return sameNonce && sameRound && samePreviousHash
+	headerHashToSet := lastExecutionResult.GetHeaderHash()
```
AdoAdoAdo: check this (headerHashToSet) is the header.prevHash and avoid iterating on pending.

process/asyncExecution/executionManager/executionManager.go
```diff
@@ -90,26 +89,48 @@ func (em *executionManager) AddPairForExecution(pair queue.HeaderBodyPair) error
...
+	executionResultToSet := lastExecutionResult
+	for i := len(pendingExecutionResults) - 1; i >= 0; i-- {
+		if pendingExecutionResults[i].GetHeaderNonce() <= prevNonce {
+			headerHashToSet = pendingExecutionResults[i].GetHeaderHash()
+			executionResultToSet = pendingExecutionResults[i]
+			break
+		}
+	}
+
```
AdoAdoAdo: if header.prevHash does not match return an error.

process/asyncExecution/executionManager/executionManager.go
```diff
@@ -90,26 +89,48 @@ func (em *executionManager) AddPairForExecution(pair queue.HeaderBodyPair) error
...
+	headerToSet, err := em.headers.GetHeaderByHash(headerHashToSet)
+	if err != nil {
+		return err
+	}
+
+	em.blockChain.SetLastExecutedBlockHeaderAndRootHash(headerToSet, headerHashToSet, executionResultToSet.GetRootHash())
+	em.blockChain.SetLastExecutionResult(executionResultToSet)
+
```
AdoAdoAdo: cleanup pending execution results on top of the set last execution result, from execution results tracker.

process/asyncExecution/executionManager/executionManager.go
```diff
@@ -90,26 +89,48 @@ func (em *executionManager) AddPairForExecution(pair queue.HeaderBodyPair) error
...
-	// todo: make sure the lastExecutedBlockHeader from blockchain is only set in blockchain if
-	// the block has passed consensus (was committed)
+
 	return em.blocksQueue.AddOrReplace(pair)
 }
 
-func areSameHeaders(header1, header2 data.HeaderHandler) bool {
-	if check.IfNil(header1) || check.IfNil(header2) {
-		return false
+func (em *executionManager) updateContextForReplacedHeader(header data.HeaderHandler) error {
```
mariusmihaic, line 104: I think we can only pass the nonce here

process/block/baseProcess.go
```diff
@@ -3123,6 +3124,61 @@ func (bp *baseProcessor) addExecutionResultsOnHeader(header data.HeaderHandler)
...
+	for i, res := range pendingExecutionResults {
+		h := res.GetHeaderHash()
+		pendingHashToIndex[string(h)] = i
+	}
+
+	// Walk the canonical chain of previous hashes and check membership in the pending set.
+	prevHash := currentHeader.GetPrevHash()
+	prevNonce := currentHeader.GetNonce() - 1
+
+	for {
```
mariusmihaic: [question]
Is there any way to early exit this if we find out that we are iterating "way too much in the "past"  and it's likely we are not going to return from it? 
Or we just rely on the fact that the pool will not have "too old" headers and we break this for?

process/asyncExecution/executionManager/executionManager.go
AdoAdoAdo: I have updated, using also the prevHash

sstanculeanu (approved):

process/asyncExecution/executionManager/executionManager_test.go
```diff
@@ -143,24 +144,355 @@ func TestExecutionManager_SetHeadersExecutor(t *testing.T) {
...
+			Header: &block.Header{
+				Nonce:    9,
+				PrevHash: []byte("lastNotarizedExecResultHash"),
+			},
+			Body: &block.Body{},
+		}
+
+		err := em.AddPairForExecution(pair)
+		require.Equal(t, errExpected, err)
+		require.Equal(t, 2, counter)
```
mariusmihaic, line 374: [nitpicking]

Checking this counter instead of actually checking that the expected functions have been called could be misleading.
You could have for example here `SetLastExecutionResultCalled` being called twice instead of calling `SetLastExecutedBlockHeaderAndRootHashCalled` + `SetLastExecutionResultCalled`

process/block/baseProcess_test.go
```diff
@@ -5205,3 +5205,115 @@ func TestBaseProcessor_CacheIntraShardMiniBlocks(t *testing.T) {
...
+		}
+
+		sanitizedPendingExecResults := bp.ExcludeRevertedExecutionResultsForHeader(
+			header,
+			pendingExecutionResults,
+		)
+		require.Equal(t, 0, len(sanitizedPendingExecResults))
+	})
+
+	t.Run("should remove last execution result if if getting the execution result's header from storage fails", func(t *testing.T) {
```
mariusmihaic, line 5250:
```suggestion
	t.Run("should remove last execution result if getting the execution result's header from storage fails", func(t *testing.T) {
```

mariusmihaic (approved):

PR #7519, opened 2025-12-09
allow 600m gas per tx in one mb

sstanculeanu:
- one tx per mini block that has more gas than the mini block limit should be allowed
  
## Proposed changes
- added special handling

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/process/block/gasConsumption.go
+++ b/process/block/gasConsumption.go
@@ -233,8 +233,6 @@ func (gc *gasConsumption) checkGasConsumedByMiniBlock(mb data.MiniBlockHeaderHan
 		if check.IfNil(tx) {
 			continue
 		}
-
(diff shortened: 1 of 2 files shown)


axenteoctavian (dismissed):

process/block/gasConsumption.go
AdoAdoAdo: is there also a check that gasLimit is not exceeding the maximum allowed? not only for the receiver, but the total.

sstanculeanu: this method is only called for incoming mini blocks, thus only the gas in receiver's shard is considered here. the check for the total consumed is done on the method handling outgoing txs

AdoAdoAdo (approved):

axenteoctavian (approved):

PR #7525, opened 2025-12-10
Improve AddTx in list for sender

danielradu10:
- At the moment, when we insert a new transaction in pool, we also have to insert the transaction in the list for sender. Currently, the list of transactions for a specific sender is a doubly linked list. This means that for finding the insertion place of a new txs, we have to iterate over the list. In the worst scenario, this might not be the best solution.

**This PR introduces a new data structure, a Red Black Tree (a balanced BST) which allows us finding the insertion place of a new element in O(log n).**
  
## Proposed changes
- added transactions red black tree
- make use of the tree
- unit tests, benchmarks

## Pre-requisites


txcache/transactionsRedBlackTree.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+
+	node, found := trb.tree.Floor(e)
+	if found {
+		// our new element should be placed after this one
+		return (node.Key).(*list.Element), nil
+	}
+
+	return nil, nil
```
sasurobert, line 97: return nil, nil should not happen.

txcache/txListForSender.go
```diff
@@ -44,9 +45,11 @@ func (listForSender *txListForSender) AddTx(tx *WrappedTransaction, tracker *sel
 	}
 
 	if insertionPlace == nil {
-		listForSender.items.PushFront(tx)
+		element := listForSender.items.PushFront(tx)
```
sasurobert, line 48: I get this is how it was used, but returning nil, nil from findInsertionPlace is a bad design, you could return and extra bool, and solve with that.

sstanculeanu:
Different approach implemented on #7613

PR #7527, opened 2025-12-10
implemented direct sent transactions cache

sstanculeanu:
- some transactions might come from direct send, which means they will not be available on all nodes
  
## Proposed changes
- store the transactions received via direct send in a new cache
- when proposing a header, if there are some porposed transactions received via direct send, broadcast them on the network

## Testing procedure
- with feat branch + specific binary to allow direct send

## Pre-requisites


consensus/spos/bls/v2/subroundBlock.go
```diff
@@ -323,12 +325,38 @@ func (sr *subroundBlock) sendBlockHeader(
...
 	headerOutput, err := common.PrettifyStruct(headerHandler)
 	if err == nil {
-		log.Debug("Proposed header sent", "header", headerOutput)
+		log.Debug("proposed header sent", "header", headerOutput)
 	}
 
 	return true
 }
 
+func (sr *subroundBlock) sendDirectSentTransactions(
```
sasurobert, line 345: do we need this kind of feature ? The answer is probably yes, as this open up possibilities of private pools, and different mechanics for that. 

I wonder whether there is a moment, earlier than this, in which the broadcasting can happen.

process/interceptors/processor/txInterceptorProcessor.go
```diff
@@ -46,7 +53,7 @@ func (txip *TxInterceptorProcessor) Validate(data process.InterceptedData, _ cor
 }
 
 // Save will save the received data into the cacher
-func (txip *TxInterceptorProcessor) Save(data process.InterceptedData, peerOriginator core.PeerID, _ string) (bool, error) {
+func (txip *TxInterceptorProcessor) Save(data process.InterceptedData, peerOriginator core.PeerID, _ string, broadcastMethod p2p.BroadcastMethod) (bool, error) {
```
sasurobert, line 56: can't you find some better method on doing this. Like having broadcast method pushed on interceptor.save does not look good. 

Also, we have peerOriginator which is used only for printing out the trace.

sstanculeanu: on this interceptor, yes, peerOriginator is only for print.. but it is useful for heartbeat related messages

indeed, having the broadcast method on Save seems odd, but only on Save we know that the intercepted data is valid

consensus/spos/bls/v2/subroundBlock.go
sstanculeanu: yes, this should cover the private pools scenario

we decided to broadcast these transactions on the leader, with the proposed header.. not sure when can be done earlier

AdoAdoAdo (dismissed):

sasurobert (dismissed):

sasurobert (approved):

AdoAdoAdo (approved):

PR #7531, opened 2025-12-11
add shard trigger v3

ssd04:
- Add shard trigger registry v3 

## Testing procedure
- System tests with restarts

## Pre-requisites

### DIFF
--- a/epochStart/shardchain/triggerRegistry.go
+++ b/epochStart/shardchain/triggerRegistry.go
@@ -46,7 +46,12 @@ func (t *trigger) LoadState(key []byte) error {
 
 // UnmarshalTrigger unmarshalls the trigger
 func UnmarshalTrigger(marshaller marshal.Marshalizer, data []byte) (data.TriggerRegistryHandler, error) {
-	trig, err := unmarshalTriggerV2(marshaller, data)
+	trig, err := unmarshalTriggerV3(marshaller, data)
+	if err == nil {
+		return trig, nil
+	}
+
+	trig, err = unmarshalTriggerV2(marshaller, data)
 	if err == nil {
 		return trig, nil
 	}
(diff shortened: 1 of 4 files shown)


epochStart/shardchain/triggerRegistry.go
```diff
@@ -46,7 +46,12 @@ func (t *trigger) LoadState(key []byte) error {
 
 // UnmarshalTrigger unmarshalls the trigger
 func UnmarshalTrigger(marshaller marshal.Marshalizer, data []byte) (data.TriggerRegistryHandler, error) {
```
mariusmihaic, line 48: Perhaps add a unit test for backwards compatibility ?

AdoAdoAdo (dismissed):

epochStart/shardchain/triggerRegistry_test.go
```diff
@@ -135,30 +136,167 @@ func TestTrigger_LoadStateBackwardsCompatibility(t *testing.T) {
...
+
+		err = epochStartTrigger2.LoadState(key)
+		require.Nil(t, err)
+		require.Equal(t, epochStartTrigger1, epochStartTrigger2)
+	})
+
+	t.Run("header v1", func(t *testing.T) {
+		t.Parallel()
+
+		marshaller := &marshal.GogoProtoMarshalizer{}
```
mariusmihaic: I think you can move this in `createMockShardEpochStartTriggerArguments`

epochStart/shardchain/triggerRegistry_test.go
```diff
@@ -135,30 +136,167 @@ func TestTrigger_LoadStateBackwardsCompatibility(t *testing.T) {
...
+
+		err = bootStorer.Put(trigInternalKey, d)
+		require.Nil(t, err)
+
+		err = epochStartTrigger2.LoadState(key)
+		require.Nil(t, err)
+		require.Equal(t, epochStartTrigger1, epochStartTrigger2)
+	})
+
+	t.Run("header v1", func(t *testing.T) {
```
mariusmihaic, line 167: I'm personally not a big fan of this much duplicated code.
You could've made a base test func with interfaces for Header + trigger and pass actual pointers to v1,v2,v3

epochStart/shardchain/triggerRegistry_test.go
ssd04: right, it was already proto there

epochStart/shardchain/triggerRegistry_test.go
ssd04: removed some of the duplicated code; i prefer it like this, to have it stated explicitly (even if there is some duplicated code in test) not to have too many abstractions in tests, especially for the later changes.
For example, for this tests, it took me quite a while to figure out the issue due to the createDummy trigger and cloneTrigger funcs which are used only in tests

AdoAdoAdo (approved):

epochStart/shardchain/triggerRegistry_test.go
mariusmihaic: It is still 99% copy-pasted code with just a different struct init ...
Can be left as it is

mariusmihaic (approved):

PR #7540, opened 2025-12-12
Supernova TxPool: Fixed removing blocks on selection

danielradu10:
- At the moment, on selection **we remove all blocks with nonce greater than the given one.** The current parameter on selection is the nonce of the block **on which** the selection is built. However, it should be the nonce of the block **for which** the selection is built, which means that **we have to remove all blocks with nonce greater or equal with the given nonce.**
  
## Proposed changes
- fixed removing, fixed unit tests, integration tests, comments

## Pre-requisites

### DIFF
--- a/txcache/selectionTracker.go
+++ b/txcache/selectionTracker.go
@@ -414,7 +414,7 @@ func (st *selectionTracker) deriveVirtualSelectionSession(
 	defer st.mutTracker.Unlock()
 
 	if !shouldRemoveTrackedBlocks {
-		err := st.removeBlocksAboveNonceNoLock(nonce)
+		err := st.removeBlocksAboveOrEqualToNonceNoLock(nonce)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


txcache/selectionTracker.go
```diff
@@ -448,11 +449,11 @@ func (st *selectionTracker) deriveVirtualSelectionSession(
 	return computer.createVirtualSelectionSession(globalAccountsBreadcrumbs)
 }
 
-// removeBlocksAboveNonceNoLock removes blocks with nonce higher than the given nonce.
+// removeBlocksAboveNonceNoLock removes blocks with nonce higher or equal than the given nonce.
 // The removeBlocksAboveNonceNoLock is used on the deriveVirtualSelectionSession flow.
 func (st *selectionTracker) removeBlocksAboveNonceNoLock(nonce uint64) error {
 	for blockHash, tb := range st.blocks {
-		if tb.hasHigherNonce(nonce) {
+		if tb.hasSameNonceOrHigherThanGivenNonce(nonce) {
```
AdoAdoAdo, line 456: the name of the parent method is misleading, the method is called removeBlocksAboveNonceNoLock, but here you are trying to remove also the nonce, not only above it.

danielradu10: fixed

AdoAdoAdo (approved):

PR #7556, opened 2025-12-16
Late broadcast header and proof

miiu96:
- Late broadcast header and proof --> shards have to include the meta header in the next round not in the same round ( similar behavior how the real chain works)

## Pre-requisites

### DIFF
--- a/node/chainSimulator/chainSimulator.go
+++ b/node/chainSimulator/chainSimulator.go
@@ -394,13 +394,33 @@ func (s *simulator) ForceChangeOfEpoch() error {
 }
 
 func (s *simulator) allNodesCreateBlocks() error {
+	headers := make(map[uint32]*dtos.BroadcastData, len(s.handlers))
 	for _, node := range s.handlers {
 		// TODO MX-15150 remove this when we remove all goroutines
 		time.Sleep(2 * time.Millisecond)
 
-		err := node.CreateNewBlock()
+		pair, err := node.CreateNewBlock()
 		if err != nil {
 			return err
+		}
+		if pair == nil {
+			continue
+		}
+
+		headers[pair.Header.GetShardID()] = pair
+	}
+
+	for shardID, pair := range headers {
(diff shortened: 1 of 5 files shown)


node/chainSimulator/process/processor.go
```diff
@@ -202,65 +203,58 @@ func (creator *blocksCreator) CreateNewBlock() error {
...
-
 	err = messenger.BroadcastMiniBlocks(miniBlocks, leader.PubKey())
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	return messenger.BroadcastTransactions(transactions, leader.PubKey())
+	err = messenger.BroadcastTransactions(transactions, leader.PubKey())
+	return &dtos.BroadcastData{
```
raduchis, line 245: some broadcasts happended before this. the order of broadcasts changed, is it ok?

node/chainSimulator/chainSimulator.go
```diff
@@ -394,14 +394,34 @@ func (s *simulator) ForceChangeOfEpoch() error {
...
+			continue
+		}
+
+		headers[pair.Header.GetShardID()] = pair
+	}
+
+	for shardID, pair := range headers {
+		err := s.nodes[shardID].GetBroadcastMessenger().BroadcastHeader(pair.Header, pair.LeaderKey)
+		if err != nil {
+			return err
```
raduchis, line 418: before, if one shard errored, the next would be broadcasted.

node/chainSimulator/process/processor.go
miiu96: Yes, now the header and proof is broadcasted after all the shard creates blocks

node/chainSimulator/chainSimulator.go
miiu96: in case of an error we return, we should not have error here.

raduchis (approved):

axenteoctavian (approved):

PR #7574, opened 2025-12-18
Update root hash  chain simulator

miiu96:
- Fixes

## Pre-requisites

### DIFF
--- a/node/chainSimulator/process/processor.go
+++ b/node/chainSimulator/process/processor.go
@@ -79,9 +79,7 @@ func (creator *blocksCreator) createBlock(header data.HeaderHandler) (data.Heade
 }
 
 func (creator *blocksCreator) updateHeader(header data.HeaderHandler) error {
-	if check.IfNil(header) ||
-		check.IfNil(header.GetLastExecutionResultHandler()) ||
-		len(header.GetExecutionResultsHandlers()) == 0 {
+	if check.IfNil(header) || len(header.GetExecutionResultsHandlers()) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


node/chainSimulator/process/processor.go
```diff
@@ -62,16 +62,51 @@ func (creator *blocksCreator) createHeaderBasedOnRound(round uint64, nonce uint6
...
+	rootHash, err := creator.nodeHandler.GetStateComponents().AccountsAdapter().RootHash()
+	if err != nil {
+		return err
+	}
+
+	switch h := header.(type) {
+	case *dataBlock.HeaderV3:
+		if check.IfNil(h) || h.LastExecutionResult == nil {
+			return nil
+		}
```
axenteoctavian: - maybe check header is nil first thing in the function
- and maybe return some error because header should have execution results for header v3
```suggestion
		if h.LastExecutionResult == nil || len(h.ExecutionResults) == 0 {
			return SOME_ERROR
		}
```

node/chainSimulator/process/processor.go
```diff
@@ -63,16 +63,51 @@ func (creator *blocksCreator) createHeaderBasedOnRound(round uint64, nonce uint6
...
+		if len(h.ExecutionResults) > 0 {
+			h.ExecutionResults[len(h.ExecutionResults)-1].BaseExecutionResult.RootHash = rootHash
+			h.LastExecutionResult.ExecutionResult.RootHash = rootHash
+		}
+
+	case *dataBlock.MetaBlockV3:
+		if check.IfNil(h) || h.LastExecutionResult == nil {
+			return nil
```
raduchis: also here return error because you entered from a check with header.IsHeaderV3()

raduchis (changes requested):

node/chainSimulator/process/processor.go
miiu96: refactored

node/chainSimulator/process/processor.go
miiu96: refactored

raduchis (approved):

PR #7579, opened 2025-12-22
Fixes synchronized state

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


ssd04 (dismissed):

process/sync/baseSync.go
```diff
@@ -1160,6 +1131,58 @@ func (boot *baseBootstrap) prepareForSyncIfNeeded(syncingNonce uint64) error {
...
+	err = txPool.OnExecutedBlock(lastNotarizedExecutedHeader, rootHash)
+	if err != nil {
+		txPool.ResetTracker()
+		return 0, err
+	}
+
+	lastExecutionResultNonce := lastNotarizedExecutedHeader.GetNonce()
+
+	// in case there is a more recent execution result available, use it
+	lastExecutionResult := boot.chainHandler.GetLastExecutionResult()
```
sstanculeanu, line 1176: is there any chance that this happens right before we call txPool.OnExecutedBlock on L1167? would that be an ambiguous case where tx pool has last executed different than what this method returns?

sasurobert (dismissed):

process/sync/baseSync.go
AdoAdoAdo: I don't think that can happen.

The onExecute needs to be done with the latest notarized execution result not with the last available execution result. The last notarized execution result cannot change for the node if the node is not committing a new block that has passed consensus, either on the consensus flow, or on the sync flow. 

This here is preceding the sync flow.

process/missingData/missingDataResolver.go
```diff
@@ -219,15 +222,34 @@ func (r *Resolver) WaitForMissingData(timeout time.Duration) error {
 		return haveTime
 	}
 
+	err := r.blockDataRequester.IsDataPreparedForProcessing(stepHaveTime(checkMissingDataStep))
```
sasurobert, line 225: i do not think you need this line. Inside the for, it would be enough.

sstanculeanu (approved):

ssd04 (approved):

sasurobert (approved):

PR #7581, opened 2025-12-23
Fix multiple headers not being referenced

sstanculeanu:
- when some pending mini blocks from a different meta were added into the proposal, after selecting the outgoing transactions, the second meta header was not referenced on the proposal
  
## Proposed changes
- save more pending mini blocks and reference all new meta headers where needed

## Testing procedure
- with feat branch

## Pre-requisites


process/block/gasConsumption.go
```diff
@@ -125,6 +126,19 @@ func (gc *gasConsumption) AddIncomingMiniBlocks(
 			for _, mb := range miniBlocks[i:] {
 				hashStr := string(mb.GetHash())
 				gc.transactionsForPendingMiniBlocks[hashStr] = transactions[hashStr]
+
+				transactionsForMB, found := transactions[string(mb.GetHash())]
```
AdoAdoAdo: can you merge this line and L128?

both appear to be reading the same transactions

e.g do the gc.transactionsForPendingMiniBlocks[hashStr] = transactionsForMB after the !found check?
or do we need to set nil explicitly in the map?

process/block/gasConsumption.go
```diff
@@ -292,6 +308,13 @@ func (gc *gasConsumption) addPendingIncomingMiniBlocks() ([]data.MiniBlockHeader
 
 		addedMiniBlocks = append(addedMiniBlocks, mb)
 		lastIndexAdded = i
+
+		gasConsumedByIncomingAfter := gc.totalGasConsumed[incoming]
+		gasConsumedByMiniBlock := gasConsumedByIncomingAfter - gasConsumedByIncomingBefore
+		if gasConsumedByMiniBlock < gc.totalGasConsumed[pending] {
```
AdoAdoAdo: for the last pending miniblock I think this condition would be equal.

this relates to the comment below

// should never be false

but for the last one, it will be false.

maybe the conditon should be  `<=` gc.totalGasConsumed[pending]

process/block/gasConsumption.go
```diff
@@ -125,6 +126,19 @@ func (gc *gasConsumption) AddIncomingMiniBlocks(
...
+
+				transactionsForMB, found := transactions[string(mb.GetHash())]
+				if !found {
+					log.Warn("could not find transaction for pending mini block", "hash", mb.GetHash())
+					continue
+				}
+				gasConsumedByPendingMb, errCheckPending := gc.checkGasConsumedByMiniBlock(mb, transactionsForMB)
+				if errCheckPending != nil {
+					log.Warn("failed to check gas consumed by pending mini block", "hash", mb.GetHash(), "error", errCheckPending)
+					continue
```
AdoAdoAdo: I think it should break instead of continuing, as that means that it can create miniblock gaps when there's a problem, which would not be ok. 

Same with the above continue, I think it should break there as well, as we don't want to create gaps.

Or even better returning error instead of breaking, as we rely now on the computed pending miniblocks consumed gas in the next steps.

process/block/shardblockProposal.go
```diff
@@ -585,26 +607,51 @@ func (sp *shardProcessor) createProposalMiniBlocks(
...
+		extraHeadersReferenced[string(headerHash)] = header
+	}
+
+	err := sp.miniBlocksSelectionSession.AddMiniBlocksAndHashes(extraMiniBlocksAdded)
+	if err != nil {
+		return err
+	}
+
+	for referencedHash, referencedHeader := range extraHeadersReferenced {
+		sp.miniBlocksSelectionSession.AddReferencedHeader(referencedHeader, []byte(referencedHash))
```
AdoAdoAdo: the miniblockSelectionSession assumes the referenced headers are added in order, and here you are iterating over a map which does not ensure a specific iteration order.

Can you ensure that the AddReferencedHeader is called in the order of referencing the metablocks? maybe adding the reference into the above loop?

process/block/gasConsumption.go
```diff
@@ -471,6 +494,32 @@ func (gc *gasConsumption) TotalGasConsumedInShard(shard uint32) uint64 {
...
+	}
+
+	maxGasLimitPerBlock := gc.maxGasLimitPerBlock(incoming, gc.shardCoordinator.SelfId())
+	if maxGasLimitPerBlock <= totalGasConsumed {
+		return false
+	}
+
+	gasLeft := maxGasLimitPerBlock - totalGasConsumed
+	maxGasLimitPerMiniBlock := gc.maxGasLimitPerMiniBlock(gc.shardCoordinator.SelfId())
+	return totalGasToBeConsumedByPending+maxGasLimitPerMiniBlock < gasLeft
```
AdoAdoAdo: why do we add the maxGasLimitPerMiniBlock to the total gas to be consumed by pending?

process/block/shardblockProposal.go
```diff
@@ -585,26 +607,51 @@ func (sp *shardProcessor) createProposalMiniBlocks(
...
+	}
+
+	return nil
+}
+
+func getIndexOfPendingMiniBlock(
```
AdoAdoAdo: this is called getIndexOfPending.... but it is not getting any index right?

process/block/gasConsumption.go
sstanculeanu: updated, missed it

process/block/gasConsumption.go
sstanculeanu: right, updated

process/block/gasConsumption.go
sstanculeanu: updated

process/block/gasConsumption.go
sstanculeanu: this method checks if there is space for one more mini block left in the block.. checking against the max gas limit, as it does not receive a specific mini block to know how much it would consume

process/block/shardblockProposal.go
sstanculeanu: renamed

process/block/shardblockProposal.go
```diff
@@ -530,22 +538,36 @@ func (sp *shardProcessor) selectIncomingMiniBlocks(
 			return nil, errCreated
 		}
 
-		pendingMiniBlocks = append(pendingMiniBlocks, createIncomingMbsResult.PendingMiniBlocks...)
+		pendingMiniBlocks = append(pendingMiniBlocks, &pendingMiniBlocksAfterSelection{
+			headerHash:                 currentMetaBlockHash,
+			header:                     currentMetaBlock,
+			pendingMiniBlocksAndHashes: createIncomingMbsResult.PendingMiniBlocks,
+		})
```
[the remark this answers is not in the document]
sstanculeanu: should be ok

process/block/shardblockProposal.go
sstanculeanu: updated as suggested

process/block/gasConsumption.go
AdoAdoAdo: but the miniblock does not necessarily need to be of max size, it can be of any size below the maximum size.

sstanculeanu: removed it

AdoAdoAdo (approved):

ssd04 (approved):

PR #7622, opened 2026-01-16
Enable epochs metrics v2

sstanculeanu:
- new enable epochs are often forgotten to be added as metrics
  
## Proposed changes
- added an automatic mechanism to return all of them

## Testing procedure
- check network/enable-epochs-v2 endpoint matches everything from enableEpochs.toml

## Pre-requisites


api/groups/networkGroup.go
```diff
@@ -22,6 +22,7 @@ const (
 	getStatusPath          = "/status"
 	economicsPath          = "/economics"
 	enableEpochsPath       = "/enable-epochs"
+	enableEpochsV2Path     = "/enable-epochs-v2"
```
ssd04, line 25: :+1:

api/groups/networkGroup.go
```diff
@@ -210,6 +216,20 @@ func (ng *networkGroup) getEnableEpochs(c *gin.Context) {
...
 }
 
+// getEnableEpochsV2 returns all enable epoch flags with their activation epochs
+func (ng *networkGroup) getEnableEpochsV2(c *gin.Context) {
+	enableEpochsMetrics := ng.getFacade().StatusMetrics().EnableEpochsMetricsV2()
+
+	c.JSON(
+		http.StatusOK,
+		shared.GenericAPIResponse{
+			Data:  gin.H{"enableEpochs": enableEpochsMetrics},
```
ssd04, line 226: maybe "enableEpochsV2"? this part will be returned in the output?

common/enablers/enableEpochsHandler.go
```diff
@@ -52,818 +53,954 @@ func (handler *enableEpochsHandler) createAllFlagsMap() {
 			isActiveInEpoch: func(epoch uint32) bool {
 				return epoch >= handler.enableEpochsConfig.SCDeployEnableEpoch
 			},
-			activationEpoch: handler.enableEpochsConfig.SCDeployEnableEpoch,
+			activationEpoch:     handler.enableEpochsConfig.SCDeployEnableEpoch,
+			activationEpochName: "SCDeployEnableEpoch",
```
ssd04, line 57: suggestion to consider removing `EnableEpoch` sufix from all metric names

api/groups/networkGroup.go
sstanculeanu: should be ok enableEpochs. yes, it is returned in the output, the main json field

common/enablers/enableEpochsHandler.go
sstanculeanu: the reason why I tried to keep it like this was to match 100% the toml file. also, some of them are DisableEpoch

ssd04 (approved):

common/enablers/enableEpochsHandler.go
```diff
@@ -18,8 +18,9 @@ var log = logger.GetOrCreate("common/enablers")
 type flagEnabledInEpoch = func(epoch uint32) bool
 
 type flagHandler struct {
-	isActiveInEpoch flagEnabledInEpoch
-	activationEpoch uint32
+	isActiveInEpoch     flagEnabledInEpoch
+	activationEpoch     uint32
+	activationEpochName string
```
mariusmihaic, line 23: [Not necessarily to be done in this PR, since I think it would require a big refactor]

Perhaps we couldn've made an `Activation Epoch` type that provides both `Value() uint32` and `Name() string` so we don't need to fill both this types everywhere

mariusmihaic (approved):

PR #7623, opened 2026-01-16
fix allPendingMbsAreProposed

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/consensus/spos/subround.go
+++ b/consensus/spos/subround.go
@@ -1,6 +1,7 @@
 package spos
 
 import (
+	"bytes"
 	"context"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


sstanculeanu (approved):

sstanculeanu (dismissed):

consensus/spos/subround.go
```diff
@@ -239,6 +240,31 @@ func (sr *Subround) ConsensusChannel() chan bool {
...
+	}
+	competingBlockNonce := prevBlock.GetNonce() + 1
+	proof, err := sr.EquivalentProofsPool().GetProofByNonce(competingBlockNonce, sr.ShardCoordinator().SelfId())
+	if err != nil || check.IfNil(proof) {
+		return false
+	}
+
+	consensusBlockHash := sr.GetData()
+	if len(consensusBlockHash) == 0 {
+		return true
```
ssd04, line 257: is this ok to be `true`? should we consider that there is a proof if no header sent or received?

sstanculeanu (dismissed):

consensus/spos/subround.go
AdoAdoAdo: this would be the case where the proof is already available but the consensus state doesn't yet have the header for this consensus round, so I would say it needs to be true (e.g skip new proposer proposing a block if there is already a proof for a competing block)

ssd04 (approved):

sstanculeanu (approved):

PR #7625, opened 2026-01-19
MX-16926 Allign the inclusion  estimator tBase with  round start

mradian1:
- align the base time for a pending execution result with the round time
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/estimator/executionResultInclusionEstimator.go
```diff
@@ -63,6 +63,11 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
 		return allowed
 	}
 
+	safetyMargin := uint64(erie.cfg.SafetyMargin)
+	if erie.cfg.SafetyMargin > 100 {
```
ssd04: add a const here?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -63,6 +63,11 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
...
+		safetyMargin -= 100
```
ssd04: what is the usage here? why reducing value to 1 if it's 101?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -99,29 +104,34 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
...
+
+		// Align previous estimated tBase with start of round if needed
+		estimatedTBase = max(estimatedTBase, tRoundNs)
+		estimatedTBase, overflow = bits.Add64(estimatedTBase, currentEstimatedTime, 0)
 		if overflow != 0 {
 			log.Debug("ExecutionResultInclusionEstimator: overflow detected in block transactions time estimation",
-				"estimatedTime", estimatedTime,
+				"estimatedTBase", estimatedTBase,
 				"currentEstimatedTime", currentEstimatedTime)
 			return i
```
ssd04: maybe update the name of the variable `i` here? it's not very clear at this point what is the meaning/usage? is it an index or execution result order?

process/estimator/executionResultInclusionEstimator_test.go
ssd04, line 1: add unit test for safety margin check?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -63,6 +63,11 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
 		return allowed
 	}
 
+	safetyMargin := uint64(erie.cfg.SafetyMargin)
```
mariusmihaic: Redundant cast here
```suggestion
	safetyMargin := erie.cfg.SafetyMargin
```

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -31,6 +32,14 @@ func TestEstimatorCreation(t *testing.T) {
 		require.False(t, erie.IsInterfaceNil(), "IsInterfaceNil() should return false for non-nil interface")
 	})
 
+	t.Run("Nil RoundHandler", func(t *testing.T) {
+		t.Parallel()
+		var erie *ExecutionResultInclusionEstimator
+		cfg := config.ExecutionResultInclusionEstimatorConfig{}
+		erie = NewExecutionResultInclusionEstimator(cfg, nil)
+		require.True(t, erie.IsInterfaceNil(), "IsInterfaceNil() should return true if RoundHandler is nil")
+	})
```
mariusmihaic, line 41: This seems wrong, the constructor logic will return nil instance instead of error

```go
func NewExecutionResultInclusionEstimator(cfg config.ExecutionResultInclusionEstimatorConfig, roundHandler RoundHandler) *ExecutionResultInclusionEstimator {
	if check.IfNil(roundHandler) {
		log.Error("NewExecutionResultInclusionEstimator: nil roundHandler")
		return nil
	}
```

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -234,8 +243,8 @@ func TestOverflowProtection(t *testing.T) {
 		}
 		erie := NewExecutionResultInclusionEstimator(cfg, roundHandler)
 		pending := []data.BaseExecutionResultHandler{
-			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 1, HeaderRound: 1, GasUsed: math.MaxUint64 / erie.tGas}}, // This will bring estimatedTime close to max
-			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 2, HeaderRound: 2, GasUsed: 2}},                          // This will cause overflow
+			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 1, HeaderRound: 1, GasUsed: (math.MaxUint64 - 1) / erie.tGas}}, // This will bring estimatedTime close to max
```
mariusmihaic: What happens in a test with old input variables? (can we also add that and check everything is expected?)

```go

GasUsed: math.MaxUint64 / erie.tGas
```

process/estimator/executionResultInclusionEstimator_test.go
```diff
@@ -372,6 +381,61 @@ func TestDecide_EdgeCases(t *testing.T) {
...
+			ProposedInRound:  0,
+		}
+		// third execution result doesn't finish in time since it is available only in round 4 and with margin goes beyond round 5 start
+		pending := []data.BaseExecutionResultHandler{
+			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 2, HeaderRound: 2, GasUsed: 50000000}},
+			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 3, HeaderRound: 3, GasUsed: 20000000}},
+			&block.ExecutionResult{BaseExecutionResult: &block.BaseExecutionResult{HeaderNonce: 4, HeaderRound: 4, GasUsed: 100000000}},
+		}
+		got := erie.Decide(lastNotarised, pending, roundNow)
+		require.Equal(t, 2, got)
```
mariusmihaic, line 442: Can we actually check which ones have been selected?

mariusmihaic (commented):
+ comments from @ssd04.
I also suggest that @AdoAdoAdo takes a look into this

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -99,29 +104,34 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
 			return i
 		}
 
-		estimatedTime, overflow = bits.Add64(estimatedTime, currentEstimatedTime, 0)
+		// Round timestamp for current execution result, since it is the time when the execution result becomes available
```
AdoAdoAdo, line 108: .. since it is the time when the transactions that produced this execution result become available.

process/estimator/executionResultInclusionEstimator.go
AdoAdoAdo: so now you don't want to compute the estimated time together with margin, but compute only the margin time and handle it separately?

mradian1: safety margin provided is in fact 100 + delta (110 for example)
This was because the initial algorithm  used to apply the margin to the total estimated time (current pending execution result as well as all the pending execution results before it) therefore:
estimated time += current
estimated time with safety = estimated time * (100 + delta) /100

The new algorithm accumulates the moment in time in order to compare it to the start of round (it is not enough to add the estimated times, we must take into account the round when the transaction has become available), so we cannot apply the margin in the same way. Moreover, now we apply the safety margin only to the current execution result and since it has already been added to the accumulated value, what needs to be considered in order to calculate the time with margin is the delta * current estimated time   

estimated time += current (needed for the next step of the algorithm)
safety = current * delta /100
estimated time with safety = estimated time + current *delta /100

process/estimator/executionResultInclusionEstimator_test.go
mradian1: we know which ones are selected: the erie processes the pending execution results in the order they were provided and returns the number of execution results that can be included, so the answer is: the first "got" execution results.

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator_test.go
mradian1: this is how it was intended: the estimator doesn't return error. it returns nil and the processor creator is the one that verifies and issues an error if that is the case

process/estimator/executionResultInclusionEstimator.go
mradian1: removed

process/estimator/executionResultInclusionEstimator_test.go
mradian1: it overflows in a different place than where it is intended

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: done. introduced nominalSafetyMargin = 100

process/estimator/executionResultInclusionEstimator_test.go
mradian1: done

process/estimator/executionResultInclusionEstimator_test.go
mradian1: with the old values, the overflow is reached in a different point, which is covered by another test

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
mradian1: done

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -63,9 +63,10 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
 		return allowed
 	}
 
-	safetyMargin := uint64(erie.cfg.SafetyMargin)
-	if erie.cfg.SafetyMargin > 100 {
-		safetyMargin -= 100
+	nominalSafetyMargin := uint64(100)
```
mariusmihaic, line 66: Could've made this a const and also reuse it for tests

mariusmihaic (approved):

AdoAdoAdo (approved):

ssd04 (approved):

PR #7635, opened 2026-01-23
Fixes unexecutable transaction

miiu96:
• Remove unexecutable transactions from the transaction cache once the block execution result is notarized
• Return the correct transaction status on the /transaction/:txHash endpoint for this case

## Pre-requisites

### DIFF
--- a/node/external/transactionAPI/apiTransactionProcessor.go
+++ b/node/external/transactionAPI/apiTransactionProcessor.go
@@ -1,6 +1,7 @@
 package transactionAPI
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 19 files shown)


node/external/transactionAPI/apiTransactionProcessor.go
```diff
@@ -748,20 +749,53 @@ func (atp *apiTransactionProcessor) computeTimestampForRoundAsMs(round uint64) i
...
-		return errG
+		return false, errG
+	}
+
+	executionResultsBytes, err := executionResultsStorer.GetFromEpoch(headerHash, miniblockMetadata.GetEpoch())
+	if err != nil {
+		return false, err
+	}
+
+	if atp.shardCoordinator.SelfId() == core.MetachainShardId {
```
sstanculeanu, line 769: move this after epoch check? before GetStorer

node/external/transactionAPI/apiTransactionProcessor.go
```diff
@@ -748,20 +749,53 @@ func (atp *apiTransactionProcessor) computeTimestampForRoundAsMs(round uint64) i
...
+	}
+	mbHeaders, err := atp.getMbHeadersFromExecutionResultBytes(executionResultsBytes)
+	if err != nil {
+		return false, err
 	}
 
-	_, err := executionResultsStorer.GetFromEpoch(headerHash, miniblockMetadata.GetEpoch())
-	return err
+	// check if the transaction miniblock metadata has a mb header on execution result
+	// if yes - the transaction was execution
```
sstanculeanu:
```suggestion
	// if yes - the transaction was executed
```

process/block/baseProcess.go
```diff
@@ -1245,6 +1245,21 @@ func (bp *baseProcessor) cleanupPools(headerHandler data.HeaderHandler) {
...
+
+func (bp *baseProcessor) cleanupUnexecutableTxsFromPool(headerHash []byte) {
+	unexecutableTxHashes, err := common.GetCachedUnexecutableTxHashes(bp.dataPool.PostProcessTransactions(), headerHash)
+	if err != nil {
+		log.Warn("bp.cleanupPools cannot get unexecutable tx hashes", "err", err)
+		return
+	}
+
+	for _, txHash := range unexecutableTxHashes {
+		cacheID := process.ShardCacherIdentifier(bp.shardCoordinator.SelfId(), bp.shardCoordinator.SelfId())
```
sstanculeanu, line 1261: will this work for cross shard? 🤔

node/external/transactionAPI/apiTransactionProcessor.go
miiu96: This function also verifies that the transaction returned by the endpoint is included in a block that has an execution results.
If no execution result is found, the function returns an error, since the transaction may not have been executed yet.
So, If I move the check upper the check will be skipped.

node/external/transactionAPI/apiTransactionProcessor.go
miiu96: fixed

process/block/baseProcess.go
miiu96: this will work also for the cross shard transactions, because the miniblock with the proposed transactions always is an intra shard miniblocks. 
after execution we will have new miniblocks that can be cross shard.

process/block/baseProcess.go
```diff
@@ -1827,17 +1842,17 @@ func (bp *baseProcessor) saveTxsToStorage(dataPool dataRetriever.ShardedDataCach
 	for _, txHash := range txHashes {
```
AdoAdoAdo, line 1842: can we still at least log the error?

process/block/baseProcess.go
sstanculeanu: the miniblock yes, but the transactions I think are still in the proper pool, not all of them in self-self

process/block/baseProcess.go
miiu96: fixed

sstanculeanu (approved):

process/block/baseProcess.go
AdoAdoAdo: there is a common cache for all transactions with senders in own shard, and only those transactions can be un-executable.

AdoAdoAdo (approved):

PR #7636, opened 2026-01-23
FIX:  Economics metric

mariusmihaic:
- FIX: economics metric for fees and dev fees
  
## Proposed changes
- Last execution results for epoch change block that was used to get the metrics did not contain the actual economics change data. 
- Only the epoch proposed block contains that data in previous execution result, so we had the take the metrics from there

## Testing procedure
- 
- 
- 

## Pre-requisites


axenteoctavian (approved):

sstanculeanu (approved):

AdoAdoAdo (commented):
There is still an edge case that is not covered when there is at least one block between epoch change proposed and start of epoch block.

Will add the fix in another PR, and will be merged together.

AdoAdoAdo:
changes are now part of #7654

PR #7665, opened 2026-02-04
hasPendingChangeGuardianTransaction on global breadcrumbs

sstanculeanu:
- with the latest tx pool optimization, a second selection won't iterate through already selected transactions, causing improper selection of post-change guardian transactions
  
## Proposed changes
- pending change guardian flag should be global

## Testing procedure
- with feat branch

## Pre-requisites


txcache/globalAccountBreadcrumb.go
```diff
@@ -40,6 +41,7 @@ func (gab *globalAccountBreadcrumb) resetNonces() {
 // updateOnAddedBreadcrumb updates a global breadcrumb when a tracked block is added,
 // but should be used only after the validation of the proposed block passed.
 func (gab *globalAccountBreadcrumb) updateOnAddedBreadcrumb(receivedBreadcrumb *accountBreadcrumb) {
+	gab.hasPendingChangeGuardianTransaction = receivedBreadcrumb.hasPendingChangeGuardianTransaction
```
AdoAdoAdo, line 44: does this rely on the fact that we cannot have new txs selected for an account with the pending guardian change set/tracked?

In that case we need to make sure that the verification is properly done here:

`lastNoncePerSender, err := st.validateTrackedBlocksAndCompileBreadcrumbsNoLock(blockBody, tBlock, accountsProvider, latestExecutedHash)` where it should return error if e.g there are txs on a new breadcrumb when the previous global breadcrumb already has the flag set.

txcache/globalAccountBreadcrumb.go
```diff
@@ -64,6 +66,7 @@ func (gab *globalAccountBreadcrumb) extendRightNonceRange(receivedBreadcrumb *ac
 
 // updateOnRemovedBreadcrumbWithSameNonceOrBelow updates the global account breadcrumb when a block is removed on the OnExecutedBlock notification
 func (gab *globalAccountBreadcrumb) updateOnRemovedBreadcrumbWithSameNonceOrBelow(receivedBreadcrumb *accountBreadcrumb) (bool, error) {
+	gab.hasPendingChangeGuardianTransaction = receivedBreadcrumb.hasPendingChangeGuardianTransaction
```
AdoAdoAdo: The received breadcrumb should be the one being removed, so we should check if this one has the flag `set`, and only then `reset` it on the global account breadcrumb, otherwise nothing to do on the flag.

txcache/globalAccountBreadcrumb.go
```diff
@@ -97,6 +100,7 @@ func (gab *globalAccountBreadcrumb) reduceLeftNonceRange(receivedBreadcrumb *acc
 // updateOnRemoveBreadcrumbWithSameNonceOrAbove updates the global account breadcrumb when a block is removed on the OnProposedBlock notification,
 // but should be used only after the validation of the proposed block passed.
 func (gab *globalAccountBreadcrumb) updateOnRemoveBreadcrumbWithSameNonceOrAbove(receivedBreadcrumb *accountBreadcrumb) (bool, error) {
+	gab.hasPendingChangeGuardianTransaction = receivedBreadcrumb.hasPendingChangeGuardianTransaction
```
AdoAdoAdo: same here, the received breadcrumb should be the one being removed, so we should check if this one has the flag `set`, and only then `reset` it on the global account breadcrumb, otherwise nothing to do on the flag.

txcache/globalAccountBreadcrumb.go
sstanculeanu: updated

txcache/selectionTracker.go
```diff
@@ -231,6 +237,20 @@ func (st *selectionTracker) validateTrackedBlocksAndCompileBreadcrumbsNoLock(
 	return lastNoncePerSender, nil
 }
 
+func (st *selectionTracker) validateTransactionsInBlock(
+	txs []*WrappedTransaction,
+) error {
+	globalAccountsBreadcrumbs := st.globalBreadcrumbsCompiler.getGlobalBreadcrumbs()
+	for _, tx := range txs {
```
AdoAdoAdo, line 244: is it possible to move the check on the breadcrumb validation instead?

PR #7666, opened 2026-02-04
Signal execution for sync block

raduchis:
- meta synchronizing processed only 2-3 blocks per second because of the 400ms timeout on the sync
  
## Proposed changes
- add a channel to signal when the processing finished a new header

## Testing procedure
- run an internal testnet and remove dbs / stop and restart observers and validators and the synchronization shoudl catch up faster than 2-3 blocks per second

## Pre-requisites


sstanculeanu (dismissed):

process/sync/baseSync.go
```diff
@@ -2257,6 +2271,11 @@ func (boot *baseBootstrap) GetNodeState() common.NodeState {
 	return common.NsNotSynchronized
 }
 
+// GetSignalProcessCompletionChan returns the channel used to signal the sync loop after execution completes
+func (boot *baseBootstrap) GetSignalProcessCompletionChan() chan uint64 {
```
ssd04: this does not seem to be used, the chan is set here from execution manager, no need to export it from here

raduchis: yes, removed - it was only used in a test

danidrasovean (dismissed):

ssd04 (approved):

AdoAdoAdo (approved):

sstanculeanu (approved):

PR #7669, opened 2026-02-04
Check guarded relayer on interceptor

sstanculeanu:
- guarded relayer check can be added on interceptor as well
  
## Proposed changes
- add the check

## Testing procedure
- with feat branch, relayed tx v3 with guarded relayer

## Pre-requisites

### DIFF
--- a/process/transaction/interceptedTransaction.go
+++ b/process/transaction/interceptedTransaction.go
@@ -16,6 +16,7 @@ import (
 	"github.com/multiversx/mx-chain-go/common"
 	"github.com/multiversx/mx-chain-go/process"
 	"github.com/multiversx/mx-chain-go/sharding"
+	"github.com/multiversx/mx-chain-go/state"
 	logger "github.com/multiversx/mx-chain-logger-go"
 )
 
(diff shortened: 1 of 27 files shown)


process/transaction/interceptedTransaction.go
```diff
@@ -289,6 +301,28 @@ func (inTx *InterceptedTransaction) verifyIfRelayedTxV3(tx *transaction.Transact
 	return nil
 }
 
+func (inTx *InterceptedTransaction) checkGuardedRelayer(tx *transaction.Transaction) error {
```
AdoAdoAdo: can you instead move the checks into `func (txv *txValidator) CheckTxValidity(interceptedTx process.InterceptedTransactionHandler) error` ?

this should already be called on the intercepted transactions, and has access to the accounts adapter, so no need to add extra dependencies

sstanculeanu: indeed an easier approach.. pushed

AdoAdoAdo (approved):

ssd04 (approved):

PR #7672, opened 2026-02-05
State snapshot refactor

BeniaminDrasovean:
- If there are multiple dbs which contain the state, during a snapshot it will check for trie nodes starting from `[n-1, n-2, ..., n-k]`. If  `k` is high, this will translate in a lot of read accesses. 
  
## Proposed changes
- In order to reduce read accesses during state snapshot, if a trie node is found in n-k, start the search for all the following children nodes from n-k, not from n-1.
- Remove redundant mutex Lock/Unlock

## Testing procedure
- Internal testnets and on some mainnet observers

## Pre-requisites


trie/branchNode.go
```diff
@@ -291,56 +291,58 @@ func (bn *branchNode) commitDirty(level byte, maxTrieLevelInMemory uint, originD
 	return nil
 }
 
+// TODO refactor long parameter list
```
raduchis, line 294: 👍

trie/branchNode.go
```diff
@@ -291,56 +291,58 @@ func (bn *branchNode) commitDirty(level byte, maxTrieLevelInMemory uint, originD
...
+		err = child.commitSnapshot(db, maxEpochToSearchFromForChild, leavesChan, missingNodesChan, ctx, stats, idleProvider, encChild, depthLevel+1)
+		if err != nil {
+			return err
+		}
 
-func (bn *branchNode) saveToStorage(targetDb common.BaseStorer, stats common.TrieStatisticsHandler, depthLevel int) error {
-	nodeSize, err := encodeNodeAndCommitToDB(bn, targetDb)
-	if err != nil {
-		return err
+		err = db.PutInEpochWithoutCache(bn.EncodedChildren[i], encChild)
```
ssd04: this part looks very similar with the part in extension node, can we extract the common code?

trie/branchNode.go
```diff
@@ -291,56 +291,58 @@ func (bn *branchNode) commitDirty(level byte, maxTrieLevelInMemory uint, originD
...
+		if err != nil {
+			return err
+		}
 	}
 
-	stats.AddBranchNode(depthLevel, uint64(nodeSize))
-
-	bn.removeChildrenPointers()
```
ssd04, line 343: this is not needed anymore?

trie/trieStorageManager.go
```diff
@@ -158,8 +158,8 @@ func (tsm *trieStorageManager) cleanupChans() {
 
 // Get checks all the storers for the given key, and returns it if it is found
 func (tsm *trieStorageManager) Get(key []byte) ([]byte, error) {
-	tsm.storageOperationMutex.Lock()
-	defer tsm.storageOperationMutex.Unlock()
+	tsm.storageOperationMutex.RLock()
+	defer tsm.storageOperationMutex.RUnlock()
 
 	if tsm.closed {
```
ssd04: if mutex protection is for `tsm.closed`, unlock it specifically without defer? to avoid mutex protection over storer operation

ssd04: same for the other methods

ssd04: or just use `IsClosed()` ? or an atomic bool?

trie/branchNode.go
```diff
@@ -291,56 +291,58 @@ func (bn *branchNode) commitDirty(level byte, maxTrieLevelInMemory uint, originD
...
-		err = bn.children[i].commitSnapshot(db, leavesChan, missingNodesChan, ctx, stats, idleProvider, depthLevel+1)
+		child, err := decodeNode(encChild, bn.marsh, bn.hasher)
 		if err != nil {
 			return err
 		}
-	}
 
-	return bn.saveToStorage(db, stats, depthLevel)
-}
+		maxEpochToSearchFromForChild := foundInEpoch + 1
```
raduchis: do you need the +1?

trie/branchNode.go
raduchis: 👍

trie/extensionNode.go
```diff
@@ -211,48 +211,50 @@ func (en *extensionNode) commitDirty(level byte, maxTrieLevelInMemory uint, orig
...
 	if err != nil {
 		return err
 	}
 
-	if !childIsMissing {
-		err = en.child.commitSnapshot(db, leavesChan, missingNodesChan, ctx, stats, idleProvider, depthLevel+1)
-		if err != nil {
-			return err
-		}
+	maxEpochToSearchFromForChild := foundInEpoch + 1
```
raduchis: do you need epoch + 1?

state/snapshotStatistics.go
```diff
@@ -89,6 +91,11 @@ func (ss *snapshotStatistics) GetSnapshotNumNodes() uint64 {
 	return ss.trieStatisticsCollector.GetNumNodes()
 }
 
+// IncrementThrottlerWaits increments the number of throttler waits. This should not be called in a concurrent context
+func (ss *snapshotStatistics) IncrementThrottlerWaits() {
+	ss.numThrottlerWaits++
```
raduchis: maybe use an Atomic Int or a mutex? There is a description but who knows

raduchis (commented):
small suggestions

trie/branchNode.go
BeniaminDrasovean: Done

trie/branchNode.go
BeniaminDrasovean: No, because we don't set the children pointers anymore, so no need to remove them.

trie/trieStorageManager.go
BeniaminDrasovean: Removed the mutex protection entirely for the storer operation. This can be done because the mainStorer handles concurrent accesses anyway.

trie/branchNode.go
BeniaminDrasovean: Yes, because `GetFromOldEpochsWithoutAddingToCache` will retrieve starting from n-1. So if a node is found in n, we should start from n. 
Refactored so that `GetFromOldEpochsWithoutAddingToCache` will start from n, so "magic numbers" are not needed anymore

trie/extensionNode.go
BeniaminDrasovean: Refactored

state/snapshotStatistics.go
BeniaminDrasovean: Done

trie/snapshotTrieStorageManager.go
```diff
@@ -79,33 +76,35 @@ func (stsm *snapshotTrieStorageManager) putInPreviousStorerIfAbsent(key []byte,
...
 }
 
-// Put adds the given value to the main storer
-func (stsm *snapshotTrieStorageManager) Put(key, data []byte) error {
-	stsm.storageOperationMutex.RLock()
-	if stsm.closed {
-		stsm.storageOperationMutex.RUnlock()
+// PutInEpochWithoutCache adds the given (key, data) in the current epoch without adding it to cache
+func (stsm *snapshotTrieStorageManager) PutInEpochWithoutCache(key, data []byte) error {
+	stsm.tsm.storageOperationMutex.RLock()
```
ssd04: this seems to be the only place where we still use the mutex for closed check, is it still needed?

ssd04: there is also one check in `trieStorageManagerInEpoch.Get`, can we remove it also?

BeniaminDrasovean: Can be removed, missed them the first time.

danidrasovean (approved):

ssd04 (approved):

raduchis (approved):

PR #7674, opened 2026-02-05
added commitInMemory method for synching optimization

raduchis:
- synching blocks on supernova is fast -> 10 blocks per second but it is still kinda slow
  
## Proposed changes
- looking in the logs for a meta node, commit account state took ~50-70 ms. Created an commitInMemory every 9 out of 10 blocks and the 10th block do a normal commit.

## Testing procedure
- Remove DB with and without the current branch and see that the current branch synchronizez ~5x faster

## Pre-requisites


process/block/baseProcess.go
```diff
@@ -2109,21 +2124,99 @@ func (bp *baseProcessor) RevertAccountsDBToSnapshot(accountsSnapshot map[state.A
...
 	return bp.commit()
 }
 
+// shouldUseSyncCommitOptimization checks if the node is syncing and should use
+// the in-memory commit optimization to improve sync speed.
+func (bp *baseProcessor) shouldUseSyncCommitOptimization(headerHandler data.HeaderHandler) bool {
+	// Disabled if syncCommitInterval is 0
+	if bp.syncCommitInterval == 0 {
+		return false
+	}
```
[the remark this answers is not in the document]
BeniaminDrasovean: Or if the `SetSyncCommitInterval` is meant to be used only in tests, move it to the `export_test` file.

state/accountsDB.go
```diff
@@ -915,6 +915,52 @@ func (adb *AccountsDB) Commit() ([]byte, error) {
...
+	adb.mutOp.Lock()
+	defer func() {
+		adb.mutOp.Unlock()
+		adb.loadCodeMeasurements.resetAndPrint()
+	}()
+
+	return adb.commitInMemory()
+}
+
+func (adb *AccountsDB) commitInMemory() ([]byte, error) {
```
BeniaminDrasovean, line 931: There a couple issues with this approach:
1. The dataTries holder reset drops the reference to all of the changed data tries. When an account is loaded, it will have the new root hash, so it will try to load the data trie with that root hash from the db, but since the dataTrie was not commited, the recreate from db will fail. To fix this, you can remove the reset, but
2. If a revert occurs, a recreateFromDb will happen - and will fail

BeniaminDrasovean (changes requested):

state/accountsDB.go
BeniaminDrasovean: Another solution would be to commit the data tries, but into a cache.

process/block/baseProcess.go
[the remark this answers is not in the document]
raduchis: moved to export_test

state/accountsDB.go
raduchis: Removed the reset from the commitInMemory. It should only happen on the normal Commit. Reverts should not happen because we are on sync and we do not do the commitInMemory if we are closer than 50 nonces from the consensus.

process/block/baseProcess.go
```diff
@@ -2109,21 +2124,99 @@ func (bp *baseProcessor) RevertAccountsDBToSnapshot(accountsSnapshot map[state.A
...
 			"time [s]", elapsedTime,
 			"header nonce", headerHandler.GetNonce(),
+			"in memory", inMemory,
 		)
 	}()
 
 	if headerHandler.IsStartOfEpochBlock() {
+		bp.resetSyncCommitCounter()
 		return bp.commitInLastEpoch(headerHandler.GetEpoch())
 	}
```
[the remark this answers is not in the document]
raduchis: fixed

state/accountsDB.go
BeniaminDrasovean: Ok. This can be tested with an observer node started on mainnet with StartInEpoch, right?

BeniaminDrasovean (commented):
Another issue came to mind: The in-epoch pruning will not work with this approach. Also, state snapshot might be incompatible, if a snapshot is triggered while the tries are not yet committed to db.

PR #7701, opened 2026-02-11
Import db check in round handler

ssd04:
- When starting the node we have to determine if we start with supernova parameters in round handler or with initial params
- This is done based on supernova activation time
  
## Proposed changes
- Added additional check to not consider supernova force activation at start if in import db mode

## Testing procedure
- ImportDB test

## Pre-requisites

### DIFF
--- a/epochStart/metachain/economics.go
+++ b/epochStart/metachain/economics.go
@@ -492,7 +492,15 @@ func (e *economics) adjustRewardsPerBlockWithLeaderPercentage(
 
 // compute inflation rate from genesisTotalSupply and economics settings for that year
 func (e *economics) computeInflationBeforeSupernova(currentRound uint64, epoch uint32) float64 {
-	roundsPerDay := numberOfSecondsInDay / uint64(e.roundTime.TimeDuration().Seconds())
+	roundDurationInSec := e.roundTime.TimeDuration().Seconds()
+	if roundDurationInSec <= 0 {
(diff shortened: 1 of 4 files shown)


epochStart/metachain/economics.go
```diff
@@ -492,7 +492,15 @@ func (e *economics) adjustRewardsPerBlockWithLeaderPercentage(
 
 // compute inflation rate from genesisTotalSupply and economics settings for that year
 func (e *economics) computeInflationBeforeSupernova(currentRound uint64, epoch uint32) float64 {
-	roundsPerDay := numberOfSecondsInDay / uint64(e.roundTime.TimeDuration().Seconds())
+	roundDurationInSec := e.roundTime.TimeDuration().Seconds()
+	if roundDurationInSec <= 0 {
+		// this means that round duration is sub-seconds
+		// set it to 1 second
+		log.Error("computeInflationBeforeSupernova: sub second round time before supernova activation")
+		roundDurationInSec = 1
```
ssd04, line 500: - added special treatment to avoid divide by zero at any cost
- this case should not happen, in case it happens it should be treated as an error and fixed

AdoAdoAdo: for before Supernova, if we anyway set a default value, we might as well set it with 6s

AdoAdoAdo (approved):

sstanculeanu (approved):

PR #7705, opened 2026-02-11
MEDIUM-007: Committed-hash guard map can grow without bound on non-V3 commit path

sstanculeanu:
- for non v3 headers, the map was only filled with data but never deleted
  
## Proposed changes
- added exception for v3

## Testing procedure
- with feat branch

## Pre-requisites

### DIFF
--- a/process/asyncExecution/executionTrack/executionResultsTracker.go
+++ b/process/asyncExecution/executionTrack/executionResultsTracker.go
@@ -247,13 +247,20 @@ func (ert *executionResultsTracker) removeExecutionResultsFromMaps(executionResu
 // If the pending execution result for the given nonce has a different hash than the one that
 // passed consensus, remove it and all higher nonces from the tracker.
 // It also records the committed hash to prevent stale results from being added later.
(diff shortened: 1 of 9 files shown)


process/asyncExecution/executionTrack/executionResultsTracker.go
```diff
@@ -247,13 +247,20 @@ func (ert *executionResultsTracker) removeExecutionResultsFromMaps(executionResu
...
+		return
+	}
+
 	ert.mutex.Lock()
 	defer ert.mutex.Unlock()
-	// record the committed hash to prevent stale results being added later
-	ert.consensusCommittedHashes[headerNonce] = headerHash
 
-	pendingExecutionResult, err := ert.getPendingExecutionResultsByNonce(headerNonce)
+	if header.IsHeaderV3() {
```
ssd04, line 258: i don't think we need to call `CleanOnConsensusReached` on non-v3 patch in baseProcess.go, in `setCurrentBlockInfo`

ssd04 (approved):

AdoAdoAdo (approved):

PR #7737, opened 2026-02-18
Optimize Supernova proposal phase: batch trie reads, lightweight accounts, guardian fast-path

mihaieremia:
- The Supernova proposal phase requires account state lookups for 10k+ unique senders per block. The sequential trie read pattern (N individual lock acquisitions) creates significant mutex contention.
- Guardian checks during selection trigger full account loads (TrackableDataTrie + DataTrieLeafParser) even when only nonce/balance/guarded-status are needed, adding ~5-6ms of unnecessary allocation overhead at 10k accounts.

## Proposed changes
- **Batch trie reads**: `GetBatch()` on Patricia Merkle Trie — single lock acquisition for N keys instead of N individual locks. Reduces mutex overhead from O(N) to O(1).
- **Lightweight accounts**: New `lightAccountInfo` read-only type (nonce + balance + codeMetadata + rootHash) that skips `TrackableDataTrie` and `DataTrieLeafParser` allocation.
- **3-tier batch prefetch**: `PrefetchAccounts` tries light batch -> full batch -> sequential fallback. Proactively upgrades guarded accounts after light batch to avoid per-account trie reads during selection.
- **Two-phase guardian check**: `IsIncorrectlyGuarded` checks codeMetadata first (works on light accounts), only upgrades to full account when `VerifyGuardian` needs data trie access. `CreateFullAccountFromLight` skips redundant main trie re-read.
- **Parallel data retrieval**: Concurrent batch requests and chunk sending in the resolver/requester layer. `MaxBulkTransactionSize` increased from 256KB to 1MB (fewer round-trips for 3000+ tx blocks).
- **Broadcast timing**: Reduced `ExtraDelayForBroadcastBlockInfo` (120->20ms) and `ExtraDelayBetweenBroadcastMbsAndTxs` (100->10ms) for faster block propagation.

### Trie operation savings

At 10k unique senders per block (production Supernova scenario):

**Mutex lock acquisitions (trie `mutOperation`):**
| Path | Before | After | Savings |
|---|---|---|---|
| First selection (prefetch) | 10,000 individual locks | 1 batch lock | **99.99% fewer lock/unlock cycles** |
| Second selection (virtual session) | 10,000 individual locks | 1 batch lock | **99.99% fewer lock/unlock cycles** |

**Trie read operations:**
| Path | Before | After | Savings |
|---|---|---|---|
| Non-existent accounts (new senders) | 10k `Get()` calls returning nil | 1 `GetBatch()` call, nil entries returned in-place | **9,999 fewer lock cycles** |
| Existing accounts (nonce+balance only) | 10k `Get()` + 10k `CreateAccount` + 10k `Unmarshal` | 1 `GetBatch()` + 10k lightweight unmarshal (no `TrackableDataTrie`/`DataTrieLeafParser`) | **~30k fewer allocations** |
| Guardian upgrade (40% guarded) | 4k main trie re-reads + 4k data trie loads | 0 main trie re-reads + 4k data trie loads | **4,000 fewer main trie reads** |

**Object allocation savings (per selection round at 10k senders):**
| Object | Before (per account) | After (light path) | Savings |
|---|---|---|---|
| `TrackableDataTrie` | 10,000 | 0 (light accounts) | **10k allocations eliminated** |
| `DataTrieLeafParser` | 10,000 | 0 (light accounts) | **10k allocations eliminated** |
| `userAccount` wrapper | 10,000 | 0 (replaced by `lightAccountInfo`) | **10k allocations eliminated** |
| Total per round | ~30k heavy allocs | ~10k lightweight structs | **~67% fewer allocations** |

**Benchmark results (second selection, 10k senders):**

| Scenario | Time |
|---|---|
| Baseline (HEAD) | ~26.9ms |
| Optimized (0% guarded) | ~24.1ms (-10%) |
| Optimized (40% guarded, 4k accounts) | ~26.9ms (no regression) |

## Testing procedure
- If you could test it better on some internal networks and so on would be great, I was limited by the chain simulator + tested on my mac not a proper server...
- All existing unit tests pass (`state/...`, `process/block/preprocess/...`, `state/factory/...`)
- New unit tests: `CreateFullAccountFromLight`, `UnmarshalLightAccount`, `IsAccountGuarded`, `GetUserAccount` upgrade path
- Integration benchmark: `TestBenchmark_RealisticSecondSelectionWithRealTrie` (0% guarded)
- Integration benchmark: `TestBenchmark_RealisticSecondSelectionGuarded` (40% guarded, exercises `upgradeGuardedAccounts` + `CreateFullAccountFromLight`)
- Broadcast delay reductions need multi-node testnet validation (TODOs added in `common/constants.go`) - no idea how these delays were added but in general I guess the term "delay" in blockchain should not be there if we want low latency/real time :) I changed them just maybe so we can test with such values especially the size change for requesting transactions.

## Pre-requisites

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -66,12 +66,16 @@ const DisabledShardIDAsObserver = uint32(0xFFFFFFFF) - 7
 // in order to mark the transaction as valid.
 const MaxTxNonceDeltaAllowed = 100
 
-// MaxBulkTransactionSize specifies the maximum size of one bulk with txs which can be send over the network
+// MaxBulkTransactionSize specifies the maximum size of one bulk with txs which can be send over the network.
+// A typical simple EGLD transfer is ~200-250 bytes serialized, while a smart contract call can reach ~500 bytes.
+// With blocks containing up to 3000+ txs (~1.5MB at ~500 bytes/tx), a 1MB bulk size means at most 2 chunks
+// per response instead of 7 with the previous 256KB limit, significantly reducing round-trips during
+// consensus when validators need to fetch missing transactions from peers.
 // TODO convert this const into a var and read it from config when this code moves to another binary
-const MaxBulkTransactionSize = 1 << 18 // 256KB bulks
+const MaxBulkTransactionSize = 1 << 20 // 1MB bulks
 
 // MaxTxsToRequest specifies the maximum number of txs to request
-const MaxTxsToRequest = 1000
+const MaxTxsToRequest = 10000
 
 // NodesSetupJsonFileName specifies the name of the json file which contains the setup of the nodes
 const NodesSetupJsonFileName = "nodesSetup.json"
@@ -1021,11 +1025,14 @@ const MaxSoftwareVersionLengthInBytes = 10
 
 // ExtraDelayForBroadcastBlockInfo represents the number of seconds to wait since a block has been broadcast and the
 // moment when its components, like mini blocks and transactions, would be broadcast too
-const ExtraDelayForBroadcastBlockInfo = 120 * time.Millisecond
+// TODO: validate in multi-node testnet — reduced from 120ms to 20ms for Supernova; validators with >20ms
+// network latency may experience message ordering issues.
+const ExtraDelayForBroadcastBlockInfo = 20 * time.Millisecond
 
 // ExtraDelayBetweenBroadcastMbsAndTxs represents the number of seconds to wait since miniblocks have been broadcast
 // and the moment when theirs transactions would be broadcast too
-const ExtraDelayBetweenBroadcastMbsAndTxs = 100 * time.Millisecond
+// TODO: validate in multi-node testnet — reduced from 100ms to 10ms for Supernova.
+const ExtraDelayBetweenBroadcastMbsAndTxs = 10 * time.Millisecond
 
 // ExtraDelayForRequestBlockInfo represents the number of seconds to wait since a block has been received and the
 // moment when its components, like mini blocks and transactions, would be requested too if they are still missing
--- a/common/interface.go
+++ b/common/interface.go
@@ -465,6 +465,19 @@ type AccountNonceAndBalanceProvider interface {
 	IsInterfaceNil() bool
 }
 
+// TrieBatchGetter is an optional interface for tries that support batch reads with a single lock acquisition.
+// This reduces mutex overhead from O(N) lock/unlock cycles to O(1) for N keys.
+type TrieBatchGetter interface {
+	GetBatch(keys [][]byte) ([][]byte, []uint32, error)
+}
+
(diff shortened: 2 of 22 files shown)


mihaieremia:
Unrelated to the above, but I've been investigating how transactions flow across P2P and how validators request missing data during consensus. I also did some research into how Solana, NEAR, Sui, and Ethereum handle this.

It turns out we're the only chain that relies on a reactive request/response model for missing transaction hashes during consensus. This worked fine at low transaction counts, but the higher we push throughput, the worse it gets and the current design has a fundamental gap in the leader's transaction re-broadcast logic.

Dropping here a summary from my investigation:

---

**The problem with `ProposedDirectSentTransactionsToBroadcast`**

The leader currently only re-broadcasts intra-shard transactions that were received via direct-send (P2P request/response). This filter is too restrictive and misses the majority of real-world transaction ingestion paths:

1. **REST API submissions** use `BroadcastOnChannel` (pubsub gossip), not direct-send — they never enter the `directSentTransactionsCache`.
2. **External clients** (light clients, custom P2P relayers) streaming transactions via gossip produce `p2p.Broadcast`-tagged messages, also excluded from the cache.
3. The `directSentTransactionsCache` is **only populated when a node fetches missing data** via the request/response resolver system (`SendToConnectedPeer` → `p2p.Direct`) — this is the on-demand fallback path, not the primary ingestion path.

The underlying assumption — "if the leader received it via gossip, all validators have it too" — is wrong. Gossip propagation is probabilistic. Under load, a significant fraction of validators may not have all transactions by the time the block is proposed. In production logs, blocks with 3000+ txs caused **43% of validators** (227/400 in the consensus group) to fail signature gathering because they couldn't fetch missing transactions within the 85% processing budget (~510ms in a 600ms round).

**How other chains avoid this entirely**

Every major blockchain ensures validators have full transaction data *with* the block proposal — none of them rely on reactive fetching during consensus:

- **Solana**: The leader streams erasure-coded shreds via Turbine to all validators in real-time as transactions are executed. No request/response cycle exists.
- **Ethereum**: The proposer broadcasts the complete `SignedBeaconBlock` (up to 10MB) including all transactions via gossipsub. Validators never request missing txs.
- **NEAR**: The chunk producer generates a self-contained `ChunkStateWitness` with all transactions and distributes erasure-coded parts to validators.
- **Sui**: The DAG-based mempool (Narwhal) ensures data availability before consensus references it.

**Proposed change**

The leader should broadcast **all** intra-shard transactions in the proposed block to the consensus group, regardless of how they were received. The `shouldSkipTransaction` filter based on `directSentTransactionsCache` should be removed. The anti-flooding concern is secondary to consensus liveness — a missed round and its cascading delays cost far more than a few duplicate P2P messages.

PR #7756, opened 2026-02-24
add closingNodeStarted flag to stop pruning

BeniaminDrasovean:
- There is an edgecase when the node is stopping in which pruning should not be done, but because of the delay in closing all the components, some trie hashes are removed
  
## Proposed changes
- When the node starts to close, use a flag to block pruning

## Testing procedure
- Normal tests - tested locally

## Pre-requisites


node/nodeRunner.go
```diff
@@ -1584,7 +1593,13 @@ func closeAllComponents(
 	httpServer shared.UpgradeableHttpServerHandler,
 	node *Node,
 	chanCloseComponents chan struct{},
+	closingNodeStarted *atomic.Bool,
+	closeComponentsDelay time.Duration,
 ) {
+	closingNodeStarted.Store(true)
+	// stop pruning, but wait a bit before closing the components to let the node finish processing the current block
+	time.Sleep(closeComponentsDelay)
```
ssd04, line 1601: there is a `SignalProcessCompletionChan` using in async headers executor, which is used in sync flow to check if the current execution finished, i'm thinking if it might help to use that here

node/nodeRunner.go
```diff
@@ -566,6 +567,10 @@ func (nr *nodeRunner) executeOneComponentCreationCycle(
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
+	// closeComponentsDelay is needed because of pruning. To avoid removing data that will be needed when the node restarts,
+	// block pruning and then wait for a duration of one round before closing the components so that the info that is needed
+	// for the restart is persisted to storage.
+	closeComponentsDelay := time.Duration(uint64(time.Millisecond) * managedCoreComponents.GenesisNodesSetup().GetRoundDuration())
```
ssd04: this will wait for the initial round duration which is 6s, not 600ms, do we want that? either way i think it's better to use the config from chainParameters holder

BeniaminDrasovean: Used the round from `chainParameters`

node/nodeRunner.go
BeniaminDrasovean: To avoid blocking because of reading from multiple goroutines, will use the delay.

raduchis (dismissed):

node/nodeRunner.go
```diff
@@ -566,6 +567,10 @@ func (nr *nodeRunner) executeOneComponentCreationCycle(
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
+	// closeComponentsDelay is needed because of pruning. To avoid removing data that will be needed when the node restarts,
+	// block pruning and then wait for a duration of one round before closing the components so that the info that is needed
+	// for the restart is persisted to storage.
+	closeComponentsDelay := time.Duration(uint64(time.Millisecond) * managedCoreComponents.ChainParametersHandler().CurrentChainParameters().RoundDuration)
```
ssd04:
```suggestion
	closeComponentsDelay := time.Millisecond * time.Duration(managedCoreComponents.ChainParametersHandler().CurrentChainParameters().RoundDuration)
```
*optional: convert directly the other filed, not duration units

ssd04 (dismissed):

ssd04 (approved):

raduchis (approved):

PR #7789, opened 2026-03-11
Extra logging for execution result mismatch

ssd04:
- In case of execution results does not match error there is no relevant info in the error log
  
## Proposed changes
- Add a prettified output for each execution result that does not match

## Testing procedure
- Standard system test

## Pre-requisites

### DIFF
--- a/process/block/executionResultsVerifier.go
+++ b/process/block/executionResultsVerifier.go
@@ -2,12 +2,12 @@ package block
 
 import (
 	"fmt"
-	"reflect"
 
 	"github.com/multiversx/mx-chain-core-go/core/check"
 	"github.com/multiversx/mx-chain-core-go/data"
 	"github.com/multiversx/mx-chain-core-go/data/block"
 
+	"github.com/multiversx/mx-chain-go/common"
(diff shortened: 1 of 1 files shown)


process/block/executionResultsVerifier.go
```diff
@@ -77,15 +77,17 @@ func (erc *executionResultsVerifier) verifyExecutionResults(
 		}
 	}
 
-	for i, er := range executionResults {
-		if !er.Equal(pendingExecutionResults[i]) {
+	for i, headerExecRes := range executionResults {
+		if !headerExecRes.Equal(pendingExecutionResults[i]) {
+			headerExecResOutput, _ := common.PrettifyStruct(headerExecRes)
+			pendingExecResOutput, _ := common.PrettifyStruct(pendingExecutionResults[i])
```
ssd04, line 83: error case here should not happen; in case there is an error with the json marshalling, `PrettifyStruct` will return "nil" string

sstanculeanu (approved):

miiu96 (approved):

PR #7810, opened 2026-03-30
wait for proof before broadcasting

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


consensus/broadcast/delayedBroadcast.go
```diff
@@ -266,36 +285,67 @@ func (dbb *delayedBlockBroadcaster) Close() {
...
+
+	headerHash := proof.GetHeaderHash()
+	header, err := dbb.headersPool.GetHeaderByHash(headerHash)
+	if err != nil {
+		log.Trace("delayedBlockBroadcaster.receivedProof: header not found in pool, will be handled by headerReceived",
+			"headerHash", headerHash,
+		)
+		return
+	}
+
```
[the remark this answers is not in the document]
AdoAdoAdo: proofs won't be accepted if not enabled.

consensus/broadcast/delayedBroadcast.go
```diff
@@ -266,36 +285,67 @@ func (dbb *delayedBlockBroadcaster) Close() {
...
+	if common.IsProofsFlagEnabledForHeader(dbb.enableEpochsHandler, headerHandler) {
+		if !dbb.proofsPool.HasProof(headerHandler.GetShardID(), headerHash) {
+			return
+		}
+	}
+
+	dbb.processMetachainHeaderBroadcast(headerHandler, headerHash)
+}
+
+func (dbb *delayedBlockBroadcaster) receivedProof(proof data.HeaderProofHandler) {
```
sstanculeanu, line 309: there is a slight chance both methods will trigger processMetachainHeaderBroadcast for the same block, if header and proof are received roughly at the same time, leading to double broadcast.. should we worry about this?

AdoAdoAdo: right, problem would be duplicated goroutines, and some processing done twice. It gets deduplicated further down when dealing with the alarm, but I have added now deduplication at the processMetachainHeaderBroadcast.

sstanculeanu (approved):

consensus/broadcast/delayedBroadcast.go
```diff
@@ -266,36 +293,75 @@ func (dbb *delayedBlockBroadcaster) Close() {
...
+			"headerHash", headerHash,
+		)
+		return
+	}
+
+	dbb.processMetachainHeaderBroadcast(header, headerHash)
+}
+
+func (dbb *delayedBlockBroadcaster) processMetachainHeaderBroadcast(headerHandler data.HeaderHandler, headerHash []byte) {
+	has, _ := dbb.cacheProcessedMetaHeaders.HasOrAdd(headerHash, struct{}{}, 0)
```
raduchis, line 330: this seems too early to add the hash... 
if the if below returns, nothing happened and next time it will return early on the "has" 

```
if len(dbb.delayedBroadcastData) == 0 && len(dbb.valBroadcastData) == 0 {
		return
	}
```

consensus/broadcast/delayedBroadcast.go
```diff
@@ -266,36 +293,75 @@ func (dbb *delayedBlockBroadcaster) Close() {
...
+func (dbb *delayedBlockBroadcaster) receivedProof(proof data.HeaderProofHandler) {
+	if check.IfNil(proof) {
+		return
+	}
+	if proof.GetHeaderShardId() != core.MetachainShardId {
+		return
+	}
+
+	headerHash := proof.GetHeaderHash()
+	header, err := dbb.headersPool.GetHeaderByHash(headerHash)
```
raduchis, line 318: could the header be  received and then evicted before the proof comes? You get header H, 100 more headers, evict H, and then you get the proof. The proof waits for the headerReceived, but it already happened.

raduchis (approved):

consensus/broadcast/delayedBroadcast.go
AdoAdoAdo: this means "no data was stored", so no matter what meta header was received, there is nothing to be done by the node.

PR #7823, opened 2026-04-06
load data tries only when needed

BeniaminDrasovean:
- When an account is loaded, it's data trie is also loaded. But not all account operations need access to the data trie.
  
## Proposed changes
- Do not load the data trie when the account is loaded, but rather when the data trie needs to be accessed. This will reduce the number of db accesses

## Testing procedure
- Normal test

## Pre-requisites


node/chainSimulator/components/testOnlyProcessingNode.go
```diff
@@ -528,19 +529,19 @@ func (node *testOnlyProcessingNode) SetStateForAddress(address []byte, addressSt
 		return err
 	}
 
-	rootHash, err := base64.StdEncoding.DecodeString(addressState.RootHash)
+	accountsAdapter := node.StateComponentsHolder.AccountsAdapter()
+	err = accountsAdapter.SaveAccount(userAccount)
 	if err != nil {
 		return err
 	}
-	if len(rootHash) != 0 {
```
miiu96, line 535: after this update is mandatory to provide also the `roothash` ?

state/errors.go
```diff
@@ -174,3 +174,6 @@ var ErrStateAccessesIndexOutOfBounds = errors.New("state access index out of bou
 
 // ErrNilStateAccessesStorer signals that a nil state accesses storer has been given
 var ErrNilStateAccessesStorer = errors.New("nil state accesses storer")
+
+// ErrNilDataTriesHolder signals that a nil data tries holder has been given
+var ErrNilDataTriesHolder = errors.New("nil data tries holder")
```
ssd04: `ErrNilDataTriesHolder` is defined also in errors package

state/accountsDB.go
```diff
@@ -759,7 +721,15 @@ func (adb *AccountsDB) getAccount(address []byte, mainTrie common.Trie) (vmcommo
...
 	}
 
-	return acnt, nil
+	baseAcc, ok := acnt.(baseAccountHandler)
+	if !ok {
+		return acnt, nil
+	}
+	// This will also set the rootHash inside the trackableDataTrie. It is needed in case the data trie needs to be
+	// accessed so that the trackableDataTrie knows the rootHash from which to recreate the data trie.
+	baseAcc.SetRootHash(baseAcc.GetRootHash())
```
ssd04: can we have this explicit for what is needed in trackableDataTrie? maybe a separate method for this? otherwise it might be confusing, especially in the other places where there is not comment for this

state/trackableDataTrie/trackableDataTrie.go
```diff
@@ -34,52 +35,108 @@ type dirtyData struct {
...
+		stateAccessesCollector: args.StateAccessesCollector,
+		dataTriesHolder:        args.DataTriesHolder,
+		dataTrieCreator:        args.DataTrieCreator,
 	}, nil
 }
 
+func (tdt *trackableDataTrie) loadTrie() error {
+	// the trie is already loaded
+	if !check.IfNil(tdt.tr) {
+		return nil
```
ssd04:
```suggestion
	
	if !check.IfNil(tdt.tr) {
	    // the trie is already loaded
		return nil
```
?

ssd04: or update comment

state/trackableDataTrie/trackableDataTrie_test.go
```diff
@@ -939,7 +966,12 @@ func TestTrackableDataTrie_SaveDirtyData(t *testing.T) {
 				return nil
 			},
 		}
-		tdt.SetDataTrie(trie)
+		args.DataTriesHolder = &trieMock.TriesHolderStub{
+			GetCalled: func(key []byte) common.Trie {
+				return trie
```
ssd04, line 971: in most of the tests i see a trie returned here, is the path with nil trie and non-empty rootHash and then recreate covered in the unit tests? if not, maybe add a unit test for this path

node/chainSimulator/components/testOnlyProcessingNode.go
BeniaminDrasovean: The problem here is that if the data trie is not yet built and the rootHash is set, it will try to recreate the trie from db. If the rootHash is not set, it will build a new dataTrie and insert the dirty data into it. I added the check after (L542) to verify that indeed the built data trie has the needed root hash.

state/errors.go
BeniaminDrasovean: Removed from here.

state/accountsDB.go
BeniaminDrasovean: Done.

state/trackableDataTrie/trackableDataTrie.go
BeniaminDrasovean: Changed as suggested

state/trackableDataTrie/trackableDataTrie_test.go
BeniaminDrasovean: `loadTrie()` func is not tested yet directly, will add tests.

miiu96 (dismissed):

state/accountsDB.go
```diff
@@ -811,12 +772,8 @@ func (adb *AccountsDB) GetAccountFromBytes(address []byte, accountBytes []byte)
 		return acnt, nil
 	}
 
-	err = adb.loadDataTrieConcurrentSafe(baseAcc, adb.getMainTrie())
-	if err != nil {
-		return nil, err
-	}
-
-	return acnt, nil
+	baseAcc.SetRootHash(baseAcc.GetRootHash())
```
ssd04: can this be also changed to use `SetDataTrieRootHash` ?

ssd04 (approved):

miiu96 (approved):

PR #7856, opened 2026-05-25
Extra checks epoch change proposed

miiu96:
- Added extra checks for epoch change proposed 

## Pre-requisites

### DIFF
--- a/process/block/metablockProposal.go
+++ b/process/block/metablockProposal.go
@@ -233,12 +233,15 @@ func (mp *metaProcessor) VerifyBlockProposal(
 	}
 
 	shouldProposeEpochChange := mp.epochStartTrigger.ShouldProposeEpochChange(headerHandler.GetRound(), headerHandler.GetNonce())
-	if header.IsEpochChangeProposed() != shouldProposeEpochChange {
+	isEpochChangeProposed := header.IsEpochChangeProposed()
+	// The header flag must match the trigger state in both directions:
+	// it is invalid if the header proposes an epoch change too early or misses one when required.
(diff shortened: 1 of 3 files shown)


process/block/metablockProposal.go
```diff
@@ -232,6 +232,17 @@ func (mp *metaProcessor) VerifyBlockProposal(
 		return process.ErrWrongTypeAssertion
 	}
 
+	shouldProposeEpochChange := mp.epochStartTrigger.ShouldProposeEpochChange(headerHandler.GetRound(), headerHandler.GetNonce())
+	if header.IsEpochChangeProposed() != shouldProposeEpochChange {
```
ssd04: how about the case when header is not epoch change proposed, but it is in the change trigger window?

ssd04: especially in the window between proposed and executed epoch start header

miiu96: changed the condition to match the trigger state for metachain blocks.
for shard blocks, epoch correctness is already validated through `checkEpochCorrectnessCrossChain`.

sstanculeanu (approved):

process/block/metablockProposal.go
ssd04: still not sure if it's fine; we have offset 2 for supernova, so trigger will return as `true` for at least 2 rounds (plus the time for the proposal to be executed) so if first block has EpochChangeProposed, the next block will not have it, it would fail on the check, because trigger will still return true

ssd04 (approved):

danidrasovean (approved):

PR #7880, opened 2026-06-15
TODOs

sstanculeanu:
- TODOs
  
## Proposed changes
- TODOs fixed/removed:
   - `// TODO add also max estimated block gas capacity` + `// TODO add also max estimated block gas capacity - used gas must be lower than this` -> implemented, now uses the block capacity from economics data
   - `// TODO: check if miniblocks and txs are set in a deterministic way (check if there are map iterations that can generate non-deterministic results)` -> checked with @ssd04 , should be safe to remove todo
   - `// todo: extract the epoch change proposal execution result here` -> fixed, now all execution results are checked to find the proper epoch changed proposed header
   - `// TODO: analyse error handling here` -> checked with @ssd04 , this is old code and should be safe as before
   - `// TODO use the right object, not a disabled one` -> fixed, integrated the real transactions processor
   - `// TODO use the right information below` -> fixed, now data is fetched from chain handler
   - `// TODO adjust this method if needed for Supernova` -> removed todo, execution state should be correctly loaded at bootstrap from the headers
   - `// TODO: the trigger would need to be changed upon commit of a block with the epoch start results` -> removed todo, already handled
   - `// TODO: set proper finalized header in outport` -> checked with @BeniaminDrasovean , proper data is already set
   - `// TODO: maybe rename this to reflect last execution results` -> removed todo, not necessarily needed
   - `// todo: check empty mini blocks vs nil. Same for block.Body.MiniBlocks` -> removed todo, GetMiniBlockHeaderHandlers and GetMiniBlocks already normalizez empty slice to nil
   - `// TODO: sanity check use the verify execution results method` -> fixed, added call to VerifyHeaderExecutionResults
   - `// TODO: this account adapter may be required to be changed as the roothash should be the last execution result roothash` -> fixed, now trie is recreated everywhere

## Testing procedure
- with feat branch

## Pre-requisites


process/block/baseProcess.go
```diff
@@ -3075,7 +3075,18 @@ func (bp *baseProcessor) OnExecutedBlock(header data.HeaderHandler, rootHash []b
 }
 
 func (bp *baseProcessor) recreateTrieIfNeeded() error {
-	rootHash := bp.blockChain.GetCurrentBlockRootHash()
+	prevHeader := bp.blockChain.GetCurrentBlockHeader()
```
miiu96: maybe you can extract this code in a new function ?

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -164,6 +167,22 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
 			return pendingExecutionIndex
 		}
 
+		newAccumulatedGas, overflow := bits.Add64(accumulatedGas, executionResultMeta.GetGasUsed(), 0)
```
AdoAdoAdo: we don't need to verify the accumulated gas between multiple execution results, those refer to multiple blocks.

The verification should be the used gas in each execution result to not be above the cap.

process/estimator/executionResultInclusionEstimator.go
```diff
@@ -164,6 +167,22 @@ func (erie *ExecutionResultInclusionEstimator) Decide(
...
+		if overflow != 0 {
+			log.Debug("ExecutionResultInclusionEstimator: overflow detected in accumulated gas",
+				"accumulatedGas", accumulatedGas,
+				"gasUsed", executionResultMeta.GetGasUsed())
+			return pendingExecutionIndex
+		}
+		if erie.maxBlockGasCapacity > 0 && newAccumulatedGas > erie.maxBlockGasCapacity {
```
AdoAdoAdo: this max block gas capacity should be dynamic, but here it depends on the restart moment of the node, which loads the corresponding config (epoch based).

epochStart/metachain/economics.go
```diff
@@ -667,17 +673,30 @@ func (e *economics) startNoncePerShardFromEpochStart(epoch uint32) (map[uint32]u
...
+func (e *economics) getPrevEpochMetaStartNonceForEconomics(previousEpochStartMeta data.MetaHeaderHandler) (uint64, error) {
 	if !previousEpochStartMeta.IsHeaderV3() {
 		return previousEpochStartMeta.GetNonce(), nil
 	}
-	// todo: extract the epoch change proposal execution result here
-	lastNotarizedResult, err := common.GetLastBaseExecutionResultHandler(previousEpochStartMeta)
-	if err != nil {
-		return 0, err
+
+	for _, execResult := range previousEpochStartMeta.GetExecutionResultsHandlers() {
```
AdoAdoAdo, line 681: I think we have something similar in extractEpochChangeProposedHeader.

Can you check if we can extract a common exported func in process/common.go
 
```go
func ExtractEpochChangeProposedMetaHeader(
      startOfEpochMeta data.MetaHeaderHandler,
      headersPool dataRetriever.HeadersPool,
      marshaller marshal.Marshalizer,
      store dataRetriever.StorageService,
  ) (data.MetaHeaderHandler, error)
```
and then use that in both places here and in shardblockProposal.go L981

epochStart/metachain/epochStartData.go
```diff
@@ -418,7 +418,6 @@ func (e *epochStartData) computePendingMiniBlockList(
 
 	epochStartIdentifier := core.EpochStartIdentifier(prevEpoch)
 
-	// TODO: analyse error handling here
 	previousEpochStartMeta, _ := process.GetMetaHeaderFromStorage([]byte(epochStartIdentifier), e.marshalizer, e.store)
```
AdoAdoAdo: can you log the error if any instead? (LogIfError for the currently suppressed error)

factory/processing/blockProcessorCreator.go
```diff
@@ -605,10 +606,14 @@ func (pcf *processComponentsFactory) newShardBlockProcessor(
...
 	}
 
+	maxBlockGasCapacity := pcf.coreData.EconomicsData().MaxGasLimitPerBlock(pcf.bootstrapComponents.ShardCoordinator().SelfId())
+	overestimationFactor := pcf.coreData.EconomicsData().BlockCapacityOverestimationFactor()
+	maxBlockGasCapacity = maxBlockGasCapacity * overestimationFactor / 100
 	inclusionEstimator, err := estimator.NewExecutionResultInclusionEstimator(
 		pcf.config.ExecutionResultInclusionEstimator,
 		pcf.coreData.RoundHandler(),
 		execResSizeComputationHandler,
+		maxBlockGasCapacity,
```
AdoAdoAdo: this reflects the block gas capacity for the current epoch, should probably get a handler that can provide the gas capacity depending on the epoch.

factory/processing/blockProcessorCreator.go
```diff
@@ -1188,10 +1195,14 @@ func (pcf *processComponentsFactory) newMetaBlockProcessor(
 		return nil, err
 	}
 
+	maxBlockGasCapacity := pcf.coreData.EconomicsData().MaxGasLimitPerBlock(pcf.bootstrapComponents.ShardCoordinator().SelfId())
```
AdoAdoAdo: same as my previous comment.

process/block/baseProcess.go
```diff
@@ -3097,7 +3097,18 @@ func (bp *baseProcessor) OnExecutedBlock(header data.HeaderHandler, rootHash []b
...
```
AdoAdoAdo: I don't think this is correct in case of legacy with scheduled transactions. The roothash on the block would be the one for regular execution, the roothash after scheduled execution would not be captured there.

AdoAdoAdo: I don't think this change is required

process/estimator/executionResultInclusionEstimator.go
sstanculeanu: fixed

process/estimator/executionResultInclusionEstimator.go
sstanculeanu: indeed, fixed

process/block/baseProcess.go
sstanculeanu: reverted

process/block/baseProcess.go
sstanculeanu: reverted

epochStart/metachain/economics.go
sstanculeanu: extractEpochChangeProposedHeader relies on headersForCurrentBlock for legacy headers, but here get them from pool/store

AdoAdoAdo (approved):

miiu96 (approved):

PR #7881, opened 2026-06-16
Subrounds timing by epoch

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


consensus/spos/bls/v1/blsSubroundsFactory.go
ssd04, line 1: *optional: keep the constants for v1, and update the factory only for v2

cmd/node/config/config.toml
```diff
@@ -106,8 +106,51 @@
...
-        { EnableEpoch = 2, NumRoundsToWaitBeforeSignalingChronologyStuck = 100 },
+        {
+            EnableEpoch = 0,
+            NumRoundsToWaitBeforeSignalingChronologyStuck = 10,
+            SubroundsTiming = {
+                SubroundStartStartTime = 0.0,
+                SubroundStartEndTime = 0.05,
+                SubroundBlockStartTime = 0.05,
+                SubroundBlockEndTime = 0.25,
+                SubroundSignatureStartTime = 0.25,
```
ssd04: *optional: remove Subround prefix

consensus/spos/bls/v1/blsSubroundsFactory.go
sstanculeanu: tried to keep similar flow

cmd/node/config/config.toml
sstanculeanu: I don't think it would be clear enough, StartStartTime, BlockStartTime, etc

consensus/spos/bls/v2/subroundBlock.go
```diff
@@ -390,7 +390,7 @@ func (sr *subroundBlock) triggerCreateSignaturesForManagedKeys(
 
 	currentEpoch := headerHandler.GetEpoch()
 
-	sigSubroundEndTime := time.Duration(float64(sr.RoundHandler().TimeDuration()) * srSignatureEndTime)
+	sigSubroundEndTime := time.Duration(sr.EndTime())
```
AdoAdoAdo: I think this needs to be set back to the srSignatureEndTime, as here we are triggering optimistically the signing that ends at the end of subround signature.

cmd/node/config/config.toml
```diff
@@ -106,8 +106,51 @@
...
+                SubroundSignatureEndTime = 0.85,
+                SubroundEndStartTime = 0.85,
+                SubroundEndEndTime = 0.95,
+                ProcessingThresholdPercent = 85
+            }
+        },
+        {
+            EnableEpoch = 1,
+            NumRoundsToWaitBeforeSignalingChronologyStuck = 10,
```
AdoAdoAdo: if this is identical with the previous config, we can probably remove it from the config?

cmd/node/config/config.toml
```diff
@@ -106,8 +106,51 @@
...
+                SubroundBlockEndTime = 0.25,
+                SubroundSignatureStartTime = 0.25,
+                SubroundSignatureEndTime = 0.85,
+                SubroundEndStartTime = 0.85,
+                SubroundEndEndTime = 0.95,
+                ProcessingThresholdPercent = 85
+            }
+        },
+        {
+            EnableEpoch = 2,
```
AdoAdoAdo: For Supernova I think we need also the round activation for the timing change - so only on both conditions we should change the timing there.

cmd/node/config/config.toml
```diff
@@ -106,8 +106,51 @@
     ]
 
     ConsensusConfigsByEpoch = [
-        { EnableEpoch = 0, NumRoundsToWaitBeforeSignalingChronologyStuck = 10 },
-        { EnableEpoch = 2, NumRoundsToWaitBeforeSignalingChronologyStuck = 100 },
+        {
```
AdoAdoAdo, line 114: can you create a validation func for each epoch config where it is loaded into the component requiring it/init?
e.g
- non negative values
- every subround has a start and end value, start<end
- no overlapping range between subrounds
- increasing values between subrounds (start subround n+1 > end subround n)
- max value < 1.

cmd/node/config/config.toml
sstanculeanu: updated

cmd/node/config/config.toml
sstanculeanu: added

consensus/spos/bls/v1/blsSubroundsFactory.go
```diff
@@ -100,28 +110,31 @@ func (fct *factory) SetOutportHandler(driver outport.OutportHandler) {
...
-func (fct *factory) GenerateSubrounds(_ uint32) error {
+func (fct *factory) GenerateSubrounds(epoch uint32) error {
 	fct.initConsensusThreshold()
 	fct.consensusCore.Chronology().RemoveAllSubrounds()
 	fct.worker.RemoveAllReceivedMessagesCalls()
 	fct.worker.RemoveAllReceivedHeaderHandlers()
 
-	err := fct.generateStartRoundSubround()
+	round := uint64(fct.consensusCore.RoundHandler().Index())
+	timing := fct.commonConfigsHandler.GetSubroundsTimingByRound(round)
```
ssd04: in RoundConfirmed we are setting timings based on received `round`, and here based on current consensus round, i'm thinking if they might not match; if RoundConfirmed is triggered from the processing flow, then they might be fetched for different rounds

sstanculeanu: good catch, fixed

AdoAdoAdo (dismissed):

ssd04:
no other things from me

consensus/spos/bls/proxy/subroundsHandler.go
```diff
@@ -51,19 +56,58 @@ type SubroundsHandler struct {
...
 	if err != nil {
 		log.Error("SubroundsHandler.EpochConfirmed: cannot initialize subrounds", "error", err)
 	}
 }
 
+// RoundConfirmed is called when a new round is confirmed
+func (s *SubroundsHandler) RoundConfirmed(round uint64, _ uint64) {
+	s.mutReInit.Lock()
+	defer s.mutReInit.Unlock()
+
```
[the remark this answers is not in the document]
sstanculeanu: added

AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

ssd04 (approved):

PR #7890, opened 2026-06-24
when transitioning to Andromeda, request missing proofs

BeniaminDrasovean:
This branch fixes an edge case in metachain/shard-header finality attestation sync where a finality-attesting header is already available locally, but its equivalent proof is missing.

## Pre-requisites

### DIFF
--- a/process/block/baseProcess.go
+++ b/process/block/baseProcess.go
@@ -1216,15 +1216,12 @@ func (bp *baseProcessor) requestMissingFinalityAttestingHeaders(
 		}
 
 		for index := range headers {
-			hdrHash := headersHashes[index]
-			hdr := headers[index]
-
-			bp.hdrsForCurrBlock.hdrHashAndInfo[string(hdrHash)] = &hdrInfo{
-				hdr:         hdr,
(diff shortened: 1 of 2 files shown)


sstanculeanu (dismissed):

process/block/baseProcess.go
```diff
@@ -1216,10 +1216,15 @@ func (bp *baseProcessor) requestMissingFinalityAttestingHeaders(
 		}
 
 		for index := range headers {
-			bp.hdrsForCurrBlock.hdrHashAndInfo[string(headersHashes[index])] = &hdrInfo{
-				hdr:         headers[index],
+			hdrHash := headersHashes[index]
```
AdoAdoAdo: was there a reason to do an alloc here instead of using the value directly besides readability?

BeniaminDrasovean: No, changed.

sstanculeanu (approved):

AdoAdoAdo (approved):

danidrasovean (approved):

PR #7897, opened 2026-07-03
added new cache for quarantined headers on late proofs

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -619,6 +619,12 @@
     Type = "SizeLRU"
     SizeInBytes = 52428800 # 50MB
 
+[QuarantinedHeadersCache]
+    Name = "QuarantinedHeadersCache"
+    Capacity = 500
+    Type = "SizeLRU"
+    SizeInBytes = 52428800 # 50MB
+
 [PostProcessTransactionsCache]
     Name = "PostProcessTransactionsCache"
     Capacity = 250000
--- a/cmd/node/config/config.toml
+++ b/cmd/node/config/config.toml
@@ -621,9 +621,9 @@
 
 [QuarantinedHeadersCache]
     Name = "QuarantinedHeadersCache"
-    Capacity = 500
+    Capacity = 100
     Type = "SizeLRU"
-    SizeInBytes = 52428800 # 50MB
+    SizeInBytes = 10485760 # 10MB
 
 [PostProcessTransactionsCache]
     Name = "PostProcessTransactionsCache"
--- a/process/block/metablockProposal.go
+++ b/process/block/metablockProposal.go
@@ -1205,6 +1205,10 @@ func (mp *metaProcessor) checkHeadersSequenceCorrectness(hdrsForShard []ShardHea
 			continue
 		}
 
+		if mp.blockTracker.IsHeaderQuarantined(shardHdrInfo.Hash) {
+			return fmt.Errorf("%w with hash %s", errIncludedQuarantinedHeader, shardHdrInfo.Hash)
+		}
+
 		err = mp.headerValidator.IsHeaderConstructionValid(shardHdrInfo.Header, lastNotarizedHeaderInfoForShard.Header)
 		if err != nil {
 			return err
--- a/process/block/metablockProposal.go
+++ b/process/block/metablockProposal.go
@@ -1206,7 +1206,7 @@ func (mp *metaProcessor) checkHeadersSequenceCorrectness(hdrsForShard []ShardHea
 		}
 
 		if mp.blockTracker.IsHeaderQuarantined(shardHdrInfo.Hash) {
-			return fmt.Errorf("%w with hash %s", errIncludedQuarantinedHeader, shardHdrInfo.Hash)
+			return fmt.Errorf("%w with hash %x", errIncludedQuarantinedHeader, shardHdrInfo.Hash)
 		}
 
 		err = mp.headerValidator.IsHeaderConstructionValid(shardHdrInfo.Header, lastNotarizedHeaderInfoForShard.Header)
--- a/process/block/shardblockProposal.go
+++ b/process/block/shardblockProposal.go
@@ -817,12 +817,16 @@ func (sp *shardProcessor) checkMetaHeadersValidityAndFinalityProposal(header dat
 	if err != nil {
 		return err
 	}
-	_, usedMetaHeaders, err := sp.getReferencedMetaHeadersFromPool(header)
+	usedMetaHashes, usedMetaHeaders, err := sp.getReferencedMetaHeadersFromPool(header)
 	if err != nil {
 		return fmt.Errorf("%w : checkMetaHeadersValidityAndFinalityProposal -> getReferencedMetaHeadersFromPool", err)
 	}
 
-	for _, metaHeader := range usedMetaHeaders {
+	for idx, metaHeader := range usedMetaHeaders {
+		if sp.blockTracker.IsHeaderQuarantined(usedMetaHashes[idx]) {
+			return fmt.Errorf("%w with hash %s", errIncludedQuarantinedHeader, usedMetaHashes[idx])
+		}
+
 		err = sp.headerValidator.IsHeaderConstructionValid(metaHeader, lastCrossNotarizedHeader)
 		if err != nil {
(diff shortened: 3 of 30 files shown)


cmd/node/config/config.toml
```diff
@@ -619,6 +619,12 @@
     Type = "SizeLRU"
     SizeInBytes = 52428800 # 50MB
 
+[QuarantinedHeadersCache]
+    Name = "QuarantinedHeadersCache"
+    Capacity = 500
+    Type = "SizeLRU"
+    SizeInBytes = 52428800 # 50MB
```
ssd04: i think it can be even lower in size, since it records only hashes, and there shouldn't be many at a time

process/track/baseBlockTrack.go
```diff
@@ -482,9 +485,28 @@ func (bbt *baseBlockTrack) CheckProofAgainstRoundHandler(proof data.HeaderProofH
...
+	if bbt.roundHandler.Index() != int64(proof.GetHeaderRound())+1 {
+		return
+	}
+
+	if bbt.shardCoordinator.SelfId() == proof.GetHeaderShardId() {
+		return
+	}
+
+	hash := proof.GetHeaderHash()
+	bbt.quarantinedHeaders.Put(hash, struct{}{}, 0)
```
ssd04: the cache will be fetched/used in a later PR, right?

process/track/baseBlockTrack.go
```diff
@@ -482,9 +485,28 @@ func (bbt *baseBlockTrack) CheckProofAgainstRoundHandler(proof data.HeaderProofH
...
+
 	return bbt.checkAgainstRoundHandler(proof.GetHeaderRound())
 }
 
+func (bbt *baseBlockTrack) quarantineIfLateProof(proof data.HeaderProofHandler) {
+	if bbt.roundHandler.Index() != int64(proof.GetHeaderRound())+1 {
+		return
+	}
+
+	if bbt.shardCoordinator.SelfId() == proof.GetHeaderShardId() {
```
ssd04, line 554: will this cover the cross shard proofs on meta? i think cross shard proofs should be covered

process/track/baseBlockTrack.go
```diff
@@ -482,9 +485,28 @@ func (bbt *baseBlockTrack) CheckProofAgainstRoundHandler(proof data.HeaderProofH
 		return process.ErrNilHeaderProof
 	}
 
+	bbt.quarantineIfLateProof(proof)
+
 	return bbt.checkAgainstRoundHandler(proof.GetHeaderRound())
 }
 
+func (bbt *baseBlockTrack) quarantineIfLateProof(proof data.HeaderProofHandler) {
+	if bbt.roundHandler.Index() != int64(proof.GetHeaderRound())+1 {
```
ssd04, line 550: i'm thinking if we should consider matching with `checkAgainstFinal` check; older proofs than final are not accepted, but if there is a proof higher than final (and lower/equal than current round) it will be accepted; should we add to quarantine also older (than last round) proofs if they are accepted in the pool? (or maybe this gap should not happen anyway due to another reasons)

process/track/baseBlockTrack.go
sstanculeanu: exactly, https://github.com/multiversx/mx-chain-go/pull/7898

process/track/baseBlockTrack.go
sstanculeanu: should be covered

process/track/baseBlockTrack.go
sstanculeanu: I think that would quarantine all headers during sync

cmd/node/config/config.toml
sstanculeanu: lowered

BeniaminDrasovean (dismissed):

process/track/baseBlockTrack.go
```diff
@@ -482,9 +485,28 @@ func (bbt *baseBlockTrack) CheckProofAgainstRoundHandler(proof data.HeaderProofH
 		return process.ErrNilHeaderProof
 	}
 
+	bbt.quarantineIfLateProof(proof)
```
AdoAdoAdo: I think this is not a good place to call this, there are multiple checks that are skipped if added to quarantine here.

Should be called instead on interceptedEquivalentProof CheckValidity, at the moment of adding the proof into pool (we also want to do the quarantine on the first proof receive, not on duplicated proofs that get dismissed afterwards).

Or could also be called inside baseBlockTrack.receivedProof, as that will be called as soon as there is a verified proof added to the pool.

ssd04 (dismissed):

process/track/baseBlockTrack.go
AdoAdoAdo: maybe do a check also for proof of nonce N+1 ? 
maybe the next block proof is already available, so we can immediately settle it.

process/track/baseBlockTrack.go
sstanculeanu: added into receivedProof

process/track/baseBlockTrack.go
sstanculeanu: added

process/track/baseBlockTrack.go
```diff
@@ -161,9 +175,52 @@ func (bbt *baseBlockTrack) receivedProof(proof data.HeaderProofHandler) {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			currentRound := bbt.roundHandler.Index()
+			for _, key := range bbt.quarantinedHeaders.Keys() {
+				val, _ := bbt.quarantinedHeaders.Get(key)
+				headerRound, ok := val.(int64)
```
AdoAdoAdo: this type assertion always fails as the round was added as a uint64

This means that it will never do the sweep

process/track/baseBlockTrack.go
```diff
@@ -161,9 +175,52 @@ func (bbt *baseBlockTrack) receivedProof(proof data.HeaderProofHandler) {
...
+				if !ok {
+					continue
+				}
+
+				if currentRound-headerRound < maxNonceDifference {
```
AdoAdoAdo: this maxNonceDifference is reused here for rounds difference. Can we have a dedicated constant instead?

process/block/metablockProposal.go
```diff
@@ -1205,6 +1205,10 @@ func (mp *metaProcessor) checkHeadersSequenceCorrectness(hdrsForShard []ShardHea
 			continue
 		}
 
+		if mp.blockTracker.IsHeaderQuarantined(shardHdrInfo.Hash) {
+			return fmt.Errorf("%w with hash %s", errIncludedQuarantinedHeader, shardHdrInfo.Hash)
```
AdoAdoAdo: use %x instead of %s?

process/block/shardblockProposal.go
```diff
@@ -817,12 +817,16 @@ func (sp *shardProcessor) checkMetaHeadersValidityAndFinalityProposal(header dat
...
-	_, usedMetaHeaders, err := sp.getReferencedMetaHeadersFromPool(header)
+	usedMetaHashes, usedMetaHeaders, err := sp.getReferencedMetaHeadersFromPool(header)
 	if err != nil {
 		return fmt.Errorf("%w : checkMetaHeadersValidityAndFinalityProposal -> getReferencedMetaHeadersFromPool", err)
 	}
 
-	for _, metaHeader := range usedMetaHeaders {
+	for idx, metaHeader := range usedMetaHeaders {
+		if sp.blockTracker.IsHeaderQuarantined(usedMetaHashes[idx]) {
+			return fmt.Errorf("%w with hash %s", errIncludedQuarantinedHeader, usedMetaHashes[idx])
```
AdoAdoAdo: use %x instead of %s ?

process/track/baseBlockTrack.go
sstanculeanu: fixed

process/track/baseBlockTrack.go
sstanculeanu: updated, missed that

AdoAdoAdo (approved):

ssd04 (approved):

PR #7898, opened 2026-07-03
settle quarantined headers

sstanculeanu:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/track/baseBlockTrack.go
+++ b/process/track/baseBlockTrack.go
@@ -164,7 +164,23 @@ func (bbt *baseBlockTrack) receivedProof(proof data.HeaderProofHandler) {
 		"hash", proof.GetHeaderHash(),
 	)
 
+	bbt.settleQuarantinedParentIfNeeded(header, headerHash)
+
 	bbt.receivedHeader(header, headerHash)
+}
+
+func (bbt *baseBlockTrack) settleQuarantinedParentIfNeeded(header data.HeaderHandler, headerHash []byte) {
+	prevHash := header.GetPrevHash()
+	if !bbt.quarantinedHeaders.Has(prevHash) {
+		return
+	}
+
+	bbt.quarantinedHeaders.Remove(prevHash)
+
+	log.Debug("settled quarantined header",
+		"quarantinedHash", prevHash,
+		"confirmationHash", headerHash,
+		"confirmationNonce", header.GetNonce())
 }
 
 func (bbt *baseBlockTrack) requestHeaderForProof(proof data.HeaderProofHandler) {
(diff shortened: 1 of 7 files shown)


process/track/baseBlockTrack.go
```diff
@@ -164,9 +164,25 @@ func (bbt *baseBlockTrack) receivedProof(proof data.HeaderProofHandler) {
 		"hash", proof.GetHeaderHash(),
 	)
 
+	bbt.settleQuarantinedParentIfNeeded(header, headerHash)
```
AdoAdoAdo, line 169: I think in a race between receiving proof and receiving header, there is a small window where this may not get settled.

I would call the settle quarantined parent also from receivedHeader() at line 206 (a check that proof is available is done just before).

process/track/blockProcessor.go
```diff
@@ -297,6 +297,11 @@ func (bp *blockProcessor) getNextHeader(
 			break
 		}
 
+		if bp.blockTracker.IsHeaderQuarantined(sortedHeadersHashes[i]) {
+			log.Trace("getNextHeader: skipping quarantined header", "hash", sortedHeadersHashes[i])
+			continue
```
ssd04, line 302: break here? if current header is quarantined, i think the next headers should not be considered. right? if they are sorted by nonce

process/track/blockProcessor.go
```diff
@@ -297,6 +297,11 @@ func (bp *blockProcessor) getNextHeader(
 			break
 		}
 
+		if bp.blockTracker.IsHeaderQuarantined(sortedHeadersHashes[i]) {
```
AdoAdoAdo, line 300: this only solves the proposal/proposer side, not the validators side.

For the validators I think adding the check in checkShardHeadersValidtiyAndFinalityProposal/ checkMetaHeadersValidityAndFinalityProposal for not quarantined headers would be needed.

process/track/export_test.go
```diff
@@ -130,6 +130,11 @@ func (bbt *baseBlockTrack) SetQuarantinedHeaders(quarantinedHeaders storage.Cach
 	bbt.quarantinedHeaders = quarantinedHeaders
 }
 
+// SettleQuarantinedParentIfNeeded -
+func (bbt *baseBlockTrack) SettleQuarantinedParentIfNeeded(header data.HeaderHandler, headerHash []byte) {
```
AdoAdoAdo, line 139: The settlement currently can have some larger delays, in case of syncing at the tip, when some followup headers/proofs are missed. 

For that I am thinking if it is not better to add a default expiry of quarantine after some rounds delta (we anyway have that at the interceptor level) so we can have something like >3 rounds -> already out of quarantine.

This would need to register also the round of the header in the quarantine, and maybe handle the cleanup together with the check.

process/track/blockProcessor.go
sstanculeanu: updated to break.. with continue, the next iteration would break at previous check, so both work

process/track/baseBlockTrack.go
sstanculeanu: updated

process/track/blockProcessor.go
sstanculeanu: updated

PR #7903, opened 2026-07-07
cross round quorum evidence

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/consensus/spos/bls/v2/proofSelfAssembly.go
+++ b/consensus/spos/bls/v2/proofSelfAssembly.go
@@ -34,6 +34,7 @@ func trySelfAssembleProof(sr *spos.Subround, store signatureEvidenceHandler, ev 
 
 	bitmap, agg, err := aggregateAndVerifyEvidence(sr, ev)
 	if err != nil {
+		stripInvalidShares(sr, ev)
 		if ev.getCount() < ev.threshold {
 			// inflated signal was adversarial; the decision tiers demote to the wait tier
 			store.DropRetained(ev.nonce)
@@ -99,8 +100,7 @@ func abortOnExistingProof(sr *spos.Subround, store signatureEvidenceHandler, ev 
 	return true
 }
 
-// aggregateAndVerifyEvidence aggregates the snapshotted shares and verifies the result; on
-// failure it strips the invalid shares so a retry or demotion can follow
+// aggregateAndVerifyEvidence aggregates the snapshotted shares and verifies the result
 func aggregateAndVerifyEvidence(sr *spos.Subround, ev *roundSignatureEvidence) ([]byte, []byte, error) {
 	bitmap, shares := ev.getAggregationData()
 
@@ -108,17 +108,16 @@ func aggregateAndVerifyEvidence(sr *spos.Subround, ev *roundSignatureEvidence) (
 	if err == nil {
 		err = sr.SigningHandler().VerifyAggregatedSigWithKeys(ev.consensusGroup, bitmap, ev.headerHash, agg, ev.epoch)
 	}
-	if err == nil {
-		return bitmap, agg, nil
+	if err != nil {
+		return nil, nil, err
 	}
 
-	stripInvalidShares(sr, ev, bitmap, shares)
-
-	return nil, nil, err
+	return bitmap, agg, nil
 }
 
 // stripInvalidShares verifies each share in parallel and drops the invalid ones from the evidence
-func stripInvalidShares(sr *spos.Subround, ev *roundSignatureEvidence, bitmap []byte, shares [][]byte) {
+func stripInvalidShares(sr *spos.Subround, ev *roundSignatureEvidence) {
+	bitmap, shares := ev.getAggregationData()
 	invalidIndices := make([]int, 0)
 	var mut sync.Mutex
 
--- a/consensus/spos/bls/v2/proofSelfAssembly.go
+++ b/consensus/spos/bls/v2/proofSelfAssembly.go
@@ -1,0 +1,185 @@
+package v2
+
+import (
+	"bytes"
+	"math/rand"
+	"sync"
+
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data/block"
+
+	commonConsensus "github.com/multiversx/mx-chain-go/common/consensus"
+	"github.com/multiversx/mx-chain-go/consensus/spos"
+)
+
+// trySelfAssembleProof aggregates the snapshotted signature shares into a proof for the previous
+// round's block; it never creates a second proof: any existing proof at the nonce aborts it
+func trySelfAssembleProof(sr *spos.Subround, store signatureEvidenceHandler, ev *roundSignatureEvidence) {
+	if !ev.assemblyStarted.CompareAndSwap(false, true) {
+		return
+	}
+	if !shouldNodeSendProofForGroup(sr, ev.consensusGroup) {
+		return
+	}
+	if abortOnExistingProof(sr, store, ev, "at entry") {
+		return
+	}
+
+	bitmap, agg, err := aggregateAndVerifyEvidence(sr, ev)
+	if err != nil {
+		if ev.getCount() < ev.threshold {
+			// inflated signal was adversarial; the decision tiers demote to the wait tier
+			store.DropRetained(ev.nonce)
+			log.Debug("trySelfAssembleProof: evidence demoted below threshold after stripping invalid shares",
+				"nonce", ev.nonce, "count", ev.getCount(), "threshold", ev.threshold)
+			return
+		}
+
+		bitmap, agg, err = aggregateAndVerifyEvidence(sr, ev)
+		if err != nil {
+			log.Warn("trySelfAssembleProof: aggregation failed after stripping invalid shares",
+				"nonce", ev.nonce, "error", err.Error())
+			return
+		}
+	}
+
+	proof := &block.HeaderProof{
+		PubKeysBitmap:       bitmap,
+		AggregatedSignature: agg,
+		HeaderHash:          ev.headerHash,
+		HeaderEpoch:         ev.epoch,
+		HeaderNonce:         ev.nonce,
+		HeaderShardId:       ev.shardID,
+		HeaderRound:         ev.headerRound,
+		IsStartOfEpoch:      ev.isStartOfEpoch,
+	}
+
+	// re-check right before publish, aggregation took time
+	if abortOnExistingProof(sr, store, ev, "before publish") {
+		return
+	}
+	if !sr.EquivalentProofsPool().AddProof(proof) {
+		return
+	}
+	store.DropRetained(ev.nonce)
+
+	sender := equivalentProofSenderForGroup(sr, ev.consensusGroup)
+	err = sr.BroadcastMessenger().BroadcastEquivalentProof(proof, []byte(sender))
+	if err != nil {
+		log.Warn("trySelfAssembleProof: failed to broadcast proof", "nonce", ev.nonce, "error", err.Error())
+		return
+	}
+
+	log.Debug("trySelfAssembleProof: self-assembled proof has been sent",
+		"nonce", ev.nonce, "headerHash", ev.headerHash)
+}
+
+func abortOnExistingProof(sr *spos.Subround, store signatureEvidenceHandler, ev *roundSignatureEvidence, stage string) bool {
+	existing, err := sr.EquivalentProofsPool().GetProofByNonce(ev.nonce, ev.shardID)
+	if err != nil || check.IfNil(existing) {
+		return false
+	}
+
+	if !bytes.Equal(existing.GetHeaderHash(), ev.headerHash) {
+		log.Warn("trySelfAssembleProof: competing proof already exists at nonce, aborting "+stage,
+			"nonce", ev.nonce, "ownHash", ev.headerHash, "existingHash", existing.GetHeaderHash())
+	}
+	store.DropRetained(ev.nonce) // nonce settled either way
+
+	return true
+}
+
+// aggregateAndVerifyEvidence aggregates the snapshotted shares and verifies the result; on
+// failure it strips the invalid shares so a retry or demotion can follow
+func aggregateAndVerifyEvidence(sr *spos.Subround, ev *roundSignatureEvidence) ([]byte, []byte, error) {
+	bitmap, shares := ev.getAggregationData()
+
+	agg, err := sr.SigningHandler().AggregateSigsWithKeys(ev.consensusGroup, bitmap, shares, ev.epoch)
+	if err == nil {
+		err = sr.SigningHandler().VerifyAggregatedSigWithKeys(ev.consensusGroup, bitmap, ev.headerHash, agg, ev.epoch)
+	}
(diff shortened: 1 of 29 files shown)


consensus/spos/bls/v2/signatureEvidence.go
```diff
@@ -0,0 +1,185 @@
...
+		return ev, true
+	}
+
+	ev = store.retained
+	if ev != nil && ev.getCount() >= ev.threshold && !store.isNonceSettled(ev) {
+		return ev, true
+	}
+
+	return nil, false
+}
```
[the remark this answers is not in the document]
AdoAdoAdo: not needed

consensus/spos/bls/v2/subroundStartRound.go
```diff
@@ -109,6 +118,77 @@ func (sr *subroundStartRound) doStartRoundJob(_ context.Context) bool {
...
+	sr.signatureEvidence.Capture(sr.buildSignatureEvidence())
+
+	ev, ok := sr.signatureEvidence.GetAssemblyCandidate()
+	if !ok {
+		return
+	}
+
+	// allow one assembly attempt per round while the nonce stays unsettled
+	ev.assemblyStarted.Store(false)
+	go trySelfAssembleProof(sr.Subround, sr.signatureEvidence, ev)
```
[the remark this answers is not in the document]
AdoAdoAdo: fixed

consensus/spos/bls/v2/signatureEvidence.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	if replaced.getCount() < replaced.threshold {
+		return
+	}
+	if store.isNonceSettled(replaced) {
+		return
+	}
+
+	store.retained = replaced
+}
```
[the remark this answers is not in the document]
AdoAdoAdo: not needed

consensus/spos/bls/v2/subroundStartRound.go
```diff
@@ -109,6 +118,75 @@ func (sr *subroundStartRound) doStartRoundJob(_ context.Context) bool {
...
+		return nil
+	}
+
+	group := sr.ConsensusGroup()
+	bitmapSize := len(group) / 8
+	if len(group)%8 != 0 {
+		bitmapSize++
+	}
+
+	bitmap := make([]byte, bitmapSize)
```
ssd04: for bitmap we already have GenerateBitmap, i think we can use that, and iterate only to get the signature shares

consensus/spos/bls/v2/subroundStartRound.go
```diff
@@ -109,6 +118,75 @@ func (sr *subroundStartRound) doStartRoundJob(_ context.Context) bool {
 	return true
 }
 
+// captureSignatureEvidence snapshots the signature shares observed for the previous round's
+// block and triggers proof self-assembly when quorum-level evidence exists for an unsettled nonce
+func (sr *subroundStartRound) captureSignatureEvidence() {
+	sr.signatureEvidence.Capture(sr.buildSignatureEvidence())
```
ssd04, line 124: will `buildSignatureEvidence` be triggered even if there is already a proof and the previous round has completed successfully? add pre-checks to avoid calculating the evidence twice

consensus/spos/bls/v2/signatureEvidence.go
```diff
@@ -0,0 +1,188 @@
...
+	DropRetained(nonce uint64)
+	IsInterfaceNil() bool
+}
+
+type signatureEvidenceStore struct {
+	proofsPool consensus.EquivalentProofsPool
+
+	mut      sync.RWMutex
+	previous *roundSignatureEvidence
+	retained *roundSignatureEvidence
```
ssd04: *optional: rename `retained` to something more explicit, how is it different from `previous`? is it the last evidence that passed the threshold count?

consensus/spos/bls/v2/subroundSignature.go
```diff
@@ -427,6 +437,34 @@ func (sr *subroundSignature) getPkForCompetingBlock(nonce uint64, currentHash []
...
+		ev, ok = sr.signatureEvidence.GetRetainedQuorumEvidence(nonce)
+	}
+	if !ok || bytes.Equal(ev.headerHash, currentHash) {
+		return false
+	}
+
+	count := ev.getCount()
+	switch {
+	case count >= ev.threshold:
+		go trySelfAssembleProof(sr.Subround, sr.signatureEvidence, ev)
```
ssd04, line 442: should there be a time context for `tryAssembleProof`? i'm thinking on the case when there are a log of invalid signatures

consensus/spos/bls/v2/signatureEvidence.go
```diff
@@ -0,0 +1,188 @@
...
+	return ev, true
+}
+
+// GetRetainedQuorumEvidence returns the retained quorum slot for the nonce, if any
+func (store *signatureEvidenceStore) GetRetainedQuorumEvidence(nonce uint64) (*roundSignatureEvidence, bool) {
+	store.mut.RLock()
+	defer store.mut.RUnlock()
+
+	ev := store.retained
+	if ev == nil || ev.nonce != nonce {
```
sstanculeanu, line 146: should this also check if the nonce is settled?

consensus/spos/bls/v2/subroundSignature.go
```diff
@@ -427,6 +437,34 @@ func (sr *subroundSignature) getPkForCompetingBlock(nonce uint64, currentHash []
...
+	if !ok {
+		ev, ok = sr.signatureEvidence.GetRetainedQuorumEvidence(nonce)
+	}
+	if !ok || bytes.Equal(ev.headerHash, currentHash) {
+		return false
+	}
+
+	count := ev.getCount()
+	switch {
+	case count >= ev.threshold:
```
ssd04: suggestion: use `if` conditions, instead of switch .. case

dataRetriever/dataPool/proofsCache/proofsPool.go
```diff
@@ -73,18 +73,50 @@ func (pp *proofsPool) AddProof(
...
 }
 
-func (pp *proofsPool) addProof(
+// AddProofIfNoneAtNonce will add the provided proof only if its (nonce, shard) slot is free or
+// holds the same header hash; unlike AddProof, it never overwrites the pre-existing proof
+func (pp *proofsPool) AddProofIfNoneAtNonce(
 	headerProof data.HeaderProofHandler,
-) bool {
-	shardID := headerProof.GetHeaderShardId()
+) (bool, data.HeaderProofHandler) {
```
ssd04, line 80: suggestions: 
- reverse returns vars `(data.HeaderProofHandler, bool)` to have it consistent to other places, same in proofsCache
- rename method to `IfNotSet`

consensus/spos/bls/v2/proofSelfAssembly.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		if ev.getCount() < ev.threshold {
+			// inflated signal was adversarial; the decision tiers demote to the wait tier
+			store.DropRetained(ev.nonce)
+			log.Debug("trySelfAssembleProof: evidence demoted below threshold after stripping invalid shares",
+				"nonce", ev.nonce, "count", ev.getCount(), "threshold", ev.threshold)
+			return
+		}
+
+		bitmap, agg, err = aggregateAndVerifyEvidence(sr, ev)
```
ssd04, line 46: suggestion: this looks duplicated with `aggregateAndVerifyEvidence` called twice, maybe `stripInvalidShares` explicitly; there is no need for second `stripInvalidShares`, and i think there is need for another threshold check after agg sigs retry

consensus/spos/bls/v2/subroundSignature.go
```diff
@@ -427,6 +437,34 @@ func (sr *subroundSignature) getPkForCompetingBlock(nonce uint64, currentHash []
 	return nil
 }
 
+// shouldAbortOnSignatureEvidence refuses to sign a competing block on quorum-level evidence for
+// the previous round's block (triggering proof self-assembly) and waits on significant evidence
+func (sr *subroundSignature) shouldAbortOnSignatureEvidence(ctx context.Context, nonce uint64, currentHash []byte) bool {
+	ev, ok := sr.signatureEvidence.GetPreviousRoundEvidence(nonce, sr.RoundHandler().Index())
```
ssd04, line 432: the evidence info is gathered before start round (only one evidence per round), after that there will be no additional info for the evidence; at start round we try to assemble the proof; if there is not update to the signature shares on the evidence, does it help to check the evidence here in signatures subround, and in block subround?

factory/crypto/signingHandler.go
```diff
@@ -323,6 +324,95 @@ func (sh *signingHandler) AggregateSigs(bitmap []byte, epoch uint32) ([]byte, er
...
+// selectByBitmap returns the signatures and public keys whose index is set in the bitmap;
+// sigShares may be nil when only the public keys are needed
+func (sh *signingHandler) selectByBitmap(pubKeys []string, bitmap []byte, sigShares [][]byte) ([][]byte, []crypto.PublicKey, error) {
+	if bitmap == nil {
+		return nil, nil, ErrNilBitmap
+	}
+	if len(bitmap)*8 < len(pubKeys) {
+		return nil, nil, ErrBitmapMismatch
+	}
+	if sigShares != nil && len(sigShares) != len(pubKeys) {
```
ssd04: split into separate checks? and avoid additional nil check for sigShares below?

consensus/spos/bls/v2/signatureEvidence.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Capture overwrites the previous-round slot (nil clears it); the replaced evidence is
+// promoted to the retained slot when it holds quorum for a still unsettled nonce.
+func (store *signatureEvidenceStore) Capture(ev *roundSignatureEvidence) {
+	store.mut.Lock()
+	defer store.mut.Unlock()
+
+	replaced := store.previous
+	store.previous = ev
```
sstanculeanu, line 111: intended to store nil when evidence is nil, right?

dataRetriever/dataPool/proofsCache/proofsCache.go
```diff
@@ -80,6 +80,34 @@ func (pc *proofsCache) addProof(proof data.HeaderProofHandler) {
 	pc.proofsByHash[newHash] = proof
 }
 
+// addProofIfNoneAtNonce adds the proof only if its nonce slot is free or already holds the same
+// header hash; returns the added flag and the pre-existing proof for the nonce, if any
+func (pc *proofsCache) addProofIfNoneAtNonce(proof data.HeaderProofHandler) (bool, data.HeaderProofHandler) {
```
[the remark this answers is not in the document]
AdoAdoAdo: fixed

dataRetriever/dataPool/proofsCache/proofsPool.go
```diff
@@ -73,18 +73,50 @@ func (pp *proofsPool) AddProof(
 	return pp.addProof(headerProof)
 }
 
-func (pp *proofsPool) addProof(
+// AddProofIfNoneAtNonce will add the provided proof only if its (nonce, shard) slot is free or
+// holds the same header hash; unlike AddProof, it never overwrites the pre-existing proof
+func (pp *proofsPool) AddProofIfNoneAtNonce(
```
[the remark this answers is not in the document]
AdoAdoAdo: fixed

consensus/spos/bls/v2/proofSelfAssembly.go
AdoAdoAdo: fixed

consensus/spos/bls/v2/signatureEvidence.go
AdoAdoAdo: renamed to retainedQuorum.
it keeps the previous round quorum evidence for the still unsettled nonce.

the capture is done for the exact state of the previous quorum, no matter the number (if number of shares >= 1). The evaluation will be done in the next round if it needs to wait, (>= 0.5*threshold), or if it can start the assembly.

consensus/spos/bls/v2/signatureEvidence.go
AdoAdoAdo: yes, if it is nil, it will clear the evidence.

consensus/spos/bls/v2/signatureEvidence.go
AdoAdoAdo: with the existing calls, this is not reached in case of settled nonce.

consensus/spos/bls/v2/subroundSignature.go
AdoAdoAdo: Indeed the node does not collect new signatures in the new round for the previous round consensus (dropped from worker).

But the wait is still required as the proof may become available from a different node, and as well the refusal to sign is still needed in case a quorum was observed but the proof is not yet available.

consensus/spos/bls/v2/subroundSignature.go
AdoAdoAdo: done

consensus/spos/bls/v2/subroundSignature.go
AdoAdoAdo: it was intended to not add a deadline to tryAssembleProof. It will exit if not enough valid signatures.

consensus/spos/bls/v2/subroundStartRound.go
AdoAdoAdo: right, added a check for proof.

consensus/spos/bls/v2/subroundStartRound.go
AdoAdoAdo: there was a slight difference, but re-used now the method and only clear the bit if the share is missing.

dataRetriever/dataPool/proofsCache/proofsPool.go
AdoAdoAdo: I think the explicit IfNoneAtNonce is ok as we already do a dedup by hash (default)

The order of return values seems ok to me, where do we have them in a different order? here the first return value is the important one (added bool), the second one is the existing proof when not added.

factory/crypto/signingHandler.go
AdoAdoAdo: fixed

dataRetriever/dataPool/proofsCache/proofsPool.go
ssd04: the behaviour is like "get if existing or add if not"; in caches we usually have `val, ok` on get, or `added` as second parameter; but optional suggestion

consensus/spos/bls/v2/subroundSignature.go
ssd04, line 131: do we still need both? i think the checks based on signature evidence are stronger than based on signaturesTracker

AdoAdoAdo: I would leave both for now, at least there is no signing two competing blocks without waiting.

sstanculeanu (approved):

ssd04 (approved):

PR #7912, opened 2026-07-13
multi proof pool

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


process/track/contendedHeadersProofPuller.go
```diff
@@ -0,0 +1,179 @@
...
+		if state == nil || state.nonce != head.GetNonce() {
+			state = &proofPullState{nonce: head.GetNonce(), nextPullRound: currentRound, backoffRounds: 1}
+			bbt.proofPullPerShard[shardID] = state
+		}
+
+		if currentRound < state.nextPullRound {
+			continue
+		}
+
+		log.Debug("pulling proofs for contended head",
```
ssd04, line 82: rename to use `header` ?

process/track/contendedHeadersProofPuller.go
```diff
@@ -0,0 +1,179 @@
...
+
+const maxProofPullBackoffRounds = 8
+
+type proofPullState struct {
+	nonce         uint64
+	nextPullRound int64
+	backoffRounds int64
+}
+
+type headCandidate struct {
```
ssd04, line 25: what is the meaning of `head` here? the highest nonce? or it's just for header

process/track/contendedHeadersProofPuller.go
AdoAdoAdo: This refers to the head of the chain, I will change to chain tip, so it removes the ambiguity.

process/track/contendedHeadersProofPuller.go
AdoAdoAdo: It is the highest nonce yes. Will adopt tip instead of head here as well.

PR #7915, opened 2026-07-14
contended blocks finality

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/block/metablock.go
+++ b/process/block/metablock.go
@@ -1697,6 +1697,11 @@ func (mp *metaProcessor) updateState(metaBlock data.MetaHeaderHandler, metaBlock
 
 	// for header v3, trie prnning is triggered in async mode from headers executor
 
+	if mp.enableEpochsHandler.IsFlagEnabledInEpoch(common.SupernovaFlag, metaBlock.GetEpoch()) {
+		mp.signalNewlyFinalBlocks(metaBlock, metaBlockHash, rootHash, prevMetaBlock)
+		return
+	}
+
 	outportFinalizedHeaderHash := metaBlockHash
 	if !common.IsFlagEnabledAfterEpochsStartBlock(metaBlock, mp.enableEpochsHandler, common.AndromedaFlag) {
 		outportFinalizedHeaderHash = metaBlock.GetPrevHash()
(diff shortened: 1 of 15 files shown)


process/block/metablock.go
```diff
@@ -1706,6 +1711,46 @@ func (mp *metaProcessor) updateState(metaBlock data.MetaHeaderHandler, metaBlock
...
+		mp.lastSignaledFinalNonce = metaBlock.GetNonce()
+		return
+	}
+
+	if finalNonce == prevMetaBlock.GetNonce() {
+		prevRootHash, _, err := mp.getRootHashAndValidatorRootHash(prevMetaBlock)
+		if err != nil {
+			log.Warn("signalNewlyFinalBlocks: cannot get root hash of the newly final block", "error", err.Error())
+		} else {
+			mp.blockChain.SetFinalBlockInfo(prevMetaBlock.GetNonce(), metaBlock.GetPrevHash(), prevRootHash)
```
ssd04, line 1747: combine first and last conditions? `SetFinalBlockInfo` if prev is the new final?

process/block/shardblock.go
```diff
@@ -1302,12 +1302,81 @@ func (sp *shardProcessor) updateState(headers []data.HeaderHandler, currentHeade
...
+		sp.setFinalizedHeaderHashInIndexer(newlyFinalHashes[i])
+	}
+	sp.lastSignaledFinalNonce = finalNonce
+
+	if finalNonce != currentHeader.GetNonce() {
+		return
+	}
+
+	scheduledHeaderRootHash, _ := sp.scheduledTxsExecutionHandler.GetScheduledRootHashForHeader(currentHeaderHash)
+	sp.setFinalBlockInfo(currentHeader, currentHeaderHash, scheduledHeaderRootHash)
```
ssd04, line 1336: i don't think this is still needed, on v3 this will just return early

process/block/metablock.go
```diff
@@ -1697,6 +1697,11 @@ func (mp *metaProcessor) updateState(metaBlock data.MetaHeaderHandler, metaBlock
 
 	// for header v3, trie prnning is triggered in async mode from headers executor
 
+	if mp.enableEpochsHandler.IsFlagEnabledInEpoch(common.SupernovaFlag, metaBlock.GetEpoch()) {
```
ssd04, line 1700: should we check here by round activation? same for the other places

process/block/shardblock.go
```diff
@@ -1302,12 +1302,81 @@ func (sp *shardProcessor) updateState(headers []data.HeaderHandler, currentHeade
...
+	header := currentHeader
+	if !bytes.Equal(finalHash, currentHeaderHash) {
+		var err error
+		header, err = process.GetShardHeader(finalHash, sp.dataPool.Headers(), sp.marshalizer, sp.store)
+		if err != nil {
+			log.Warn("signalNewlyFinalBlocks: cannot load final header, signaling only its hash",
+				"final nonce", finalNonce, "error", err.Error())
+			return hashes
+		}
+	}
```
ssd04, line 1362: is this needed? if current header is not found return just final hash?

process/sync/storageBootstrap/baseStorageBootstrapper.go
```diff
@@ -455,6 +462,16 @@ func (st *storageBootstrapper) applyBootInfos(bootInfos []bootstrapStorage.Boots
 		st.forkDetector.SetFinalToLastCheckpoint()
 	}
 
+	isSupernovaHead := st.enableEpochsHandler.IsFlagEnabledInEpoch(common.SupernovaFlag, bootInfos[0].LastHeader.Epoch)
```
ssd04, line 465:
```suggestion
	isSupernovaHeader := st.enableEpochsHandler.IsFlagEnabledInEpoch(common.SupernovaFlag, bootInfos[0].LastHeader.Epoch)
```

ssd04: or isSupernovaEnabled

process/sync/shardForkDetector.go
```diff
@@ -226,10 +231,53 @@ func (sfd *shardForkDetector) computeFinalCheckpoint() {
...
+	sfd.mutHeaders.RLock()
+	for {
+		child := sfd.getCleanProcessedChild(finalCheckpoint)
+		if child == nil {
+			break
+		}
+
+		finalCheckpoint = &checkpointInfo{nonce: child.nonce, round: child.round, hash: child.hash}
+		advanced = true
+	}
```
ssd04, line 258: should this be blocking? or block with timeout

process/block/metablock.go
AdoAdoAdo: done

process/block/shardblock.go
AdoAdoAdo: it is under Supernova epoch flag, so not reduced to V3 only (before round activation we still have V2). Got refactored into setSettledBlockInfo, but with the same guard, skipped for V3, still used for V2.

process/block/metablock.go
AdoAdoAdo: I don't think it is needed, the finality model linked only to the epoch change was intentional.

process/block/shardblock.go
AdoAdoAdo: yes, if we don't find it we just miss the signal to the indexer about the finality of that range of headers, if we find it then we don't. But this would practically never happen, as it needs a storage corruption, or pruning to already be done over those headers.

process/sync/storageBootstrap/baseStorageBootstrapper.go
AdoAdoAdo: renamed to isSupernovaLastHeader

process/sync/shardForkDetector.go
AdoAdoAdo: this should be not expensive (even with large set/lag - microseconds)
it is in memory traversing fork detector's own map.

PR #7916, opened 2026-07-14
rollback v3 above final

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/sync/baseSync.go
+++ b/process/sync/baseSync.go
@@ -1685,7 +1685,8 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
 	var currBody data.BodyHandler
 
 	defer func() {
-		if !roleBackOneBlockExecuted {
+		isHeaderV3 := !check.IfNil(currHeader) && currHeader.IsHeaderV3()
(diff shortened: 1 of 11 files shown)


process/sync/baseSync.go
```diff
@@ -1735,13 +1738,22 @@ func (boot *baseBootstrap) rollBack(revertUsingForkNonce bool) error {
...
 
-		currBody, err = boot.rollBackOneBlock(
-			currHeaderHash,
-			currHeader,
-			prevHeaderHash,
-			prevHeader,
-		)
-		roleBackOneBlockExecuted = true
+		if currHeader.IsHeaderV3() {
+			currBody, err = boot.rollBackOneBlockV3(
```
ssd04, line 1742: should we set prepared for sync to false in case of rollback on v3?

AdoAdoAdo: it is done on the defer realignAfterV3RollBack.
will refactor a bit to have the realign called also in case of error, where it would have been missed.

PR #7932, opened 2026-07-22
implement meta finality view

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/process/track/metaFinalityView.go
+++ b/process/track/metaFinalityView.go
@@ -1,0 +1,252 @@
+package track
+
+import (
+	"bytes"
+
+	"github.com/multiversx/mx-chain-core-go/core"
+	"github.com/multiversx/mx-chain-core-go/core/check"
+	"github.com/multiversx/mx-chain-core-go/data"
+
+	"github.com/multiversx/mx-chain-go/common"
+	"github.com/multiversx/mx-chain-go/dataRetriever"
+	"github.com/multiversx/mx-chain-go/process"
+)
+
+const maxMetaBlocksScannedForInclusion = 16
+const maxOwnDescendantsScannedForInclusion = 8
+
+// ArgsMetaFinalityView holds the pools the meta finality view reads from
+type ArgsMetaFinalityView struct {
+	HeadersPool dataRetriever.HeadersPool
+	ProofsPool  dataRetriever.ProofsPool
+}
+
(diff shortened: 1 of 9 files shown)


sstanculeanu (approved):

process/track/metaFinalityView.go
```diff
@@ -0,0 +1,252 @@
...
+	metaNonces := mfv.headersPool.Nonces(core.MetachainShardId)
+	if len(metaNonces) == 0 {
+		return false
+	}
+
+	branch := mfv.ownBranchHashes(shardID, headerHash, nonce)
+	highestMetaNonce := highestNonce(metaNonces)
+
+	for scanned := uint64(0); scanned < maxMetaBlocksScannedForInclusion; scanned++ {
+		metaNonce := highestMetaNonce - scanned
```
ssd04, line 98: check for highestMetaNonce >= scanned ?

ssd04: + move check `metaNonce == 0` here?

process/track/metaFinalityView.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return false
+}
+
+func (mfv *metaFinalityView) ownBranchHashes(shardID uint32, headerHash []byte, nonce uint64) [][]byte {
+	branch := [][]byte{headerHash}
+	parents := branch
+
+	for depth := uint64(1); depth <= maxOwnDescendantsScannedForInclusion; depth++ {
+		children, childrenHashes, err := mfv.headersPool.GetHeadersByNonceAndShardId(nonce+depth, shardID)
```
ssd04, line 145: so here we are looking forward start from nonce? it might depend also on how `IsIncludedInHeldFinalMetaBlock` is called/used, but i expected to be looking backwards from the last final meta

process/track/metaFinalityView.go
AdoAdoAdo: this is clearer on the final branch anchored-inclusion-scan

process/track/metaFinalityView.go
AdoAdoAdo: there are two searches one backwards and one forward, the forward lookup here is intentional. Here we are checking if a held final meta references a descendant which means that the header was also accepted.

ssd04 (approved):

PR #7935, opened 2026-07-23
use block tracker for anchored inclusion scan

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites


epochStart/shardchain/trigger.go
```diff
@@ -1221,6 +1243,108 @@ func (t *trigger) RevertStateToBlock(header data.HeaderHandler) error {
...
+	t.forgetEpochStartHeader(deadEpochStartHash)
+	t.removeStoredEpochStartMeta(epoch)
+
+	if bytes.Equal(t.epochMetaBlockHash, deadEpochStartHash) {
+		t.restorePreActivationState()
+	}
+
+	err := t.saveState(t.triggerStateKey)
+	if err != nil {
+		log.Warn("DisarmDeadEpochStartActivation saveState", "error", err)
```
sstanculeanu, line 1272: should this return false here?

process/sync/baseForkDetector.go
```diff
@@ -357,6 +357,61 @@ func (bfd *baseForkDetector) ReconcileFinalCheckpoint(nonce uint64) {
...
+
+	bfd.mutHeaders.Lock()
+	for hdrNonce := range bfd.headers {
+		if hdrNonce >= nonce {
+			delete(bfd.headers, hdrNonce)
+		}
+	}
+	bfd.mutHeaders.Unlock()
+
+	bfd.mutFork.Lock()
```
sstanculeanu, line 384: should we consider here one mutex only?

process/sync/baseForkDetector.go
```diff
@@ -357,6 +357,61 @@ func (bfd *baseForkDetector) ReconcileFinalCheckpoint(nonce uint64) {
...
+	loweredFinal := bfd.fork.finalCheckpoint.nonce >= nonce
+	if loweredFinal {
+		bfd.fork.finalCheckpoint = newFinal
+	}
+	bfd.mutFork.Unlock()
+
+	probableHighestNonce := bfd.computeProbableHighestNonce()
+	bfd.setProbableHighestNonce(probableHighestNonce)
+
+	log.Error("forkDetector.ReconcileFinalCheckpointBelow: finality regressed on dead cross-notarization evidence",
```
sstanculeanu, line 407: maybe downgrade this to warn, as it is basically the success path

process/sync/baseSync.go
```diff
@@ -2461,6 +2481,124 @@ func (boot *baseBootstrap) reconcileEvidenceStillApplies(evidence *reconcileEvid
...
+		return false
+	}
+
+	headHash := boot.chainHandler.GetCurrentBlockHeaderHash()
+	earliestDeadNonce, deadOwnHashes, collected := boot.collectOwnBlocksReferencing(deadMetaHash)
+	if !collected {
+		return false
+	}
+
+	boot.disarmDeadEpochStartIfNeeded(deadMeta, deadMetaHash)
```
sstanculeanu: not sure if this should be before the log.Error on L2515

process/block/metablockProposal.go
```diff
@@ -1158,6 +1175,137 @@ func (mp *metaProcessor) getArbitrationCandidate(parentInfo ShardHeaderInfo) (da
...
+	return newMetaAncestryView(parentHeader, parentHash)
+}
+
+func (mp *metaProcessor) newVerifyAncestryView(metaHeaderHandler data.MetaHeaderHandler) (*metaAncestryView, error) {
+	if !mp.enableEpochsHandler.IsFlagEnabled(common.SupernovaFlag) {
+		return nil, nil
+	}
+
+	prevHash := metaHeaderHandler.GetPrevHash()
+	if !check.IfNil(mp.blockChain) {
```
ssd04, line 1225: is the nil check needed? there is the check in constructor and it's used in a lot places in process/block without nil check

ssd04: same for other places

process/block/metablockProposal.go
```diff
@@ -1158,6 +1175,137 @@ func (mp *metaProcessor) getArbitrationCandidate(parentInfo ShardHeaderInfo) (da
...
+		genesisHeader := mp.blockChain.GetGenesisHeader()
+		if !check.IfNil(genesisHeader) && bytes.Equal(prevHash, mp.blockChain.GetGenesisHeaderHash()) {
```
ssd04, line 1227: i think the same with genesisHeader nil check, in all other places we trust blockChain that genesis header is not nil

process/block/metablockProposal.go
```diff
@@ -1344,25 +1497,35 @@ func (mp *metaProcessor) checkShardInfoValidity(metaHeaderHandler data.MetaHeade
...
 			continue
 		}
 
-		err = mp.checkShardHeaderContention(shardHdrInfo.Header, shardHdrInfo.Hash, lastNotarizedHeaderInfoForShard)
+		err = mp.checkShardHeaderContention(shardHdrInfo.Header, shardHdrInfo.Hash, lastNotarizedHeaderInfoForShard, ancestryView)
+		if err != nil {
+			return err
+		}
+
+		err = mp.checkReferencedMetaAncestry(shardHdrInfo.Header, ancestryView)
```
ssd04, line 1528: can we avoid multiple ancestry checks in case on unsettled parent header?

process/block/metablockProposal.go
```diff
@@ -1344,25 +1497,35 @@ func (mp *metaProcessor) checkShardInfoValidity(metaHeaderHandler data.MetaHeade
 func (mp *metaProcessor) verifyUsedShardHeadersValidity(
 	usedShardHeaders map[uint32][]ShardHeaderInfo,
 	lastCrossNotarizedHeader map[uint32]ShardHeaderInfo,
+	ancestryView *metaAncestryView,
 ) error {
 	var err error
 	for shardID, hdrsForShard := range usedShardHeaders {
-		err = mp.checkHeadersSequenceCorrectness(hdrsForShard, lastCrossNotarizedHeader[shardID])
+		err = mp.checkHeadersSequenceCorrectness(hdrsForShard, lastCrossNotarizedHeader[shardID], ancestryView)
```
ssd04, line 1504: is the ancestry view mutated between different shard header walks?

epochStart/shardchain/trigger.go
AdoAdoAdo: here the disarm was already done, the only problem was a save of the trigger state normally (should not happen and we already have a warn for it) - which means that it would affect us in case of a restart, before another change.

process/sync/baseForkDetector.go
AdoAdoAdo: I don't think it is needed, these are not kept at the same time in other places to deserve a single mutex.

process/block/metablockProposal.go
AdoAdoAdo: yes, by adding blocks to it.
I think there is a better way to just check membership without building the entire ancestry if the referenced meta is below the final, will come with an update on that.

process/sync/baseSync.go
AdoAdoAdo: moved it

process/block/metablockProposal.go
AdoAdoAdo: improved the check, and the first pass populates the set that will be verified -> becomes cheap to verify.

sstanculeanu (approved):

ssd04 (approved):

PR #7962, opened 2026-08-14
add header hash support to block proposal state management

AdoAdoAdo:
- 
- 
- 
  
## Proposed changes
- 
- 
- 

## Testing procedure
- 
- 
- 

## Pre-requisites

### DIFF
--- a/outport/process/outportDataProvider.go
+++ b/outport/process/outportDataProvider.go
@@ -3,6 +3,7 @@ package process
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/big"
 
@@ -18,8 +19,9 @@ import (
 	"github.com/multiversx/mx-chain-core-go/data/transaction"
 	"github.com/multiversx/mx-chain-core-go/hashing"
 	"github.com/multiversx/mx-chain-core-go/marshal"
-	"github.com/multiversx/mx-chain-go/storage"
 	logger "github.com/multiversx/mx-chain-logger-go"
+
+	"github.com/multiversx/mx-chain-go/storage"
 
 	"github.com/multiversx/mx-chain-go/common"
 	"github.com/multiversx/mx-chain-go/dataRetriever"
@@ -227,7 +229,7 @@ func (odp *outportDataProvider) getStateAccesses(
 			rootHash = scheduledRootHash
 		}
 
-		stateAccesses := odp.getStateAccessForRootHash(rootHash)
+		stateAccesses := odp.getStateAccessForHeader(headerHash, rootHash)
 		stateAccessesForBlock[hex.EncodeToString(headerHash)] = &outportcore.StateAccessesForBlock{StateAccesses: stateAccesses}
 		// for backward compatibility, in case some driver is still using StateAccesses instead of StateAccessesForBlock
 		return stateAccessesForBlock, stateAccesses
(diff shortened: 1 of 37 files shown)


outport/process/outportDataProvider.go
```diff
@@ -720,7 +733,7 @@ func (odp *outportDataProvider) IsInterfaceNil() bool {
 func (odp *outportDataProvider) getIntraShardMiniBlocks(bodyHandler data.BodyHandler, headerHandler data.HeaderHandler) ([]*block.MiniBlock, error) {
 	if headerHandler.IsHeaderV3() {
 		// skip intra-shard miniblocks.
-		// they are returned later when this block’s execution result is included in a future block.
+		// they are returned later when this block\u2019s execution result is included in a future block.
```
BeniaminDrasovean: Change the unicode back to `

state/stateAccesses/collector.go
```diff
@@ -88,53 +93,71 @@ func (c *collector) GetAccountChanges(oldAccount, account vmcommon.AccountHandle
...
+func (c *collector) TakeStateAccessesForHeader(headerHash, expectedRootHash []byte) (map[string]*data.StateAccesses, error) {
+	c.retainedMut.Lock()
+	retained, ok := c.stateAccessesForHeader[string(headerHash)]
 	if !ok {
-		log.Warn("stateAccessesForRootHash not found in stateAccessesForBlock", "rootHash", rootHash)
-		return nil
+		c.retainedMut.Unlock()
+		return nil, nil
+	}
+	delete(c.stateAccessesForHeader, string(headerHash))
```
BeniaminDrasovean, line 134: Is it ok to delete here, even if it is not equal to expected? Is it possible that expected is wrong, and this will be needed in the future?

outport/process/outportDataProvider.go
AdoAdoAdo: fixed

state/stateAccesses/collector.go
AdoAdoAdo: there should be no production path where the roothash does not match for the given hash, when reaching commit. Also the clean up here is safe, as there is no retrial if we reached this point, and a potential mismatch has already been logged.

process/asyncExecution/executionTrack/executionResultsTracker.go
```diff
@@ -32,14 +32,37 @@ type executionResultsTracker struct {
...
 		nonceHash:                newNonceHash(),
 		consensusCommittedHashes: make(map[uint64][]byte),
+		dismissedHandler:         dismissedHandler,
+	}, nil
+}
+
+func (ert *executionResultsTracker) unlockAndNotifyDismissed() {
+	notifications := ert.dismissedNotifications
+	ert.dismissedNotifications = nil
+	ert.mutex.Unlock()
```
ssd04: i think it's better to keep them separated, `Unlock` and the NotifyDismissed, to keep Lock/Unlock in the same context func

state/stateAccesses/collector.go
```diff
@@ -235,39 +258,59 @@ func stateAccessToString(stateAccess *data.StateAccess) string {
...
-	c.stateAccessesMut.Unlock()
 
-	return c.storer.Store(collectedStateAccesses)
+	if len(headerHash) == 0 {
+		return nil
+	}
+
+	c.retainedMut.Lock()
+	c.stateAccessesForHeader[string(headerHash)] = &committedStateAccesses{
+		rootHash: append([]byte(nil), rootHash...),
```
ssd04, line 295: is this needed? we should not end up with a nil rootHash at this point, maybe early return?

state/stateAccesses/collector.go
```diff
@@ -88,53 +93,71 @@ func (c *collector) GetAccountChanges(oldAccount, account vmcommon.AccountHandle
...
 }
 
-// GetStateAccessesForRootHash will return the collected state accesses
-func (c *collector) GetStateAccessesForRootHash(rootHash []byte) map[string]*data.StateAccesses {
-	c.stateAccessesForBlockMut.RLock()
-	defer c.stateAccessesForBlockMut.RUnlock()
+// EndExecution clears the identity when the generation is still current
+func (c *collector) EndExecution(generation uint64) {
+	c.headerScopeMut.Lock()
+	if c.headerGeneration == generation {
```
ssd04: is `headerGeneration` needed? add a header hash check?

state/stateAccesses/collector.go
```diff
@@ -235,39 +258,59 @@ func stateAccessToString(stateAccess *data.StateAccess) string {
...
 
-	c.stateAccessesForBlockMut.Lock()
-	_, alreadyCollected := c.stateAccessesForBlock[string(rootHash)]
-	if !alreadyCollected {
-		c.stateAccessesForBlock[string(rootHash)] = collectedStateAccesses
-		log.Trace("state accesses collected", "numStateAccesses", len(collectedStateAccesses), "rootHash", rootHash)
-	} else {
-		log.Debug("state accesses already collected for root hash, keeping previous accesses", "rootHash", rootHash)
+	c.headerScopeMut.RLock()
+	headerHash := append([]byte(nil), c.headerHash...)
```
raduchistest, line 264: c.headerHash could be replaced by BeginExecution while the commit is running. 
maybe pass the headerHash explicitly to CommitCollectedAccesses alongside the rootHash

state/errors.go
```diff
@@ -174,3 +174,23 @@ var ErrStateAccessesIndexOutOfBounds = errors.New("state access index out of bou
...
+// StateAccessesRootMismatchError contains state-access identity diagnostics
+type StateAccessesRootMismatchError struct {
+	HeaderHash   []byte
+	ExpectedRoot []byte
+	ActualRoot   []byte
+}
+
+// Error returns the root mismatch message
+func (e *StateAccessesRootMismatchError) Error() string {
+	return ErrStateAccessesRootMismatch.Error()
```
raduchistest: could add HeaderHash, ExpectedRoot, ActualRoot to the error for debugging purposes

state/stateAccesses/collector.go
raduchis, line 264: c.headerHash could be replaced by BeginExecution while the commit is running.
maybe pass the headerHash explicitly to CommitCollectedAccesses alongside the rootHash

state/errors.go
raduchis: could add HeaderHash, ExpectedRoot, ActualRoot to the error for debugging purposes

raduchis (changes requested):

state/stateAccesses/collector.go
AdoAdoAdo: There should be only synchronous calls for this, valid as a robustness against future changes, but currently there is no path that would call it asynchronously/in parallel.

I would postpone a robustness change, as it would touch multiple places.

state/errors.go
AdoAdoAdo: done

process/asyncExecution/executionTrack/executionResultsTracker.go
AdoAdoAdo: done

state/stateAccesses/collector.go
AdoAdoAdo: the roothash is not nil at this point, as the only caller is AccountsDB.commit, and it does an error check before (and RootHash() never returns nil without an error).
the append is used here to copy the roothash and not use the trie internal one.

state/stateAccesses/collector.go
AdoAdoAdo: right, the header check is the better protection, especially since we don't allow overlapping collections.

ssd04 (approved):

raduchis (approved):

